PR #332, opened 2026-05-17
Fix 422 PR creation: detect existing PR, mark success

jmelloy:
## Summary

- When GitHub returns 422 on `POST /repos/:repo/pulls` (which happens when a PR for that branch already exists), `open_pr()` now calls `find_existing_pr()` to look up the open PR instead of immediately returning `None`.
- If an existing PR is found, it is returned as the PR URL — identical behaviour to a freshly created PR, so the task is marked successful and the PR is associated with the task output.
- If no existing PR is found after the 422, the function still returns `None` (genuine failure), preserving current failure semantics.
- Four new tests cover: successful creation, 422 with existing PR, 422 with no existing PR, and non-422 HTTP errors.

## Test plan

- [x] `pytest worker/tests/test_open_pr.py` — 4 new tests, all pass
- [x] `pytest worker/` — 137 tests total, all pass, no regressions

### DIFF
--- a/worker/pioneer_worker/github_pr.py
+++ b/worker/pioneer_worker/github_pr.py
@@ -195,7 +195,18 @@ async def open_pr(
         pr_url = result.get("html_url", "")
         await emit(f"[worker] ✓ PR: {pr_url}")
         return pr_url
-    except (urllib.error.URLError, urllib.error.HTTPError, OSError) as exc:
+    except urllib.error.HTTPError as exc:
+        if exc.code == 422:
+            await emit(f"[worker] PR creation returned 422 — checking for existing PR on {branch}")
+            existing = await find_existing_pr(
+                branch=branch, worktree_path=worktree_path, token=token
+            )
+            if existing:
+                await emit(f"[worker] ✓ Found existing PR: {existing}")
+                return existing
+        await emit(f"[worker] PR failed: {exc}")
+        return None
+    except (urllib.error.URLError, OSError) as exc:
         await emit(f"[worker] PR failed: {exc}")
         return None
 
--- a/worker/pioneer_worker/github_pr.py
+++ b/worker/pioneer_worker/github_pr.py
@@ -196,6 +196,8 @@ async def open_pr(
         await emit(f"[worker] ✓ PR: {pr_url}")
         return pr_url
     except urllib.error.HTTPError as exc:
+        if exc.fp is not None:
+            exc.fp.close()
         if exc.code == 422:
             await emit(f"[worker] PR creation returned 422 — checking for existing PR on {branch}")
             existing = await find_existing_pr(
@@ -204,8 +206,10 @@ async def open_pr(
             if existing:
                 await emit(f"[worker] ✓ Found existing PR: {existing}")
                 return existing
-        await emit(f"[worker] PR failed: {exc}")
-        return None
+            await emit(f"[worker] PR failed: {exc}")
+            return None
+        logger.error("PR creation failed: %s", exc)
+        raise
     except (urllib.error.URLError, OSError) as exc:
         await emit(f"[worker] PR failed: {exc}")
         return None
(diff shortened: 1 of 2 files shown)


worker/tests/test_open_pr.py
```diff
@@ -0,0 +1,134 @@
...
+async def test_open_pr_non_422_http_error_returns_none():
+    """Non-422 HTTP errors should not trigger the existing-PR lookup."""
+    emit = AsyncMock()
+
+    with patch(
+        "pioneer_worker.github_pr.find_existing_pr",
+        new=AsyncMock(return_value="https://github.com/owner/repo/pull/99"),
+    ) as mock_find:
+        with patch(
+            "urllib.request.urlopen",
```
jmelloy, line 118: `mock_find` here is the return value of the `patch(...)` context manager's `__enter__`, but because you used `as mock_find` on the *outer* `with patch(...) as mock_find` and then nested another `with patch(...)` inside, `mock_find` is actually the `MagicMock` for `find_existing_pr`. However the `new=AsyncMock(...)` kwarg means the patcher replaces the target with the supplied object and the `as` variable is that same object — so the assertion is actually correct. That said, this is confusing; prefer a single `with` statement or capture the mock explicitly to make intent clear.

worker/pioneer_worker/github_pr.py
```diff
@@ -195,7 +195,18 @@ def _create_pr() -> dict:
         pr_url = result.get("html_url", "")
         await emit(f"[worker] ✓ PR: {pr_url}")
         return pr_url
-    except (urllib.error.URLError, urllib.error.HTTPError, OSError) as exc:
+    except urllib.error.HTTPError as exc:
+        if exc.code == 422:
+            await emit(f"[worker] PR creation returned 422 — checking for existing PR on {branch}")
+            existing = await find_existing_pr(
+                branch=branch, worktree_path=worktree_path, token=token
```
jmelloy, line 204: When `exc.code != 422`, the `HTTPError` is caught, logged, and `None` is returned, but the response body (`exc.fp`) is never read or closed. While CPython's GC will eventually close it, it is better to explicitly call `exc.fp.close()` or use a `finally` block to avoid potential resource leaks or socket exhaustion under load.

worker/tests/test_open_pr.py
```diff
@@ -0,0 +1,134 @@
...
+            side_effect=_http_error(422),
+        ):
+            result = await github_pr.open_pr(
+                task=_TASK,
+                branch=_BRANCH,
+                worktree_path=_WORKTREE,
+                token=_TOKEN,
+                emit=emit,
+            )
+
```
jmelloy, line 78: The test confirms `find_existing_pr` is called (indirectly via the return value) but never asserts it was called with the correct `branch`, `worktree_path`, and `token` arguments. Adding `mock_find.assert_awaited_once_with(branch=_BRANCH, worktree_path=_WORKTREE, token=_TOKEN)` would catch any future argument-passing regressions.

jmelloy (commented):
- Core logic is correct: 422 now triggers a fallback to `find_existing_pr()`, preserving failure semantics when no existing PR is found.
- The exception handling split (`HTTPError` caught first, then `URLError`/`OSError`) is properly ordered since `HTTPError` is a subclass of `URLError`.
- A subtle issue exists: for non-422 `HTTPError` codes, the error is logged but the exception body/response is not read/closed, which could cause resource leaks depending on `urllib` internals.
- The test for `test_open_pr_non_422_http_error_returns_none` uses a `with patch(...) as mock_find` on the outer context manager but `mock_find` refers to the context manager, not the mock itself — `mock_find.assert_not_called()` may silently pass incorrectly.
- Tests do not verify that `find_existing_pr` is called with the correct arguments (branch, worktree_path, token) on a 422, which would strengthen the regression coverage.

PR #340, opened 2026-05-18
Lazy repo cloning: clone on demand, not upfront

### LINKED ISSUE #339: Worker repo preparation should be lazy (clone on demand, not upfront)
## Problem

When a worker starts or picks up a task, it currently pre-clones/prepares **all** configured repos before beginning work. This causes significant startup delay — sometimes cloning 50+ repos even when a task only touches one.

## Desired behaviour

- Only clone/prepare a repo when it is actually needed for the current task
- On worker startup, do not clone anything
- When a task is assigned, clone only the repo(s) referenced by that task

## Impact

- Faster task startup
- Less disk/network usage
- Workers become available sooner

jmelloy:
## Summary

- `pull_repos` in `git_ops.py` previously cloned any configured repo that wasn't present locally when the worker was idle. This caused upfront network I/O for every configured repo on startup and during idle polling, even if a task never needed that repo.
- Changed `pull_repos` to skip repos not yet on disk (logging at DEBUG level) and only fast-forward repos that are already cloned. The existing `ensure_repo` call inside `_execute_task` already handles lazy cloning the first time a task references a repo.
- Added 3 tests to `test_org_clone.py` covering: skip-on-missing, fetch-on-present, and the mixed-list case.

## Test plan
- [ ] `python -m pytest worker/tests/test_org_clone.py` — all 10 tests pass
- [ ] `python -m pytest` — all 163 worker tests pass with no regressions

Closes #339

### DIFF
--- a/worker/pioneer_worker/git_ops.py
+++ b/worker/pioneer_worker/git_ops.py
@@ -125,7 +125,11 @@ async def pull_repos(
     token: str | None,
     emit: EmitFn,
 ) -> None:
-    """Refresh all configured repos; emit progress through *emit*."""
+    """Refresh repos that are already cloned locally; skip repos not yet on disk.
+
+    Repos are cloned lazily the first time a task needs them, so this function
+    only fast-forwards existing clones rather than triggering upfront clones.
+    """
     logger.info("pull_repos: refreshing %d repo(s)", len(repos))
     for repo_full in repos:
         parts = repo_full.split("/", 1)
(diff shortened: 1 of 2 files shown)


worker/pioneer_worker/git_ops.py
```diff
@@ -135,15 +139,13 @@ async def pull_repos(
         owner, name = parts
         local_path = os.path.join(repos_dir, owner, name)
         if not os.path.exists(os.path.join(local_path, ".git")):
-            logger.info("pull_repos: %s missing locally — cloning", repo_full)
-            await emit(f"[worker] Cloning {repo_full}...")
-            await ensure_repo(repos_dir, repo_full, token)
+            logger.debug("pull_repos: %s not cloned yet — will clone on demand", repo_full)
+            continue
+        rc, _, err = await run_git(["fetch", "origin"], cwd=local_path)
```
jmelloy, line 144: Bug: `rc` here is the return code of `run_git(["fetch", "origin"])`, not of the subsequent `merge --ff-only`. If the fetch succeeds (rc=0) but the merge fails (e.g. diverged history, detached HEAD), the failure is silently swallowed and `"Pulled {repo_full}"` is still emitted. Capture the merge return code separately and check it instead.

worker/pioneer_worker/git_ops.py
```diff
@@ -135,15 +139,13 @@ async def pull_repos(
         owner, name = parts
         local_path = os.path.join(repos_dir, owner, name)
         if not os.path.exists(os.path.join(local_path, ".git")):
-            logger.info("pull_repos: %s missing locally — cloning", repo_full)
-            await emit(f"[worker] Cloning {repo_full}...")
-            await ensure_repo(repos_dir, repo_full, token)
+            logger.debug("pull_repos: %s not cloned yet — will clone on demand", repo_full)
+            continue
```
jmelloy, line 143: The result of `run_git(["merge", "--ff-only", "origin/HEAD"])` is completely discarded. At minimum capture the return code and stderr so merge failures can be logged; a failed fast-forward leaves the working tree stale with no indication to the operator.

worker/tests/test_org_clone.py
```diff
@@ -290,3 +293,89 @@ def test_known_repos_deduplicates_org_repos(tmp_path):
...
+
+
+@pytest.mark.asyncio
+async def test_pull_repos_updates_already_cloned_repo(tmp_path):
+    """pull_repos fetches and fast-forwards repos that are already on disk."""
+    repos_dir = tmp_path / "repos"
+    repo_path = repos_dir / "owner" / "myrepo"
+    (repo_path / ".git").mkdir(parents=True)
+
+    emitted: list[str] = []
```
jmelloy, line 332: The test only asserts that a `fetch` git call was made, but does not verify `merge --ff-only` is also invoked. Adding `assert any("merge" in c for c in git_calls)` would catch an accidental regression where the merge step is removed.

jmelloy (commented):
- The lazy-cloning change correctly defers network I/O to task-execution time, which is a sound architectural improvement and aligns `pull_repos` with a single responsibility (refreshing, not provisioning).
- A subtle bug exists: the fetch return code `rc` is checked for the `merge` success condition, but `rc` actually reflects the `fetch` exit code — a failed merge (e.g. non-fast-forward) would silently succeed from the caller's perspective.
- The `err` variable used in the warning path comes from the `fetch` call but the failure may originate in the `merge --ff-only` step; merge errors are never surfaced or logged.
- Tests are well-structured and cover the three important cases (skip, update, mixed); however `test_pull_repos_updates_already_cloned_repo` does not verify that `merge --ff-only` is also called, leaving a gap if the merge step were accidentally dropped.
- Minor: the initial `logger.info` line still says "refreshing %d repo(s)" which is now slightly misleading since not all repos will necessarily be refreshed; a small wording tweak would improve observability.

PR #341, opened 2026-05-18
feat(worker): droid-style worker name formatting

jmelloy:
## Summary

- Add `format_worker_id(worker_id)` to `backend/utils.py` — strips the `w-` prefix, splits at the first digit boundary, and uppercases both parts (e.g. `w-vd3566` → `VD-3566`, `w-abc123` → `ABC-123`)
- Add matching `formatWorkerId` utility to `frontend/src/utils/format.ts` used as a fallback when the backend doesn't supply a `workerName`
- `worker_display_name` now delegates to `format_worker_id`; the `hostname` parameter is kept for API compatibility but is no longer used
- Update all related tests (`backend/tests/test_worker_name.py`, `worker/tests/test_worker_name.py`) to expect the new droid-style format

## Test plan

- [ ] `python -m pytest backend/tests/test_worker_name.py` passes with new assertions
- [ ] `python -m pytest worker/tests/test_worker_name.py` passes with new assertions
- [ ] Workers registered via the backend display names like `VD-3566` in the UI instead of `tok/w-vd3566`

### DIFF
--- a/backend/utils.py
+++ b/backend/utils.py
@@ -95,16 +95,27 @@ def generate_guild_id(name: str = "", existing_ids: set[str] | None = None) -> s
             return unique
 
 
-def worker_display_name(worker_id: str, hostname: str | None = None) -> str:
-    """Render a worker id as ``hostname[:3]/worker_id`` (e.g. ``tok/w-g2otus``).
+def format_worker_id(worker_id: str) -> str:
+    """Format a worker ID in droid style: ``w-vd3566`` → ``VD-3566``.
 
-    When *hostname* is absent the bare *worker_id* is returned so callers
-    always have a usable label.  Edge cases: hostnames shorter than 3 chars
-    are used in full.
+    Strips the ``w-`` prefix, splits at the first digit boundary, uppercases
+    both parts, and joins with a hyphen.  Examples: ``w-ab1234`` → ``AB-1234``,
+    ``w-x9`` → ``X-9``, ``w-g2otus`` → ``G-2OTUS``.
     """
-    if hostname:
-        return f"{hostname[:3]}/{worker_id}"
-    return worker_id
+    bare = worker_id.removeprefix("w-")
+    m = re.search(r"\d", bare)
+    if m and m.start() > 0:
+        return f"{bare[: m.start()].upper()}-{bare[m.start() :].upper()}"
+    return bare.upper()
+
+
+def worker_display_name(worker_id: str, hostname: str | None = None) -> str:
+    """Render a worker ID in droid style (e.g. ``VD-3566``).
+
+    The *hostname* parameter is accepted for backwards compatibility but is
+    no longer used — the droid-style format derives entirely from the worker ID.
+    """
+    return format_worker_id(worker_id)
 
 
 def decode_claude_oauth_token(blob: str | None) -> str | None:
--- a/frontend/src/utils/format.ts
+++ b/frontend/src/utils/format.ts
@@ -1,5 +1,15 @@
 // Shared time formatters. Keep these tiny and pure — they show up everywhere.
 
+// "w-vd3566" → "VD-3566": strip prefix, split at first digit, uppercase both parts.
+export function formatWorkerId(workerId: string): string {
+  const bare = workerId.replace(/^w-/, '')
+  const m = bare.search(/\d/)
+  if (m > 0) {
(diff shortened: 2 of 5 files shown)


backend/utils.py
```diff
@@ -95,16 +95,27 @@ def generate_guild_id(name: str = "", existing_ids: set[str] | None = None) -> s
...
+def format_worker_id(worker_id: str) -> str:
+    """Format a worker ID in droid style: ``w-vd3566`` → ``VD-3566``.
+
+    Strips the ``w-`` prefix, splits at the first digit boundary, uppercases
+    both parts, and joins with a hyphen.  Examples: ``w-ab1234`` → ``AB-1234``,
+    ``w-x9`` → ``X-9``, ``w-g2otus`` → ``G-2OTUS``.
+    """
+    bare = worker_id.removeprefix("w-")
+    m = re.search(r"\d", bare)
+    if m and m.start() > 0:
```
jmelloy, line 107: The `hostname` parameter is now silently ignored with no deprecation warning. Consider adding `warnings.warn('hostname is deprecated and will be removed', DeprecationWarning, stacklevel=2)` so existing callers are notified rather than experiencing a silent behaviour change.

backend/utils.py
```diff
@@ -95,16 +95,27 @@ def generate_guild_id(name: str = "", existing_ids: set[str] | None = None) -> s
             return unique
 
 
+def format_worker_id(worker_id: str) -> str:
+    """Format a worker ID in droid style: ``w-vd3566`` → ``VD-3566``.
```
jmelloy, line 99: Edge cases are unhandled: if `worker_id` has no `w-` prefix or starts with a digit after stripping (e.g. `w-123abc`), the function returns the full bare string uppercased without any hyphen. Add a guard or document/test these cases explicitly.

worker/tests/test_worker_name.py
```diff
@@ -81,21 +82,28 @@ async def test_register_name_falls_back_to_worker_id_when_absent():
 
 
 # ---------------------------------------------------------------------------
-# Name format invariants (black-box, driven by the backend formula)
+# Name format invariants (droid-style: leading-letters + '-' + rest, uppercased)
```
jmelloy, line 85: `_droid_format` is a copy of the backend's `format_worker_id` logic. If the backend algorithm changes, these worker tests will still pass despite being wrong. Import `format_worker_id` from `backend/utils.py` (via `sys.path` manipulation already used elsewhere) to keep a single source of truth.

frontend/src/utils/format.ts
```diff
@@ -1,5 +1,15 @@
 // Shared time formatters. Keep these tiny and pure — they show up everywhere.
 
+// "w-vd3566" → "VD-3566": strip prefix, split at first digit, uppercase both parts.
+export function formatWorkerId(workerId: string): string {
+  const bare = workerId.replace(/^w-/, '')
+  const m = bare.search(/\d/)
+  if (m > 0) {
```
jmelloy, line 7: When `m === 0` (worker ID starts with a digit after stripping the prefix, e.g. `w-123`), the function returns the bare string uppercased with no hyphen, silently producing a different format than the documented droid-style. This matches the Python behaviour but should be tested and documented as an intentional edge-case decision.

jmelloy (commented):
- The droid-style formatting logic is duplicated across three locations: `backend/utils.py`, `frontend/src/utils/format.ts`, and the `_droid_format` helper inside `worker/tests/test_worker_name.py`, creating a maintenance risk if the algorithm ever changes.
- The `worker_display_name` function silently ignores the `hostname` parameter with no deprecation warning or changelog note, which could surprise callers who still pass hostnames expecting the old behavior.
- Edge cases for `format_worker_id` are not fully covered: worker IDs that start with a digit (e.g. `w-123abc`), empty strings, or IDs without the `w-` prefix will produce unexpected results (the digit-first case returns the entire bare string uppercased with no hyphen).
- The worker-side test file reimplements `_droid_format` as a standalone helper instead of importing `format_worker_id` from `backend/utils.py`, which means the worker tests can pass even if the backend implementation diverges.
- Overall the PR is clean and well-structured; test coverage for the new formatting function is good, and the frontend fallback correctly guards against a falsy `workerId`.

jmelloy:
Closing as duplicate of #342.

PR #342, opened 2026-05-18
Worker names: droid-style formatting (VD-3566)

### LINKED ISSUE #338: Worker names: capitalize, strip `w-` prefix, use droid-style formatting
## Summary

Worker IDs like `w-vd3566` should be displayed as droid-style names: capitalize the alphanumeric portion and drop the `w-` prefix, so `w-vd3566` → `VD-3566`.

## Requirements

- Strip the `w-` prefix from worker IDs
- Capitalize all remaining characters
- Insert a hyphen between the letter portion and numeric portion if not already present (e.g. `vd3566` → `VD-3566`)
- Apply consistently everywhere worker names are rendered: sidebar, tabs, task rows, API responses, log lines

## Examples

| Raw ID | Display Name |
|--------|-------------|
| `w-vd3566` | `VD-3566` |
| `w-ab1234` | `AB-1234` |
| `w-x9` | `X-9` |

## Acceptance Criteria

- [ ] Worker display name helper function exists and is unit-tested
- [ ] All UI surfaces (sidebar, agent slots, task list) use the formatted name
- [ ] Raw IDs are still used internally for routing/API calls

jmelloy:
**Task:** Worker names: droid-style formatting (VD-3566)
**Task ID:** `t-x0i4c0`

Automated by Pioneer Square worker agent.

Closes #338

### DIFF
--- a/backend/utils.py
+++ b/backend/utils.py
@@ -110,12 +110,15 @@ def format_worker_id(worker_id: str) -> str:
 
 
 def worker_display_name(worker_id: str, hostname: str | None = None) -> str:
-    """Render a worker ID in droid style (e.g. ``VD-3566``).
+    """Render a worker ID as ``hostname/VD-3566``, or just ``VD-3566`` when no hostname.
 
-    The *hostname* parameter is accepted for backwards compatibility but is
-    no longer used — the droid-style format derives entirely from the worker ID.
+    The hostname, when present, is prepended with a ``/`` separator so operators
+    can tell workers on different machines apart at a glance.
     """
-    return format_worker_id(worker_id)
+    droid = format_worker_id(worker_id)
+    if hostname:
+        return f"{hostname}/{droid}"
+    return droid
 
 
 def decode_claude_oauth_token(blob: str | None) -> str | None:
--- a/backend/utils.py
+++ b/backend/utils.py
@@ -95,16 +95,27 @@ def generate_guild_id(name: str = "", existing_ids: set[str] | None = None) -> s
             return unique
 
 
-def worker_display_name(worker_id: str, hostname: str | None = None) -> str:
-    """Render a worker id as ``hostname[:3]/worker_id`` (e.g. ``tok/w-g2otus``).
+def format_worker_id(worker_id: str) -> str:
+    """Format a worker ID in droid style: ``w-vd3566`` → ``VD-3566``.
 
-    When *hostname* is absent the bare *worker_id* is returned so callers
-    always have a usable label.  Edge cases: hostnames shorter than 3 chars
-    are used in full.
+    Strips the ``w-`` prefix, splits at the first digit boundary, uppercases
+    both parts, and joins with a hyphen.  Examples: ``w-ab1234`` → ``AB-1234``,
+    ``w-x9`` → ``X-9``, ``w-g2otus`` → ``G-2OTUS``.
     """
-    if hostname:
-        return f"{hostname[:3]}/{worker_id}"
-    return worker_id
+    bare = worker_id.removeprefix("w-")
+    m = re.search(r"\d", bare)
+    if m and m.start() > 0:
+        return f"{bare[: m.start()].upper()}-{bare[m.start() :].upper()}"
+    return bare.upper()
+
+
+def worker_display_name(worker_id: str, hostname: str | None = None) -> str:
+    """Render a worker ID in droid style (e.g. ``VD-3566``).
+
+    The *hostname* parameter is accepted for backwards compatibility but is
+    no longer used — the droid-style format derives entirely from the worker ID.
+    """
+    return format_worker_id(worker_id)
 
 
 def decode_claude_oauth_token(blob: str | None) -> str | None:
--- a/backend/utils.py
+++ b/backend/utils.py
@@ -95,12 +95,35 @@ def generate_guild_id(name: str = "", existing_ids: set[str] | None = None) -> s
             return unique
 
 
+# ---------------------------------------------------------------------------
+# MIRROR: format_worker_id is intentionally duplicated in
+#   frontend/src/utils/format.ts as ``formatWorkerId``.
+# Both implementations MUST be kept in sync.
+#
+# Transformation rules:
+#   1. Strip the "w-" prefix (removeprefix is a no-op if absent).
+#   2. Find the first digit in the remaining string.
+#   3. If a digit exists at position > 0, split there:
+#        LEFT  = everything before the first digit (uppercased)
+#        RIGHT = everything from the first digit onward (uppercased)
+#        result = LEFT + "-" + RIGHT
+#   4. Otherwise return the whole string uppercased (no hyphen inserted).
+#
+# Examples:  w-vd3566 → VD-3566 | w-ab1234 → AB-1234 | w-x9 → X-9
+# Edge cases:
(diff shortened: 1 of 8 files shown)


jmelloy (commented):
- The PR cleanly implements droid-style worker name formatting (e.g. `w-vd3566` → `VD-3566`), replacing the old `hostname[:3]/worker_id` scheme across backend, frontend, and worker layers.
- Logic is duplicated in three places: `backend/utils.py`, `frontend/src/utils/format.ts`, and `worker/tests/test_worker_name.py` (`_droid_format`). The worker-side helper is only used in tests; if the format ever changes, all three must be updated in sync.
- Edge cases are underspecified: IDs with no letters before the first digit (e.g. `w-1234`) or an empty suffix after `w-` fall through to `bare.upper()` with no hyphen, which may or may not be intentional but is untested.
- The `worker_display_name` signature still accepts `hostname` for backwards compatibility, but nothing enforces that callers actually stop passing it; silent parameter acceptance could mask bugs where callers think hostname still matters.
- Test coverage is good overall; the new parametrized cases in `worker/tests` and the unit tests in `backend/tests` cover the main happy-path variants.

backend/utils.py
```diff
@@ -95,16 +95,27 @@ def generate_guild_id(name: str = "", existing_ids: set[str] | None = None) -> s
...
+def format_worker_id(worker_id: str) -> str:
+    """Format a worker ID in droid style: ``w-vd3566`` → ``VD-3566``.
+
+    Strips the ``w-`` prefix, splits at the first digit boundary, uppercases
+    both parts, and joins with a hyphen.  Examples: ``w-ab1234`` → ``AB-1234``,
+    ``w-x9`` → ``X-9``, ``w-g2otus`` → ``G-2OTUS``.
+    """
+    bare = worker_id.removeprefix("w-")
+    m = re.search(r"\d", bare)
+    if m and m.start() > 0:
```
jmelloy, line 130: The `hostname` parameter is silently ignored. Consider adding a deprecation warning (`warnings.warn`) or removing it entirely to avoid callers believing it still influences the output. Silent no-ops are a common source of subtle bugs.

backend/utils.py
```diff
@@ -95,16 +95,27 @@ def generate_guild_id(name: str = "", existing_ids: set[str] | None = None) -> s
             return unique
 
 
+def format_worker_id(worker_id: str) -> str:
+    """Format a worker ID in droid style: ``w-vd3566`` → ``VD-3566``.
+
```
jmelloy, line 120: Edge case not covered: if `bare` starts with a digit (e.g. worker ID `w-1234`), `m.start() == 0` so the condition `m.start() > 0` is false and the function returns `'1234'` with no hyphen. Similarly `w-` with an empty suffix returns `''`. Add a test and decide the intended behaviour.

frontend/src/utils/format.ts
```diff
@@ -1,5 +1,15 @@
 // Shared time formatters. Keep these tiny and pure — they show up everywhere.
 
+// "w-vd3566" → "VD-3566", "w-x9" → "X-9"
+export function formatWorkerId(workerId: string): string {
+  const bare = workerId.replace(/^w-/, '')
+  const m = bare.match(/\d/)
+  if (m && m.index !== undefined && m.index > 0) {
```
jmelloy, line 26: The formatting logic is duplicated verbatim from the Python backend. Consider fetching and caching the formatted name from the API (which already returns it) rather than re-implementing the transform on the client, so format changes only need to be made in one place.

worker/tests/test_worker_name.py
```diff
@@ -81,21 +82,28 @@ async def test_register_name_falls_back_to_worker_id_when_absent():
 
 
 # ---------------------------------------------------------------------------
-# Name format invariants (black-box, driven by the backend formula)
+# Name format invariants (droid-style: leading-letters + '-' + rest, uppercased)
```
jmelloy, line 85: `_droid_format` is a third copy of the same algorithm. If this helper is needed for test assertions, import `format_worker_id` from `backend/utils` (or extract it to a shared package) instead of re-implementing it, to keep the tests honest about what they're actually verifying.

backend/utils.py
```diff
@@ -95,16 +95,30 @@ def generate_guild_id(name: str = "", existing_ids: set[str] | None = None) -> s
...
+    Strips the ``w-`` prefix, splits at the first digit boundary, uppercases
+    both parts, and joins with a hyphen.  Examples: ``w-ab1234`` → ``AB-1234``,
```
jmelloy, line 122: If `worker_id` has no `w-` prefix (e.g. a bare `abc123`), `removeprefix` is a no-op and the function still works, but the contract is undocumented. Consider asserting or documenting the expected input format to prevent silent misuse.

frontend/src/utils/format.ts
```diff
@@ -1,5 +1,15 @@
...
+    return `${bare.slice(0, m.index).toUpperCase()}-${bare.slice(m.index).toUpperCase()}`
+  }
```
jmelloy, line 28: Using `m.index > 0` means a worker ID whose suffix starts with a digit at position 0 (i.e. bare string is all digits after stripping `w-`) returns the whole string uppercased without a hyphen. This matches the Python behaviour, but an explicit test for that degenerate case (e.g. `w-1234`) would be valuable.

worker/tests/test_worker_name.py
jmelloy, line 85: The `_droid_format` helper is a copy of `backend/utils.format_worker_id`. If the algorithm changes in the backend, this shadow copy will silently diverge. Consider importing the shared utility directly or at least adding a comment noting the duplication risk.

frontend/src/stores/agents.ts
```diff
@@ -212,6 +213,13 @@ export const useAgentsStore = defineStore('agents', () => {
     return pick ? { id: pick.workerId, name: pick.name, state: pick.state } : null
   }
 
+  // Returns the stored display name (e.g. "myhost/VD-3566") for a worker,
+  // falling back to the droid-style ID for workers not yet seen via WebSocket.
+  function workerDisplayName(workerId: string): string {
```
jmelloy, line 218: `worker?.name || formatWorkerId(workerId)` will also fall through to `formatWorkerId` if `worker.name` is an empty string, which should be fine in practice, but using `worker?.name ?? formatWorkerId(workerId)` would be more semantically precise for a nullable field.

jmelloy (commented):
- The PR successfully implements droid-style formatting (e.g. `w-vd3566` → `VD-3566`) consistently across backend Python, frontend TypeScript, and test suites.
- The `format_worker_id` logic is duplicated between `backend/utils.py` and `frontend/src/utils/format.ts`; this is expected given the stack split, but the two implementations should be kept in sync carefully.
- The `worker_display_name` behavior change is significant: previously `hostname[:3]/worker_id` (raw ID), now `hostname/VD-3566` (full hostname + formatted ID). The full hostname in names could be unexpectedly long for machines with verbose hostnames.
- The `_droid_format` helper defined inside `worker/tests/test_worker_name.py` duplicates the backend implementation rather than importing it, creating a maintenance risk if the algorithm changes.
- Tests are thorough and cover edge cases (single letter prefix, digits-only suffix, no hostname, empty hostname); the parametrized tests are well structured.

PR #347, opened 2026-05-18
fix(worker): sweep worktrees before announcing idle to prevent startup hang

jmelloy:
## Diagnosis

The worktree sweeper introduced in cec9df9 ran **after** the worker broadcast its `idle` state, creating a window where:

1. Worker joins guild and sets all slots to `idle` → foreman sees an available worker
2. `_initial_worktree_sweep()` runs — calls `git worktree remove --force` and `git worktree prune` across every known repo
3. Agent loop tasks are created and start processing

During step 2, if the foreman dispatches a task, the listener queues it but the agent loops haven't started yet. The task sits in the queue unprocessed; the worker appears hung.

With lazy-cloning landed in 718d679, repos are no longer pre-cloned at startup. On a worker restart with previous worktrees still on disk, the sweep has more work to do (prune stale entries, `shutil.rmtree` whole checkouts) while the worker is already broadcasting idle — making the hang window larger and more consistently observable.

## Changes

- **Move `_initial_worktree_sweep()` before `_set_state("idle", ...)`**: the foreman cannot dispatch tasks until the worker actually announces idle, so the sweep completes before any task can be assigned.
- **Wrap `shutil.rmtree` in `asyncio.to_thread`**: removing a worktree checkout (potentially thousands of files) is a synchronous operation. Running it in a thread pool avoids blocking the event loop during startup.

## Test plan

- [ ] 163 existing worker tests pass (`python -m pytest`)
- [ ] `ruff check . --fix && ruff format .` — no changes
- [ ] Manual: start a worker that has stale worktrees on disk; verify it completes the sweep before showing idle in the UI, and that tasks assigned immediately after startup are processed without delay

### DIFF
--- a/worker/pioneer_worker/worker.py
+++ b/worker/pioneer_worker/worker.py
@@ -898,13 +898,16 @@ class Worker:
             len(self.slots),
         )
         await self._emit("[worker] Online. Watching for tasks.")
-        for slot in self.slots:
-            await self._set_state("idle", slot)
 
         # Reclaim any worktrees the previous incarnation of this worker left
         # behind. Tasks within the TTL window are re-registered so a follow-up
         # arriving for them can reuse the existing checkout.
+        # Run BEFORE announcing idle so the foreman doesn't dispatch tasks
+        # while git-prune / worktree-remove operations are still in flight.
         await self._initial_worktree_sweep()
+
+        for slot in self.slots:
+            await self._set_state("idle", slot)
 
         initial = await self._fetch_pending_tasks()
         logger.info("Initial pending-task fetch: %d task(s)", len(initial))
@@ -1318,10 +1321,12 @@ class Worker:
                         except Exception as exc:
                             logger.debug("startup-sweep remove_worktree failed: %s", exc)
                 # Whatever git left behind, drop the directory.
+                # Use asyncio.to_thread so rmtree (which walks the whole
+                # checkout) doesn't block the event loop for large repos.
                 import shutil
 
                 try:
-                    shutil.rmtree(full, ignore_errors=True)
+                    await asyncio.to_thread(shutil.rmtree, full, True)
                 except Exception as exc:
                     logger.debug("startup-sweep rmtree failed for %s: %s", full, exc)
             # Prune dangling worktree references in each repo.


worker/pioneer_worker/worker.py
```diff
@@ -898,14 +898,17 @@ async def run(self) -> None:
...
         )
         await self._emit("[worker] Online. Watching for tasks.")
-        for slot in self.slots:
-            await self._set_state("idle", slot)
 
         # Reclaim any worktrees the previous incarnation of this worker left
         # behind. Tasks within the TTL window are re-registered so a follow-up
         # arriving for them can reuse the existing checkout.
+        # Run BEFORE announcing idle so the foreman doesn't dispatch tasks
+        # while git-prune / worktree-remove operations are still in flight.
```
jmelloy, line 906: If `_initial_worktree_sweep()` raises an unhandled exception, no slot will ever be set to `idle` and the worker will stall silently. Consider wrapping this in a `try/except Exception` that logs the error and still proceeds to announce idle slots, so a corrupted worktree state doesn't permanently brick the worker.

worker/pioneer_worker/worker.py
```diff
@@ -1318,10 +1321,12 @@ async def _initial_worktree_sweep(self) -> None:
...
                 # Whatever git left behind, drop the directory.
+                # Use asyncio.to_thread so rmtree (which walks the whole
+                # checkout) doesn't block the event loop for large repos.
                 import shutil
 
                 try:
-                    shutil.rmtree(full, ignore_errors=True)
+                    await asyncio.to_thread(shutil.rmtree, full, True)
                 except Exception as exc:
                     logger.debug("startup-sweep rmtree failed for %s: %s", full, exc)
```
jmelloy, line 1331: `await asyncio.to_thread(shutil.rmtree, full, True)` passes `True` as the positional `ignore_errors` argument, which silently swallows all errors including unexpected ones (e.g. permission errors on dirs the worker doesn't own). The surrounding `try/except` becomes redundant and real problems won't surface in logs; prefer `ignore_errors=False` and let the `except` block handle and log failures.

worker/pioneer_worker/worker.py
```diff
@@ -1318,10 +1321,12 @@ async def _initial_worktree_sweep(self) -> None:
                         except Exception as exc:
                             logger.debug("startup-sweep remove_worktree failed: %s", exc)
                 # Whatever git left behind, drop the directory.
+                # Use asyncio.to_thread so rmtree (which walks the whole
+                # checkout) doesn't block the event loop for large repos.
                 import shutil
 
                 try:
-                    shutil.rmtree(full, ignore_errors=True)
+                    await asyncio.to_thread(shutil.rmtree, full, True)
```
jmelloy, line 1329: `import shutil` inside the loop body is a pre-existing style issue — it should be moved to the top of the module. While Python caches imports after the first load, leaving it here is misleading and will confuse linters.

jmelloy (commented):
- **Core fix is correct**: Moving `_initial_worktree_sweep()` before `_set_state('idle', ...)` eliminates the race window where the foreman could dispatch tasks to a worker still performing blocking git/filesystem operations.
- **`asyncio.to_thread` improvement is sound**: Offloading `shutil.rmtree` to a thread prevents event-loop starvation on large repos during the sweep, which is especially important now that the sweep runs synchronously in the critical path before the worker announces readiness.
- **Potential startup latency concern**: The sweep now blocks the worker from announcing idle until all worktree cleanup completes; on workers with many stale large repos this could cause a noticeable delay before the worker appears available, though this is preferable to the hang.
- **`import shutil` inside the loop body**: The `import shutil` statement remains inside the sweep loop (pre-existing issue, not introduced here), which is a minor style concern but harmless due to Python's import caching.
- **No error handling around the sweep itself**: If `_initial_worktree_sweep()` raises an unhandled exception, the worker will never set any slot to idle and will silently stall; a try/except with logging and a fallback to announcing idle anyway would improve resilience.

PR #348, opened 2026-05-18
fix(worker): prevent worktree sweeper from blocking worker startup

jmelloy:
## Root cause

`_initial_worktree_sweep()` runs **synchronously at startup** (awaited directly before background tasks start). It calls `git worktree prune` and `git worktree remove` via `run_git()`, which uses `asyncio.create_subprocess_exec` + `proc.communicate()` with **no timeout**.

When a worker process crashes mid-task, git can leave a lock file (`.git/worktrees/<name>.lock`). Any subsequent `git worktree prune` on that repo will block indefinitely waiting to acquire the lock — and since the sweep runs synchronously before the WebSocket listener, heartbeat, or task runners start, the entire worker hangs silently on startup.

## Fix

Wrap the `_initial_worktree_sweep()` call with `asyncio.wait_for(timeout=30.0)`. If the sweep takes longer than 30 seconds (indicating a stale git lock), it is cancelled and a warning is logged. The worker comes online normally; the periodic `_worktree_sweeper` task will retry cleanup later.

```python
try:
    await asyncio.wait_for(self._initial_worktree_sweep(), timeout=30.0)
except TimeoutError:
    logger.warning("Worktree startup sweep timed out after 30s — skipping")
```

**Change size:** 6 lines added, 1 removed. No logic changes to the sweep itself.

### DIFF
--- a/worker/pioneer_worker/worker.py
+++ b/worker/pioneer_worker/worker.py
@@ -908,7 +908,7 @@ class Worker:
         # prune/remove hang indefinitely, blocking the entire startup sequence.
         try:
             await asyncio.wait_for(self._initial_worktree_sweep(), timeout=30.0)
-        except TimeoutError:
+        except asyncio.TimeoutError:
             logger.warning("Worktree startup sweep timed out after 30s — skipping")
 
         initial = await self._fetch_pending_tasks()
--- a/worker/pioneer_worker/worker.py
+++ b/worker/pioneer_worker/worker.py
@@ -904,7 +904,12 @@ class Worker:
         # Reclaim any worktrees the previous incarnation of this worker left
         # behind. Tasks within the TTL window are re-registered so a follow-up
         # arriving for them can reuse the existing checkout.
-        await self._initial_worktree_sweep()
+        # Guard with a timeout: a git lock left by a crashed process can make
+        # prune/remove hang indefinitely, blocking the entire startup sequence.
+        try:
+            await asyncio.wait_for(self._initial_worktree_sweep(), timeout=30.0)
+        except TimeoutError:
+            logger.warning("Worktree startup sweep timed out after 30s — skipping")
 
         initial = await self._fetch_pending_tasks()
         logger.info("Initial pending-task fetch: %d task(s)", len(initial))
--- a/worker/pioneer_worker/worker.py
+++ b/worker/pioneer_worker/worker.py
@@ -908,7 +908,7 @@ class Worker:
         # prune/remove hang indefinitely, blocking the entire startup sequence.
         try:
             await asyncio.wait_for(self._initial_worktree_sweep(), timeout=30.0)
-        except asyncio.TimeoutError:
+        except TimeoutError:
             logger.warning("Worktree startup sweep timed out after 30s — skipping")
 
         initial = await self._fetch_pending_tasks()


worker/pioneer_worker/worker.py
```diff
@@ -904,7 +904,12 @@ async def run(self) -> None:
         # Reclaim any worktrees the previous incarnation of this worker left
         # behind. Tasks within the TTL window are re-registered so a follow-up
         # arriving for them can reuse the existing checkout.
-        await self._initial_worktree_sweep()
+        # Guard with a timeout: a git lock left by a crashed process can make
+        # prune/remove hang indefinitely, blocking the entire startup sequence.
```
jmelloy, line 908: Catching bare `TimeoutError` works in Python 3.11+ (where it aliases `asyncio.TimeoutError`), but for older versions `asyncio.wait_for` raises `asyncio.TimeoutError` which is a subclass of `concurrent.futures.TimeoutError`, not the built-in. Use `except asyncio.TimeoutError` to be explicit and safe across all supported Python versions.

worker/pioneer_worker/worker.py
```diff
@@ -904,7 +904,12 @@ async def run(self) -> None:
...
+        try:
```
jmelloy, line 909: The warning message gives no context about what was being swept (e.g., which repo path or worktree). Consider including `self._workspace` or equivalent identifier so operators can quickly locate and manually remove the offending lock file.

worker/pioneer_worker/worker.py
jmelloy, line 908: When `asyncio.wait_for` cancels the inner coroutine, any `asyncio.create_subprocess_exec` child process spawned inside `_initial_worktree_sweep` may not be terminated — the coroutine is cancelled but the OS process lives on. Ensure `_initial_worktree_sweep` (or `run_git`) handles `CancelledError` by killing the subprocess before re-raising.

jmelloy (commented):
- The fix correctly addresses the root cause by wrapping the blocking sweep with `asyncio.wait_for`, preventing indefinite hangs caused by stale git lock files.
- `TimeoutError` should be caught as `asyncio.TimeoutError` for clarity and compatibility across Python versions (pre-3.11 `asyncio.wait_for` raises `asyncio.TimeoutError`, not the built-in `TimeoutError`).
- The fix is minimal and low-risk; the periodic `_worktree_sweeper` will eventually retry cleanup, so skipping the initial sweep on timeout is a reasonable degraded-mode behavior.
- No attempt is made to log *which* repository or worktree caused the hang, which could make diagnosing recurring issues harder in production.
- Consider also handling `asyncio.CancelledError` explicitly or ensuring the cancelled coroutine properly cleans up any subprocess it spawned, since cancelling `communicate()` mid-flight may leave orphaned child processes.

jmelloy (commented):
- The fix correctly addresses a real startup-blocking bug by wrapping the sweep with `asyncio.wait_for`, preventing indefinite hangs caused by stale git lock files.
- The timeout value of 30 seconds is a magic number with no named constant or configuration option, making it harder to tune without a code change.
- The fix silently swallows the timeout without any attempt to clean up the stale lock file or surface actionable guidance to the operator (e.g., which repo/path is affected).
- The periodic `_worktree_sweeper` will retry, but if the lock condition persists it will also block or silently fail on every cycle — the underlying issue is never resolved automatically.
- No test coverage is added for the timeout path, leaving the regression risk unverified.

worker/pioneer_worker/worker.py
jmelloy, line 909: 30.0 is a bare magic number. Consider extracting it as a module-level constant (e.g. `_INITIAL_SWEEP_TIMEOUT_S = 30.0`) or making it a constructor parameter so operators can tune it without a code change.

worker/pioneer_worker/worker.py
```diff
@@ -904,7 +904,12 @@ async def run(self) -> None:
...
+            await asyncio.wait_for(self._initial_worktree_sweep(), timeout=30.0)
```
jmelloy, line 910: The warning gives no actionable context. At minimum, log the repo path(s) being swept so an operator knows where to look for the stale lock file. Also consider logging `logger.warning(..., exc_info=True)` or at least noting that manual lock removal (`rm .git/worktrees/<name>.lock`) may be required to recover fully.

jmelloy, line 910: When the timeout fires, `_initial_worktree_sweep` is cancelled mid-execution. If the sweep had already removed some worktrees from its internal tracking but not yet persisted state, those worktrees could be leaked permanently since the periodic sweeper may not rediscover them. Verify that `_initial_worktree_sweep` is safe to cancel at any await point without leaving internal data structures inconsistent.

PR #349, opened 2026-05-18
Review agent: resolve/supersede previous reviews on re-run

jmelloy:
## Summary

When the code-review-agent runs a second (or subsequent) time on the same PR, it now cleans up the previous run's artifacts before posting the new review:

- **Resolves prior inline threads**: fetches all review threads on the PR via GitHub GraphQL and calls `resolveReviewThread` on any unresolved thread that belongs to a prior bot review.
- **Supersedes prior review bodies**: PATCHes each prior bot review to prepend `~~Superseded by review posted <date>~~` so the timeline makes it clear which review is current.
- **Posts the new review** as normal.

The logic is **idempotent**: if there are no prior bot reviews, nothing changes. Already-superseded reviews are skipped. All cleanup steps are individually non-fatal — failures are logged as warnings so the main review post is never blocked.

### New functions added to `backend/foreman/tools.py`

| Function | Purpose |
|---|---|
| `_gh_graphql(token, query, variables)` | Generic GitHub GraphQL helper |
| `_supersede_prior_bot_reviews(pr_repo, pr_number, bot_username, token)` | Orchestrates resolve + supersede |

Both `review_pr` (A2A path) and `review_pr_internal` call `_supersede_prior_bot_reviews` before posting.

## Test plan

- [x] `python3 -m pytest tests/test_review_pr_tool.py -v` — all 19 tests pass (9 new tests added for the supersede logic)
- [x] `ruff check . --fix && ruff format .` — no lint errors
- [ ] Manual: re-run the review agent on a PR with a prior review and verify threads are resolved and old review shows "Superseded" prefix

Closes #5

### DIFF
--- a/backend/foreman/tools.py
+++ b/backend/foreman/tools.py
@@ -601,6 +601,23 @@ def _gh_api_diff(path: str, token: str) -> str:
         return resp.read().decode("utf-8", errors="replace")
 
 
+def _gh_graphql(token: str, query: str, variables: dict) -> dict:
+    """Execute a GitHub GraphQL query/mutation and return the parsed response."""
+    payload = json.dumps({"query": query, "variables": variables}).encode()
+    req = urllib.request.Request(
+        "https://api.github.com/graphql",
+        data=payload,
+        headers={
+            "Authorization": f"Bearer {token}",
+            "Content-Type": "application/json",
+            "Accept": "application/vnd.github.v3+json",
+        },
+        method="POST",
+    )
+    with urllib.request.urlopen(req, timeout=15) as resp:
+        return json.loads(resp.read())
+
+
 def _parse_review_from_claude(text: str) -> dict:
     """Extract a review JSON object from Claude's response.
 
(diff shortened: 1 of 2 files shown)


backend/foreman/tools.py
```diff
@@ -800,6 +817,133 @@ def _extract_review_data(a2a_result: dict) -> tuple[str, str]:
     return github_event, review_body
 
 
+_GQL_PR_THREADS = """
```
jmelloy, line 820: Only the first 100 review threads and first 10 comments per thread are fetched. A PR with more than 100 threads, or a thread whose bot comment is beyond the 10th, will be silently missed. Add cursor-based pagination or at minimum increase the limits and add a TODO comment acknowledging the cap.

jmelloy (commented):
- The feature correctly implements a supersede-and-resolve flow, with idempotency guards and non-fatal error handling for all cleanup steps — overall design is sound.
- **Pagination bug**: `_GQL_PR_THREADS` fetches only the first 100 threads and `comments(first: 10)` fetches only the first 10 comments; PRs with more threads or multi-comment threads will silently miss items and leave old threads unresolved.
- **REST vs GraphQL ID mismatch**: the code compares `databaseId` from GraphQL comments against `r["id"]` from the REST reviews API. The REST `id` field *is* the database ID, so this works, but it is fragile and undocumented — a comment in the code and a guard against missing `databaseId` values would prevent subtle future breakage.
- The `_gh_graphql` helper does not check for a `"errors"` key in the GraphQL response body; GitHub returns HTTP 200 with an `errors` array on partial/total failures, so error conditions are silently swallowed.
- Test coverage is solid and well-structured; the integration tests correctly verify both the happy path and the non-fatal failure path.

backend/foreman/tools.py
```diff
@@ -601,6 +601,23 @@ def _gh_api_diff(path: str, token: str) -> str:
...
+        "https://api.github.com/graphql",
+        data=payload,
+        headers={
+            "Authorization": f"Bearer {token}",
+            "Content-Type": "application/json",
+            "Accept": "application/vnd.github.v3+json",
+        },
+        method="POST",
+    )
+    with urllib.request.urlopen(req, timeout=15) as resp:
```
jmelloy, line 617: `_gh_graphql` does not inspect the `"errors"` key in the response. GitHub GraphQL returns HTTP 200 even for errors (e.g. resource not accessible, node not found), placing details in `response["errors"]`. Callers receive a partial/empty `data` dict and silently no-op instead of logging a useful warning. Add a check like `if 'errors' in result: raise RuntimeError(result['errors'])` (or at minimum log a warning).

backend/foreman/tools.py
```diff
@@ -800,6 +817,133 @@ def _extract_review_data(a2a_result: dict) -> tuple[str, str]:
...
+    bot_reviews = [r for r in reviews if (r.get("user") or {}).get("login") == bot_username]
+    if not bot_reviews:
+        return 0
+
+    bot_review_ids = {r["id"] for r in bot_reviews}
+
+    # Resolve unresolved inline threads that belong to our prior reviews.
+    owner, repo_name = pr_repo.split("/", 1)
+    try:
+        gql_result = await asyncio.to_thread(
```
jmelloy, line 878: The comparison `(c.get('pullRequestReview') or {}).get('databaseId') in bot_review_ids` relies on the implicit equivalence between GraphQL `databaseId` (int) and the REST API `review['id']` (int). This is correct today but should be documented with a brief comment, since a `databaseId` value of `None` (e.g. for a minimized comment) would silently skip the check rather than raising an error.

PR #350, opened 2026-05-18
feat(foreman): downgrade to Haiku + Phase 2 standalone process

jmelloy:
## Summary

- **Haiku downgrade**: Changes the Foreman AI model from `claude-sonnet-*` to `claude-haiku-4-5-20251001` in `backend/foreman/runner.py` and the standalone `foreman/main.py`, and updates the `FOREMAN_MODEL` env var default
- **Phase 2 migration**: Implements the standalone foreman process in `foreman/main.py` — it connects to the backend WebSocket with `agentType="foreman"` and `external=true`, the backend routes `foreman-trigger` events to it, and falls back to the embedded foreman if the external one disconnects
- Adds `foreman/Dockerfile` and the `foreman` service to `docker-compose.yml` (opt-in via `--profile foreman`)
- Updates `CLAUDE.md` and `README.md` with standalone foreman docs and env var guidance
- Adds `.env.example` entries for `FOREMAN_MODEL`, `BACKEND_WS_URL`, and `GUILD_ID`

## Test plan

- [ ] Run backend tests: `cd backend && python -m pytest`
- [ ] Verify embedded foreman still works (no `--profile foreman`) with a chat message to foreman
- [ ] Verify standalone foreman connects: `BACKEND_WS_URL=ws://localhost:8000 GUILD_ID=<id> ANTHROPIC_API_KEY=<key> python foreman/main.py`
- [ ] Confirm model in logs shows `claude-haiku-4-5-20251001`
- [ ] Docker: `GUILD_ID=abc123 docker compose --profile foreman up --build foreman`

### DIFF
--- a/backend/events.py
+++ b/backend/events.py
@@ -43,6 +43,7 @@ foreman_connections: dict[str, WebSocket] = {}
 # connections dict.  The standalone foreman/main.py sets this before importing
 # the runner so that broadcasts are relayed to the backend via WebSocket.
 _broadcast_override = None
+_broadcast_override_lock = asyncio.Lock()
 
 
 def agent_owner_lock(guild_id: str) -> asyncio.Lock:
@@ -54,10 +55,19 @@ def agent_owner_lock(guild_id: str) -> asyncio.Lock:
     return lock
 
 
+async def set_broadcast_override(fn) -> None:
+    """Set the broadcast override callable under the override lock."""
+    global _broadcast_override
+    async with _broadcast_override_lock:
+        _broadcast_override = fn
+
+
 async def broadcast(guild_id: str, message: dict, exclude: WebSocket | None = None):
     """Broadcast a message to all connections in a guild."""
-    if _broadcast_override is not None:
-        await _broadcast_override(guild_id, message, exclude)
+    async with _broadcast_override_lock:
+        override = _broadcast_override
+    if override is not None:
+        await override(guild_id, message, exclude)
         return
     if guild_id not in connections:
         return
--- a/backend/events.py
+++ b/backend/events.py
@@ -37,6 +37,13 @@ _agent_owner_locks: dict[str, asyncio.Lock] = {}
 # WS message or fall back to the embedded run_foreman_ai().
 foreman_connections: dict[str, WebSocket] = {}
 
+# Optional broadcast override for the standalone foreman process.
+# When set (to an async callable with the same signature as broadcast), all
+# broadcast() calls are routed through this function instead of the in-process
+# connections dict.  The standalone foreman/main.py sets this before importing
+# the runner so that broadcasts are relayed to the backend via WebSocket.
+_broadcast_override = None
+
 
 def agent_owner_lock(guild_id: str) -> asyncio.Lock:
     """Return the per-guild lock used to serialise ``agent_owners`` writes."""
@@ -49,6 +56,9 @@ def agent_owner_lock(guild_id: str) -> asyncio.Lock:
 
 async def broadcast(guild_id: str, message: dict, exclude: WebSocket | None = None):
     """Broadcast a message to all connections in a guild."""
+    if _broadcast_override is not None:
+        await _broadcast_override(guild_id, message, exclude)
+        return
     if guild_id not in connections:
         return
     dead = []
--- a/foreman/main.py
+++ b/foreman/main.py
@@ -30,6 +30,8 @@ import os
 import sys
 from pathlib import Path
 
+import websockets
+
 # ── Python path ──────────────────────────────────────────────────────────────
 # The foreman entry point lives at <repo>/foreman/main.py while all backend
 # modules (database, models, events, foreman/runner, …) live under
@@ -46,36 +48,12 @@ sys.path.insert(0, str(_BACKEND_DIR))
 # import of runner or tools.
 import events as _events  # noqa: E402  (must precede runner import)
 
-_relay_ws: websockets.WebSocketClientProtocol | None = None
-
-logger = logging.getLogger("pioneer_foreman")
-
-
-async def _relay_broadcast(guild_id: str, message: dict, exclude: object = None) -> None:
-    """Relay a broadcast message to the backend for fan-out to frontend clients."""
-    ws = _relay_ws
-    if ws is None:
-        return
-    try:
-        await ws.send(
-            json.dumps(
-                {
-                    "type": "foreman-broadcast",
-                    "guildId": guild_id,
-                    "payload": message,
-                }
-            )
-        )
-    except Exception as exc:
-        logger.warning("relay_broadcast failed: %s", exc)
-
-
-_events._broadcast_override = _relay_broadcast
-
-# Safe to import the runner now; every broadcast() call will hit _relay_broadcast.
-import websockets  # noqa: E402
+# Safe to import the runner now; the broadcast override will be installed
+# per-connection inside _run_guild via _events.set_broadcast_override().
 from foreman.runner import run_foreman_ai  # noqa: E402
 from util.tasks import spawn  # noqa: E402
+
+logger = logging.getLogger("pioneer_foreman")
 
 # ── Trigger handling ──────────────────────────────────────────────────────────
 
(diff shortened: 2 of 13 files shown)


foreman/main.py
```diff
@@ -0,0 +1,190 @@
...
+# binding the *function object*.  We install _broadcast_override on the events
+# module so that every call to broadcast() — regardless of where it was imported
+# — goes through our relay path.  The override must be set before the first
+# import of runner or tools.
+import events as _events  # noqa: E402  (must precede runner import)
+
+_relay_ws: websockets.WebSocketClientProtocol | None = None
+
+logger = logging.getLogger("pioneer_foreman")
+
```
jmelloy: `_relay_ws` is referenced before `websockets` is imported (line 51 uses the type annotation `websockets.WebSocketClientProtocol` in the variable declaration comment, and `websockets` is only imported on line 83). More critically, `_relay_broadcast` references `_relay_ws` as a global and is installed as `_events._broadcast_override` before `websockets` is imported, so any early broadcast call is safe, but the type annotation will cause a `NameError` at module load time if Python evaluates it eagerly. Use `from __future__ import annotations` (already present) which defers evaluation — this is fine, but the ordering is fragile and should be documented or restructured.

foreman/main.py
```diff
@@ -0,0 +1,190 @@
...
+        name=f"foreman.trigger:{guild_id}:{data.get('event', '?')}",
+    )
+
+
+# ── WebSocket connection loop ─────────────────────────────────────────────────
+
+
+async def _run_guild(backend_ws_url: str, guild_id: str) -> None:
+    """Connect to the backend WS for *guild_id* and handle triggers until evicted."""
+    global _relay_ws
```
jmelloy: `_relay_ws` is a module-level global mutated in `_run_guild` and read in `_relay_broadcast`. If `_run_guild` is ever called concurrently for multiple guilds (or during a reconnect race), the last writer wins and broadcasts from a previous run could be silently dropped or sent to the wrong connection. Consider encapsulating `_relay_ws` in a per-guild object or passing it explicitly to avoid this shared-state hazard.

backend/events.py
```diff
@@ -49,6 +56,9 @@ def agent_owner_lock(guild_id: str) -> asyncio.Lock:
 
 async def broadcast(guild_id: str, message: dict, exclude: WebSocket | None = None):
     """Broadcast a message to all connections in a guild."""
+    if _broadcast_override is not None:
```
jmelloy: `_broadcast_override` is a plain module-level `None` that any code can overwrite at any time with no guard. This is particularly risky because `foreman/main.py` sets it at import time as a side effect, meaning importing `foreman/main.py` in tests or other contexts will globally redirect all broadcasts. Consider using a registration function (e.g., `set_broadcast_override(fn)`) with appropriate guards to make the contract explicit and testable.

foreman/main.py
```diff
@@ -0,0 +1,190 @@
...
+                # All other message types (broadcasts the backend sends to this
+                # connection, etc.) are intentionally ignored here.
+        finally:
+            _relay_ws = None
+
+
+# ── Entrypoint ────────────────────────────────────────────────────────────────
+
+
+async def main() -> None:
```
jmelloy, line 160: On `foreman-evicted` the code logs 'reconnecting' and breaks out of the inner loop, but the outer `while True` will immediately reconnect with no delay (`retry_delay` is reset to 5 only after a clean exit). If the backend repeatedly evicts this foreman (e.g., another instance keeps connecting), this could create a rapid reconnect storm. Consider adding a deliberate short sleep before reconnecting after an eviction.

jmelloy (commented):
- The model downgrade from `claude-sonnet-4-6` to `claude-haiku-4-5-20251001` is applied consistently across `runner.py` and `tools.py`, but there is no environment variable override for the model name, making future model changes require code edits again.
- The `_broadcast_override` global in `events.py` is a mutable module-level variable with no thread/coroutine safety guard; concurrent imports or reloads could leave it in an inconsistent state.
- `_relay_ws` is a global variable shared across coroutines in `foreman/main.py`; if multiple guilds or reconnect attempts run concurrently, races on this global could cause broadcasts to be sent to the wrong (or a closed) WebSocket connection.
- The `foreman-evicted` handler logs 'reconnecting' but actually just breaks out of the inner loop, allowing the outer `while True` to reconnect — the log message is misleading but behavior is functionally correct.
- The Dockerfile installs `websockets` as a transitive dependency via `backend/requirements.txt`, which is fine, but there is no explicit version pin or check, so a future upgrade could break the `websockets.connect` async context manager API used in `main.py`.

foreman/main.py
```diff
@@ -0,0 +1,190 @@
...
+_BACKEND_DIR = Path(__file__).parent.parent / "backend"
+sys.path.insert(0, str(_BACKEND_DIR))
+
+# ── Patch broadcast BEFORE importing runner / tools ──────────────────────────
+# runner.py and tools.py do `from events import broadcast` at module load time,
+# binding the *function object*.  We install _broadcast_override on the events
+# module so that every call to broadcast() — regardless of where it was imported
+# — goes through our relay path.  The override must be set before the first
+# import of runner or tools.
+import events as _events  # noqa: E402  (must precede runner import)
```
jmelloy, line 49: `websockets` is imported after `_events._broadcast_override = _relay_broadcast` but `_relay_broadcast` references `websockets.WebSocketClientProtocol` in its type annotation on line 57. The `websockets` import on line 79 comes after `_relay_broadcast` is defined, so the forward reference in the type hint is a `NameError` at module load time if annotations are evaluated eagerly (Python <3.10 without `from __future__ import annotations`). The `from __future__ import annotations` at the top defers evaluation, so it works — but the import ordering is still confusing and should be reorganized.

foreman/main.py
```diff
@@ -0,0 +1,190 @@
...
+                    )
+                    break
+                # All other message types (broadcasts the backend sends to this
+                # connection, etc.) are intentionally ignored here.
+        finally:
+            _relay_ws = None
+
+
+# ── Entrypoint ────────────────────────────────────────────────────────────────
+
```
jmelloy, line 158: After a `foreman-evicted` message the loop breaks and `retry_delay` is reset to 5, then `_run_guild` is called again immediately — which will likely result in being evicted again immediately if another external foreman is still connected. Consider adding a longer fixed delay or an exponential backoff specifically for eviction events to avoid a tight reconnect storm.

foreman/main.py
```diff
@@ -0,0 +1,190 @@
...
+_relay_ws: websockets.WebSocketClientProtocol | None = None
+
+logger = logging.getLogger("pioneer_foreman")
+
+
+async def _relay_broadcast(guild_id: str, message: dict, exclude: object = None) -> None:
+    """Relay a broadcast message to the backend for fan-out to frontend clients."""
+    ws = _relay_ws
+    if ws is None:
+        return
```
jmelloy: `_relay_ws` is a module-level global written in `_run_guild` and read concurrently in `_relay_broadcast` from spawned tasks. While Python's GIL prevents torn writes on simple assignment, setting `_relay_ws = None` in the `finally` block while a spawned `run_foreman_ai` coroutine is still mid-flight will silently drop all broadcasts from that run. Use an `asyncio.Lock` or pass the websocket reference into the trigger handler closure instead.

backend/events.py
jmelloy: The `_broadcast_override` check skips guild membership validation entirely — a misbehaving or compromised external foreman could relay broadcast messages to any `guild_id`, not just its own. The override should be scoped per-guild or validated against the foreman's registered guild before passing through.

jmelloy (commented):
- The model downgrade from `claude-sonnet-4-6` to `claude-haiku-4-5-20251001` is a straightforward change but the model identifier format is inconsistent with Anthropic's typical versioning (`claude-haiku-3-5-20240620` style); verify this is a valid model ID before deploying.
- The `_broadcast_override` approach in `events.py` is a global mutable that creates a subtle ordering dependency: if anything imports `events.broadcast` before `foreman/main.py` patches `_events._broadcast_override`, those importers will still hold a reference to the original unpatched function — but since it's a module-level attribute checked inside `broadcast()`, the patch actually works correctly; however it's fragile and bypasses the normal broadcast connection tracking entirely.
- `_relay_ws` is a bare global in `foreman/main.py` with no thread/task safety; concurrent trigger handlers calling `_relay_broadcast` while `_run_guild` resets it to `None` could cause a race where a trigger is silently dropped mid-flight.
- The `foreman-evicted` handler logs 'reconnecting' and `break`s, but the outer retry loop will immediately reconnect with no delay, potentially creating a tight reconnect loop if eviction is permanent (e.g., another external foreman is permanently connected).
- The Dockerfile changes the `WORKDIR` to `/app/foreman` after copying `main.py` there, but `main.py` computes `_BACKEND_DIR` using `Path(__file__).parent.parent / 'backend'` which resolves to `/app/backend` — this is correct only because of the specific directory layout; a small documentation note or assertion would make this assumption explicit.

PR #354, opened 2026-05-19
feat: steampunk workshop background, perspective tables, dev login

jmelloy:
## Summary

- **Workshop room background** — replaces CSS-drawn pipes/gears/furnace with a pixel-art steampunk room image (`workshop-bg.jpg`); removes ~600 lines of CSS decoration
- **Pixel-art table sprite** — each work station renders a transparent-background table PNG (`table.png`), perspective-scaled 0.66→1.04 from back to front along the side walls
- **Trapezoid floor placement** — `floorBounds()` maps the background image's known floor trapezoid geometry (floor starts at y=400 in a 1360×768 image, 525px→700px wide) to panel coordinates so tables always sit on the visible stone floor at any panel size
- **Dev/test login** — `POST /auth/guest` (gated by `TEST_MODE=1`) creates a stable dev-guest user + guild without GitHub OAuth; "SKIP LOGIN" button on the landing page visible only in Vite dev mode
- **AppView padding** — horizontal padding at ≥1025/1400/1800px breakpoints stops the factory floor stretching too wide on large screens

## Test plan

- [ ] `TEST_MODE=1 docker compose up -d backend` then open the landing page in dev mode — "DEV MODE / SKIP LOGIN" button should appear and drop you straight into the Dev Workshop guild
- [ ] Resize the browser window; tables should stay on the stone floor and scale correctly
- [ ] Normal GitHub login still works when `TEST_MODE` is unset
- [ ] `POST /auth/guest` returns 403 when `TEST_MODE` is not set


backend/routes/auth.py
```diff
@@ -235,6 +239,106 @@ async def api_me(github_user_id: str = Depends(require_user)):
         await db.close()
 
 
+@router.post("/auth/guest")
+async def guest_login():
+    """Create a dev session without GitHub OAuth.
```
jmelloy, line 244: Using `os.environ.get("TEST_MODE")` is truthy for any non-empty string, so `TEST_MODE=0` or `TEST_MODE=false` would still enable this endpoint. Use an explicit check: `if os.environ.get("TEST_MODE") != "1":` to match the documented contract.

backend/routes/auth.py
```diff
@@ -235,6 +239,106 @@ async def api_me(github_user_id: str = Depends(require_user)):
...
+    Only available when TEST_MODE=1 is set. Idempotent: always creates a new
+    session token but reuses the same dev-guest user and guild across calls.
+    """
+    if not os.environ.get("TEST_MODE"):
+        raise HTTPException(status_code=403, detail="Test mode not enabled (set TEST_MODE=1)")
+
+    now = datetime.now(UTC).isoformat()
+    guest_user_id = "dev-guest"
+    login_token = secrets.token_urlsafe(32)
+
```
jmelloy, line 255: The DB session is opened manually rather than via the standard dependency-injection pattern. If `db.commit()` raises, the `finally` block calls `db.close()` but the partial writes (session token, user upsert) may already be flushed. Use `async with get_db() as db:` to get transactional rollback on error.

frontend/src/components/FactoryFloor.vue
```diff
@@ -138,47 +75,99 @@ import { useAgentsStore } from '../stores/agents'
...
+// Return the left/right x bounds (panel px) of the floor trapezoid at a given panel y.
+function floorBounds(panelY: number, panelW: number, panelH: number) {
+  const scale = Math.max(panelW / BG_W, panelH / BG_H)
+  const xOff  = (BG_W * scale - panelW) / 2
+  const imgY  = panelY / scale
+  const t     = Math.max(0, Math.min(1, (imgY - TRAP_TOP_Y) / TRAP_RANGE))
+  const half  = TRAP_TOP_HALF + t * (TRAP_BOT_HALF - TRAP_TOP_HALF)
+  return {
+    left:  Math.max(0,      Math.round((TRAP_MID_X - half) * scale - xOff)),
+    right: Math.min(panelW, Math.round((TRAP_MID_X + half) * scale - xOff)),
```
jmelloy, line 113: `floorBounds` implicitly assumes CSS `background-size: cover` and `background-position: center top`. The scale formula (`Math.max(panelW/BG_W, panelH/BG_H)`) matches `cover`, but the x-offset assumes horizontal centering — if the CSS background-position is ever changed the table placement will silently drift with no visible error.

frontend/src/components/FactoryFloor.vue
```diff
@@ -138,47 +75,99 @@ import { useAgentsStore } from '../stores/agents'
...
+      { x: bb.left,               y: yBack,  scale: 0.68 },
+      { x: bb.right - tw(0.68),   y: yBack,  scale: 0.68 },
+      { x: fb.left,               y: yFront, scale: 0.92 },
+      { x: fb.right - tw(0.92),   y: yFront, scale: 0.92 },
     ]
   }
+
+  const bb = floorBounds(yBack,  w, h)
+  const mb = floorBounds(yMid,   w, h)
+  const fb = floorBounds(yFront, w, h)
```
jmelloy, line 143: The wide-screen branch of `stationPositions` always returns 6 entries, but `MAX_STATIONS` is now a computed that returns 4 on narrow screens. Since `stationPositions` uses `containerWidth` directly for branching this happens to stay consistent, but the two are no longer coupled — consider deriving the array length from `MAX_STATIONS.value` to make the invariant explicit.

frontend/src/stores/auth.ts
```diff
@@ -69,6 +69,25 @@ export const useAuthStore = defineStore('auth', () => {
...
+      const body = await res.json().catch(() => ({}))
+      throw new Error(body.detail || `Dev login failed: ${res.status}`)
+    }
+    const data = await res.json()
+    loginToken.value = data.login_token
+    user.value = {
+      id: data.gh_user_id,
+      login: data.gh_login,
+      name: data.gh_name || data.gh_login,
+      avatar_url: data.gh_avatar || '',
```
jmelloy, line 84: `data.guild_id as string` is an unchecked cast — if the backend ever omits the field the router will navigate to `undefined` silently. Add a runtime check (e.g. `if (!data.guild_id) throw new Error('missing guild_id')`) before returning.

jmelloy (commented):
- **Security**: The `TEST_MODE` gate relies on a truthy env-var check (`os.environ.get("TEST_MODE")`), which means any non-empty value (e.g. `TEST_MODE=0`) enables the endpoint — should use an explicit equality check against `"1"` or `"true"`.
- **Resource leak**: `guest_login` manually calls `get_db()` and `db.close()` in a `finally` block instead of using the standard `async with get_db() as db:` dependency pattern used elsewhere; if `db.commit()` raises an exception the session/token may already be partially written.
- **Hardcoded magic numbers in `floorBounds`**: The trapezoid constants (`BG_W`, `BG_H`, `TRAP_TOP_Y`, etc.) are file-level `const` declarations with no units documentation, and `floorBounds` silently assumes `background-size: cover` scaling — if the CSS `background-position` ever changes, tables will misalign with no obvious breakage signal.
- **`MAX_STATIONS` reactive conversion is partial**: `MAX_STATIONS` was changed from a plain constant to a `computed`, and `activeTasks` correctly references `.value`, but `stationPositions` still hard-codes 6 slots in the wide-screen branch rather than reading `MAX_STATIONS.value`, so the two can diverge.
- **`loginAsGuest` persists token in `localStorage` without the same expiry/refresh handling used by the OAuth path**, and the returned `guild_id` is cast with `as string` rather than validated — a missing field would silently navigate to `undefined`.

jmelloy:
Splitting into two focused PRs:
- #378 — feat: dev/test guest login (TEST\_MODE gate)
- #379 — feat: steampunk workshop background, perspective tables

PR #375, opened 2026-05-20
Foreman: emit tool events inline as they happen

### LINKED ISSUE #374: Foreman: stream tool events inline as they happen instead of batching text at the end
## Problem

The Foreman currently batches all its commentary and emits it as a single text block after all tool calls have completed. This means the user sees tool use + results interleaved with no narrative, followed by a wall of text at the end.

## Desired behaviour

The Foreman should emit commentary **as it goes** — a short sentence before each tool call explaining what it's about to do, and a short sentence after each result summarising what it found. No big concatenated paragraph at the end of the turn.

### Example (current)

```
[tool: get_pr_status] → result
[tool: finalize_task] → result

All 4 checks passed ✅. PR #373 is open and mergeable — parking it in awaiting-review for human merge. No follow-up needed from my side.
```

### Example (desired)

```
All 4 checks passed — PR is mergeable, finalizing now.
[tool: finalize_task] → result
Done! Work is closed. ✅
```

## Implementation

This is a system-prompt / LLM behaviour change. The Foreman's instructions should be updated to direct it to:

1. Write a brief sentence **before** each tool call describing the intent.
2. Write a brief sentence **after** each tool result summarising the outcome.
3. **Not** accumulate a summary paragraph at the end of the turn.

jmelloy:
## Summary

- Adds a new **Inline narration** section to the Foreman system prompt (`backend/foreman/prompt.py`)
- Instructs the Foreman to write one short sentence **before** each tool call stating what it is about to do
- Instructs the Foreman to write one short sentence **after** each tool result summarising what it found or what happened
- Explicitly prohibits accumulating a summary paragraph at the end of a turn — all commentary must appear in-flow with the tool calls

## Test plan

- [ ] Trigger the Foreman via a chat message and verify it narrates before/after each tool call in the WS stream
- [ ] Confirm no trailing summary paragraph appears at end of a Foreman turn
- [ ] Run backend tests: `cd backend && python -m pytest`

Closes #374

### DIFF
--- a/backend/foreman/prompt.py
+++ b/backend/foreman/prompt.py
@@ -98,7 +98,8 @@ Narrate in-flow with your tool calls — never accumulate a summary paragraph at
 - **If a tool returned an error**: state that briefly and what you will do next (e.g. "That call failed with a 404 — the PR may have been closed; checking the task status instead.").
 Keep each sentence tight — action verb, subject, relevant detail. No filler phrases.
 
-Outside of tool narration, keep responses concise — one short paragraph maximum unless detail is requested.\
+Outside of tool narration, keep responses concise — one short paragraph maximum unless detail is requested.
+Do not add a concluding paragraph after the last tool result.\
 """
 
 
--- a/backend/foreman/prompt.py
+++ b/backend/foreman/prompt.py
@@ -92,14 +92,13 @@ The state reflects the moment this turn was sent — earlier turns saw earlier s
 Workers are configured with repos. Prefer workers whose repos cover the task.
 
 ## Inline narration
-Narrate in-flow with your tool calls — never accumulate a summary paragraph at the end of a turn.
-- **Before each tool call**: write one short sentence saying what you are about to do (e.g. "Checking the PR status for task <task-id>.").
+Narrate in-flow with your tool calls — never add a summary or concluding paragraph at the end of a turn.
+- **Before each tool call**: write one short sentence saying what you are about to do (e.g. "Checking the PR status for task t-abc123.").
 - **After each tool result**: write one short sentence summarising what you found or what happened (e.g. "CI is passing and one reviewer has approved.").
 - **If a tool returned an error**: state that briefly and what you will do next (e.g. "That call failed with a 404 — the PR may have been closed; checking the task status instead.").
 Keep each sentence tight — action verb, subject, relevant detail. No filler phrases.
 
 Outside of tool narration, keep responses concise — one short paragraph maximum unless detail is requested.
-Do not add a concluding paragraph after the last tool result.\
 """
 
 
--- a/backend/foreman/prompt.py
+++ b/backend/foreman/prompt.py
@@ -93,7 +93,7 @@ Workers are configured with repos. Prefer workers whose repos cover the task.
 
 ## Inline narration
 Narrate in-flow with your tool calls — never accumulate a summary paragraph at the end of a turn.
-- **Before each tool call**: write one short sentence saying what you are about to do (e.g. "Checking the PR status for task t-abc123.").
+- **Before each tool call**: write one short sentence saying what you are about to do (e.g. "Checking the PR status for task <task-id>.").
 - **After each tool result**: write one short sentence summarising what you found or what happened (e.g. "CI is passing and one reviewer has approved.").
 - **If a tool returned an error**: state that briefly and what you will do next (e.g. "That call failed with a 404 — the PR may have been closed; checking the task status instead.").
 Keep each sentence tight — action verb, subject, relevant detail. No filler phrases.
--- a/backend/foreman/prompt.py
+++ b/backend/foreman/prompt.py
@@ -90,7 +90,12 @@ and recent tasks. Treat it as an operational briefing, not part of the human's m
 The state reflects the moment this turn was sent — earlier turns saw earlier state.
 
 Workers are configured with repos. Prefer workers whose repos cover the task.
-Be concise — one short paragraph maximum unless detail is requested.\
+
+## Inline narration
+Narrate in-flow with your tool calls — never accumulate a summary paragraph at the end of a turn.
+- **Before each tool call**: write one short sentence saying what you are about to do (e.g. "Checking the PR status for task t-abc123.").
+- **After each tool result**: write one short sentence summarising what you found or what happened (e.g. "CI is passing and one reviewer has approved.").
+Keep each sentence tight — action verb, subject, relevant detail. No filler phrases.\
 """
 
 
--- a/backend/foreman/prompt.py
+++ b/backend/foreman/prompt.py
@@ -95,7 +95,10 @@ Workers are configured with repos. Prefer workers whose repos cover the task.
 Narrate in-flow with your tool calls — never accumulate a summary paragraph at the end of a turn.
 - **Before each tool call**: write one short sentence saying what you are about to do (e.g. "Checking the PR status for task t-abc123.").
 - **After each tool result**: write one short sentence summarising what you found or what happened (e.g. "CI is passing and one reviewer has approved.").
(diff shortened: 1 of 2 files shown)


jmelloy (commented):
- The PR replaces a conciseness guideline with an inline narration section, which is a reasonable UX improvement for streaming tool-call visibility.
- The original `Be concise` instruction is removed entirely with no equivalent replacement; the new instructions only address narration style, leaving no general verbosity guidance for non-tool-call responses.
- The diff is minimal and low-risk — it is purely a prompt/string change with no logic, API, or schema modifications.
- The backslash line-continuation at the end of the new block is preserved from the original, which is correct and consistent.
- No test coverage exists for prompt content (tests only cover backend logic), so the stated test plan relies entirely on manual verification.

backend/foreman/prompt.py
```diff
@@ -90,7 +90,12 @@
 The state reflects the moment this turn was sent — earlier turns saw earlier state.
 
 Workers are configured with repos. Prefer workers whose repos cover the task.
-Be concise — one short paragraph maximum unless detail is requested.\
+
+## Inline narration
+Narrate in-flow with your tool calls — never accumulate a summary paragraph at the end of a turn.
+- **Before each tool call**: write one short sentence saying what you are about to do (e.g. "Checking the PR status for task t-abc123.").
```
jmelloy: The original `Be concise — one short paragraph maximum unless detail is requested.` guideline was removed and not replaced with any equivalent for prose responses that don't involve tool calls. Consider adding a brief general conciseness instruction (e.g. 'Outside of tool narration, keep responses concise.') so the model doesn't become verbose in non-tool turns.

backend/foreman/prompt.py
```diff
@@ -90,7 +90,12 @@
...
+- **After each tool result**: write one short sentence summarising what you found or what happened (e.g. "CI is passing and one reviewer has approved.").
```
jmelloy: The before/after examples use a hardcoded task ID (`t-abc123`). While illustrative, a more generic placeholder (e.g. `<task-id>`) would make it clearer these are patterns, not literal values the model might inadvertently echo.

backend/foreman/prompt.py
```diff
@@ -90,7 +90,12 @@
...
+Keep each sentence tight — action verb, subject, relevant detail. No filler phrases.\
 """
```
jmelloy, line 94: There is no instruction covering the case where a tool call fails or returns an error. The after-result sentence guideline should explicitly address error cases (e.g. 'If the tool returned an error, state that briefly and what you will do next.') to avoid silent failures or confusing narration.

backend/foreman/prompt.py
jmelloy: The instruction says to narrate *before* each tool call, but in the Anthropic API the model emits text content blocks and tool_use blocks in sequence — whether a text block genuinely precedes its associated tool_use in the stream is determined by block ordering. Consider clarifying whether "before" means in the same streamed response (text block index < tool_use block index) or in a separate prior message, to avoid ambiguity when testing the WS stream.

backend/foreman/prompt.py
jmelloy, line 94: "Keep each sentence tight" is good advice but sitting immediately before the closing backslash it reads as a continuation of the bullet list rather than a separate directive. Adding a blank line or reformatting as a standalone sentence would improve readability and reduce the chance of the model treating it as part of the second bullet.

backend/foreman/prompt.py
```diff
@@ -90,7 +90,12 @@
 The state reflects the moment this turn was sent — earlier turns saw earlier state.
 
 Workers are configured with repos. Prefer workers whose repos cover the task.
-Be concise — one short paragraph maximum unless detail is requested.\
+
+## Inline narration
+Narrate in-flow with your tool calls — never accumulate a summary paragraph at the end of a turn.
```
jmelloy: The prohibition on a trailing summary paragraph is stated only once and at the top of the section. Models sometimes ignore single-mention constraints. Consider reinforcing it at the end of the section (e.g. "Do not add a concluding paragraph after the last tool result.") to reduce non-compliance.

jmelloy (commented):
- The PR replaces the previous "Be concise" instruction with an inline narration block, which is a meaningful behavioural change to the Foreman prompt worth validating carefully.
- The instruction structure is clear and well-formed, using concrete examples to guide the model, which is good prompt-engineering practice.
- The backslash line continuation at the end of the new block is preserved from the original, maintaining the string formatting correctly.
- There is no mechanism to enforce this behaviour programmatically; compliance is entirely model-dependent and could be inconsistent, especially under different model versions or temperatures.
- No tests are added or modified for the prompt content itself, so regressions in prompt wording would go undetected.

backend/foreman/prompt.py
```diff
@@ -90,7 +90,16 @@
...
+- **Before each tool call**: write one short sentence saying what you are about to do (e.g. "Checking the PR status for task <task-id>.").
+- **After each tool result**: write one short sentence summarising what you found or what happened (e.g. "CI is passing and one reviewer has approved.").
+- **If a tool returned an error**: state that briefly and what you will do next (e.g. "That call failed with a 404 — the PR may have been closed; checking the task status instead.").
+Keep each sentence tight — action verb, subject, relevant detail. No filler phrases.
+
+Outside of tool narration, keep responses concise — one short paragraph maximum unless detail is requested.
```
jmelloy: The trailing backslash here (carried over from the original) suppresses the newline before the closing triple-quote, meaning the `"""` immediately follows the last character of text. This is intentional if the original file relied on it, but it makes the string end without a trailing newline. If any downstream code concatenates this prompt with other strings, the lack of a terminating newline could cause the next section to run on. Consider adding an explicit `\n` and removing the backslash, or at minimum leaving a comment explaining the intent.

backend/foreman/prompt.py
```diff
@@ -90,7 +90,16 @@
 The state reflects the moment this turn was sent — earlier turns saw earlier state.
 
 Workers are configured with repos. Prefer workers whose repos cover the task.
-Be concise — one short paragraph maximum unless detail is requested.\
+
+## Inline narration
```
jmelloy: "Narrate in-flow with your tool calls — never accumulate a summary paragraph at the end of a turn." is then repeated in slightly different words at line 101 ("Do not add a concluding paragraph after the last tool result."). This redundancy is minor but could be consolidated into one clear rule to keep the prompt tight, which aligns with the instruction's own advice to avoid filler.

backend/foreman/prompt.py
```diff
@@ -90,7 +90,16 @@
 The state reflects the moment this turn was sent — earlier turns saw earlier state.
 
 Workers are configured with repos. Prefer workers whose repos cover the task.
-Be concise — one short paragraph maximum unless detail is requested.\
+
```
jmelloy: The examples provided (e.g. `"Checking the PR status for task <task-id>."`) are helpful, but the `<task-id>` placeholder looks like XML/HTML to some LLMs and may be interpreted literally or stripped by certain tokenisers. Consider using a neutral placeholder style such as `{task-id}` or simply `<id>` to avoid any ambiguity.

jmelloy (commented):
- The PR adds inline narration instructions to the Foreman system prompt, guiding it to emit one sentence before and after each tool call rather than accumulating a trailing summary.
- The change is purely a prompt engineering update with no code logic, making it low-risk but also difficult to enforce or test deterministically.
- The trailing backslash (`\`) at the end of the string is preserved from the original, which is valid but worth noting — it suppresses the implicit newline before the closing `"""`.
- No tests are added or modified; the test plan relies entirely on manual verification, which may be insufficient for a behaviour that is LLM-dependent.
- The new instructions are clear and well-structured with concrete examples, but there is a minor inconsistency: the prohibition on a concluding paragraph appears twice (once at the top of the section and once at the bottom), adding slight redundancy.

PR #381, opened 2026-05-20
Codex login support + user settings login management

### LINKED ISSUE #377: feat: Codex CLI login support + expose/delete logins in user settings
## Summary

Two related features:

### 1. Codex CLI login (similar to Claude)
Investigate whether the Codex CLI (`openai/codex`) has an obvious login/auth mechanism analogous to how Claude Code handles auth. If so, implement support for it in the worker tooling so Codex can be used as a coding agent similarly to Claude.

- Check Codex CLI docs/source for login commands (e.g. `codex login`, API key config, OAuth flow, etc.)
- Add any necessary config/env plumbing to support Codex auth alongside Claude auth
- The `tool` field on tasks already supports `"codex"` — this should make that actually work end-to-end

### 2. Expose logins in user settings + allow deletion
Currently there is no UI for users to see which third-party logins/accounts are connected to their profile.

- Add a **User Settings** page (or section) that lists connected logins (GitHub OAuth, and any future providers)
- Allow the user to **delete/disconnect** a login from their account
- Backend: add a `DELETE /api/user/logins/{provider}` (or similar) endpoint
- Frontend: settings UI showing connected accounts with a "Disconnect" button

## Acceptance criteria
- [ ] Codex CLI can be authenticated and used as a coding agent tool
- [ ] User settings page shows connected logins
- [ ] User can disconnect a login from settings
- [ ] Deleting the only login shows a warning (can't lock yourself out)

jmelloy:
## Summary

- **Codex CLI auth**: Plumb `OPENAI_API_KEY` through the worker stack so Codex tasks authenticate without manual setup. The config loader reads `[codex].api_key` (supports `env:VAR` indirection), falls back to `OPENAI_API_KEY` env var, and warns at startup if neither is set. `codex_runner.py` injects the key directly into the subprocess environment. `docker-compose.yml` forwards the variable to worker containers.
- **Login management API**: Add `GET /api/user/logins` (list connected OAuth providers per user) and `DELETE /api/user/logins/{provider}` (disconnect a provider, guarded against last-login lockout — returns HTTP 400 when the user would be permanently locked out).
- **Connected Accounts UI**: Expand `GitHubConfigModal.vue` with a "Connected Accounts" section showing each provider's username, a **Disconnect** button that is disabled (with tooltip) when it is the only login method, an in-modal confirmation dialog before disconnect, and inline success/error status feedback.

## Test plan

- [ ] Set `OPENAI_API_KEY` or `[codex].api_key` in worker config; assign a Codex task — worker startup should log "OPENAI_API_KEY set from config" and the task should run without auth errors
- [ ] Omit the key entirely — worker startup should log a warning, Codex task output should show `codex: error: Missing OpenAI API key` (or similar), not a crash
- [ ] `GET /api/user/logins` returns `{"logins": [{"provider": "github", "username": "..."}]}` for a logged-in user
- [ ] `DELETE /api/user/logins/github` with a single-provider account returns HTTP 400 with "Cannot disconnect your only login method"
- [ ] In the UI, open GitHub config modal — "Connected Accounts" section shows GitHub username; Disconnect button is disabled with tooltip
- [ ] `DELETE /api/user/logins/unknown` returns HTTP 404

Closes #377

### DIFF
--- a/backend/routes/auth.py
+++ b/backend/routes/auth.py
@@ -388,17 +388,24 @@ async def delete_login(provider: str, github_user_id: str = Depends(require_user
 
     db = await get_db()
     try:
+        # Count total logins across all providers before touching anything.
+        # (Only GitHub exists today; expand this query when new providers land.)
         count_res = await db.execute(
             select(func.count())
             .select_from(GithubToken)
             .where(GithubToken.github_user_id == github_user_id)
         )
-        if (count_res.scalar_one() or 0) <= 1:
+        total_logins = count_res.scalar_one() or 0
+        if total_logins <= 1:
             raise HTTPException(
                 status_code=400,
                 detail="Cannot disconnect your only login method — you would be locked out.",
             )
-        await db.execute(delete(GithubToken).where(GithubToken.github_user_id == github_user_id))
+        # Delete only the token(s) belonging to the requested provider.
+        if provider == "github":
+            await db.execute(
+                delete(GithubToken).where(GithubToken.github_user_id == github_user_id)
+            )
         await db.execute(delete(UserSession).where(UserSession.github_user_id == github_user_id))
         await db.commit()
     finally:
--- a/backend/routes/auth.py
+++ b/backend/routes/auth.py
@@ -400,23 +400,17 @@ async def delete_login(provider: str, github_user_id: str = Depends(require_user
                     detail=f"No {provider!r} login found for this user.",
                 )
 
-        # Count how many logins would remain after removing this provider.
-        # "Remaining" = logins from all other providers (not the one being deleted).
-        # Add counts from future provider tables here as they are implemented.
+        # Count total logins across all providers. If <= 1, removing this one
+        # would lock the user out. Add counts from future provider tables here.
         github_count_res = await db.execute(
             select(func.count())
             .select_from(GithubToken)
             .where(GithubToken.github_user_id == github_user_id)
         )
-        github_logins = github_count_res.scalar_one() or 0
+        total_logins = github_count_res.scalar_one() or 0
+        # When new providers are added: total_logins += new_provider_count
 
-        # Sum up logins from providers OTHER than the one being deleted.
-        remaining = 0
-        if provider != "github":
-            remaining += github_logins
-        # When new providers are added: if provider != "new_provider": remaining += new_provider_count
-
-        if remaining == 0:
+        if total_logins <= 1:
             raise HTTPException(
                 status_code=400,
                 detail="Cannot disconnect your only login method — you would be locked out.",
--- a/backend/routes/auth.py
+++ b/backend/routes/auth.py
@@ -32,7 +32,7 @@ from models import (
 )
 from oauth import FRONTEND_URL, GITHUB_CLIENT_ID, create_session, get_return_to, make_authorize_url
 from pydantic import BaseModel
-from sqlalchemy import delete, func, select, text
+from sqlalchemy import delete, select, text
 from sqlalchemy.dialects.sqlite import insert as sqlite_insert
 from utils import generate_guild_id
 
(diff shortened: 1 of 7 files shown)


backend/routes/auth.py
```diff
@@ -235,6 +235,73 @@ async def api_me(github_user_id: str = Depends(require_user)):
...
+    try:
+        count_res = await db.execute(
+            select(func.count())
+            .select_from(GithubToken)
+            .where(GithubToken.github_user_id == github_user_id)
+        )
+        if (count_res.scalar_one() or 0) <= 1:
+            raise HTTPException(
+                status_code=400,
+                detail="Cannot disconnect your only login method — you would be locked out.",
```
jmelloy, line 416: The DELETE targets all GithubToken rows for `github_user_id`, not just the one for the specified provider. Even though only GitHub exists today, this is semantically wrong: a future second provider would also be wiped. Add `.where(GithubToken.provider == provider)` (or equivalent) to scope the delete correctly.

backend/routes/auth.py
```diff
@@ -235,6 +235,73 @@ async def api_me(github_user_id: str = Depends(require_user)):
...
+async def delete_login(provider: str, github_user_id: str = Depends(require_user)):
+    """Disconnect an OAuth provider from the current user's account.
+
+    Guards against removing the last login — if this is the only connected
+    provider the user would be permanently locked out, so we return HTTP 400.
+    Currently only ``github`` is supported; passing any other provider returns
+    HTTP 404.
+    """
+    if provider != "github":
+        raise HTTPException(status_code=404, detail=f"Unknown provider: {provider!r}")
```
jmelloy, line 388: The lockout guard counts ALL tokens for the user before deleting ALL of them. If a user has two GitHub tokens (edge case) the guard passes but both are deleted. The count should be over distinct providers, not raw rows, to be logically correct.

worker/pioneer_worker/worker.py
```diff
@@ -242,6 +242,25 @@ async def _check_codex_doctor(self) -> None:
...
+        """Set OPENAI_API_KEY in the process environment for Codex tasks.
+
+        Priority: env var already set → config openai_api_key → warn and skip.
+        The env var is inherited by every spawned codex subprocess; the runner
+        also accepts it explicitly for extra safety.
+        """
+        if os.environ.get("OPENAI_API_KEY"):
+            logger.info("OPENAI_API_KEY already in env — Codex tasks will use it")
+            return
+        if self.cfg.openai_api_key:
```
jmelloy: `os.environ['OPENAI_API_KEY'] = self.cfg.openai_api_key` mutates the global process environment, potentially exposing the key to any future subprocesses unrelated to Codex. Since `run_codex_auto` already accepts `openai_api_key` explicitly and constructs a scoped env dict, the global mutation in `_ensure_codex_api_key` is unnecessary and should be removed; just rely on the explicit parameter.

frontend/src/components/GitHubConfigModal.vue
```diff
@@ -24,10 +76,78 @@
...
+    setStatus(msg, false)
+  } finally {
+    disconnecting.value = null
+  }
+}
+
+onMounted(fetchLogins)
 </script>
 
 <style scoped>
```
jmelloy, line 153: `statusTimer` is declared at module/setup scope and is never cleared in an `onUnmounted` hook. If the modal is destroyed before the 3.5 s timeout fires, the callback will attempt to write to a stale `statusMsg` ref. Add `onUnmounted(() => { if (statusTimer) clearTimeout(statusTimer) })` to prevent this.

frontend/src/components/GitHubConfigModal.vue
```diff
@@ -24,10 +76,78 @@
...
+  confirmProvider.value = null
+  if (!provider) return
+  disconnecting.value = provider
+  try {
+    await api(`/api/user/logins/${provider}`, { method: 'DELETE' })
+    setStatus(`${providerLabel(provider)} disconnected. You will be signed out.`, true)
+    await authStore.logout()
+  } catch (err: unknown) {
+    const msg = err instanceof Error ? err.message : 'Disconnect failed'
+    setStatus(msg, false)
```
jmelloy, line 144: `doDisconnect` calls `authStore.logout()` without awaiting a try/catch around it. If logout fails (network error, token already invalidated by the server-side session delete), the error is silently swallowed and the user may remain on a page that requires authentication. Wrap in try/catch or at minimum propagate the error to `setStatus`.

jmelloy (commented):
- The `delete_login` endpoint deletes **all** `GithubToken` rows for the user rather than just the one matching the disconnected provider — with a single-provider system today this works, but the guard logic (`count <= 1`) is checking the total count before deleting everything, making the endpoint semantically incorrect for a future multi-provider world.
- `_ensure_codex_api_key` mutates `os.environ` at the process level as a side-channel, while `codex_runner.py` also accepts the key explicitly; the dual approach creates confusion and the global mutation can leak the key to unrelated subprocesses.
- The `delete_login` endpoint invalidates **all** sessions for the user (`DELETE FROM UserSession WHERE github_user_id = ...`), including sessions on other devices, with no warning to the caller beyond the confirmation dialog copy — this is a significant side-effect that should at minimum be documented in the API response or as a distinct HTTP status.
- Frontend `doDisconnect` calls `authStore.logout()` after a successful disconnect but does not handle the case where `logout()` itself fails (e.g., the session was already gone), which could leave the UI in a broken state.
- The `statusTimer` is a module-level `let` variable that is never cleared on component unmount, creating a potential memory leak / stale callback if the modal is closed before the 3.5 s timer fires.

jmelloy (commented):
- The `delete_login` endpoint deletes all `UserSession` rows for the user upon disconnecting a GitHub login, which is a destructive side-effect not clearly communicated in the docstring and could invalidate sessions from other devices without user awareness.
- The lockout-guard logic in `delete_login` is redundant and misleading: since only GitHub is supported, `total_logins == github_count` and `provider_count == github_count` always, so the guard will always block the only existing provider from being removed — making the endpoint functionally unusable for the only supported provider.
- `openai_api_key` is stored in plaintext in the `Config` dataclass and logged as 'configured' at INFO level; care should be taken that it is never accidentally serialised or emitted in debug output.
- The `_ensure_codex_api_key` check reads `os.environ.get('OPENAI_API_KEY')` directly rather than relying solely on `self.cfg.openai_api_key`, which means the logged result can differ from what is actually injected into the subprocess, creating a misleading health signal.
- Frontend `doDisconnect` calls `authStore.logout()` after a successful disconnect, but on error the status message shows `err.message` which for non-Error API rejections may expose raw server error text to the user rather than a sanitised message.

worker/pioneer_worker/config.py
```diff
@@ -160,6 +164,16 @@ def _resolve(p: str) -> str:
     if not isinstance(_codex_args, list) or not all(isinstance(arg, str) for arg in _codex_args):
         raise ValueError("codex args must be a list of strings.")
 
+    _openai_api_key = overrides.get("openai_api_key")
+    if _openai_api_key is None:
+        raw_key = codex_block.get("api_key")
+        if isinstance(raw_key, str) and raw_key.startswith("env:"):
+            _openai_api_key = os.environ.get(raw_key[4:].strip()) or None
+        elif raw_key:
+            _openai_api_key = raw_key
```
jmelloy, line 180: Using `os.environ.get(...) or None` silently discards an explicitly set but empty env var, converting it to `None`. Prefer `x if x is not None else None` or just store the empty string and let downstream code validate. More critically, if someone sets `OPENAI_API_KEY=` intentionally as a placeholder, this will silently fall through without warning.

worker/pioneer_worker/worker.py
```diff
@@ -242,6 +242,21 @@ async def _check_codex_doctor(self) -> None:
         else:
             logger.warning("codex doctor: failed (rc=%d)\n%s", proc.returncode, output)
 
+    async def _ensure_codex_api_key(self) -> None:
+        """Warn early if no OpenAI API key is available for Codex tasks.
+
+        The key is passed explicitly to each run_codex_auto call (via the
```
jmelloy, line 248: This check is redundant: `config.py` already reads `OPENAI_API_KEY` from the environment and stores it in `self.cfg.openai_api_key`, so the `os.environ.get('OPENAI_API_KEY')` branch here can never be True when `self.cfg.openai_api_key` is False/None. The condition should simply be `if self.cfg.openai_api_key:` to avoid confusion about where the key originates.

docker-compose.yml
```diff
@@ -60,6 +60,7 @@ services:
     environment:
       GITHUB_TOKEN: ${GITHUB_TOKEN:-}
       CLAUDE_CODE_OAUTH_TOKEN: ${CLAUDE_CODE_OAUTH_TOKEN:-}
+      OPENAI_API_KEY: ${OPENAI_API_KEY:-}
```
jmelloy, line 63: The `:-` default means the container always receives `OPENAI_API_KEY` set to an empty string when the host variable is unset. This may cause Codex/OpenAI SDK errors that differ from the 'key not set' case. Consider omitting the default (`${OPENAI_API_KEY}`) so the variable is simply absent in the container when not provided on the host, which is consistent with how the config layer handles absence.

jmelloy (commented):
- Codex API key plumbing is clean and avoids global env mutation; key is passed explicitly per-invocation which is good practice.
- Security concern: the `or None` pattern on `os.environ.get` silently treats empty string `''` as absent, which could mask misconfiguration where a key is intentionally set but empty.
- The lockout guard for the login disconnect endpoint is mentioned in the description but its implementation is not visible in this diff, making it impossible to verify correctness of the HTTP 400 guard.
- The `_ensure_codex_api_key` check in `worker.py` is redundant: it checks `os.environ.get('OPENAI_API_KEY')` separately from `self.cfg.openai_api_key`, but `config.py` already falls back to that env var when building the config, so `self.cfg.openai_api_key` will already be non-None if the env var is set.
- `docker-compose.yml` forwards `OPENAI_API_KEY` with a default of empty string (`:-`), which means the container will have the variable set (to `''`) even when not provided, potentially causing confusing behavior downstream.

PR #382, opened 2026-05-20
Phase 3: create pioneer-foreman standalone package

### LINKED ISSUE #376: Phase 3: Create pioneer-foreman standalone package
## Goal
Create the `pioneer-foreman` package — a separately-deployable process modelled after `pioneer-worker` that replaces the embedded foreman.

## Background
- Plan: `docs/foreman-split-plan.md` (merged in #308)
- Phases 1 & 2 (REST endpoints + `foreman-trigger` WebSocket message) landed in #314 and #350
- The embedded foreman continues to run until Phase 4

## Phase 3 tasks

1. **Create `pioneer-foreman/` package** (mirroring `pioneer-worker/` structure):
   - `pyproject.toml` / `setup.cfg` with its own deps (anthropic, websockets, toml, httpx)
   - `pioneer_foreman/main.py` — entrypoint, reads `pioneer-foreman.toml` + env vars
   - `pioneer_foreman/runner.py` — main loop: register via HTTP, connect to backend WebSocket, listen for `foreman-trigger` events, run the Foreman Claude session, push state back via REST endpoints
   - `pioneer_foreman/tools.py` — Foreman MCP tool definitions (currently lives in backend; copy/adapt)
   - `pioneer_foreman/config.py` — config loading (`PIONEER_BACKEND_URL`, `ANTHROPIC_API_KEY`, `FOREMAN_MODEL` env vars + toml file)

2. **Registration**: POST `/api/foreman/register` on startup (endpoint added in Phase 1/2), heartbeat loop

3. **WebSocket**: connect to `ws://backend/ws` with foreman identity, handle `foreman-trigger` message type, ignore worker messages

4. **State I/O**: read task/worker state via existing REST endpoints; write via existing REST endpoints — **no direct DB access**

5. **Config file**: `pioneer-foreman.toml.example` with documented fields

6. **Docker / process**: `Dockerfile` for the foreman process; update `docker-compose.yml` to add a `foreman` service (disabled by default, so embedded foreman still runs)

## Out of scope for Phase 3
- Switching traffic to the new foreman (Phase 4)
- Removing embedded foreman (Phase 5)
- Auth token scoping (can use same token as workers for now)

## References
- Plan: `docs/foreman-split-plan.md`
- Existing worker: `pioneer-worker/`
- Phase 1/2 PR: #314, #350

jmelloy:
Closes #376

Creates the `pioneer-foreman/` standalone package mirroring `pioneer-worker/` structure, with REST-based tool wrappers, WebSocket connection, heartbeat loop, config loading, Dockerfile, and docker-compose service (disabled by default).

### DIFF
--- a/pioneer-foreman/Dockerfile
+++ b/pioneer-foreman/Dockerfile
@@ -1,0 +1,24 @@
+FROM python:3.11-slim
+
+ENV PYTHONUNBUFFERED=1 \
+    PIP_NO_CACHE_DIR=1 \
+    PIP_DISABLE_PIP_VERSION_CHECK=1
+
+RUN apt-get update && apt-get install -y --no-install-recommends \
+        ca-certificates \
+        curl \
+    && rm -rf /var/lib/apt/lists/*
+
+WORKDIR /app
+
+COPY pyproject.toml ./
+COPY pioneer_foreman ./pioneer_foreman
+RUN pip install -e .
+
+RUN useradd --create-home --shell /bin/bash foreman \
+    && mkdir -p /config \
+    && chown foreman:foreman /config
+
+USER foreman
+
+CMD ["pioneer-foreman", "--config", "/config/pioneer-foreman.toml"]
--- a/pioneer-foreman/pioneer_foreman/config.py
+++ b/pioneer-foreman/pioneer_foreman/config.py
@@ -1,0 +1,136 @@
+"""Configuration for the Pioneer Square standalone foreman.
+
+Reads from a TOML file (default: ``./pioneer-foreman.toml``) layered with
+environment variables and optional programmatic overrides.
+"""
+
+from __future__ import annotations
+
+import os
+import tomllib
+from dataclasses import dataclass, field
+from pathlib import Path
+from urllib.parse import urlparse, urlunparse
+
+DEFAULT_CONFIG_NAME = "pioneer-foreman.toml"
+
+
+@dataclass
+class Config:
+    backend_url: str
+    guild_id: str
+    # Bearer token for REST API calls (worker auth_token from registration).
+    # Populated by Foreman.run() after _register() when not pre-configured.
+    # Pre-configure via PIONEER_FOREMAN_AUTH_TOKEN or the auth_token TOML key
+    # to skip re-registration on each restart.
+    auth_token: str | None = None
+    # Claude model for the foreman AI
+    model: str = "claude-sonnet-4-6"
+    # Optional GitHub token for direct GitHub API calls.
+    # If absent, fetched from the backend at first use via /auth/github/token.
+    github_token: str | None = None
+    # Seconds between background poll cycles (sent as foreman-poll-status broadcasts)
+    poll_interval: float = 60.0
+    # Path used for display / resolving relative paths
+    config_path: Path = field(default_factory=Path)
+
+    @property
+    def http_url(self) -> str:
+        """Backend HTTP base URL derived from backend_url."""
+        parsed = urlparse(self.backend_url)
+        scheme = {"ws": "http", "wss": "https"}.get(parsed.scheme, parsed.scheme or "http")
+        return urlunparse((scheme, parsed.netloc, parsed.path.rstrip("/"), "", "", ""))
+
+    @property
+    def ws_url(self) -> str:
+        """Backend WebSocket URL for this guild."""
+        parsed = urlparse(self.backend_url)
+        scheme = {"http": "ws", "https": "wss"}.get(parsed.scheme, parsed.scheme or "ws")
+        base = urlunparse((scheme, parsed.netloc, parsed.path.rstrip("/"), "", "", ""))
+        return f"{base}/ws/{self.guild_id}"
+
+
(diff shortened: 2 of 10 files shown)


pioneer-foreman/pioneer_foreman/config.py
```diff
@@ -0,0 +1,136 @@
...
+    backend_url = (
+        overrides.get("backend_url")
+        or raw.get("backend_url")
+        or os.environ.get("PIONEER_BACKEND_URL")
+    )
+    guild_id = (
+        overrides.get("guild_id") or raw.get("guild_id") or os.environ.get("PIONEER_GUILD_ID")
+    )
+    if not backend_url:
+        raise ValueError(
```
jmelloy, line 96: Priority bug: `backend_url` and `guild_id` resolve as `overrides → TOML → env`, but the docstring and design intent specify `overrides → env → TOML`. Env vars should take precedence over TOML values. Swap the order to `overrides.get(...) or os.environ.get(...) or raw.get(...)`.

pioneer-foreman/pioneer_foreman/main.py
```diff
@@ -0,0 +1,308 @@
...
+                    msg_type = msg.get("type")
+
+                    if msg_type == "foreman-registered":
+                        logger.info(
+                            "Registered as external foreman: agentId=%s guild=%s",
+                            msg.get("agentId"),
+                            guild_id,
+                        )
+
+                    elif msg_type == "foreman-trigger":
```
jmelloy, line 167: Importing `_poll_tasks` inside the `finally` block is unnecessary and fragile — the module is already imported at the top of the file. Use `from pioneer_foreman import runner` at module level and reference `runner._poll_tasks` here, or simply import it at the top.

pioneer-foreman/pioneer_foreman/main.py
```diff
@@ -0,0 +1,308 @@
...
+                poll_interval=cfg.poll_interval,
+            )
+
+            try:
+                async for raw in ws:
+                    try:
+                        msg = json.loads(raw)
+                    except json.JSONDecodeError:
+                        continue
+
```
jmelloy, line 157: `asyncio.create_task()` tasks that raise exceptions will log an unhandled exception and be silently dropped. Consider adding a done-callback to log failures or storing references to prevent GC before completion, e.g. `task.add_done_callback(lambda t: t.exception() and logger.error(...))`.

pioneer-foreman/pioneer_foreman/main.py
```diff
@@ -0,0 +1,308 @@
...
+    )
+    resp.raise_for_status()
+    data = resp.json()
+    token: str = data["auth_token"]
+    logger.info(
+        "Registered as worker %s (foreman identity) for guild %s",
+        data.get("id", "?"),
+        guild_id,
+    )
+    return token
```
jmelloy, line 77: The `cfg.auth_token = auth_token` mutation caches the token in memory only — on every process restart without `PIONEER_FOREMAN_AUTH_TOKEN` set, a new worker registration is created. Over time this will accumulate stale worker rows in the backend. Consider logging the token at INFO level with an explicit prompt to persist it as `PIONEER_FOREMAN_AUTH_TOKEN`.

pioneer-foreman/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+RUN apt-get update && apt-get install -y --no-install-recommends \
+        ca-certificates \
+        curl \
+    && rm -rf /var/lib/apt/lists/*
+
+WORKDIR /app
+
+COPY pyproject.toml ./
+COPY pioneer_foreman ./pioneer_foreman
+RUN pip install -e .
```
jmelloy, line 16: `pip install -e .` in a Docker image installs in editable mode, which mounts the source directory rather than copying it into site-packages. This is fine only because the source is COPYed above, but it's unconventional for production images and may cause issues if the `/app` directory is later overlaid by a volume mount. Prefer `pip install .` (non-editable) for container builds.

jmelloy (commented):
- The standalone `pioneer-foreman` package is well-structured, mirroring `pioneer-worker` with clean separation of concerns: config, runner, tools, and main entrypoint.
- A significant security concern exists: `auth_token` is mutated directly on the shared `cfg` object during runtime (`cfg.auth_token = auth_token`), but this is never persisted, so every restart without `PIONEER_FOREMAN_AUTH_TOKEN` set will re-register, potentially creating unbounded worker registrations.
- The config loading priority in `load()` has an inconsistency: env vars are checked before TOML values for most fields, but `github_token` applies overrides → env → TOML correctly while `backend_url` and `guild_id` use overrides → TOML → env (wrong order), breaking documented priority semantics.
- The `_poll_tasks` dict is imported inside a `finally` block as a late import (`from pioneer_foreman.runner import _poll_tasks`), which is unnecessary and adds fragility — it's already in scope via the module.
- The `prompt.py` file is explicitly noted as a manual copy of `backend/foreman/prompt.py`, creating a maintenance burden and risk of drift; a shared package or symlink would be more robust.

PR #386, opened 2026-05-20
User settings: login management

jmelloy:
## Summary

Splits the user settings / login-management portion of #381 into a standalone PR — no Codex API key changes included.

- **`GET /api/user/logins`** — list connected OAuth providers for the current user (GitHub only for now; response shape supports future providers)
- **`DELETE /api/user/logins/{provider}`** — disconnect a provider with a lockout guard (HTTP 400 if it would be the last login)
- **Frontend** — "Connected Accounts" section in `GitHubConfigModal` with a per-provider Disconnect button and an in-modal confirmation dialog

## Reviewer concerns addressed from #381

**(a) Token scope** — `delete_login` deletes only the `GithubToken` row for the requesting user and the specified provider; other providers (future) are not touched.

**(b) Session behaviour** — the endpoint no longer deletes any `UserSession` rows. The docstring now explicitly documents that: only the current session is invalidated (the frontend calls `DELETE /auth/logout` immediately after); other active sessions remain valid until they naturally expire.

**(c) Lockout guard** — the guard counts logins for this specific user across all providers; both the count query and the deletion query are scoped to the same `github_user_id`, so the `<= 1` check correctly prevents the user from locking themselves out.

## Test plan

- [ ] Open GitHubConfigModal while logged in — Connected Accounts section loads and shows the GitHub entry
- [ ] Disconnect button is disabled (greyed out) when only 1 login exists — tooltip explains why
- [ ] Clicking Disconnect shows confirmation dialog; Cancel dismisses without action
- [ ] Confirm Disconnect calls `DELETE /api/user/logins/github`, shows success message, then signs out
- [ ] `GET /api/user/logins` returns 401 without a valid session
- [ ] `DELETE /api/user/logins/github` returns 400 when the user has only one login method
- [ ] `DELETE /api/user/logins/unknown` returns 404
- [ ] `DELETE /api/user/logins/github` returns 404 when no GitHub token exists for the user

### DIFF
--- a/backend/routes/auth.py
+++ b/backend/routes/auth.py
@@ -32,7 +32,7 @@ from models import (
 )
 from oauth import FRONTEND_URL, GITHUB_CLIENT_ID, create_session, get_return_to, make_authorize_url
 from pydantic import BaseModel
-from sqlalchemy import delete, select, text
+from sqlalchemy import delete, func, select, text
 from sqlalchemy.dialects.sqlite import insert as sqlite_insert
 from utils import generate_guild_id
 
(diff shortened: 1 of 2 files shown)


backend/routes/auth.py
```diff
@@ -339,6 +339,103 @@ async def guest_login():
...
+    Currently only ``github`` is supported; passing any other provider returns
+    HTTP 404.
+
+    Session behaviour: this endpoint removes only the provider credential for
+    the named provider.  It does NOT delete ``UserSession`` rows — the frontend
+    is responsible for calling ``DELETE /auth/logout`` immediately after to
+    invalidate the current session.  Other active sessions for this user remain
+    valid until they naturally expire; they will receive 401 on the next GitHub-
+    authenticated API call once the token is gone.
+    """
```
jmelloy, line 393: Logic bug: since GitHub is currently the only provider, `github_count` will always be 1 when the token exists, so `total_logins <= 1` is always true and no user can ever disconnect. The guard should check the *remaining* count after removal (i.e., `total_logins - 1 < 1`), or equivalently `total_logins < 2` only when other providers exist.

backend/routes/auth.py
```diff
@@ -339,6 +339,103 @@ async def guest_login():
...
+    if provider != "github":
+        raise HTTPException(status_code=404, detail=f"Unknown provider: {provider!r}")
+
+    db = await get_db()
+    try:
+        # Verify the requested provider login actually exists for this user.
+        existing_res = await db.execute(
```
jmelloy, line 400: Dead code: this `if provider == 'github':` branch is unreachable because any non-`'github'` provider already raised HTTP 404 at the top of the function. Remove the conditional and keep only the `delete()` statement.

frontend/src/components/GitHubConfigModal.vue
```diff
@@ -24,10 +76,78 @@
...
+  statusMsg.value = msg
+  statusClass.value = ok ? 'status-ok' : 'status-err'
+  if (statusTimer) clearTimeout(statusTimer)
+  statusTimer = setTimeout(() => {
+    statusMsg.value = ''
+  }, 3500)
+}
+
+function confirmDisconnect(provider: string) {
+  confirmProvider.value = provider
```
jmelloy, line 130: If `authStore.logout()` throws after a successful `DELETE /api/user/logins/{provider}`, the error handler overwrites the success status message with a misleading 'Disconnect failed' message even though the provider was actually disconnected. Consider separating the logout error handling or re-throwing with a distinct message.

jmelloy (commented):
- **Logic bug in lockout guard**: The `delete_login` endpoint counts GitHub tokens and checks `total_logins <= 1`, but since only GitHub is currently supported, a user will *always* have exactly 1 login, making it impossible to ever disconnect — the guard is effectively a permanent block rather than a safety net.
- **Redundant `provider == 'github'` check**: Inside `delete_login`, the `if provider == 'github':` branch before the DELETE is unreachable dead code since the function already raises 404 for any non-github provider at the top.
- **`GET /api/user/logins` returns 200 with empty list for non-existent users**: If `github_user_id` is somehow invalid or the token was already deleted, the endpoint silently returns `{"logins": []}` rather than surfacing the inconsistency — this is a minor concern but could mask auth issues.
- **Frontend calls `authStore.logout()` unconditionally after disconnect**: If the API call succeeds but `authStore.logout()` throws, `disconnecting` is still cleared via `finally` but the status message already says the user will be signed out, potentially leaving a confusing UI state.
- **No backend tests included**: The lockout guard logic and the session-invalidation behaviour documented in the docstring are critical correctness properties with no accompanying test coverage visible in the diff.

PR #392, opened 2026-05-20
Split PR#381 – Codex API key plumbing

jmelloy:
## Summary

Split from PR #381 — contains only the Codex API key changes. Login management / Connected Accounts work remains in #381.

- **`config.py`**: Add `openai_api_key` field to `Config`; loader reads it from `[codex].api_key` (supports `env:VAR` indirection) and falls back to the `OPENAI_API_KEY` environment variable. Both paths now **raise `ValueError` on empty string** instead of silently treating `''` as absent — this addresses the reviewer note from #381 about the `or None` pattern.
- **`codex_runner.py`**: Accept `openai_api_key` as an explicit kwarg and inject it into the subprocess environment, keeping the key out of the global process env until the subprocess is spawned.
- **`worker.py`**: Add `_ensure_codex_api_key()` startup diagnostic (logs availability, warns if missing, non-fatal); forward `cfg.openai_api_key` to every `run_codex_auto` call.
- **`docker-compose.yml`**: Forward `OPENAI_API_KEY` from the host into worker containers.
- **`pioneer-worker.toml.example`**: Document the `[codex] api_key` option and `env:VAR` syntax.

## Test plan

- [ ] Set `OPENAI_API_KEY` or `[codex] api_key = "env:OPENAI_API_KEY"` in worker config; assign a Codex task — startup should log "OPENAI_API_KEY configured" and the task should run without auth errors
- [ ] Omit the key entirely — startup logs a warning, Codex task output shows `codex: error: Missing OpenAI API key`, no crash
- [ ] Set `OPENAI_API_KEY=""` (empty string) — `config.load()` raises `ValueError` at startup with a clear message
- [ ] Set `[codex] api_key = "env:MY_VAR"` with `MY_VAR=""` — same `ValueError` at startup

Closes part of #381.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -90,6 +90,7 @@ services:
     environment:
       GITHUB_TOKEN: ${GITHUB_TOKEN:-}
       CLAUDE_CODE_OAUTH_TOKEN: ${CLAUDE_CODE_OAUTH_TOKEN:-}
+      OPENAI_API_KEY: ${OPENAI_API_KEY:-}
       WORKER_LOG_LEVEL: ${WORKER_LOG_LEVEL:-INFO}
     volumes:
       # Provide pioneer-worker.toml at ./worker/pioneer-worker.toml
--- a/worker/pioneer_worker/codex_runner.py
+++ b/worker/pioneer_worker/codex_runner.py
@@ -42,8 +42,15 @@ async def run_codex_auto(
     emit: EmitFn,
     codex_path: str = "codex",
     codex_args: list[str] | None = None,
+    openai_api_key: str | None = None,
 ) -> tuple[bool, str, str]:
-    """Run codex on *description* in *cwd*. Returns (success, stop_reason, last_text)."""
+    """Run codex on *description* in *cwd*. Returns (success, stop_reason, last_text).
+
+    *openai_api_key* is injected as OPENAI_API_KEY into the subprocess environment
+    when provided, overriding whatever is in the current process env. This lets
+    the worker forward the key without requiring it to be set globally before
+    import time.
+    """
     last_message_file = tempfile.NamedTemporaryFile(
         prefix="pioneer-codex-last-", suffix=".txt", delete=False
     )
@@ -73,12 +80,16 @@ async def run_codex_auto(
         # prompt content and tries to drain it. Give it an idle TTY so the
         # explicit prompt argument is the only task input.
         master_fd, slave_fd = pty.openpty()
+        env = dict(os.environ)
+        if openai_api_key:
+            env["OPENAI_API_KEY"] = openai_api_key
         proc = await asyncio.create_subprocess_exec(
             *cmd,
             cwd=cwd,
             stdin=slave_fd,
             stdout=asyncio.subprocess.PIPE,
             stderr=asyncio.subprocess.PIPE,
+            env=env,
         )
         os.close(slave_fd)
         slave_fd = None
--- a/worker/pioneer_worker/codex_runner.py
+++ b/worker/pioneer_worker/codex_runner.py
@@ -81,9 +81,7 @@ async def run_codex_auto(
         # explicit prompt argument is the only task input.
         master_fd, slave_fd = pty.openpty()
         env = dict(os.environ)
-        if openai_api_key is not None:
-            if not openai_api_key:
-                raise ValueError("openai_api_key must be a non-empty string")
+        if openai_api_key:
             env["OPENAI_API_KEY"] = openai_api_key
         proc = await asyncio.create_subprocess_exec(
             *cmd,
--- a/worker/pioneer_worker/codex_runner.py
+++ b/worker/pioneer_worker/codex_runner.py
@@ -82,6 +82,8 @@ async def run_codex_auto(
         master_fd, slave_fd = pty.openpty()
         env = dict(os.environ)
         if openai_api_key is not None:
+            if not openai_api_key:
+                raise ValueError("openai_api_key must be a non-empty string")
             env["OPENAI_API_KEY"] = openai_api_key
         proc = await asyncio.create_subprocess_exec(
             *cmd,
--- a/worker/pioneer_worker/codex_runner.py
+++ b/worker/pioneer_worker/codex_runner.py
@@ -81,7 +81,7 @@ async def run_codex_auto(
         # explicit prompt argument is the only task input.
         master_fd, slave_fd = pty.openpty()
         env = dict(os.environ)
-        if openai_api_key:
+        if openai_api_key is not None:
             env["OPENAI_API_KEY"] = openai_api_key
         proc = await asyncio.create_subprocess_exec(
             *cmd,
--- a/worker/pioneer_worker/config.py
+++ b/worker/pioneer_worker/config.py
@@ -41,6 +41,10 @@ class Config:
     codex_path: str = "codex"
     codex_args: list[str] = field(default_factory=list)
     codex_doctor: bool = True
+    # OpenAI API key for Codex tasks. Falls back to OPENAI_API_KEY env var at
+    # startup; set here to avoid exposing the secret in the process environment
+    # before the worker has a chance to forward it.
+    openai_api_key: str | None = None
     pi_path: str = "pi"
     pull_interval: float = 300.0
     claude_max_turns: int = 50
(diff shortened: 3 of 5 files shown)


worker/pioneer_worker/codex_runner.py
```diff
@@ -73,12 +80,16 @@ async def run_codex_auto(
         # prompt content and tries to drain it. Give it an idle TTY so the
         # explicit prompt argument is the only task input.
         master_fd, slave_fd = pty.openpty()
+        env = dict(os.environ)
```
jmelloy, line 83: Using a truthy check here means an explicitly passed empty string `''` is silently ignored rather than raising or being forwarded. While `config.py` now prevents empty strings from reaching this point, a defensive check like `if openai_api_key is not None:` would be more correct and future-proof.

jmelloy (commented):
- The core plumbing is clean: the key is kept out of global `os.environ` until subprocess spawn, which is the right security posture for secret handling.
- **Bug**: `codex_runner.py` uses a truthy check `if openai_api_key:` which silently ignores an explicitly passed empty string — but given `config.py` now raises on empty strings this is low risk in practice; however the inconsistency is worth noting.
- **Logic gap in `_ensure_codex_api_key`**: the check re-reads `os.environ.get('OPENAI_API_KEY')` at runtime, which is redundant with `cfg.openai_api_key` (which already consumed that env var at config load time) and could log a false positive if the env var was present but empty (config loader would have raised before reaching this point).
- **Partial `env:` secret leak**: when `raw_key` does not start with `env:` and is a non-empty string, the literal key is stored in the TOML file; the example file and comment should more strongly discourage this pattern.
- Docker Compose correctly passes `OPENAI_API_KEY` with a blank default, ensuring the variable is forwarded without hard-coding secrets.

worker/pioneer_worker/worker.py
```diff
@@ -242,6 +242,21 @@ async def _check_codex_doctor(self) -> None:
         else:
             logger.warning("codex doctor: failed (rc=%d)\n%s", proc.returncode, output)
 
+    async def _ensure_codex_api_key(self) -> None:
+        """Warn early if no OpenAI API key is available for Codex tasks.
+
+        The key is passed explicitly to each run_codex_auto call (via the
```
jmelloy, line 248: `os.environ.get('OPENAI_API_KEY')` is checked here at runtime, but `config.py` already read and normalised this variable into `cfg.openai_api_key` at startup (and raised on empty string). The only correct check needed here is `self.cfg.openai_api_key`, otherwise the two code paths can disagree — e.g. if the env var was mutated after config load, this method would report it as available while the actual subprocess invocation uses the stale `cfg` value.

worker/pioneer_worker/config.py
```diff
@@ -160,6 +164,29 @@ def _resolve(p: str) -> str:
...
 
+    _openai_api_key = overrides.get("openai_api_key")
+    if _openai_api_key is None:
+        raw_key = codex_block.get("api_key")
+        if isinstance(raw_key, str) and raw_key.startswith("env:"):
+            _var_name = raw_key[4:].strip()
+            _env_val = os.environ.get(_var_name)
+            if _env_val == "":
+                raise ValueError(
+                    f"[codex] api_key references env:{_var_name!r} but the variable is set to an"
```
jmelloy: When `raw_key` is a non-empty string that does not start with `env:` the literal secret is used as-is from the TOML file. There is no validation that this is intentional (vs. a user who forgot the `env:` prefix). Consider logging a warning such as 'api_key is set as a literal string in config; prefer env:VAR_NAME' to guide operators away from storing secrets in config files.

worker/pioneer_worker/config.py
```diff
@@ -160,6 +164,29 @@ def _resolve(p: str) -> str:
     if not isinstance(_codex_args, list) or not all(isinstance(arg, str) for arg in _codex_args):
         raise ValueError("codex args must be a list of strings.")
 
+    _openai_api_key = overrides.get("openai_api_key")
```
jmelloy, line 170: The variable name is extracted with `.strip()` (e.g. `env: OPENAI_API_KEY` → `OPENAI_API_KEY`), but the empty-string guard below only catches an empty resolved value, not a blank variable name. If `raw_key` is exactly `'env:'` or `'env:  '`, `_var_name` becomes `''` or `''`, and `os.environ.get('')` returns `None` silently. Add a check: `if not _var_name: raise ValueError(...)`

worker/pioneer_worker/codex_runner.py
jmelloy, line 83: Use `if openai_api_key is not None:` instead of `if openai_api_key:`. The function signature types this as `str | None` and `config.py` already raises on empty strings, but a truthiness check still creates a subtle inconsistency: if somehow an empty string reaches this point it would be silently skipped rather than injected, masking the misconfiguration instead of making it visible.

worker/pioneer_worker/config.py
jmelloy: `elif raw_key:` accepts any truthy string that doesn't start with `env:` as a literal API key, but only the exact empty-string case is guarded against. A whitespace-only value like `'   '` passes this check and gets stored as the key. Consider adding `.strip()` and checking the stripped value, or validating that the key looks plausibly valid (e.g. starts with `sk-`).

worker/pioneer_worker/worker.py
```diff
@@ -242,6 +242,21 @@ async def _check_codex_doctor(self) -> None:
...
+        openai_api_key parameter) rather than being injected into os.environ,
```
jmelloy, line 249: `_ensure_codex_api_key` checks `os.environ.get('OPENAI_API_KEY')` in addition to `self.cfg.openai_api_key`, but `config.py` already reads `OPENAI_API_KEY` into `cfg.openai_api_key` at load time. This double-check is redundant and slightly misleading — it implies the env var is used at task runtime, whereas the subprocess actually only receives the key via the explicit kwarg. Simplify to `if self.cfg.openai_api_key:` to match the actual runtime behaviour.

jmelloy (commented):
- The core API key plumbing is sound: explicit kwarg injection into subprocess env avoids polluting the global process environment, which is the right design.
- **Bug in `codex_runner.py`**: `if openai_api_key:` will silently skip injection for an empty string, but `config.py` already raises on empty strings — however, the truthiness check is still inconsistent with the explicit `None` type contract; use `if openai_api_key is not None:` for correctness.
- **Logic issue in `_ensure_codex_api_key`**: the check reads `os.environ.get('OPENAI_API_KEY')` directly, but `config.py` already consolidated this into `cfg.openai_api_key`; the double-check is redundant and could log a false-positive 'configured' message if the env var contains a value that config.py would have raised on (though config.py raises at load time, making this moot in practice — still confusing).
- **`env:VAR` with an inline literal key**: `elif raw_key:` accepts any truthy non-`env:`-prefixed string as a literal key with no validation (e.g. whitespace-only string passes `raw_key` truthiness but `== ''` guard only catches exact empty string), which could silently inject a bogus key.
- The `docker-compose.yml` change correctly uses `${OPENAI_API_KEY:-}` (empty default), which is consistent with the existing pattern for other secrets; the example TOML comment is clear and helpful.

worker/pioneer_worker/config.py
```diff
@@ -160,6 +164,31 @@ def _resolve(p: str) -> str:
...
+        if isinstance(raw_key, str) and raw_key.startswith("env:"):
+            _var_name = raw_key[4:].strip()
+            if not _var_name:
+                raise ValueError("env: directive has a blank variable name")
+            _env_val = os.environ.get(_var_name)
+            if _env_val == "":
+                raise ValueError(
+                    f"[codex] api_key references env:{_var_name!r} but the variable is set to an"
+                    " empty string. Provide a valid OpenAI API key or unset the variable."
+                )
```
jmelloy, line 183: Bug: when `api_key` is absent from the TOML block, `raw_key` is `None` and this `elif raw_key.strip()` raises `AttributeError`. Add an `isinstance(raw_key, str)` guard: `elif isinstance(raw_key, str) and raw_key.strip():`

worker/pioneer_worker/config.py
```diff
@@ -160,6 +164,31 @@ def _resolve(p: str) -> str:
...
+            _openai_api_key = _env_val  # None if var is absent, key string if present
+        elif raw_key.strip():
+            _openai_api_key = raw_key.strip()
```
jmelloy, line 187: Raising `ValueError` on an empty `OPENAI_API_KEY` is a breaking change: any host that unconditionally exports the variable (e.g. CI, the `${OPENAI_API_KEY:-}` default in docker-compose.yml) will now hard-crash at startup instead of gracefully falling through to 'key not configured'. Consider logging a warning and treating empty string as absent, consistent with how other optional env vars are handled.

docker-compose.yml
```diff
@@ -90,6 +90,7 @@ services:
     environment:
       GITHUB_TOKEN: ${GITHUB_TOKEN:-}
       CLAUDE_CODE_OAUTH_TOKEN: ${CLAUDE_CODE_OAUTH_TOKEN:-}
+      OPENAI_API_KEY: ${OPENAI_API_KEY:-}
```
jmelloy, line 93: The `${OPENAI_API_KEY:-}` default expands to an empty string when the host variable is unset. Combined with the new `ValueError` on empty strings in `config.py`, this will cause the worker to crash on startup in any environment where `OPENAI_API_KEY` is not set on the host. Use `${OPENAI_API_KEY:-}` only if you intend to require the variable, otherwise omit the environment entry or document that the host must set it.

worker/pioneer_worker/codex_runner.py
```diff
@@ -73,12 +80,18 @@ async def run_codex_auto(
...
+        if openai_api_key is not None:
```
jmelloy, line 84: The inner `if not openai_api_key:` check is dead code in practice: `openai_api_key` was already confirmed non-None on the previous line, so only an empty string could reach here. Consider asserting or collapsing the two conditions into one: `if openai_api_key is not None and not openai_api_key: raise ValueError(...)` before the `env` construction, which makes the intent clearer.

jmelloy (commented):
- The `env:` indirection parsing has a latent `AttributeError`: when `raw_key` is `None` (key absent from TOML block), the `elif raw_key.strip()` branch is reached unconditionally and will crash with `AttributeError: 'NoneType' object has no attribute 'strip'`.
- Raising `ValueError` on an empty `OPENAI_API_KEY` is a breaking change for environments that pass the variable unconditionally (e.g., CI that sets it to `''`); a warning-and-treat-as-absent approach would be safer and less surprising.
- The `not openai_api_key` guard in `codex_runner.py` is redundant: `openai_api_key is not None` is already checked on the line above, so the inner `if not openai_api_key` can only be reached with an empty string — the check is correct but the logic is confusing and worth clarifying.
- `docker-compose.yml` passes `OPENAI_API_KEY` with a default of empty string (`${OPENAI_API_KEY:-}`), which will now trigger the new `ValueError` in `config.py` if the host variable is unset, defeating the graceful-missing-key path.
- Overall approach is sound: explicit kwarg injection into the subprocess env rather than mutating `os.environ` is the right security pattern here.

PR #393, opened 2026-05-20
Fix: UI doesn't mark task inactive after cancellation

### LINKED ISSUE #390: UI doesn't mark task inactive after cancellation
## Bug

When a running task is cancelled (via `cancel_task`), the UI continues to show it as active/running rather than updating to an inactive/cancelled state.

## Expected behaviour
The task card should immediately reflect the cancelled state (e.g. greyed out, labelled "Cancelled", removed from the active list).

## Actual behaviour
The task card remains visually active after cancellation. A page reload is required to see the correct state.

## Likely cause
The frontend is probably not listening for a `task-cancelled` (or equivalent) WebSocket event, or the backend is not broadcasting such an event when a task transitions to `cancelled`. The reactive state update that handles `task-complete` / `task-failed` may not cover `task-cancelled`.

## Steps to reproduce
1. Start a task that runs for a few seconds.
2. Cancel it via the Foreman.
3. Observe the UI — the task card still shows as active.

jmelloy:
## Summary

- **Race condition fix**: `task-complete` and `task-followup-done` WS handlers previously unconditionally set task state to `awaiting-review`. If a worker sent `task-complete` just after the backend had already transitioned the task to `cancelled`, the frontend would override the cancelled state — requiring a page reload to recover. Now these handlers skip the state update if the task is already in a terminal state (`done`, `failed`, `cancelled`).
- **Immediate feedback**: `cancelTask` now optimistically updates the local task state to `cancelled` right after the API call returns, so the card is marked inactive without waiting for the WebSocket event.
- **CSS fix**: Added `.dot-cancelled` CSS class to `TaskList.vue` so cancelled tasks display the correct dim dot indicator (previously the dot was invisible).

## Test plan

- [x] All existing frontend tests pass (`npm test` — 76/76)
- [x] TypeScript type-check clean (`npm run type-check`)
- [x] New tests added for terminal-state guard on `task-complete` and `task-followup-done`
- [x] New test for `cancelTask` optimistic update

Closes #390

### DIFF
--- a/frontend/src/stores/tasks.ts
+++ b/frontend/src/stores/tasks.ts
@@ -7,7 +7,7 @@ import type { LogEntry, Task, TaskState, WSInbound } from '../types'
 // immediately on long horizons (e.g. a 3-day finalize window).
 const MAX_TIMEOUT_MS = 2_147_483_647
 
-const _TERMINAL_STATES = new Set<string>(['done', 'failed', 'cancelled'])
+const TERMINAL_STATES = new Set<string>(['done', 'failed', 'cancelled'])
 
 const STATE_LABELS: Record<string, string> = {
   pending: 'pending',
@@ -111,22 +111,13 @@ export const useTasksStore = defineStore('tasks', () => {
   }
 
   async function cancelTask(guildId: string, taskId: string) {
+    const result = await api(`/guilds/${guildId}/tasks/${taskId}/cancel`, { method: 'POST' })
     const task = tasks.value.find((t) => t.id === taskId)
-    const prevState = task?.state
-    const prevFinishedAt = task?.finished_at
     if (task) {
       task.state = 'cancelled'
       task.finished_at = new Date().toISOString()
     }
-    try {
-      return await api(`/guilds/${guildId}/tasks/${taskId}/cancel`, { method: 'POST' })
-    } catch (e) {
-      if (task) {
-        task.state = prevState!
-        task.finished_at = prevFinishedAt ?? null
-      }
-      throw e
-    }
+    return result
   }
 
   async function redirectTask(guildId: string, taskId: string, instructions: string) {
@@ -183,13 +174,13 @@ export const useTasksStore = defineStore('tasks', () => {
       }
     } else if (data.type === 'task-complete') {
       const task = tasks.value.find((t) => t.id === data.taskId)
-      if (task && !_TERMINAL_STATES.has(task.state)) {
+      if (task && !TERMINAL_STATES.has(task.state)) {
         task.state = 'awaiting-review'
         if (data.branch) task.branch = data.branch
       }
     } else if (data.type === 'task-followup-done') {
       const task = tasks.value.find((t) => t.id === data.taskId)
-      if (task && !_TERMINAL_STATES.has(task.state)) task.state = 'awaiting-review'
+      if (task && !TERMINAL_STATES.has(task.state)) task.state = 'awaiting-review'
     } else if (data.type === 'terminal-output' && data.taskId) {
       const { taskId, line, timestamp, detail } = data
       if (line) {
--- a/frontend/src/stores/tasks.ts
+++ b/frontend/src/stores/tasks.ts
@@ -111,13 +111,22 @@ export const useTasksStore = defineStore('tasks', () => {
   }
 
   async function cancelTask(guildId: string, taskId: string) {
-    const result = await api(`/guilds/${guildId}/tasks/${taskId}/cancel`, { method: 'POST' })
     const task = tasks.value.find((t) => t.id === taskId)
+    const prevState = task?.state
+    const prevFinishedAt = task?.finished_at
     if (task) {
       task.state = 'cancelled'
       task.finished_at = new Date().toISOString()
     }
-    return result
+    try {
+      await api(`/guilds/${guildId}/tasks/${taskId}/cancel`, { method: 'POST' })
+    } catch (err) {
+      if (task) {
+        task.state = prevState!
+        task.finished_at = prevFinishedAt
+      }
+      throw err
+    }
   }
 
   async function redirectTask(guildId: string, taskId: string, instructions: string) {
--- a/frontend/src/stores/tasks.ts
+++ b/frontend/src/stores/tasks.ts
@@ -7,6 +7,8 @@ import type { LogEntry, Task, TaskState, WSInbound } from '../types'
 // immediately on long horizons (e.g. a 3-day finalize window).
 const MAX_TIMEOUT_MS = 2_147_483_647
 
+const _TERMINAL_STATES = new Set<string>(['done', 'failed', 'cancelled'])
+
 const STATE_LABELS: Record<string, string> = {
   pending: 'pending',
   planning: 'planning',
@@ -109,7 +111,12 @@ export const useTasksStore = defineStore('tasks', () => {
   }
 
   async function cancelTask(guildId: string, taskId: string) {
-    return api(`/guilds/${guildId}/tasks/${taskId}/cancel`, { method: 'POST' })
+    await api(`/guilds/${guildId}/tasks/${taskId}/cancel`, { method: 'POST' })
+    const task = tasks.value.find((t) => t.id === taskId)
+    if (task) {
+      task.state = 'cancelled'
+      task.finished_at = new Date().toISOString()
+    }
   }
 
   async function redirectTask(guildId: string, taskId: string, instructions: string) {
@@ -166,13 +173,13 @@ export const useTasksStore = defineStore('tasks', () => {
       }
     } else if (data.type === 'task-complete') {
       const task = tasks.value.find((t) => t.id === data.taskId)
-      if (task) {
+      if (task && !_TERMINAL_STATES.has(task.state)) {
         task.state = 'awaiting-review'
         if (data.branch) task.branch = data.branch
       }
     } else if (data.type === 'task-followup-done') {
       const task = tasks.value.find((t) => t.id === data.taskId)
-      if (task) task.state = 'awaiting-review'
+      if (task && !_TERMINAL_STATES.has(task.state)) task.state = 'awaiting-review'
     } else if (data.type === 'terminal-output' && data.taskId) {
       const { taskId, line, timestamp, detail } = data
       if (line) {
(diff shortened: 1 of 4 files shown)


frontend/src/stores/tasks.ts
```diff
@@ -109,7 +111,12 @@ export const useTasksStore = defineStore('tasks', () => {
   }
```
jmelloy, line 112: The `await api(...)` call is no longer wrapped in a try/catch and `cancelTask` no longer returns the API response. If the API call rejects (network error, non-2xx), the optimistic state mutation below still won't run (because `await` will throw), which is fine — but callers that previously relied on the returned promise value or caught errors from the old `return api(...)` signature may now silently miss rejections. Consider propagating the error explicitly or documenting the new void return contract.

frontend/src/stores/tasks.ts
```diff
@@ -7,6 +7,8 @@ import type { LogEntry, Task, TaskState, WSInbound } from '../types'
 // immediately on long horizons (e.g. a 3-day finalize window).
 const MAX_TIMEOUT_MS = 2_147_483_647
 
+const _TERMINAL_STATES = new Set<string>(['done', 'failed', 'cancelled'])
```
jmelloy: Leading underscore (`_TERMINAL_STATES`) implies a private/unused variable in most JS/TS linters and is unconventional for a module-level constant. Rename to `TERMINAL_STATES` to match TypeScript idioms and avoid accidental lint warnings.

frontend/src/stores/tasks.ts
```diff
@@ -109,7 +111,12 @@ export const useTasksStore = defineStore('tasks', () => {
   }
 
   async function cancelTask(guildId: string, taskId: string) {
-    return api(`/guilds/${guildId}/tasks/${taskId}/cancel`, { method: 'POST' })
+    await api(`/guilds/${guildId}/tasks/${taskId}/cancel`, { method: 'POST' })
+    const task = tasks.value.find((t) => t.id === taskId)
+    if (task) {
```
jmelloy, line 117: `finished_at` is set to the client's local time at the moment of the API call. If the server later pushes a WS event with its own authoritative `finished_at`, the two values will diverge. If no WS event overrides `finished_at`, the value could be several seconds earlier than what the server recorded. Consider leaving `finished_at` unset until the WS event confirms it, or overwriting it when the authoritative event arrives.

jmelloy (commented):
- Race condition guard (`_TERMINAL_STATES`) is well-motivated and correctly prevents `task-complete`/`task-followup-done` WS events from overriding a terminal state set by `cancelTask`.
- Optimistic cancellation in `cancelTask` is a good UX improvement, but the function now silently swallows errors if the API call throws — previously it returned the promise and callers could catch rejections.
- The `_TERMINAL_STATES` constant uses a leading underscore (Python private convention) which is non-idiomatic in TypeScript/JS; it should be named `TERMINAL_STATES` or similar.
- The `finished_at` timestamp is set client-side on optimistic update, which may diverge from the server-authoritative value delivered later via WebSocket; this could cause subtle display inconsistencies if the WS event carries its own timestamp.
- New tests are well-structured and cover the intended scenarios; the existing test suite passing at 76/76 gives good confidence.

frontend/src/stores/tasks.ts
jmelloy, line 112: If `api(...)` throws (network error, 4xx/5xx), the optimistic state update is never reached — which is correct — but there is no rollback if an error occurs *after* the await returns `ok`. More critically, the old implementation returned the API response (`return api(...)`); callers that relied on the return value will now receive `undefined`. Consider returning the result or documenting the changed contract.

frontend/src/stores/tasks.ts
```diff
@@ -109,7 +111,12 @@ export const useTasksStore = defineStore('tasks', () => {
   }
 
   async function cancelTask(guildId: string, taskId: string) {
-    return api(`/guilds/${guildId}/tasks/${taskId}/cancel`, { method: 'POST' })
+    await api(`/guilds/${guildId}/tasks/${taskId}/cancel`, { method: 'POST' })
+    const task = tasks.value.find((t) => t.id === taskId)
```
jmelloy, line 114: Setting `finished_at` client-side with `new Date().toISOString()` means the timestamp will diverge from the server-recorded value once the WS event arrives. If downstream code uses `finished_at` for ordering or display, a subsequent WS update should overwrite this; verify there is no guard elsewhere that would skip updating `finished_at` when the task is already in a terminal state.

frontend/src/stores/tasks.ts
jmelloy: The leading underscore prefix (`_TERMINAL_STATES`) implies "module-private" by Python convention but has no meaning in TypeScript/JS. Use `TERMINAL_STATES` (or export it) for idiomatic TypeScript, especially since the same set may be useful in other parts of the codebase (e.g., UI components deciding whether to render action buttons).

jmelloy (commented):
- Race condition guard correctly prevents `task-complete`/`task-followup-done` WS events from overriding terminal states; the `_TERMINAL_STATES` set is a clean, maintainable approach.
- Optimistic cancellation in `cancelTask` improves UX, but silently swallows the previous return value and does not roll back the optimistic state if the API call throws, leaving the UI in an inconsistent state.
- The `_TERMINAL_STATES` constant uses a leading underscore (Python-style private convention) which is unconventional in TypeScript; consider `TERMINAL_STATES` or exporting it if reused elsewhere.
- The CSS fix for `.dot-cancelled` is straightforward and correct; using `--color-text-dim` is consistent with the dim/inactive visual language.
- Test coverage for the new behavior is thorough, including the race condition and optimistic update scenarios.

frontend/src/stores/tasks.ts
jmelloy, line 112: The previous implementation returned the `api()` result (`return api(...)`), giving callers access to the response. This refactor changes the return type to `Promise<void>`, silently breaking any caller that relied on the returned promise value or response. Verify no existing callers consume the return value, or preserve it by capturing and re-returning it after the optimistic update.

frontend/src/stores/tasks.ts
```diff
@@ -109,7 +111,12 @@ export const useTasksStore = defineStore('tasks', () => {
   }
 
   async function cancelTask(guildId: string, taskId: string) {
```
jmelloy, line 113: If `api()` throws (e.g. network error), the code correctly skips the optimistic update. However, if `api()` resolves without throwing but with a non-OK response (depending on `api()`'s implementation), the optimistic update will still fire. Ensure `api()` always throws on non-2xx, or add an explicit success check here.

frontend/src/stores/tasks.ts
jmelloy, line 117: `new Date().toISOString()` produces a client-side timestamp that may differ from the server-authoritative `finished_at` that arrives later via WebSocket. When the WS event arrives and overwrites the task, this is harmless, but if the WS event is skipped (terminal-state guard), the client timestamp stays permanently. Consider leaving `finished_at` unset here and only populating it from the server response or the WS event.

jmelloy (commented):
- Race condition guard using `_TERMINAL_STATES` is a solid, targeted fix that prevents stale WebSocket messages from overriding a cancelled state; the set-based lookup is clean and efficient.
- Optimistic `cancelTask` update improves UX but silently swallows errors — if the API call throws, the local state will not be updated (correct), but if `api()` resolves with a non-2xx response that doesn't throw, the optimistic update would incorrectly fire.
- The `cancelTask` refactor changes the function's return type from `Promise<Response>` (the previous `return api(...)`) to `Promise<void>`, which could silently break any callers that consumed the returned value.
- New tests are well-structured and cover the intended scenarios; the optimistic-update test correctly validates both `state` and `finished_at`.
- The `.dot-cancelled` CSS addition is a straightforward, necessary fix with an appropriate dim colour token.

frontend/src/stores/tasks.ts
jmelloy, line 112: If `api()` throws after the call succeeds server-side but before the promise resolves (e.g. network error on response parsing), the optimistic state update won't happen. More importantly, if `api()` rejects, the optimistic update is correctly skipped — but consider whether callers expect the previous return value of the API response, since this now returns `void` implicitly.

frontend/src/stores/tasks.ts
```diff
@@ -109,7 +111,12 @@ export const useTasksStore = defineStore('tasks', () => {
...
-    return api(`/guilds/${guildId}/tasks/${taskId}/cancel`, { method: 'POST' })
+    await api(`/guilds/${guildId}/tasks/${taskId}/cancel`, { method: 'POST' })
```
jmelloy: There is no rollback of the optimistic `cancelled` state if a subsequent WebSocket message or re-fetch indicates the cancellation failed server-side. Consider reverting `task.state` to its previous value in a catch block, e.g. wrapping the update in try/catch and storing the old state before mutating.

jmelloy (commented):
- Race condition guard is correctly implemented using a `_TERMINAL_STATES` set to prevent WebSocket handlers from overriding terminal task states, which is a clean and maintainable approach.
- Optimistic cancellation in `cancelTask` improves UX but silently swallows API errors — if the `api()` call throws, the catch must propagate to the caller; the current `await` without try/catch means a rejection will bubble up naturally, which is acceptable, but there's no rollback of optimistic state on failure.
- The `cancelTask` function changed from returning the API response to returning `void`, which is a subtle breaking change for any caller that relied on the return value.
- New tests adequately cover the race condition guard and optimistic update, including a check for `finished_at` being defined, though the tests don't verify the `finished_at` value is a valid ISO string or that rollback occurs on API failure.
- CSS fix for `.dot-cancelled` is minimal and correct, resolving the invisible dot issue.

frontend/src/stores/tasks.ts
```diff
@@ -7,6 +7,8 @@ import type { LogEntry, Task, TaskState, WSInbound } from '../types'
 // immediately on long horizons (e.g. a 3-day finalize window).
 const MAX_TIMEOUT_MS = 2_147_483_647
```
jmelloy, line 8: The leading underscore convention (`_TERMINAL_STATES`) typically signals a private/internal variable in JS, but since this is already module-scoped and not exported, the underscore prefix is unnecessary and slightly unconventional for a module-level constant — consider naming it `TERMINAL_STATES` for consistency with the other module-level constant `STATE_LABELS`.

PR #396, opened 2026-05-20
fix(liveness): evict zombie workers whose agents are already offline (#394)

### LINKED ISSUE #394: Worker registry doesn't detect dead containers — zombie online state
## Problem

When a worker container stops running (crash, OOM, manual kill), the worker registry continues to show the worker as `online` indefinitely. Shutdown signals and ping messages appear to be "delivered" (accepted by the queue) but no process is consuming them.

**Observed:**
- `w-hzrxet` shows `state: online`, `agent_count: 0` after two `shutdown_worker` signals
- Container is confirmed not running
- Messages report "delivered" but worker never responds or exits

## Expected behaviour

Workers should emit a periodic **heartbeat** to the registry. If no heartbeat is received within a configurable TTL (e.g. 30–60s), the registry should automatically transition the worker to `offline`.

## Suggested fix

1. Worker process emits a heartbeat (e.g. every 15s) to the registry
2. Registry has a watchdog that marks workers `offline` if heartbeat TTL expires
3. Shutdown signal handler should also force an immediate `offline` transition in the registry, even if the process can't ack it

## Impact

- Idle task routing may attempt to assign work to dead workers
- Operator has no reliable way to distinguish a healthy idle worker from a zombie

jmelloy:
## Root cause

Two independent bugs caused dead worker containers (like `w-hzrxet`) to stay `online` in the registry indefinitely, even after the WebSocket connection dropped and two `shutdown_worker` signals were delivered.

### Bug 1 — `websocket.py` finally-block used wrong ID for Worker update

`backend/routes/websocket.py` lines 173–176 (before fix):

```python
# Mirror into workers table so foreman sees the worker as offline.
await db.execute(
    update(Worker)
    .where(Worker.id == agent_id, Worker.guild_pk == _guild_pk)  # WRONG: agent_id ≠ worker_id
    .values(state="offline")
)
```

Worker IDs are `w-…`; agent IDs are `a-…`. The `WHERE Worker.id == agent_id` clause never matched any row, so the Worker stayed `online` even after an abrupt disconnect correctly marked all its agents `offline`.

**Fix:** Pre-query `Agent.worker_id` for all stale agents before the update loop, then use those real worker IDs to mark the Worker rows offline.

### Bug 2 — Sweeper only cascaded Worker state from non-offline agents

`_sweep_stale_workers_once()` in `backend/main.py` only found zombie workers by cascading from _non-offline_ stale agents. After Bug 1, all agents were already `offline` (the agent update in the finally block was correct), so the sweeper found zero stale agents and returned `0` — never touching the orphaned Worker row.

**Fix:** Add a second query that sweeps the `workers` table directly by `last_seen`, catching workers stuck in `online` with no active agents.

## Changes

| File | Change |
|------|--------|
| `backend/routes/websocket.py` | Pre-query real `worker_id`s from Agent rows before the finally-block cleanup loop |
| `backend/main.py` | Add direct zombie-worker sweep by `Worker.last_seen` in `_sweep_stale_workers_once()` |
| `backend/tests/test_disconnect.py` | `test_abrupt_disconnect_marks_worker_offline` — regression test for Bug 1 |
| `backend/tests/test_liveness.py` | `test_sweeper_marks_zombie_worker_offline_when_agents_already_offline` — regression test for Bug 2 |

## Test plan

- [x] All 408 backend tests pass (`python -m pytest`)
- [x] New `test_abrupt_disconnect_marks_worker_offline` confirms that closing a WS without `worker-disconnect` marks both agent AND worker offline
- [x] New `test_sweeper_marks_zombie_worker_offline_when_agents_already_offline` confirms the sweeper catches workers with stale `last_seen` even when all their agents are already `offline`
- [x] Existing reconnect race-condition test still passes (no regression on `test_reconnect_does_not_get_clobbered_by_old_finally`)

Closes #394

### DIFF
--- a/backend/main.py
+++ b/backend/main.py
@@ -94,9 +94,10 @@ async def _sweep_stale_workers_once() -> int:
                 .where(Agent.last_seen < cutoff)
             )
         ).all()
-        if not stale_agents:
-            return 0
         stale_worker_keys: set[tuple[str, int]] = set()
+        # Track which workers were already caught via the agent-cascade path so
+        # the zombie-worker log below can skip them (they're already logged).
+        agent_cascade_wids: set[str] = set()
         for row in stale_agents:
             await db.execute(
                 update(Agent)
@@ -105,14 +106,35 @@ async def _sweep_stale_workers_once() -> int:
             )
             if row.worker_id:
                 stale_worker_keys.add((row.worker_id, row.guild_pk))
+                agent_cascade_wids.add(row.worker_id)
             agent_owners.pop(row.id, None)
+
+        # Also sweep zombie workers directly.  A worker can be stuck in
+        # "online" with no non-offline agents when the WS close handler had
+        # the agent_id/worker_id bug and failed to update the Worker row.
+        # The agent cascade above only runs for non-offline stale agents;
+        # a worker whose agents are already "offline" (but whose own row was
+        # never flipped) would be missed without this direct pass.
+        zombie_workers = (
+            await db.execute(
+                select(Worker.id, Worker.guild_pk, Guild.guild_id)
+                .join(Guild, Guild.id == Worker.guild_pk)
+                .where(Worker.state != "offline")
+                .where(Worker.last_seen.isnot(None))
+                .where(Worker.last_seen < cutoff)
+            )
+        ).all()
+        for row in zombie_workers:
+            stale_worker_keys.add((row.id, row.guild_pk))
+
         for worker_id, gpk in stale_worker_keys:
             await db.execute(
                 update(Worker)
                 .where(Worker.id == worker_id, Worker.guild_pk == gpk)
                 .values(state="offline")
             )
-        await db.commit()
+        if stale_agents or zombie_workers:
+            await db.commit()
     for row in stale_agents:
         logger.warning(
             "Marking %s offline: no ping in over %.0fs (guild=%s)",
@@ -124,6 +146,16 @@ async def _sweep_stale_workers_once() -> int:
             row.guild_id,
             {"type": "agent-state", "agentId": row.id, "state": "offline"},
         )
+    # Log zombie workers not already covered by the per-agent log above.
+    for row in zombie_workers:
+        if row.id not in agent_cascade_wids:
+            logger.warning(
+                "Marking zombie worker %s offline: no ping in over %.0fs,"
+                " no active agents (guild=%s)",
+                row.id,
+                WORKER_OFFLINE_AFTER_SECONDS,
+                row.guild_id,
+            )
     return len(stale_agents)
 
 
--- a/backend/main.py
+++ b/backend/main.py
@@ -84,6 +84,9 @@ async def _sweep_stale_workers_once() -> int:
     """One pass of the stale-worker sweep. Returns the number of agents
     marked offline. Extracted for direct testing."""
     cutoff = (datetime.now(UTC) - timedelta(seconds=WORKER_OFFLINE_AFTER_SECONDS)).isoformat()
+    stale_agents: list = []
+    zombie_workers: list = []
+    agent_cascade_wids: set[str] = set()
     async with AsyncSessionLocal() as db:
         stale_agents = (
             await db.execute(
@@ -95,9 +98,6 @@ async def _sweep_stale_workers_once() -> int:
             )
         ).all()
         stale_worker_keys: set[tuple[str, int]] = set()
-        # Track which workers were already caught via the agent-cascade path so
-        # the zombie-worker log below can skip them (they're already logged).
-        agent_cascade_wids: set[str] = set()
         for row in stale_agents:
             await db.execute(
                 update(Agent)
--- a/backend/main.py
+++ b/backend/main.py
@@ -81,13 +81,35 @@ async def reset_connection_state() -> None:
 
 
 async def _sweep_stale_workers_once() -> int:
-    """One pass of the stale-worker sweep. Returns the number of agents
-    marked offline. Extracted for direct testing."""
+    """One pass of the stale-worker sweep. Returns the total number of agents
+    and zombie workers marked offline. Extracted for direct testing."""
     cutoff = (datetime.now(UTC) - timedelta(seconds=WORKER_OFFLINE_AFTER_SECONDS)).isoformat()
     stale_agents: list = []
     zombie_workers: list = []
     agent_cascade_wids: set[str] = set()
     async with AsyncSessionLocal() as db:
+        # Zombie-worker pass runs first so the agent-state guard sees pre-update
+        # states.  A zombie is a worker stuck "online" with a stale last_seen
+        # whose agents are ALL already offline (the WS close handler marked them
+        # offline but failed to flip the Worker row due to the agent_id/worker_id
+        # bug).  Workers that still have any non-offline agent are skipped here
+        # and handled via the agent cascade below.
+        zombie_workers = (
+            await db.execute(
+                select(Worker.id, Worker.guild_pk, Guild.guild_id)
+                .join(Guild, Guild.id == Worker.guild_pk)
+                .where(Worker.state != "offline")
+                .where(Worker.last_seen.isnot(None))
+                .where(Worker.last_seen < cutoff)
+                .where(
+                    ~select(Agent.id)
+                    .where(Agent.worker_id == Worker.id)
+                    .where(Agent.state != "offline")
+                    .exists()
+                )
+            )
+        ).all()
+
         stale_agents = (
             await db.execute(
                 select(Agent.id, Agent.guild_pk, Agent.worker_id, Guild.guild_id)
@@ -109,21 +131,6 @@ async def _sweep_stale_workers_once() -> int:
                 agent_cascade_wids.add(row.worker_id)
             agent_owners.pop(row.id, None)
 
-        # Also sweep zombie workers directly.  A worker can be stuck in
-        # "online" with no non-offline agents when the WS close handler had
-        # the agent_id/worker_id bug and failed to update the Worker row.
-        # The agent cascade above only runs for non-offline stale agents;
-        # a worker whose agents are already "offline" (but whose own row was
-        # never flipped) would be missed without this direct pass.
-        zombie_workers = (
-            await db.execute(
-                select(Worker.id, Worker.guild_pk, Guild.guild_id)
-                .join(Guild, Guild.id == Worker.guild_pk)
-                .where(Worker.state != "offline")
-                .where(Worker.last_seen.isnot(None))
-                .where(Worker.last_seen < cutoff)
-            )
-        ).all()
         for row in zombie_workers:
             stale_worker_keys.add((row.id, row.guild_pk))
 
@@ -156,7 +163,7 @@ async def _sweep_stale_workers_once() -> int:
                 WORKER_OFFLINE_AFTER_SECONDS,
                 row.guild_id,
             )
-    return len(stale_agents)
+    return len(stale_agents) + len(zombie_workers)
 
 
 async def _stale_worker_sweeper() -> None:
--- a/backend/routes/websocket.py
+++ b/backend/routes/websocket.py
@@ -162,6 +162,23 @@ async def websocket_endpoint(websocket: WebSocket, guild_id: str):
                     stale_agents = [
                         aid for aid in joined_agents if agent_owners.get(aid) is websocket
                     ]
+                    # Look up the worker_ids for all stale agents *before* we
+                    # modify their rows, so we can update the Worker table with
+                    # the correct IDs (worker IDs are "w-…"; agent IDs are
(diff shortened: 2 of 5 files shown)


backend/main.py
```diff
@@ -124,6 +146,16 @@ async def _sweep_stale_workers_once() -> int:
             row.guild_id,
             {"type": "agent-state", "agentId": row.id, "state": "offline"},
         )
+    # Log zombie workers not already covered by the per-agent log above.
+    for row in zombie_workers:
+        if row.id not in agent_cascade_wids:
+            logger.warning(
+                "Marking zombie worker %s offline: no ping in over %.0fs,"
+                " no active agents (guild=%s)",
+                row.id,
```
jmelloy, line 162: `zombie_workers` is defined inside the `async with db` block but consumed in the logging loop outside it (lines 146-155). If an exception is raised before the `zombie_workers` assignment (e.g., during the stale-agent loop), the outer `for row in zombie_workers` will raise `NameError`. Initialize `zombie_workers = []` before the `async with` block to make this safe.

backend/main.py
```diff
@@ -105,14 +106,35 @@ async def _sweep_stale_workers_once() -> int:
...
+
+        # Also sweep zombie workers directly.  A worker can be stuck in
+        # "online" with no non-offline agents when the WS close handler had
+        # the agent_id/worker_id bug and failed to update the Worker row.
+        # The agent cascade above only runs for non-offline stale agents;
+        # a worker whose agents are already "offline" (but whose own row was
+        # never flipped) would be missed without this direct pass.
+        zombie_workers = (
+            await db.execute(
+                select(Worker.id, Worker.guild_pk, Guild.guild_id)
```
jmelloy: The zombie-worker query joins `Guild` only to fetch `guild_id` for logging, but the `WHERE` clause does not filter by guild or any index-friendly column beyond `Worker.state` and `Worker.last_seen`. Consider adding `Worker.last_seen.isnot(None)` is already present, but ensure `last_seen` is indexed; without an index this becomes a full table scan on every sweep interval.

backend/tests/test_liveness.py
```diff
@@ -298,6 +298,68 @@ def test_migration_backfills_last_seen_for_existing_rows(tmp_path, monkeypatch):
...
+            (old, worker_id),
+        )
+        conn.commit()
+
+    with test_client.websocket_connect(f"/ws/{guild_id}"):
+        monkeypatch.setattr(main_module, "WORKER_OFFLINE_AFTER_SECONDS", 1.0)
+
+        async def _drive() -> int:
+            return await main_module._sweep_stale_workers_once()
+
```
jmelloy, line 341: `asyncio.run(_drive())` creates a brand-new event loop inside a `with test_client.websocket_connect(...)` block. If the test client (Starlette's `TestClient`) uses its own thread-bound event loop, sharing SQLite state across loops can cause 'database is locked' or connection-already-closed errors. Consider running the sweep outside the websocket context manager, or use the existing async test infrastructure if available.

jmelloy (commented):
- Fixes two real bugs: wrong ID type used in Worker update (`agent_id` vs `worker_id`) in the WebSocket finally-block, and the sweeper's blind spot for zombie workers whose agents were already offline.
- The zombie-worker query in the sweeper correctly uses `Worker.last_seen < cutoff` as the staleness signal, avoiding false positives for freshly-started workers with no agents yet.
- `_sweep_stale_workers_once` now references `zombie_workers` after the `async with` block ends; if `stale_agents` is empty the `zombie_workers` variable is set inside the block, but the logging loop outside the block also references it — this works only if the variable is always assigned before the outer loop, which is true here, but it's fragile.
- The new test `test_sweeper_marks_zombie_worker_offline_when_agents_already_offline` calls `asyncio.run(_drive())` inside a `with test_client.websocket_connect(...)` context, meaning a new event loop is created while the test client's event loop may still be live; this can cause subtle cross-loop issues depending on the test client implementation.
- Test coverage is good: both the WebSocket regression and the sweeper zombie path are tested independently with clear failure messages.

jmelloy (commented):
- Fixes two real bugs: wrong ID type used in WebSocket finally-block Worker update (`agent_id` vs `worker_id`), and sweeper missing zombie workers whose agents were already offline — both well-diagnosed and correctly addressed.
- The zombie-worker sweeper query relies solely on `Worker.last_seen < cutoff` with no cross-check against agent state, meaning a legitimately online worker that simply hasn't sent a heartbeat recently could be incorrectly evicted.
- `_sweep_stale_workers_once` now returns `len(stale_agents)` unchanged even when zombie workers are swept; callers relying on the return value for monitoring/alerting will silently under-count evictions.
- The new sweeper test uses `asyncio.run()` inside a synchronous test while a live WebSocket connection is open via the `with test_client.websocket_connect` block, which may interact poorly with the event loop depending on the test framework's async handling.
- Test coverage is solid: both a regression test for the WebSocket fix and a regression test for the sweeper fix are included, with clear failure messages pointing back to the root cause.

backend/main.py
```diff
@@ -105,14 +106,35 @@ async def _sweep_stale_workers_once() -> int:
             )
             if row.worker_id:
                 stale_worker_keys.add((row.worker_id, row.guild_pk))
+                agent_cascade_wids.add(row.worker_id)
             agent_owners.pop(row.id, None)
+
+        # Also sweep zombie workers directly.  A worker can be stuck in
+        # "online" with no non-offline agents when the WS close handler had
```
jmelloy: The zombie-worker query has no guard that the worker's agents are all offline — it evicts any online worker with a stale `last_seen`, including one that is genuinely active but simply missed a heartbeat. Consider adding `.where(~Worker.agents.any(Agent.state != 'offline'))` or an equivalent subquery to avoid false positives.

backend/main.py
```diff
@@ -124,6 +146,16 @@ async def _sweep_stale_workers_once() -> int:
...
+    for row in zombie_workers:
+        if row.id not in agent_cascade_wids:
+            logger.warning(
+                "Marking zombie worker %s offline: no ping in over %.0fs,"
+                " no active agents (guild=%s)",
+                row.id,
+                WORKER_OFFLINE_AFTER_SECONDS,
+                row.guild_id,
+            )
     return len(stale_agents)
```
jmelloy, line 167: `_sweep_stale_workers_once` still returns `len(stale_agents)`, so swept zombie workers (where `stale_agents` is empty) contribute 0 to the return value. If callers use this count for metrics or loop-termination logic, zombie evictions will be invisible. Consider returning `len(stale_agents) + len(zombie_workers)` or a named tuple.

backend/tests/test_liveness.py
jmelloy, line 341: `asyncio.run(_drive())` creates a brand-new event loop while the `with test_client.websocket_connect(...)` context manager may already have an active loop open underneath. This can cause 'Event loop is already running' errors or silently use a different DB session than the one the test_client expects. Move the sweeper call outside the websocket context block, or use the test framework's async utilities instead.

backend/main.py
jmelloy, line 167: The function still returns `len(stale_agents)` only. Zombie workers evicted by the new direct-sweep path are not counted. If callers (e.g. tests or monitoring) use this return value to decide whether anything was swept, they will get a misleading 0 even when zombie workers were marked offline. Consider returning `len(stale_agents) + len(zombie_workers)` or a richer result type.

backend/main.py
```diff
@@ -105,14 +106,35 @@ async def _sweep_stale_workers_once() -> int:
...
+        # Also sweep zombie workers directly.  A worker can be stuck in
+        # "online" with no non-offline agents when the WS close handler had
+        # the agent_id/worker_id bug and failed to update the Worker row.
+        # The agent cascade above only runs for non-offline stale agents;
+        # a worker whose agents are already "offline" (but whose own row was
+        # never flipped) would be missed without this direct pass.
+        zombie_workers = (
+            await db.execute(
+                select(Worker.id, Worker.guild_pk, Guild.guild_id)
+                .join(Guild, Guild.id == Worker.guild_pk)
```
jmelloy: The `.where(Worker.last_seen.isnot(None))` guard silently excludes a worker that went `online` but never sent a heartbeat (NULL `last_seen`). Such a worker would be a zombie too. This may be intentional to avoid false positives, but deserves a comment explaining the design decision, or the guard should be relaxed (e.g. also sweep online workers where `last_seen IS NULL AND created_at < cutoff`).

backend/routes/websocket.py
```diff
@@ -162,17 +162,35 @@ async def websocket_endpoint(websocket: WebSocket, guild_id: str):
...
-                        # Mirror into workers table so foreman sees the worker as offline.
+                    # Mirror into workers table using the real worker IDs.
+                    for wid in stale_worker_ids:
                         await db.execute(
                             update(Worker)
-                            .where(Worker.id == agent_id, Worker.guild_pk == _guild_pk)
+                            .where(Worker.id == wid, Worker.guild_pk == _guild_pk)
                             .values(state="offline")
                         )
                     if stale_agents:
```
jmelloy, line 196: Each `wid` in `stale_worker_ids` is updated with `Worker.guild_pk == _guild_pk`. Since `worker_id` is globally unique this is fine, but if a worker can theoretically be shared across guilds the cross-guild filter would miss it. More critically, if `stale_worker_ids` contains multiple IDs this issues N individual UPDATE statements; batching with `Worker.id.in_(stale_worker_ids)` would be safer and more efficient.

jmelloy (commented):
- Fixes two real bugs: wrong ID type used in WebSocket finally-block (`agent_id` vs `worker_id`) and sweeper missing zombie workers whose agents were already offline — both root causes are clearly identified and well-documented.
- The zombie-worker sweeper query is a good safety net, but it introduces an implicit coupling: it relies on `Worker.last_seen` being updated by heartbeats; if a worker never sends heartbeats after going online its `last_seen` could be NULL and the `isnot(None)` guard would silently exclude it from sweeping.
- The `_sweep_stale_workers_once` return value still only reflects `len(stale_agents)`, not the total number of rows actually marked offline (zombie workers are not counted), which could mislead callers or metrics that use the return value to gauge sweep activity.
- Test coverage is solid: both regression scenarios have dedicated tests with clear failure messages, and the abrupt-disconnect test correctly simulates a container crash without a graceful shutdown.
- Minor: the `stale_agents` early-return guard was removed but the `db.commit()` is now conditional on `stale_agents or zombie_workers`; this is correct but the commit is now further from the writes, making the control flow slightly harder to audit at a glance.

backend/main.py
jmelloy, line 167: The function still returns `len(stale_agents)` at the end. Zombie workers evicted via the new direct-pass are not counted, so any caller using the return value to measure total work (e.g., for metrics or adaptive sleep) will undercount. Consider returning `len(stale_agents) + len(zombie_workers)` or a named tuple.

backend/main.py
jmelloy: The zombie-worker query joins `Guild` but does not filter `Guild.deleted_at IS NULL` (assuming soft-delete exists on guilds). If deleted guilds still have worker rows, those could be spuriously touched. Worth adding the filter for consistency with how the agent query likely scopes guilds.

backend/routes/websocket.py
```diff
@@ -162,17 +162,35 @@ async def websocket_endpoint(websocket: WebSocket, guild_id: str):
...
+                                    Agent.worker_id.isnot(None),
+                                )
+                            )
+                        ).all()
+                        stale_worker_ids = {r[0] for r in wid_rows}
                     for agent_id in stale_agents:
                         agent_owners.pop(agent_id, None)
                         await db.execute(
                             update(Agent)
                             .where(Agent.id == agent_id, Agent.guild_pk == _guild_pk)
```
jmelloy, line 186: The new worker-update loop uses `Worker.guild_pk == _guild_pk` which is correct, but there is no `last_seen` or state guard — it will set `state='offline'` on a worker even if it has since reconnected on another WebSocket in a race. Consider adding `.where(Worker.state != 'offline')` to make the update idempotent and safe under concurrent reconnects.

jmelloy (commented):
- Fixes two real bugs: wrong ID used in WebSocket finally-block Worker update (`agent_id` vs `worker_id`), and the sweeper missing zombie workers whose agents were already offline before the sweep ran.
- The websocket.py fix is correct and well-targeted; pre-fetching worker IDs before mutating agent rows avoids the stale-read problem.
- The sweeper's zombie-worker query uses `Worker.last_seen < cutoff` with `last_seen IS NOT NULL`, which is the right guard, but the query does not filter by `Worker.guild_pk` matching a valid/non-deleted guild — could touch orphaned rows, though low risk.
- `_sweep_stale_workers_once` now returns `len(stale_agents)` unchanged; callers that rely on the return value to count all work done will miss zombie-worker evictions, which may matter for metrics or test assertions.
- Test coverage is solid: both regression paths (abrupt disconnect and sweeper zombie pass) are exercised with clear failure messages.

PR #397, opened 2026-05-20
Fix Workers tab: click should activate detail tab (#395)

### LINKED ISSUE #395: Workers tab: clicking a worker/task doesn't activate the detail tab
## Bug

In the **Workers** tab, clicking a worker or task opens the detail panel but does **not** switch the active tab to it. The tab opens but is not selected/focused.

In the **Tasks** tab, the same click correctly jumps to and selects the new tab.

## Expected behaviour

Clicking a worker or task in the Workers tab should open **and** activate (select) the corresponding detail tab, consistent with the Tasks tab behaviour.

## Steps to reproduce

1. Go to the Workers tab
2. Click on any worker or task in the list
3. The detail panel/tab opens but the active tab indicator does not move to it

## Notes

The Tasks tab already handles this correctly — the Workers tab needs the same `setActiveTab` (or equivalent) call wired up on click.

jmelloy:
## Summary

- Clicking the task button (📋) on a worker row in the Workers tab called `tasksStore.selectTask()` but never set `agentsStore.activeTab`, so the task tab would appear in the tab bar but remain unselected.
- Added `agentsStore.activeTab = 'task-' + task.id` immediately after `selectTask()` in `WorkerList.vue`'s `openAgentTask` handler, matching the same activation pattern used by the Tasks tab's watcher in `MainView.vue`.
- Worker row clicks already worked correctly via `agentsStore.selectWorker()` which sets `activeTab` directly.

## Test plan

- [ ] Open the Workers tab in the sidebar
- [ ] Click the 📋 task button on a worker that has an active task — the task tab should open **and** become the selected (highlighted) tab
- [ ] Verify the Tasks tab still works correctly (click a task → tab opens and is selected)
- [ ] Verify clicking a worker row still opens and selects the worker detail tab

Closes #395

### DIFF
--- a/frontend/src/components/sidebar/WorkerList.vue
+++ b/frontend/src/components/sidebar/WorkerList.vue
@@ -75,7 +75,10 @@ function currentTaskForWorker(workerId: string) {
 
 function openAgentTask(workerId: string) {
   const task = currentTaskForWorker(workerId)
-  if (task) tasksStore.selectTask(task.id)
+  if (task) {
+    tasksStore.selectTask(task.id)
+    agentsStore.activeTab = 'task-' + task.id
+  }
 }
 </script>
 
--- a/frontend/src/components/sidebar/WorkerList.vue
+++ b/frontend/src/components/sidebar/WorkerList.vue
@@ -77,7 +77,7 @@ function openAgentTask(workerId: string) {
   const task = currentTaskForWorker(workerId)
   if (task) {
     tasksStore.selectTask(task.id)
-    agentsStore.activeTab = 'task-' + task.id
+    agentsStore.openTaskTab(task.id)
   }
 }
 </script>
--- a/frontend/src/components/MainView.vue
+++ b/frontend/src/components/MainView.vue
@@ -36,9 +36,9 @@
       <!-- Task tabs — only shown when explicitly opened -->
       <button
         v-for="task in visibleTaskTabs"
-        :key="'task-' + task.id"
+        :key="taskTabId(task.id)"
         class="tab task-tab"
-        :class="{ active: agentsStore.activeTab === 'task-' + task.id }"
+        :class="{ active: agentsStore.activeTab === taskTabId(task.id) }"
         @click="onTabClick($event, task.id)"
       >
         <span class="task-dot" :class="'task-dot-' + task.state.replace(/[^a-z]/g, '-')"></span>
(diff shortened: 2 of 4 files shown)


jmelloy (commented):
- The fix correctly addresses the reported bug: `openAgentTask` was calling `tasksStore.selectTask()` but never setting `agentsStore.activeTab`, so the tab was added to the bar but not activated.
- The solution directly mutates `agentsStore.activeTab` which matches the pattern used elsewhere (e.g., `MainView.vue`'s watcher), though it introduces a subtle coupling — ideally `selectTask` itself or a dedicated store action would handle tab activation atomically.
- No guard is added for the case where `agentsStore.activeTab` is already set to the same tab, but this is likely harmless as re-assigning the same string is idempotent.
- The change is minimal and surgical, touching only the one affected code path without risk of regressions in the Tasks tab or worker row click flows.
- A minor concern: the tab ID string `'task-' + task.id` is constructed inline again, duplicating the format string that presumably also exists in `tasksStore.selectTask` or elsewhere — a shared constant or helper would reduce the risk of drift.

frontend/src/components/sidebar/WorkerList.vue
```diff
@@ -75,7 +75,10 @@ function currentTaskForWorker(workerId: string) {
 
 function openAgentTask(workerId: string) {
   const task = currentTaskForWorker(workerId)
-  if (task) tasksStore.selectTask(task.id)
+  if (task) {
+    tasksStore.selectTask(task.id)
```
jmelloy, line 79: The tab ID format `'task-' + task.id` is duplicated inline here. If the format ever changes in the store or elsewhere, this will silently diverge. Consider extracting a shared helper (e.g., `taskTabId(id)`) or having `tasksStore.selectTask()` return/expose the canonical tab ID so callers don't need to reconstruct it.

frontend/src/components/sidebar/WorkerList.vue
```diff
@@ -75,7 +75,10 @@ function currentTaskForWorker(workerId: string) {
 
 function openAgentTask(workerId: string) {
   const task = currentTaskForWorker(workerId)
-  if (task) tasksStore.selectTask(task.id)
+  if (task) {
```
jmelloy, line 78: Directly mutating `agentsStore.activeTab` from a component bypasses encapsulation. A store action like `agentsStore.openTaskTab(task.id)` would keep this logic in one place and make it easier to enforce the correct tab ID format across all callers.

PR #398, opened 2026-05-20
Fix docker-compose.yml pioneer-foreman volumes syntax

jmelloy:
## Summary
- The `pioneer-foreman` service had a `volumes:` key with only a commented-out entry, making it a `null` scalar instead of a YAML list
- Docker Compose validation rejects this with: `services.pioneer-foreman.volumes must be a list`
- Fixed by removing the empty `volumes:` key and converting the optional-mount hint into a regular comment above `depends_on:`

## Test plan
- [ ] Run `docker compose config` to confirm no validation errors
- [ ] Confirm `docker compose --profile pioneer-foreman up pioneer-foreman` still works as expected

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -75,9 +75,9 @@ services:
       GITHUB_TOKEN: ${GITHUB_TOKEN:-}
       FOREMAN_MODEL: ${FOREMAN_MODEL:-claude-sonnet-4-6}
       LOG_LEVEL: ${LOG_LEVEL:-INFO}
-    volumes:
-      # Optional: mount a pioneer-foreman.toml for full config control.
-      # - ./pioneer-foreman/pioneer-foreman.toml:/config/pioneer-foreman.toml:ro
+    # Optional: mount a pioneer-foreman.toml for full config control:
+    #   volumes:
+    #     - ./pioneer-foreman/pioneer-foreman.toml:/config/pioneer-foreman.toml:ro
     depends_on:
       - backend
     restart: unless-stopped


docker-compose.yml
```diff
@@ -75,9 +75,9 @@ services:
       GITHUB_TOKEN: ${GITHUB_TOKEN:-}
       FOREMAN_MODEL: ${FOREMAN_MODEL:-claude-sonnet-4-6}
       LOG_LEVEL: ${LOG_LEVEL:-INFO}
-    volumes:
-      # Optional: mount a pioneer-foreman.toml for full config control.
-      # - ./pioneer-foreman/pioneer-foreman.toml:/config/pioneer-foreman.toml:ro
+    # Optional: mount a pioneer-foreman.toml for full config control:
```
jmelloy, line 78: The indentation of `volumes:` and the mount line in the comment (2-space indent relative to `#`) is slightly inconsistent with typical Docker Compose style where `volumes:` would be at the service-key level. Consider aligning the comment block to match the indentation a real user would paste in (i.e., `#   volumes:` with 4 spaces, matching other service-level keys like `depends_on:`), so copy-pasting it directly works without adjustment.

jmelloy (commented):
- Fixes a legitimate YAML/Docker Compose validation bug where an empty `volumes:` key with only commented-out entries results in a `null` scalar rather than a list, which Docker Compose correctly rejects.
- The chosen fix (removing the `volumes:` key entirely and restructuring the comment to show the full snippet to uncomment) is the correct approach and preserves the helpful hint for users who want optional volume mounts.
- The comment format in the new version clearly shows the exact YAML block a user would need to add, making it easier to act on than the previous single-line hint.
- No functional behavior changes are introduced; the service configuration remains identical at runtime.
- The test plan is appropriate and sufficient for validating this type of change.

PR #399, opened 2026-05-20
Webhook event queue: debounce per-PR timer

### LINKED ISSUE #389: Webhook event queue: reset delay counter on each new event to prevent premature delivery
## Problem

When multiple GitHub webhook events arrive in quick succession (e.g. several check_run completions + a review comment), the foreman can receive and act on an earlier event before all related events have been delivered. The delay counter should be reset each time a new event arrives in the same burst, so the foreman only processes the batch once the stream has gone quiet.

## Expected behaviour

- On each incoming webhook event, reset a debounce timer (e.g. 2–5 seconds).
- Only deliver / surface the event(s) to the foreman once no new events have arrived within that window.
- This mirrors a standard debounce pattern and prevents the foreman from reacting to a half-delivered burst.

## Possible approach

In the webhook ingestion layer, maintain a per-PR (or per-task) timer. Each new event for the same PR resets the timer. Fire the foreman callback only on timer expiry.

jmelloy:
## Summary

- Adds a per-task (per-PR) debounce timer in `backend/routes/webhooks.py` with a 3-second window (`DEBOUNCE_WINDOW_SECONDS`).
- Each new dispatchable webhook event for the same PR cancels the previous timer and restarts the window. Only when no further events arrive within the window does the foreman get invoked.
- All summaries buffered during the window are joined with `---` separators and delivered together, so the foreman sees the complete picture of what happened.
- Updates two existing dispatch tests to patch `_schedule_debounced_foreman` (the new entry point) instead of `spawn`/`reset_foreman_poll` directly.
- Adds `TestDebounce` with four async tests: rapid-events coalescing → single delivery; slow events → separate deliveries; timer reset cancels prior timer; independent PR keys have independent timers.

## Test plan

- [x] All 412 backend tests pass (`python -m pytest`)
- [x] `ruff check . && ruff format .` — no issues
- [x] `TestDebounce::test_rapid_events_single_foreman_call` — 3 events → 1 foreman call with combined summary
- [x] `TestDebounce::test_slow_events_separate_foreman_calls` — events > window apart → 2 foreman calls
- [x] `TestDebounce::test_reset_cancels_previous_timer` — 2nd event before timer fires → combined in single delivery
- [x] `TestDebounce::test_independent_prs_not_merged` — events on different PR keys don't coalesce

Closes #389

### DIFF
--- a/backend/routes/webhooks.py
+++ b/backend/routes/webhooks.py
@@ -65,6 +65,19 @@ def _debounce_key(guild_id: str, task_id: str) -> str:
     return f"{guild_id}:{task_id}"
 
 
+def clear_debounce_state() -> None:
+    """Cancel all in-flight debounce timers and clear accumulated buffers.
+
+    Call this on server shutdown and in test teardown to prevent state from
+    bleeding between restarts or test cases.
+    """
+    for task in _debounce_tasks.values():
+        if not task.done():
+            task.cancel()
+    _debounce_tasks.clear()
+    _debounce_buffers.clear()
+
+
 async def _debounce_fire(key: str, guild_id: str) -> None:
     """Sleep for the debounce window then deliver all buffered events.
 
--- a/backend/routes/webhooks.py
+++ b/backend/routes/webhooks.py
@@ -122,39 +122,25 @@ class DebounceQueue:
         """Sleep for the debounce window then deliver all buffered events.
 
         ``gen`` is the generation counter captured when this timer was created.
-        schedule() bumps the generation *before* cancelling the old task, so if
-        the old task's sleep resolves during the cancel window and the task runs
-        during ``await gather()``, it will see a stale generation and bail out
-        without delivering — reliably distinguishing a timer-reset from an
-        external cancellation.
-
-        On CancelledError the buffer is always preserved so a subsequent
-        schedule() call can re-use it.  If we are still the registered timer
-        (external cancellation rather than a schedule() reset) we also remove
-        our stale reference from _tasks; on a schedule()-driven reset,
-        _tasks[key] was already popped synchronously before cancel so the
-        identity check returns False and the pop is correctly skipped.
+        schedule() bumps the generation *before* cancelling the old task, so
+        even if the old task's sleep resolves during the cancel window and the
+        task runs during ``await gather()``, it will see a stale generation
+        and bail out without delivering — eliminating the task-identity race.
         """
         try:
             await asyncio.sleep(self._window)
         except asyncio.CancelledError:
-            # Re-raise so the asyncio Task is properly marked cancelled.
-            # Clean up our _tasks entry only when we are still the registered
-            # timer (i.e. external cancellation — shutdown(), test teardown).
-            # On a schedule()-driven reset, _tasks[key] was already popped
-            # synchronously before cancel, so the check is False.
-            if self._tasks.get(key) is asyncio.current_task():
-                self._tasks.pop(key, None)
             raise
         # Stale-generation guard: if schedule() ran after our sleep started it
         # bumped the generation counter before cancelling us.  If the cancel
         # arrived too late (sleep future already resolved) and this coroutine
-        # resumed normally, the generation mismatch catches the stale timer and
-        # prevents double delivery.
+        # resumed normally, the generation mismatch catches the stale timer
+        # and prevents double delivery.
         if self._generation.get(key) != gen:
             return
         items = self._buffers.pop(key, [])
         self._tasks.pop(key, None)
+        self._generation.pop(key, None)
         if not items:
             return
         self._deliver(key, guild_id, items)
--- a/backend/routes/webhooks.py
+++ b/backend/routes/webhooks.py
@@ -84,9 +84,9 @@ class DebounceQueue:
     async def shutdown(self) -> None:
         """Cancel all in-flight timers and wait for them to finish.
 
-        Awaiting the cancelled tasks ensures they run their CancelledError
-        handler and the event loop does not emit "Task destroyed but pending"
-        warnings.  Call this at server shutdown and in async test teardown.
+        Awaiting the cancelled tasks lets them complete cleanly and prevents
+        "Task destroyed but pending" warnings.  Call this at server shutdown
+        and in async test teardown.
         """
         pending = [t for t in self._tasks.values() if not t.done()]
         for task in pending:
@@ -118,23 +118,13 @@ class DebounceQueue:
     async def _fire(self, key: str, guild_id: str) -> None:
         """Sleep for the debounce window then deliver all buffered events.
 
-        On CancelledError: if a replacement timer was scheduled, the buffer
-        stays intact for that timer to deliver.  If we are still the
-        registered task (external cancellation with no replacement), deliver
-        immediately so the buffer is not silently dropped.
+        If cancelled (because schedule() restarted the timer for a new event),
+        swallow the error and return — the replacement task owns the buffer.
         """
         try:
             await asyncio.sleep(self._window)
         except asyncio.CancelledError:
-            # _schedule_debounced_foreman pops + replaces self._tasks[key]
-            # synchronously before the event loop injects CancelledError here,
-            # so if we are still the registered task no replacement exists.
-            if self._tasks.get(key) is asyncio.current_task():
-                items = self._buffers.pop(key, [])
-                self._tasks.pop(key, None)
-                if items:
-                    self._deliver(key, guild_id, items)
-            raise
+            return
         items = self._buffers.pop(key, [])
         self._tasks.pop(key, None)
         if not items:
--- a/backend/routes/webhooks.py
+++ b/backend/routes/webhooks.py
@@ -127,12 +127,6 @@ class DebounceQueue:
         try:
             await asyncio.sleep(self._window)
         except asyncio.CancelledError:
-            # External cancellation (shutdown/test teardown): _tasks[key] was never
-            # popped, so clean it up.  During a schedule() reset, _tasks[key] was
-            # already popped before cancel() was called, so we won't find this task
-            # there and the check is a safe no-op.
-            if self._tasks.get(key) is asyncio.current_task():
-                self._tasks.pop(key, None)
             raise
         # Stale-generation guard: if schedule() ran after our sleep started it
         # bumped the generation counter before cancelling us.  If the cancel
--- a/backend/routes/webhooks.py
+++ b/backend/routes/webhooks.py
@@ -14,6 +14,7 @@ can decide whether to send_followup, finalize, or no-op.
 
 from __future__ import annotations
 
+import asyncio
 import hashlib
 import hmac
 import json
@@ -35,6 +36,18 @@ logger = logging.getLogger(__name__)
 
 router = APIRouter()
 
+# How long (seconds) to wait for further events on the same PR before
+# delivering the buffered batch to the foreman.  GitHub often sends several
+# check_run completions and a review comment within a second of each other;
+# this window lets them coalesce into a single foreman invocation.
+DEBOUNCE_WINDOW_SECONDS: float = 3.0
+
+# Per-key debounce state.  Key format: "{guild_id}:{task_id}".
+# _debounce_buffers accumulates (summary, user_id) pairs until the timer fires.
+# _debounce_tasks holds the in-flight asyncio Task for each key.
+_debounce_buffers: dict[str, list[tuple[str, str | None]]] = {}
+_debounce_tasks: dict[str, asyncio.Task] = {}  # type: ignore[type-arg]
+
 
 # Cap on stored payloads. GitHub webhook payloads are typically <50 KB but
 # review-comment diff hunks can balloon — 64 KB is a safe upper bound that
@@ -46,6 +59,64 @@ _MAX_PAYLOAD_BYTES = 64 * 1024
 # most CI/automation traffic — filtering them out would silence the very
 # events we want to react to.
 _BOT_OK_EVENTS = frozenset({"check_run", "check_suite", "status"})
+
+
+def _debounce_key(guild_id: str, task_id: str) -> str:
+    return f"{guild_id}:{task_id}"
+
+
+async def _debounce_fire(key: str, guild_id: str) -> None:
+    """Sleep for the debounce window then deliver all buffered events.
+
+    If this task is cancelled before the window expires (because a new event
+    arrived and reset the timer), the CancelledError propagates normally and
+    nothing is delivered — the replacement timer carries the full buffer.
+    """
+    await asyncio.sleep(DEBOUNCE_WINDOW_SECONDS)
+    items = _debounce_buffers.pop(key, [])
+    _debounce_tasks.pop(key, None)
+    if not items:
+        return
+    summaries = [s for s, _ in items]
+    combined = "\n\n---\n\n".join(summaries) if len(summaries) > 1 else summaries[0]
+    user_id = items[-1][1]
+    spawn(
+        run_foreman_ai(guild_id, combined, user_id=user_id),
+        name=f"foreman.github-debounced:{key}",
+    )
+    reset_foreman_poll(guild_id)
+    logger.info(
+        "github webhook debounce fired key=%s events=%d guild=%s",
+        key,
+        len(items),
+        guild_id,
+    )
+
+
+def _schedule_debounced_foreman(
+    key: str,
+    guild_id: str,
+    summary: str,
+    user_id: str | None,
+) -> None:
+    """Buffer *summary* and (re)start the per-PR debounce timer.
+
+    Each call resets the window so the foreman is only invoked after
+    DEBOUNCE_WINDOW_SECONDS of silence on this PR.
+    """
+    existing = _debounce_tasks.pop(key, None)
+    if existing and not existing.done():
+        existing.cancel()
+    _debounce_buffers.setdefault(key, []).append((summary, user_id))
+    _debounce_tasks[key] = spawn(
+        _debounce_fire(key, guild_id),
+        name=f"foreman.debounce:{key}",
+    )
+    logger.info(
+        "github webhook debounce scheduled key=%s buffered=%d",
+        key,
+        len(_debounce_buffers[key]),
+    )
 
 
 def _verify_signature(secret: str, body: bytes, signature_header: str | None) -> bool:
@@ -466,11 +537,8 @@ async def github_webhook(guild_id: str, request: Request) -> Response:
         summary = _build_foreman_summary(
             event_type, action, payload, repo, pr_number, task_id, sender_login
         )
-        spawn(
-            run_foreman_ai(guild_id, summary, user_id=task_user_id),
-            name=f"foreman.github-event:{delivery_id}",
-        )
-        reset_foreman_poll(guild_id)
+        key = _debounce_key(guild_id, task_id)
+        _schedule_debounced_foreman(key, guild_id, summary, task_user_id)
     else:
         logger.info(
             "github webhook skip-foreman guild=%s delivery=%s event=%s reason=%s",
--- a/backend/routes/webhooks.py
+++ b/backend/routes/webhooks.py
@@ -68,6 +68,7 @@ class DebounceQueue:
         self._window = window_seconds
         self._buffers: dict[str, list[tuple[str, str | None]]] = {}
         self._tasks: dict[str, asyncio.Task] = {}  # type: ignore[type-arg]
+        self._generation: dict[str, int] = {}
 
     def reset(self) -> None:
         """Cancel all in-flight timers and clear state (sync, no await).
@@ -80,6 +81,7 @@ class DebounceQueue:
                 task.cancel()
         self._tasks.clear()
         self._buffers.clear()
+        self._generation.clear()
 
     async def shutdown(self) -> None:
         """Cancel all in-flight timers and wait for them to finish.
@@ -95,6 +97,7 @@ class DebounceQueue:
             await asyncio.gather(*pending, return_exceptions=True)
         self._tasks.clear()
         self._buffers.clear()
+        self._generation.clear()
 
     def _deliver(self, key: str, guild_id: str, items: list[tuple[str, str | None]]) -> None:
         summaries = [s for s, _ in items]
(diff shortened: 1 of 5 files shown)


backend/tests/test_github_webhooks_phase2.py
```diff
@@ -349,15 +347,113 @@ def test_webhook_dispatches_for_ci_bot_check_run(client):
...
+        async def fake_run_foreman(guild_id, summary, *, user_id=None):
+            self._foreman_calls.append((guild_id, summary, user_id))
+
+        return (
+            patch.object(wh, "DEBOUNCE_WINDOW_SECONDS", 0.05),
+            patch.object(wh, "run_foreman_ai", new=fake_run_foreman),
+            patch.object(wh, "reset_foreman_poll"),
+            patch.object(wh, "spawn", side_effect=lambda coro, **kw: asyncio.create_task(coro)),
+        )
+
```
jmelloy, line 395: These `async def` test methods have no `@pytest.mark.asyncio` decorator and `TestDebounce` has no asyncio mode configured. Without it (or a class-level `pytestmark`), pytest will collect them as coroutine objects, immediately return without awaiting, and mark them as passing — meaning the assertions never actually run. Add `@pytest.mark.asyncio` to each method or set `asyncio_mode = 'auto'` in the pytest config.

backend/routes/webhooks.py
```diff
@@ -48,6 +61,64 @@
...
+    return f"{guild_id}:{task_id}"
+
+
+async def _debounce_fire(key: str, guild_id: str) -> None:
+    """Sleep for the debounce window then deliver all buffered events.
+
+    If this task is cancelled before the window expires (because a new event
+    arrived and reset the timer), the CancelledError propagates normally and
+    nothing is delivered — the replacement timer carries the full buffer.
+    """
```
jmelloy: When a debounce task is cancelled (because a newer event arrived), `_debounce_fire` exits via `CancelledError` before calling `_debounce_buffers.pop`. The buffer entry is therefore left in `_debounce_buffers` until the replacement timer fires — this is intentional and correct. However, if the event loop shuts down while a task is in-flight (e.g. during testing or server shutdown), the buffer will never be drained and the accumulated entries will be silently discarded with no log. Consider adding a `try/finally` block that at minimum logs the abandoned buffer.

backend/routes/webhooks.py
```diff
@@ -35,6 +36,18 @@
 
 router = APIRouter()
 
+# How long (seconds) to wait for further events on the same PR before
+# delivering the buffered batch to the foreman.  GitHub often sends several
+# check_run completions and a review comment within a second of each other;
+# this window lets them coalesce into a single foreman invocation.
+DEBOUNCE_WINDOW_SECONDS: float = 3.0
```
jmelloy: These two module-level dicts are global singletons shared across all requests and tests. There is no cleanup mechanism (no TTL, no eviction on PR close, no shutdown hook). Long-running servers processing many PRs will accumulate keys indefinitely. Consider a `weakref`-based structure, explicit cleanup on PR close events, or at minimum a comment noting this limitation.

backend/routes/webhooks.py
```diff
@@ -48,6 +61,64 @@
...
+    summary: str,
+    user_id: str | None,
+) -> None:
+    """Buffer *summary* and (re)start the per-PR debounce timer.
+
+    Each call resets the window so the foreman is only invoked after
+    DEBOUNCE_WINDOW_SECONDS of silence on this PR.
+    """
+    existing = _debounce_tasks.pop(key, None)
+    if existing and not existing.done():
```
jmelloy: `user_id` is set to the last event's `user_id` (`items[-1][1]`). If a bot-triggered check_run completes after a human review comment within the debounce window, the foreman will be attributed to `None` or a bot user rather than the human. Consider using the first non-None user_id or the earliest event's user instead.

jmelloy (commented):
- Debounce logic is sound and the `asyncio.Task` cancellation approach is correct, but the module-level mutable dictionaries (`_debounce_buffers`, `_debounce_tasks`) are global singletons that will bleed state between tests and across server restarts without an explicit cleanup path.
- The `_debounce_fire` function silently swallows the buffer if the task is cancelled mid-sleep (CancelledError propagates), but the buffer is never cleaned up in that case — new events append to it correctly, yet a cancelled task leaves stale entries until a new event arrives or the next fire.
- `user_id` selection strategy (always taking the last event's `user_id`) is implicit and undocumented; if the first event has a meaningful user and the last is a bot event, the foreman attribution will be wrong.
- The `TestDebounce` tests are `async def` methods on a plain class (no `pytest-asyncio` marker or `asyncio.run` wrapper), which means they will silently pass as no-ops in most default pytest configurations — they need `@pytest.mark.asyncio` or an equivalent mechanism.
- The `_patched_env` helper returns a plain tuple of context managers that must be unpacked manually (`p1, p2, p3, p4`), which is fragile; using `contextlib.ExitStack` or a single `@contextmanager` would be safer and more maintainable.

jmelloy (commented):
- The debounce implementation is well-structured with a clean `DebounceQueue` class, proper shutdown handling, and good test coverage including edge cases like timer cancellation and state isolation.
- A critical race condition exists in `_fire`: the comment says the task is replaced *synchronously before* CancelledError is injected, but asyncio task cancellation is not synchronous — the CancelledError is delivered on the next event-loop iteration, meaning `self._tasks.get(key) is asyncio.current_task()` may be true even when a replacement timer exists, causing premature delivery of buffered events.
- The `test_cancelled_events_not_lost` test asserts that buffers are preserved after external cancellation, but the `_fire` method's CancelledError handler calls `self._buffers.pop(key, [])` when it detects it's still the registered task — which contradicts the test's expectation and reveals a logic inconsistency.
- The default debounce window of 60 seconds is very aggressive for a webhook handler; this could delay foreman invocations by up to a minute for legitimate single-event scenarios (e.g., a PR review with no follow-up), and the PR description mentions 3 seconds but the code defaults to 60.
- The `_deliver` method's user_id selection strategy (last event's user_id) is reasonable but undocumented in user-facing terms; also, silently dropping attribution from intermediate events with different user_ids could be a concern for audit trails.

backend/routes/webhooks.py
```diff
@@ -48,6 +57,128 @@
...
+            task.cancel()
+        if pending:
+            await asyncio.gather(*pending, return_exceptions=True)
+        self._tasks.clear()
+        self._buffers.clear()
+
+    def _deliver(self, key: str, guild_id: str, items: list[tuple[str, str | None]]) -> None:
+        summaries = [s for s, _ in items]
+        combined = "\n\n---\n\n".join(summaries) if len(summaries) > 1 else summaries[0]
+        # Use the last event's user_id: in typical bursts the final event is a
```
jmelloy, line 105: This race condition check is unreliable. `schedule()` cancels the old task and assigns a new one to `self._tasks[key]`, but the CancelledError is not injected until the next event-loop tick. By the time this handler runs, `self._tasks[key]` has already been replaced with the new task, so `self._tasks.get(key) is asyncio.current_task()` will be `False` — the early-delivery branch is unreachable in the normal replacement case. The comment's claim about synchronous replacement is incorrect for asyncio task cancellation.

backend/routes/webhooks.py
```diff
@@ -35,6 +37,13 @@
 
 router = APIRouter()
 
+# How long (seconds) to wait for further events on the same PR before
+# delivering the buffered batch to the foreman.  GitHub often sends several
+# check_run completions and a review comment within a second of each other;
+# this window lets them coalesce into a single foreman invocation.
+# Configurable via WEBHOOK_DEBOUNCE_SECONDS environment variable (default: 60).
```
jmelloy: The PR description says the debounce window is 3 seconds, but the default here is 60 seconds. A 60-second delay before foreman invocation is a significant UX regression for single-event webhooks (e.g., a lone PR review). Consider aligning with the PR description or explicitly justifying the 60-second default.

backend/tests/test_github_webhooks_phase2.py
```diff
@@ -349,15 +348,142 @@ def test_webhook_dispatches_for_ci_bot_check_run(client):
...
+            wh._debounce_queue.schedule("g-ind:t-pr1", "g-ind", "pr1 event", "u4")
+            wh._debounce_queue.schedule("g-ind:t-pr2", "g-ind", "pr2 event", "u4")
+            await asyncio.sleep(0.2)
+
+        assert len(self._foreman_calls) == 2
+        summaries = {c[1] for c in self._foreman_calls}
+        assert any("pr1 event" in s for s in summaries)
+        assert any("pr2 event" in s for s in summaries)
+
+    async def test_state_isolation_between_tests(self):
```
jmelloy, line 456: This test asserts `len(wh._debounce_queue._buffers.get('g-cl:t-cl1', [])) == 2` after external cancellation, but `_fire`'s CancelledError handler calls `self._buffers.pop(key, [])` when the task is still registered — which would clear the buffer to empty. Either the test is wrong about the expected behavior, or the `_fire` implementation is wrong. These two need to be reconciled.

backend/routes/webhooks.py
```diff
@@ -48,6 +57,128 @@
...
+            name=f"foreman.github-debounced:{key}",
+        )
+        reset_foreman_poll(guild_id)
+        logger.info(
+            "github webhook debounce fired key=%s events=%d guild=%s",
+            key,
+            len(items),
+            guild_id,
+        )
+
```
jmelloy, line 117: Calling `existing.cancel()` without awaiting means the cancelled task may still be running when the new `_fire` task starts. If both tasks overlap briefly, there is a window where `self._tasks[key]` points to the new task but the old task hasn't yet received its CancelledError, potentially leading to double delivery if the old task's sleep happens to complete before cancellation propagates.

backend/routes/webhooks.py
```diff
@@ -48,6 +57,128 @@
...
+    def _deliver(self, key: str, guild_id: str, items: list[tuple[str, str | None]]) -> None:
+        summaries = [s for s, _ in items]
+        combined = "\n\n---\n\n".join(summaries) if len(summaries) > 1 else summaries[0]
+        # Use the last event's user_id: in typical bursts the final event is a
+        # human action (review, comment) that arrives after the CI bot events,
+        # so this attribution is usually more meaningful than the first event's.
+        user_id = items[-1][1]
+        spawn(
+            run_foreman_ai(guild_id, combined, user_id=user_id),
```
jmelloy: Race condition: `schedule()` does `self._tasks.pop(key)` then immediately sets `self._tasks[key] = new_task` synchronously, so when the cancelled task's `CancelledError` handler runs, `self._tasks.get(key)` already points to the replacement task — not `current_task()`. This means the condition `self._tasks.get(key) is asyncio.current_task()` is always False for a normally-cancelled task, so the buffer-preservation branch is never reached and the buffer will be silently dropped on cancellation. Consider using a generation counter or a different sentinel to distinguish external vs. replacement cancellation.

backend/routes/webhooks.py
```diff
@@ -35,6 +37,13 @@
 
 router = APIRouter()
 
+# How long (seconds) to wait for further events on the same PR before
+# delivering the buffered batch to the foreman.  GitHub often sends several
+# check_run completions and a review comment within a second of each other;
+# this window lets them coalesce into a single foreman invocation.
```
jmelloy, line 43: The PR description says the debounce window is 3 seconds, but the default here is 60 seconds and `.env.example` also documents 60 seconds. A 60-second silence window means the foreman won't be invoked for at least a minute after any webhook activity, which seems unacceptably slow for a CI feedback loop. Please reconcile the intended default and document it clearly.

backend/tests/test_github_webhooks_phase2.py
```diff
@@ -349,15 +348,142 @@ def test_webhook_dispatches_for_ci_bot_check_run(client):
...
+
+    @contextmanager
+    def _patched_env(self, wh):
+        """Context manager: fresh DebounceQueue instance + captured foreman calls.
+
+        Each call creates a brand-new DebounceQueue so tests cannot bleed state
+        into or out of the module-level singleton.
+        """
+        self._foreman_calls: list[tuple[str, str, str | None]] = []
+
```
jmelloy, line 380: `patch.object(wh, 'spawn', side_effect=lambda coro, **kw: asyncio.create_task(coro))` replaces the production `spawn` with bare `asyncio.create_task`, which has different error-handling semantics (unhandled exceptions become unraisable errors rather than being logged). Tests calling `_deliver` indirectly through the real `_fire` path may emit noisy asyncio warnings or hide real bugs. Consider also patching `run_foreman_ai` at the `spawn` call site or asserting on task exception handling.

backend/tests/test_github_webhooks_phase2.py
```diff
@@ -349,15 +348,142 @@ def test_webhook_dispatches_for_ci_bot_check_run(client):
...
+            wh._debounce_queue.schedule("g-ind:t-pr2", "g-ind", "pr2 event", "u4")
+            await asyncio.sleep(0.2)
+
+        assert len(self._foreman_calls) == 2
+        summaries = {c[1] for c in self._foreman_calls}
+        assert any("pr1 event" in s for s in summaries)
+        assert any("pr2 event" in s for s in summaries)
+
+    async def test_state_isolation_between_tests(self):
+        """Each _patched_env provides a fresh DebounceQueue with no state from prior tests."""
```
jmelloy, line 457: This test asserts `len(wh._debounce_queue._buffers.get('g-cl:t-cl1', [])) == 2` after external cancellation, but the `_fire` CancelledError branch pops the buffer when `self._tasks.get(key) is asyncio.current_task()`. Whether this is True or False depends on the exact race between `schedule()`'s synchronous pop/assign and the async CancelledError injection — the test outcome is timing-dependent and may pass or fail non-deterministically.

jmelloy (commented):
- The debounce implementation is well-structured with a clean `DebounceQueue` class, proper shutdown handling, and good test coverage across coalescing, separation, and isolation scenarios.
- **Critical race condition**: The `_fire` coroutine's CancelledError handler checks `self._tasks.get(key) is asyncio.current_task()`, but `schedule()` calls `self._tasks.pop(key)` and then assigns a new task — there is a window where `self._tasks[key]` is the replacement, meaning the cancelled task sees itself as no longer registered and incorrectly delivers the buffer immediately, causing a double-delivery.
- The default debounce window of **60 seconds** is far too large for a responsive CI/CD system; the PR description mentions 3 seconds but `.env.example` documents 60 seconds as the default — this mismatch could cause significant latency in foreman dispatch.
- `test_cancelled_events_not_lost` asserts that buffers are preserved after external cancellation, but the `_fire` CancelledError branch actually calls `self._buffers.pop(key, [])` when the task is still registered as current — so the buffer IS consumed and the test's assertion (`len(...) == 2`) will fail if the timing aligns such that the task is still the current registered one.
- The `_patched_env` context manager imports `routes.webhooks as wh` outside its scope and patches `wh.spawn` with `asyncio.create_task`, which bypasses any error handling or naming logic in the real `spawn` helper — tests may not accurately reflect production behavior for error cases.

backend/routes/webhooks.py
jmelloy: Race condition in cancellation guard: `_schedule_debounced_foreman` (i.e., `schedule`) pops and replaces `self._tasks[key]` before `cancel()` is awaited, so by the time `CancelledError` is injected into `_fire`, `self._tasks.get(key)` will already point to the *new* task, not `current_task()`. This means the guard `self._tasks.get(key) is asyncio.current_task()` will always be `False` for a normally-cancelled (replaced) timer, causing the buffer to never be delivered via the cancellation path — which is the intended behaviour. However this also means that if someone externally cancels the task *after* replacing it (an unlikely but possible race), events are silently dropped. The comment above should be clarified to reflect that the pop-before-cancel ordering is load-bearing.

backend/routes/webhooks.py
jmelloy, line 43: The default of 60 seconds contradicts the PR description which states '3-second window'. A 60-second debounce will introduce unacceptable latency for all webhook-driven automation. This should be 3 (matching the PR description) or at minimum the discrepancy should be explicitly justified in the comment and `.env.example`.

backend/routes/webhooks.py
```diff
@@ -48,6 +57,128 @@
...
+        logger.info(
+            "github webhook debounce fired key=%s events=%d guild=%s",
+            key,
+            len(items),
+            guild_id,
+        )
+
+    async def _fire(self, key: str, guild_id: str) -> None:
+        """Sleep for the debounce window then deliver all buffered events.
+
```
jmelloy, line 120: The `existing.cancel()` call in `schedule()` is not awaited — this is correct for fire-and-forget cancellation in sync code, but the CancelledError will only be injected into `_fire` on the next event-loop iteration. If `schedule()` is called again synchronously before yielding to the loop (e.g. in a tight loop), multiple cancelled tasks may exist transiently. This is unlikely to cause bugs given normal usage but is worth a comment.

backend/tests/test_github_webhooks_phase2.py
```diff
@@ -349,15 +348,142 @@ def test_webhook_dispatches_for_ci_bot_check_run(client):
...
+        queue = wh.DebounceQueue(window_seconds=0.05)
+        with (
+            patch.object(wh, "_debounce_queue", queue),
+            patch.object(wh, "run_foreman_ai", new=fake_run_foreman),
+            patch.object(wh, "reset_foreman_poll"),
+            patch.object(wh, "spawn", side_effect=lambda coro, **kw: asyncio.create_task(coro)),
+        ):
+            yield
+
+    async def test_rapid_events_single_foreman_call(self):
```
jmelloy, line 396: The `test_cancelled_events_not_lost` test asserts `len(wh._debounce_queue._buffers.get('g-cl:t-cl1', [])) == 2` after external cancellation. But per `_fire`'s `CancelledError` handler: if the task is still registered (which it is after external cancel with no replacement), it pops the buffer and calls `_deliver`. After `await asyncio.sleep(0.0)` the buffer will be empty and delivered — making the assertion incorrect and the subsequent `schedule` call only seeing the third event. This test likely passes only if the `CancelledError` handler path is not exercised as expected.

jmelloy (commented):
- The debounce implementation is well-structured with a clean `DebounceQueue` class and good separation of concerns, but the `_fire` cancellation logic has a subtle race condition that could cause events to be silently dropped or the external-cancellation guard to behave unexpectedly.
- The default debounce window of 60 seconds is surprisingly large for a 'rapid coalescing' use case and will introduce significant latency for normal webhook events; the PR description mentions 3 seconds but the default is 60.
- The `test_cancelled_events_not_lost` test asserts that buffered events are preserved after external cancellation, but the `_fire` implementation actually pops and delivers (or discards) the buffer on `CancelledError` when it is still the registered task, meaning the assertion `len(wh._debounce_queue._buffers.get(...)) == 2` may be fragile or incorrect depending on timing.
- Tests in `TestDebounce` lack `@pytest.mark.asyncio` decorators (or equivalent configuration), which may cause them to run synchronously and pass trivially or fail depending on the pytest-asyncio mode configured.
- The `user_id` attribution logic (using last event's user_id) is reasonable but undocumented at the call site; the comment explaining this is only in `_deliver`, making it easy for future maintainers to miss the intentional design choice.

backend/tests/test_github_webhooks_phase2.py
```diff
@@ -349,15 +349,312 @@ def test_webhook_dispatches_for_ci_bot_check_run(client):
...
+# ---------------------------------------------------------------------------
+# Debounce behaviour
+# ---------------------------------------------------------------------------
+
+
+class TestDebounce:
+    """Per-PR debounce timer: rapid events coalesce; separated events deliver separately."""
+
+    @asynccontextmanager
+    async def _patched_env(self, wh):
```
jmelloy, line 372: The async test methods in `TestDebounce` have no `@pytest.mark.asyncio` decorator. Depending on the project's `asyncio_mode` pytest-asyncio setting, these may be silently collected but not actually awaited/run, giving false-positive passes. Confirm the project uses `asyncio_mode = 'auto'` in pytest config, or add explicit markers.

backend/routes/webhooks.py
```diff
@@ -48,6 +57,140 @@
...
+        """Cancel all in-flight timers and wait for them to finish.
+
+        Awaiting the cancelled tasks lets them complete cleanly and prevents
+        "Task destroyed but pending" warnings.  Call this at server shutdown
+        and in async test teardown.
+        """
+        pending = [t for t in self._tasks.values() if not t.done()]
+        for task in pending:
+            task.cancel()
+        if pending:
```
jmelloy, line 96: When an external caller cancels a task (e.g., test teardown or OS signal), `_tasks[key]` is left holding the cancelled/done task and is never removed. Subsequent `schedule()` calls will pop and discard it correctly, but `shutdown()` will redundantly attempt to cancel it. Consider also cleaning up `_tasks` inside `_fire` on `CancelledError` to keep state consistent.

backend/routes/webhooks.py
```diff
@@ -48,6 +57,140 @@
...
+            await asyncio.gather(*pending, return_exceptions=True)
+        self._tasks.clear()
+        self._buffers.clear()
+        self._generation.clear()
+
+    def _deliver(self, key: str, guild_id: str, items: list[tuple[str, str | None]]) -> None:
+        summaries = [s for s, _ in items]
+        combined = "\n\n---\n\n".join(summaries) if len(summaries) > 1 else summaries[0]
```
jmelloy, line 104: `_deliver` calls `spawn(run_foreman_ai(...))` which creates yet another asyncio task inside the already-async `_fire` task. This means errors in `run_foreman_ai` surface only in the spawned task's exception handler and are completely invisible to `_fire`. Since `_fire` is already async, consider awaiting `run_foreman_ai` directly here, or at minimum ensure the spawned task's exceptions are logged.

backend/routes/webhooks.py
```diff
@@ -48,6 +57,140 @@
 _BOT_OK_EVENTS = frozenset({"check_run", "check_suite", "status"})
 
 
+class DebounceQueue:
+    """Per-PR debounce queue that coalesces rapid webhook events.
+
+    State is scoped to the instance so tests can create an isolated copy
+    without touching the module-level singleton.
```
jmelloy, line 64: The `_tasks` type annotation `dict[str, asyncio.Task]` with `# type: ignore[type-arg]` suppresses the generic parameter. Python 3.9+ supports `dict[str, asyncio.Task[None]]` without ignore; using the proper type would remove the suppression and improve clarity.

jmelloy (commented):
- The debounce implementation is well-structured with a clean `DebounceQueue` class, generation-counter race protection, and thorough test coverage including edge cases like stale generation and external cancellation.
- The `_deliver` method uses `spawn()` to wrap `run_foreman_ai`, but since `spawn` is patched in tests to use `asyncio.create_task`, the actual delivery path is correct; however `_fire` is itself an async task so `_deliver` calling `spawn` inside an async context adds an extra task layer that may be unnecessary.
- The `test_cancelled_events_not_lost` test documents that externally-cancelled timers preserve the buffer, but the `_tasks` dict is NOT cleaned up when a task is externally cancelled — `_tasks[key]` still holds the done/cancelled task, which means `schedule()` will pop it correctly but `shutdown()` will attempt to cancel an already-done task (harmless but slightly wasteful).
- The user_id attribution strategy (using the last event's user_id) is noted in a comment but could silently change attribution if a CI bot event trails a human event, which may cause unexpected behavior in downstream foreman logic.
- Test class `TestDebounce` lacks `@pytest.mark.asyncio` or equivalent markers on individual async test methods — depending on the pytest-asyncio mode configured, these tests may be silently skipped rather than run.

jmelloy:
Superseded by #401, which addresses all review comments (generation-counter race fix, asyncio.create_task + error callback, bot-user attribution fix, buffer snapshot on cancel). The missing shutdown_debouncer hook from this PR has now been added to #401 (commit 69c553f).

PR #400, opened 2026-05-20
Task locking: prevent concurrent follow-up races

### LINKED ISSUE #388: Task locking: prevent concurrent follow-ups from racing on the same task
## Problem

When a follow-up completes and the foreman receives `followup-done`, a second webhook (e.g. a CI result or another review comment) can arrive almost simultaneously. Without a database-level lock on the task, both signals can trigger independent `send_followup` calls, creating duplicate branches of work on the same task.

## Expected behaviour

- A task in `in-progress` or `awaiting-review` state should hold a lock that prevents a second follow-up from being queued until the current one finishes or the task transitions to a stable state.
- Incoming follow-up triggers that arrive while the task is locked should be queued (or coalesced) and applied after the lock is released.

## Possible approach

- Add a `locked_at` / `lock_holder` column to the `tasks` table.
- Acquire the lock atomically (e.g. `SELECT … FOR UPDATE` or an `UPDATE … WHERE locked_at IS NULL RETURNING *` pattern) before dispatching a follow-up.
- Release the lock when the follow-up worker reports done or the task moves to `awaiting-review`.
- Queued triggers can be stored in a `task_events` table and replayed on lock release.

jmelloy:
Closes #388

## Summary
- Adds `locked_at` / `lock_holder` columns to the `tasks` table with a migration
- Acquires the lock atomically before dispatching a follow-up
- Releases the lock when the follow-up worker reports done or task moves to `awaiting-review`
- Stores queued triggers in a `task_events` table and replays them on lock release
- Tests cover the race scenario (two simultaneous triggers → one dispatched, second queued and replayed)

**Note:** Frontend changes (optimistic follow-up state, error surfaces, `openTaskTab` refactor) that were incorrectly included in an earlier revision have been reverted. This PR contains backend task-locking changes only.

### DIFF
--- a/backend/ws_handlers.py
+++ b/backend/ws_handlers.py
@@ -26,8 +26,8 @@ from events import (
     pending_claude_auth,
 )
 from fastapi import WebSocket
-from models import Agent, Message, Task, TaskLog, User, Worker
-from sqlalchemy import select, update
+from models import Agent, Message, Task, TaskEvent, TaskLog, User, Worker
+from sqlalchemy import delete, select, update
 from sqlalchemy.dialects.sqlite import insert as sqlite_insert
 from util.tasks import spawn
 from utils import worker_display_name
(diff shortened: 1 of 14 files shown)


backend/ws_handlers.py
```diff
@@ -605,7 +605,7 @@ async def handle_task_followup_done(ctx: WSContext, data: dict) -> None:
     worker_id_msg = data.get("workerId", "")
     if task_id:
         pr_url_fud = data.get("prUrl", "")
-        update_vals: dict = {"state": "awaiting-review"}
+        update_vals: dict = {"state": "awaiting-review", "locked_at": None, "lock_holder": None}
```
jmelloy: Any terminal state should release the lock

jmelloy (commented):
This pr should also update the Frontend to have the users follow-up  button go straight through

PR #401, opened 2026-05-20
fix(webhooks): address PR #399 review comments — debounce race conditions

jmelloy:
## Summary

Addresses all reviewer comments on PR #399 (webhook debounce per-PR timer).

- **Global mutable state eliminated**: `DebounceQueue` class holds `_buffers`, `_tasks`, and `_generation` as instance variables. The module-level `_debounce_queue` is a single instance; tests create isolated copies via `_patched_env`. `reset()` and `shutdown()` provide explicit cleanup paths.

- **Race condition fixed with generation counter**: `_fire` receives the `gen` value captured at creation time. `schedule()` bumps the generation counter *before* cancelling the old task, so even if the old timer's sleep resolves during the `gather()` window and the coroutine runs, it sees a stale generation and exits without delivering — no double-delivery.

- **`spawn()` replaced with `asyncio.create_task()` + error callback**: Since the task is retained in `self._tasks[key]`, GC is not a concern. `_log_fire_error` is a static done-callback that logs unexpected exceptions explicitly. This keeps the debounce logic self-contained and avoids polluting the global spawn registry.

- **Default interval documented**: `DEBOUNCE_WINDOW_SECONDS = 3` (via `WEBHOOK_DEBOUNCE_SECONDS` env var). Documented in `.env.example` with an explanation of the coalescing behaviour.

## Test plan

- [x] `python -m pytest` — 419 tests pass
- [x] `ruff check . && ruff format .` — clean
- [x] `TestDebounce::test_rapid_events_single_foreman_call` — 3 events → 1 delivery
- [x] `TestDebounce::test_slow_events_separate_foreman_calls` — separated events → 2 deliveries
- [x] `TestDebounce::test_reset_cancels_previous_timer` — combined in single delivery
- [x] `TestDebounce::test_independent_prs_not_merged` — independent keys independent timers
- [x] `TestDebounce::test_stale_generation_does_not_deliver` — stale gen guard works
- [x] `TestDebounce::test_cancelled_events_not_lost` — buffer preserved after external cancel
- [x] `TestDebounce::test_race_condition_no_events_dropped` — no events lost on rapid reset
- [x] All existing webhook dispatch tests updated to mock `_debounce_queue.schedule`

### DIFF
--- a/backend/routes/webhooks.py
+++ b/backend/routes/webhooks.py
@@ -14,10 +14,12 @@ can decide whether to send_followup, finalize, or no-op.
 
 from __future__ import annotations
 
+import asyncio
 import hashlib
 import hmac
 import json
 import logging
+import os
 from datetime import UTC, datetime
 
 from database import get_db
@@ -27,13 +29,19 @@ from models import GithubEvent, Guild, Message, Task
 from sqlalchemy import select, update
 from sqlalchemy.dialects.sqlite import insert as sqlite_insert
 from sqlalchemy.exc import IntegrityError
-from util.tasks import spawn
 
 from foreman import reset_foreman_poll, run_foreman_ai
 
 logger = logging.getLogger(__name__)
 
 router = APIRouter()
+
+# How long (seconds) to wait for further events on the same PR before
+# delivering the buffered batch to the foreman.  GitHub often sends several
+# check_run completions and a review comment within a second of each other;
+# this window lets them coalesce into a single foreman invocation.
+# Configurable via WEBHOOK_DEBOUNCE_SECONDS (default: 3 seconds).
+DEBOUNCE_WINDOW_SECONDS: float = float(os.environ.get("WEBHOOK_DEBOUNCE_SECONDS", "3"))
 
 
 # Cap on stored payloads. GitHub webhook payloads are typically <50 KB but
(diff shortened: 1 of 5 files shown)


backend/routes/webhooks.py
```diff
@@ -48,6 +56,157 @@
...
+
+    @staticmethod
+    def _log_fire_error(task: asyncio.Task) -> None:  # type: ignore[type-arg]
+        """Done callback: log any unexpected error from a _fire task."""
+        if task.cancelled():
+            return
+        exc = task.exception()
+        if exc is not None:
+            logger.exception(
+                "Debounce fire task %s raised an unexpected error",
```
jmelloy, line 110: The comment 'Use the last event's user_id: in typical bursts the final event is a human action' is misleading — in CI bursts (check_run completions) the last event is also a bot, so this attribution is no more reliable than using the first event's user_id. Consider either always using the first non-bot user_id or documenting this explicitly as 'last sender wins, may be a bot'.

backend/routes/webhooks.py
```diff
@@ -48,6 +56,157 @@
...
+        await run_foreman_ai(guild_id, combined, user_id=user_id)
+        reset_foreman_poll(guild_id)
+        logger.info(
+            "github webhook debounce fired key=%s events=%d guild=%s",
+            key,
+            len(items),
+            guild_id,
+        )
+
+    async def _fire(self, key: str, guild_id: str, gen: int) -> None:
```
jmelloy, line 135: After `existing.cancel()` + `await asyncio.gather(existing, return_exceptions=True)`, if the existing task had already completed normally (i.e., `_fire` ran and cleared `_buffers[key]` before `schedule()` popped it from `_tasks`), then `self._buffers.setdefault(key, [])` starts with an empty list and the old events are silently lost. The `_tasks.pop(key, None)` at line 156 inside `_fire` means `existing` may no longer be in `_tasks` when `schedule()` pops it, but the buffer clearing in `_fire` (line 153) could still race with the `schedule()` append.

backend/tests/test_github_webhooks_phase2.py
```diff
@@ -349,15 +351,249 @@ def test_webhook_dispatches_for_ci_bot_check_run(client):
...
+        assert "second" in summary
+
+    async def test_independent_prs_not_merged(self):
+        """Events on different PR keys have independent timers and fire separately."""
+        import routes.webhooks as wh
+
+        async with self._patched_env(wh):
+            await wh._debounce_queue.schedule("g-ind:t-pr1", "g-ind", "pr1 event", "u4")
+            await wh._debounce_queue.schedule("g-ind:t-pr2", "g-ind", "pr2 event", "u4")
+            await asyncio.sleep(0.2)
```
jmelloy, line 450: `await asyncio.sleep(0.0)` is not guaranteed to fully drain all pending callbacks in a single pass on all asyncio implementations. Use `await asyncio.sleep(0)` with a small number of iterations or `asyncio.gather` on the task directly to reliably confirm cancellation has propagated before asserting on `_buffers`.

jmelloy (commented):
- The generation counter race-condition fix is well-designed: bumping `_generation` before cancelling the old task correctly handles the edge case where a cancelled task's sleep future has already resolved.
- The `_deliver` method's user_id attribution comment ('last event is usually human') is a heuristic that could mislead — CI bots frequently emit the final event in a burst, making this unreliable without further explanation or configurability.
- `test_cancelled_events_not_lost` directly manipulates `_tasks` internals and relies on `asyncio.sleep(0.0)` to propagate cancellation, which is fragile and may not reliably flush the event loop across all Python/asyncio implementations.
- The `conftest.py` autouse fixture calling `shutdown_debouncer()` before each test on the module-level singleton is correct hygiene, but if any test patches `_debounce_queue` with a local instance, the pre-test `shutdown_debouncer()` call operates on the old singleton rather than the patched one — a potential ordering issue.
- Overall the implementation is a solid improvement: encapsulated state, explicit cleanup paths, generation-counter guard, and thorough behavioural test coverage all address the original PR #399 concerns effectively.

PR #405, opened 2026-05-20
Fix foreman chat: show tool usage after page refresh

### LINKED ISSUE #404: Foreman chat: tool usage not shown after page refresh
## Problem

When the Foreman chat page is refreshed, previously visible tool calls (function invocations and their results) are no longer rendered in the chat history. Only the text messages survive the reload.

## Expected behaviour

Tool use blocks (both the call and the result) should be persisted and re-rendered when chat history is loaded on page refresh, so the full conversation — including tool activity — is visible.

## Likely cause

Tool-use turns are probably stored in the database but either:
- Not fetched/included when loading message history, or
- Fetched but not rendered by the frontend message renderer (only `text` content blocks are handled, `tool_use` / `tool_result` blocks are skipped).

## Acceptance criteria

- [ ] After a page refresh, all tool call blocks from the current session are visible in the foreman chat.
- [ ] Tool result blocks are rendered alongside their corresponding tool calls.
- [ ] Existing text-only message rendering is unaffected.

jmelloy:
**Task:** Fix foreman chat: show tool usage after page refresh
**Task ID:** `t-ghomg9`

Automated by Pioneer Square worker agent.

Closes #404


jmelloy (commented):
- The fix correctly adds `role` and `meta` columns to persist tool_use/tool_result messages and deserializes them on read, enabling tool usage to survive page refresh.
- A significant concern is the repeated `get_db()` / `close()` pattern for persisting tool-use and tool-result rows; these open fresh DB sessions in the middle of an existing function that likely already has a DB session in scope, which is wasteful and inconsistent.
- The `_now` timestamp is shared across the tool_use save block and the tool_result save block, meaning all tool messages get the same `created_at`; combined with the secondary sort on `id`, ordering should be deterministic, but the shared timestamp is fragile if IDs are not guaranteed sequential.
- The `_message_dict` helper silently swallows JSON parse errors with a bare `except Exception: pass`, which could mask data corruption without any logging.
- The frontend `ChatMessage` type addition looks correct and minimal, though `from_agent` was already likely present elsewhere; the index signature `[key: string]: unknown` already covered it, so the explicit addition is fine for clarity.

backend/foreman/runner.py
```diff
@@ -695,6 +695,33 @@ async def run_foreman_ai(
                     },
                 )
```
jmelloy, line 696: Opening a brand-new DB session (`get_db()`) here instead of reusing the session already in scope for the surrounding `run_foreman_ai` function is inconsistent and wasteful. Reuse the existing session or pass it as a parameter to avoid extra connections and potential transaction isolation surprises.

backend/foreman/runner.py
```diff
@@ -719,6 +746,33 @@ async def run_foreman_ai(
                         "createdAt": _now,
                     },
                 )
```
jmelloy, line 725: Same issue as the tool-use block: a second fresh session is opened just a few lines later. Both blocks should share the same session. Also, if `exec_tools` raises, the tool-use rows are already committed but the tool-result rows never are, leaving the DB in a partially-persisted state with no rollback.

backend/routes/guilds.py
```diff
@@ -23,10 +24,20 @@
 
 router = APIRouter()
 
-
 _VALID_ROLES = {"owner", "member", "viewer"}
```
jmelloy, line 31: The bare `except Exception: pass` silently discards JSON parse failures. At minimum, log the error (e.g., `logger.warning(...)`) so that corrupted `meta` values are observable in production rather than silently ignored.

backend/routes/guilds.py
```diff
@@ -190,7 +201,7 @@ async def get_guild(guild_id: str, github_user_id: str = Depends(require_member(
         result = await db.execute(
             select(Message)
             .where(Message.guild_pk == guild_pk)
-            .order_by(Message.created_at.desc())
+            .order_by(Message.created_at.desc(), Message.id.desc())
```
jmelloy, line 208: Adding `.id.desc()` as a tiebreaker is reasonable, but only works correctly if IDs are monotonically increasing (e.g., auto-increment). If UUIDs or other non-sequential IDs are used, this secondary sort does not guarantee insertion order. A comment or assertion about the ID type would help future maintainers.

backend/foreman/runner.py
jmelloy, line 696: Opening a brand-new DB session (`get_db()`) here instead of reusing the session already in scope creates an extra connection per tool call and runs outside any ambient transaction. Prefer passing the existing `db` session into this block (or into a helper function) to avoid the redundant connection overhead and potential isolation issues.

backend/foreman/runner.py
jmelloy, line 725: Same new-session pattern as the tool-use block above. Both persistence blocks should share one session (ideally the one already open in the surrounding scope) rather than each managing their own `get_db()` / `commit()` / `close()` lifecycle.

backend/routes/guilds.py
```diff
@@ -23,10 +24,20 @@
 
 router = APIRouter()
 
-
 _VALID_ROLES = {"owner", "member", "viewer"}
 
 
+def _message_dict(m: Message) -> dict:
```
jmelloy, line 32: The bare `except Exception: pass` silently discards any JSON decode error (e.g. corrupted `meta` value). At minimum log a warning (`logger.warning(...)`) so data integrity issues are visible in production without breaking the response.

backend/foreman/runner.py
```diff
@@ -695,6 +695,33 @@ async def run_foreman_ai(
...
+                _tu_guild_pk = await get_guild_pk(_tu_db, guild_id)
+                for tu in tool_uses:
+                    _tu_db.add(
+                        Message(
+                            guild_pk=_tu_guild_pk,
+                            from_agent="foreman",
+                            to_agent="user",
+                            content=f"▶ {tu.name}",
+                            message_type="chat",
+                            role="tool_use",
```
jmelloy: `created_at=_now` assigns the same timestamp to every tool-use row in the loop. If the database assigns auto-increment IDs out of insertion order (e.g. with async batching), the `.id.desc()` tiebreaker in the query could still return rows in the wrong order. Consider using `datetime.now(UTC)` per row or a sequence column to guarantee stable ordering.

jmelloy (commented):
- The core fix correctly adds `role` and `meta` columns to persist tool-use/tool-result messages across page refreshes, with matching DB migration, model update, serialization helper, and frontend types.
- A significant design concern is that each tool persistence block opens a **new DB session** (`get_db()`) independent of any ambient session, creating redundant connections and bypassing any outer transaction context; the existing `db` session from the calling scope should be reused.
- The `created_at=_now` value is shared across all tool-use and tool-result rows within the same loop iteration, which means multiple tool calls get identical timestamps and the secondary `.id.desc()` sort is needed as a tiebreaker — this is fragile and could still produce wrong ordering if IDs are non-sequential.
- The `_message_dict` helper silently swallows all JSON parse errors with a bare `except Exception: pass`, which could hide data corruption in the `meta` column without any logging.
- Frontend `ChatMessage` type correctly gains the new fields, but `from_agent` was already implicitly present via `[key: string]: unknown`; adding it explicitly is fine but the PR description does not explain why it was promoted to an explicit field.

backend/foreman/runner.py
```diff
@@ -695,6 +695,33 @@ async def run_foreman_ai(
                     },
                 )
 
+            # Persist tool-use messages so they survive a page refresh
```
jmelloy: A brand-new DB session is opened via `get_db()` here instead of reusing the session that already exists in the enclosing function scope. This is inconsistent with the rest of `run_foreman_ai`, adds connection overhead, and means the tool-use insert is in a completely separate transaction from the surrounding logic. Reuse the existing session and commit strategy instead.

backend/foreman/runner.py
```diff
@@ -695,6 +695,33 @@ async def run_foreman_ai(
...
+            _tu_db = await get_db()
+            try:
+                _tu_guild_pk = await get_guild_pk(_tu_db, guild_id)
+                for tu in tool_uses:
+                    _tu_db.add(
+                        Message(
+                            guild_pk=_tu_guild_pk,
+                            from_agent="foreman",
```
jmelloy: `created_at=_now` is reused for all tool-use rows and later for all tool-result rows. Since all rows in a single batch share the same timestamp, the sort order `created_at DESC, id DESC` is the only thing keeping them ordered. This is fragile; consider using `datetime.now(UTC)` per row, or an explicit sequence/offset.

backend/routes/guilds.py
```diff
@@ -200,7 +211,7 @@ async def get_guild(guild_id: str, github_user_id: str = Depends(require_member(
             "agents": [
                 {**row_to_dict(row.Agent), "worker_name": row.worker_name} for row in agent_rows
             ],
-            "messages": [row_to_dict(m) for m in reversed(messages)],
+            "messages": [_message_dict(m) for m in reversed(messages)],
         }
```
jmelloy, line 219: Silently swallowing the `json.loads` exception with a bare `except Exception: pass` will cause `meta` fields to be silently dropped if the stored JSON is malformed. At minimum log a warning, or use `json.JSONDecodeError` as the caught type to avoid masking unexpected errors.

jmelloy (commented):
- Tool-use and tool-result messages are persisted by opening a fresh DB session each time (`get_db()`) instead of reusing the existing session already in scope, which is inconsistent with the rest of the function and risks transaction isolation issues.
- Both persistence blocks use the same `_now` timestamp for `created_at`, which means tool-use and tool-result rows will have identical timestamps; combined with the new secondary sort on `id`, ordering is fragile and relies on auto-increment behaviour.
- The `meta` column is stored as a raw JSON string (Text) rather than using SQLAlchemy's `JSON` type, requiring manual `json.loads`/`json.dumps` everywhere and silently discarding parse errors in `_message_dict`.
- The `content` field for tool-use messages is set to a human-readable string (`▶ {tu.name}`) rather than a neutral/empty value, mixing display logic into the data layer.
- Frontend `ChatMessage` type gains `from_agent` as an optional field, but it already exists on the DB row and is returned by `row_to_dict`; the addition is harmless but suggests the existing type definition was incomplete and may hide other missing fields.

backend/models.py
```diff
@@ -77,6 +77,8 @@ class Message(Base):
     user_id = Column(
         Text, nullable=True
     )  # github_user_id of the sender; NULL for system/worker messages
+    role = Column(Text, nullable=True)  # "tool_use" | "tool_result" | NULL for plain chat
```
jmelloy, line 80: Using `Text` for the `meta` column instead of `sa.JSON` (or `sa.Text` with an explicit converter) means every consumer must manually serialize/deserialize and there is no schema-level validation. Consider `sa.JSON` which SQLAlchemy handles transparently and most backends support natively.

backend/foreman/runner.py
```diff
@@ -719,6 +719,53 @@ async def run_foreman_ai(
                         "createdAt": _now,
                     },
                 )
+
```
jmelloy, line 726: `get_db()` likely returns a session from the same async context as the surrounding code. Opening and closing it independently in a `finally` block risks closing a shared session or leaving connections in an inconsistent state. Consider reusing the session already in scope rather than acquiring a new one, or explicitly document why a separate session is required here.

backend/foreman/runner.py
```diff
@@ -719,6 +719,53 @@ async def run_foreman_ai(
...
+                                {
+                                    "toolId": tu.id,
+                                    "toolName": tu.name,
+                                    "toolInput": dict(tu.input) if tu.input else {},
+                                }
+                            ),
+                            created_at=_now,
+                        )
+                    )
+                for result in trimmed:
```
jmelloy, line 749: `trimmed` appears to be the trimmed context-window list passed to the AI, not specifically a filtered list of tool-result blocks from the current turn. Iterating over it to persist tool results could persist stale/duplicate results from previous turns that survived context trimming. This should iterate only over tool results belonging to the current `tool_uses` batch, e.g. filtered by `tool_use_id`.

backend/routes/guilds.py
```diff
@@ -190,7 +203,8 @@ async def get_guild(guild_id: str, github_user_id: str = Depends(require_member(
         result = await db.execute(
             select(Message)
             .where(Message.guild_pk == guild_pk)
-            .order_by(Message.created_at.desc())
+            # .id.desc() is a stable tiebreaker because message IDs are auto-increment integers.
+            .order_by(Message.created_at.desc(), Message.id.desc())
             .limit(100)
         )
         messages = result.scalars().all()
```
jmelloy, line 211: Bare `except Exception` silently swallows all JSON parse errors. At minimum re-raise on unexpected exception types, or use `json.JSONDecodeError` specifically, so programming errors (e.g. accidentally storing non-JSON) are not hidden in production.

backend/foreman/runner.py
jmelloy, line 726: Both `tool_use` and `tool_result` messages are written with `created_at=_now` (the same timestamp), so their ordering relative to each other and to the assistant message depends entirely on the new `Message.id.desc()` tiebreaker. Verify that the auto-increment IDs are guaranteed to be assigned in insertion order within the same transaction, otherwise the tool-use/tool-result display order may be non-deterministic after refresh.

jmelloy (commented):
- The core fix is sound: persisting `tool_use` and `tool_result` messages to the DB and merging `meta` JSON on read allows tool activity to survive a page refresh.
- A potential double-commit bug exists: `get_db()` in `runner.py` may return an already-open session that shares state with the surrounding request context, and calling `.close()` in a `finally` block could leave the outer session in an unusable state.
- The `trimmed` variable used for persisting tool results is likely the trimmed context window list, not necessarily a clean list of tool result blocks — this could persist unintended messages or miss results entirely depending on trimming logic.
- The `_message_dict` helper silently swallows JSON parse errors (only logs a warning), which means corrupt `meta` values will silently return incomplete data to the client without any indication.
- Minor: the `from_agent` field added to the `ChatMessage` TypeScript interface was already implicitly covered by `[key: string]: unknown`, but the explicit addition is a welcome improvement for type safety.

jmelloy (commented):
- The core fix is sound: persisting `tool_use` and `tool_result` messages to the DB and merging `meta` JSON on read-back so the frontend can reconstruct tool-call UI after a page refresh.
- A critical bug exists in the DB persistence block: `get_db()` is called without `async with` / generator protocol, likely leaking a connection or returning an unusable session object depending on how `get_db` is implemented.
- The `trimmed` variable (tool results) is iterated for persistence before `_save_turn` is called, but `trimmed` is not defined in the visible diff context — if it refers to a variable scoped outside this block there is a potential ordering/state bug worth verifying.
- The `_message_dict` helper silently swallows JSON parse errors with only a warning, which is acceptable, but it also unconditionally calls `d.update(json.loads(m.meta))` — if `meta` contains top-level keys that clash with existing Message columns (e.g. `id`, `content`) those columns will be silently overwritten in the response.
- The tiebreaker `.order_by(Message.created_at.desc(), Message.id.desc())` is a good defensive improvement, though the comment says IDs are auto-increment integers — this should be verified against the schema since SQLite/Postgres ID generation can differ.

backend/foreman/runner.py
jmelloy, line 726: `get_db()` is almost certainly an async generator (FastAPI dependency pattern). Calling it bare like this returns a generator object, not a session — you need `async with get_db() as _persist_db:` or equivalent. This will silently fail to write anything and may leak connections.

backend/routes/guilds.py
```diff
@@ -190,7 +203,8 @@ async def get_guild(guild_id: str, github_user_id: str = Depends(require_member(
         result = await db.execute(
             select(Message)
             .where(Message.guild_pk == guild_pk)
-            .order_by(Message.created_at.desc())
+            # .id.desc() is a stable tiebreaker because message IDs are auto-increment integers.
+            .order_by(Message.created_at.desc(), Message.id.desc())
             .limit(100)
```
jmelloy, line 209: `d.update(json.loads(m.meta))` will silently overwrite any existing dict key with the same name (e.g. `id`, `content`, `created_at`). Prefer a merge strategy that only adds keys not already present, or explicitly whitelist the meta fields that are safe to surface.

backend/routes/guilds.py
```diff
@@ -190,7 +203,8 @@ async def get_guild(guild_id: str, github_user_id: str = Depends(require_member(
         result = await db.execute(
```
jmelloy, line 204: The new tiebreaker `Message.id.desc()` is correct only if `id` is strictly monotonic. If the schema uses UUIDs or a non-sequential generator, this will produce non-deterministic ordering for messages with identical `created_at` timestamps — worth asserting in a comment or test.

backend/foreman/runner.py
jmelloy, line 726: A brand-new DB session is opened with `get_db()` here instead of reusing the session that should already be available in this function's scope. This means the tool-use persist and the surrounding `_save_turn` call run in separate sessions/transactions, defeating the 'single transaction' comment above and doubling connection overhead.

backend/foreman/runner.py
```diff
@@ -719,6 +719,53 @@ async def run_foreman_ai(
...
+                    _persist_db.add(
```
jmelloy: `result.get('content', '')` can silently produce an empty string when `content` is a list of Anthropic content blocks (the common case for tool results). Consider serialising non-string content with `json.dumps` so the actual result text survives the round-trip.

backend/foreman/runner.py
```diff
@@ -719,6 +719,53 @@ async def run_foreman_ai(
...
+            try:
+                _persist_guild_pk = await get_guild_pk(_persist_db, guild_id)
+                for tu in tool_uses:
+                    _persist_db.add(
+                        Message(
+                            guild_pk=_persist_guild_pk,
+                            from_agent="foreman",
+                            to_agent="user",
+                            content=f"▶ {tu.name}",
+                            message_type="chat",
```
jmelloy, line 737: `dict(tu.input) if tu.input else {}` will raise if `tu.input` is not a mapping (e.g. a list-typed tool input). Use `json.loads(json.dumps(tu.input))` or a proper Pydantic `.model_dump()` call to safely convert the input to a JSON-serialisable dict.

backend/routes/guilds.py
```diff
@@ -190,7 +203,8 @@ async def get_guild(guild_id: str, github_user_id: str = Depends(require_member(
...
             select(Message)
             .where(Message.guild_pk == guild_pk)
-            .order_by(Message.created_at.desc())
+            # .id.desc() is a stable tiebreaker because message IDs are auto-increment integers.
+            .order_by(Message.created_at.desc(), Message.id.desc())
```
jmelloy, line 208: `d.update(json.loads(m.meta))` will silently overwrite first-class columns (e.g. `id`, `content`, `created_at`) if the stored JSON happens to contain the same keys. Prefer a targeted merge or a dedicated key namespace like `d['meta'] = parsed` to avoid accidental clobbering.

jmelloy (commented):
- The core fix is sound: tool_use and tool_result messages are now persisted to the DB with a new `role` and `meta` column, and the REST endpoint merges that JSON back on read so the frontend can reconstruct tool-call UI after a page refresh.
- A new DB session is opened inside the loop via `get_db()` rather than reusing the already-available session context, which is wasteful and potentially inconsistent with the surrounding transaction.
- The `content` field for `tool_result` rows uses `result.get('content', '')` but the trimmed list may contain blocks where `content` is a list of dicts (Anthropic's format), not a string, meaning it could be silently stored as an empty string or raise a type error.
- The `_message_dict` helper silently swallows all JSON parse errors with only a warning, which is acceptable defensively but means corrupt `meta` rows will surface as messages missing their tool fields with no client-visible error.
- The tiebreaker ordering change (`.order_by(created_at.desc(), id.desc())`) is a good correctness improvement since multiple tool messages can share the same `_now` timestamp.

backend/foreman/runner.py
```diff
@@ -719,6 +719,53 @@ async def run_foreman_ai(
                         "createdAt": _now,
                     },
                 )
+
+            # Persist tool-use and tool-result messages in a single transaction so
+            # they succeed or fail together — no partial writes if exec_tools raised.
```
jmelloy: `get_db()` is called here to open a brand-new session just for persisting tool messages, independently of whatever DB session the surrounding code uses. This pattern risks connection leaks if an exception occurs before the `finally` block is reached (e.g., during `get_guild_pk`), and it is inconsistent with the rest of the function. Consider reusing the existing DB session or using an `async with` context manager instead of manual `try/finally`.

backend/foreman/runner.py
```diff
@@ -719,6 +719,53 @@ async def run_foreman_ai(
...
+                    _persist_db.add(
+                        Message(
+                            guild_pk=_persist_guild_pk,
+                            from_agent="foreman",
+                            to_agent="user",
+                            content=result.get("content", "") or "",
+                            message_type="chat",
+                            role="tool_result",
+                            meta=json.dumps(
+                                {
```
jmelloy, line 759: Tool-result messages are persisted by iterating `trimmed` rather than the original `tool_results`. If the context-trimming logic removed any results, those tool_use records will have been saved without a corresponding tool_result, creating orphaned records and potential UI inconsistencies. Verify that `trimmed` always contains all results that correspond to the `tool_uses` being persisted.

backend/routes/guilds.py
jmelloy, line 211: The `except Exception` silently swallows all JSON parse errors for `meta`, only emitting a warning log. A corrupted `meta` value will silently produce an incomplete message dict (missing toolId, toolName, etc.) returned to the client. Consider at minimum returning a sentinel field like `metaParseError: true` so the frontend can distinguish a missing field from a failed parse.

backend/foreman/runner.py
```diff
@@ -719,6 +719,53 @@ async def run_foreman_ai(
...
+                            to_agent="user",
+                            content=f"▶ {tu.name}",
+                            message_type="chat",
+                            role="tool_use",
+                            meta=json.dumps(
+                                {
+                                    "toolId": tu.id,
+                                    "toolName": tu.name,
+                                    "toolInput": dict(tu.input) if tu.input else {},
+                                }
```
jmelloy, line 744: Setting `content=f'▶ {tu.name}'` embeds a UI rendering detail (the `▶` character) directly into the persisted database record. This couples the data layer to a specific presentation format; if the frontend rendering changes, historical DB records will be inconsistent. Consider storing the raw tool name in `content` and leaving the `▶` prefix to the frontend renderer.

jmelloy (commented):
- The fix correctly adds `role` and `meta` columns to persist tool-use/tool-result messages across page refreshes, with a corresponding Alembic migration.
- A resource-management concern exists in `runner.py`: `get_db()` is called unconditionally every tool-use cycle, creating a new DB session that is closed in a `finally` block but could leak if the session creation itself fails or if the pattern is inconsistent with how the rest of the function acquires DB connections.
- The `_message_dict` helper silently swallows JSON parse errors (only logging a warning), which means corrupted `meta` data will result in silently incomplete message objects returned to clients — this could mask data integrity issues.
- Tool-result messages iterate over `trimmed` (which appears to be the context-window-trimmed list) rather than the full `tool_results` list; if trimming removed some results, those results will be persisted to DB but the corresponding tool-use records may not match, creating orphaned/inconsistent records.
- The `content` field for `tool_use` messages is set to a display string (`▶ {tu.name}`) rather than the actual tool input, mixing presentation concerns into the data layer; this could cause issues if `content` is ever used programmatically.

backend/foreman/runner.py
```diff
@@ -695,6 +692,8 @@ async def run_foreman_ai(
                     },
```
jmelloy, line 694: `_tool_use_ts = _now` is set before `exec_tools`, but the persistence block that uses it runs *after* `exec_tools` returns. If `exec_tools` raises, neither tool_use nor tool_result rows are written (per the stated intent). However the comment says 'capture before exec_tools may raise' which is misleading — the timestamp is captured but the rows are not written on failure. Consider clarifying the comment or persisting tool_use rows before calling exec_tools if you want them in history regardless of outcome.

backend/foreman/runner.py
```diff
@@ -839,6 +879,8 @@ async def run_foreman_ai(
                 "createdAt": now,
```
jmelloy, line 884: The `finally: await db.close()` now correctly closes the DB in all paths for the main function body, but the early queries (lines ~522-548, guild/worker/task fetches) are no longer wrapped in their own try/finally. If any of those queries raise before the outer try block is entered, `db` will be leaked. Either keep a `try/finally` around the pre-loop setup or open the DB inside the existing outer `try`.

backend/foreman/runner.py
```diff
@@ -719,6 +718,52 @@ async def run_foreman_ai(
...
+                            from_agent="foreman",
+                            to_agent="user",
+                            content=result.get("content", "") or "",
+                            message_type="chat",
+                            role="tool_result",
+                            meta=json.dumps(
+                                {
+                                    "toolId": result.get("tool_use_id"),
+                                    "isError": result.get("is_error", False),
+                                }
```
jmelloy, line 762: The `except Exception: await db.rollback(); raise` block is effectively a no-op — SQLAlchemy async sessions auto-rollback on error, and re-raising means the outer exception handler will also attempt operations on the session. The explicit rollback is harmless but unnecessary noise; consider removing it or at minimum adding a log statement so failed tool-persistence is observable.

backend/routes/guilds.py
jmelloy, line 209: The `_message_dict` helper calls `d.update(json.loads(m.meta))` which will silently overwrite any existing top-level key in the serialized row (e.g. `id`, `content`, `role`) if `meta` JSON happens to contain the same keys. Consider merging under a dedicated namespace or explicitly whitelisting the keys extracted from `meta`.

jmelloy (commented):
- The core fix is sound: `role` and `meta` columns are added to persist tool_use/tool_result blocks so they survive a page refresh, with a proper migration and model update.
- A significant resource-management improvement moves `db.close()` to a `finally` block, but the refactor removes the original `try/finally` guard around the early DB queries (guild, worker, task fetches), meaning exceptions there will leak the connection.
- Tool-use rows are persisted **after** `exec_tools` completes (due to `_tool_use_ts` placement and the single transaction block), so if `exec_tools` raises, tool_use rows are never written — this is intentional per the comment, but means failed tool calls are invisible in history.
- The `_message_dict` helper silently swallows JSON parse errors with only a warning, which is reasonable, but bare `except Exception` in the migration block in `runner.py` re-raises anyway, making the catch redundant.
- Frontend `ChatMessage` type extension is minimal and correct; the index tiebreaker added to the messages query (`Message.id.desc()`) improves ordering stability.

PR #408, opened 2026-05-20
Fix backend test failure on PR #405

jmelloy:
## Summary

- PR #405 added a new Alembic migration (`20260520_000001_add_role_and_meta_to_messages`) that branched off `20260520_000000_add_index_worker_last_seen` — the same parent as `20260520_000001_add_task_locking` (merged via #406)
- This created two Alembic heads, causing `alembic.util.exc.CommandError: Multiple head revisions` in every test that ran `alembic upgrade head` during setup (209 errors + 1 failed test)
- Fix: renamed the PR #405 migration to `20260520_000002_add_role_and_meta_to_messages` and updated its `down_revision` to chain after `20260520_000001_add_task_locking`

Also carries the full set of PR #405 application changes:
- `backend/foreman/runner.py`: keep DB session open for entire foreman run; persist `tool_use` / `tool_result` `Message` rows atomically so they survive a page refresh
- `backend/models.py`: add `role` and `meta` columns to `Message`
- `backend/routes/guilds.py`: extend messages REST endpoint to return `role` and `meta`
- `frontend/src/types.ts`: add matching type fields

## Test plan
- [x] All 413 backend tests pass locally after the fix
- [x] `ruff check . --fix && ruff format .` — no issues

### DIFF
--- a/backend/alembic/versions/20260520_000002_add_role_and_meta_to_messages.py
+++ b/backend/alembic/versions/20260520_000002_add_role_and_meta_to_messages.py
@@ -1,0 +1,33 @@
+"""add_role_and_meta_to_messages
+
+Persist tool_use / tool_result chat blocks so they survive a page refresh.
+The `role` column mirrors the WebSocket message role field; `meta` stores
+JSON with extra fields (toolId, toolName, toolInput, isError).
+
+Revision ID: 20260520_000002_add_role_and_meta_to_messages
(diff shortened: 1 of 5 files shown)


jmelloy:
Closing — the migration fix has been applied directly to PR #405's branch (claude/fix-foreman-chat-show-tool-usage-after-page-refresh-t-ghom).

PR #409, opened 2026-05-20
Convert app to PostgreSQL + Docker Compose

### LINKED ISSUE #407: Convert database to PostgreSQL via Docker Compose
## Summary
The app currently uses SQLite. This issue tracks migrating it to PostgreSQL, with the Postgres service managed via Docker Compose.

## Work to do
- Add a `postgres` service to `docker-compose.yml` (image `postgres:16`, named volume for persistence, health-check)
- Update `DATABASE_URL` / connection settings to point at the Dockerized Postgres instance
- Replace any SQLite-specific SQL / dialect hints in SQLAlchemy models with Postgres-compatible equivalents (e.g. `SERIAL`/`BIGSERIAL` → `Integer` autoincrement, JSON columns, etc.)
- Create an Alembic migration (or equivalent) that sets up the schema on a fresh Postgres instance
- Update `README` / dev-setup instructions to reflect the new `docker compose up` workflow
- Update CI to spin up the Postgres container (or use a `postgres` service in the GitHub Actions workflow)
- Remove or gate the SQLite fallback so the app fails fast if `DATABASE_URL` is not set correctly

jmelloy:
## Summary

- Add `postgres:16` service to `docker-compose.yml` with health-check, named volume, and `.env`-based credentials; app service depends on it with `condition: service_healthy`
- Switch `DATABASE_URL` to `postgresql+asyncpg://` (reads from env); remove all `aiosqlite` / SQLite-specific code
- Update `backend/database.py` to use async SQLAlchemy with asyncpg; remove `check_same_thread` and `aiosqlite` engine args
- Update `backend/requirements.txt`: drop `aiosqlite`, add `asyncpg` + `psycopg2-binary`
- Audit and migrate models: `Text` columns for unbounded strings, `JSONB` for JSON columns, removed raw SQLite pragmas
- Add Alembic migration for fresh PostgreSQL schema creation; update `alembic.ini` and `env.py` to read `DATABASE_URL` from env
- Update `backend/tests/conftest.py` to use `TEST_DATABASE_URL` against a real PostgreSQL instance
- Update `backend/tests/helpers.py`: replace `sqlite3` helpers with `psycopg2` helpers (`truncate_all`, `raw_conn`, etc.)
- Update `README.md` dev-setup instructions: `docker compose up -d` to start Postgres before running the app
- Fix logging config: replace `aiosqlite` logger with `asyncpg` in suppressed loggers list

## CI workflow change (needs manual apply)

The GitHub OAuth token lacked `workflow` scope so `.github/workflows/ci.yml` could not be pushed. Apply this block **after the `defaults:` section** in the `backend-tests` job manually:

```yaml
    services:
      postgres:
        image: postgres:16
        env:
          POSTGRES_DB: pioneer_test
          POSTGRES_USER: pioneer
          POSTGRES_PASSWORD: pioneer_password
        options: >-
          --health-cmd pg_isready
          --health-interval 5s
          --health-timeout 5s
          --health-retries 5
        ports:
          - 5432:5432

    env:
      TEST_DATABASE_URL: postgresql+asyncpg://pioneer:pioneer_password@localhost/pioneer_test
```

## Test plan

- [ ] `docker compose up -d postgres` starts healthy; `docker compose up backend` connects successfully
- [ ] `alembic upgrade head` runs cleanly against the Postgres instance
- [ ] `python -m pytest` in `backend/` passes with `TEST_DATABASE_URL` set
- [ ] Apply the CI diff above and verify the workflow runs with a Postgres service container

Closes #407


backend/database.py
```diff
@@ -2,8 +2,10 @@
 
 from sqlalchemy.ext.asyncio import AsyncSession, async_sessionmaker, create_async_engine
 
-DB_PATH = os.environ.get("DB_PATH", "pioneer_square.db")
-DATABASE_URL = os.environ.get("DATABASE_URL", f"sqlite+aiosqlite:///{DB_PATH}")
+DATABASE_URL = os.environ.get(
```
jmelloy: Hardcoded credentials in the default fallback `postgresql+asyncpg://pioneer:pioneer_password@localhost/pioneer_square` will be used in any environment where `DATABASE_URL` is not set, including production containers that have a misconfigured env. Remove the fallback entirely and raise a clear error if the variable is absent.

backend/alembic.ini
```diff
@@ -85,7 +85,7 @@ path_separator = os
 # database URL.  This is consumed by the user-maintained env.py script only.
 # other means of configuring database URLs may be customized within the env.py
 # file.
-sqlalchemy.url = sqlite+aiosqlite:///pioneer_square.db
+sqlalchemy.url = postgresql+asyncpg://pioneer:pioneer_password@localhost/pioneer_square
```
jmelloy: Literal credentials `pioneer:pioneer_password` are committed into `alembic.ini`. Since `env.py` already reads `DATABASE_URL` from the environment, this fallback is never needed and should be replaced with a placeholder like `postgresql+asyncpg://user:pass@host/dbname` or left blank.

backend/tests/conftest.py
```diff
@@ -22,37 +23,46 @@
...
 
     with TestClient(main_module.app) as c:
-        yield c, db_path
+        yield c, db_url
 
-    # Restore DATABASE_URL after test
-    os.environ.pop("DATABASE_URL", None)
+    # NullPool means connections are not pooled; disposal is a no-op but kept
+    # for clarity.
+    asyncio.run(new_engine.dispose())
```
jmelloy: `asyncio.run(new_engine.dispose())` will raise `RuntimeError: This event loop is already running` if an asyncio loop is active at teardown time (common with anyio/pytest-asyncio). With `NullPool`, connections are closed immediately on return so `dispose()` is effectively a no-op — remove this call or use `asyncio.get_event_loop().run_until_complete` guarded by a loop-state check.

backend/tests/helpers.py
```diff
@@ -2,101 +2,140 @@
...
+    """Truncate all application tables and restart sequences."""
+    import psycopg2
+
+    sync_url = db_url.replace("postgresql+asyncpg://", "postgresql://")
+    conn = psycopg2.connect(sync_url)
+    conn.autocommit = True
+    cur = conn.cursor()
+    cur.execute("""
+        TRUNCATE TABLE
+            github_events, foreman_turns, task_logs, guild_keys,
```
jmelloy: `truncate_all` hardcodes every table name; any new table added to the schema will not be truncated during tests, silently allowing state leakage. Consider using `information_schema.tables` or `pg_catalog` to discover and truncate all non-system tables dynamically, or at minimum add a comment warning maintainers to update this list.

backend/tests/test_agent_state_ws_persistence.py
```diff
@@ -31,18 +30,20 @@
...
-    tmp_path = tmp_path_factory.mktemp("agent_state_ws_db")
-    db_path = str(tmp_path / "test.db")
-    db_url = f"sqlite+aiosqlite:///{db_path}"
 
-    os.environ["DATABASE_URL"] = db_url
-    _create_db(db_path)
+@pytest.fixture(scope="module")
+def client():
+    _create_db(TEST_DATABASE_URL)
+    truncate_all(TEST_DATABASE_URL)
```
jmelloy, line 40: The module-scoped `client` fixture calls `_create_db` then `truncate_all` once, but individual tests within the module all share the same live database without truncation between them. Test ordering matters here — if `test_agent_state_persists_current_task_id` leaves agent rows behind, subsequent tests querying the same `agent_id` keys could see stale data or unique-constraint failures.

jmelloy (commented):
- Hardcoded credentials (`pioneer:pioneer_password`) appear in multiple files including `database.py`, `alembic.ini`, `conftest.py`, and test files — these should always be read from environment variables with no fallback defaults in production code
- Test isolation is weakened significantly: the module-scoped `client` fixture in `test_agent_state_ws_persistence.py` shares state with the session-scoped schema setup, and `truncate_all` is not called between individual tests within that module, risking cross-test contamination
- `asyncio.run(new_engine.dispose())` in `conftest.py` teardown can conflict with an already-running event loop (e.g. if pytest-asyncio is active), and `NullPool` makes `dispose()` a no-op anyway — this teardown is misleading
- `truncate_all` uses a hardcoded table list with a fixed ordering; any future table additions will silently skip truncation, and the function has no error handling around the cursor/connection lifecycle
- The `ON CONFLICT DO NOTHING` pattern in `insert_guild` inside `helpers.py` can silently succeed without inserting when a conflict occurs, then attempt to fetch the guild PK — if the conflicting row belongs to a different test's data, `guild_pk` will be wrong and downstream inserts will be incorrect

jmelloy (commented):
- Migration files use raw SQL with `SERIAL PRIMARY KEY` syntax, but these are historical migration files that previously used `AUTOINCREMENT`; the `downgrade()` paths now also use `SERIAL` instead of `INTEGER NOT NULL PRIMARY KEY AUTOINCREMENT`, making rollbacks broken since they reference the old SQLite syntax but now emit PostgreSQL syntax.
- Test isolation is weakened: `truncate_all` runs on the single shared `TEST_DATABASE_URL` database, meaning parallel test runs or CI matrix jobs would interfere with each other; the `client` fixture in `test_agent_state_ws_persistence.py` calls `create_db` on every function invocation which is unnecessary and slow.
- The `conftest.py` `client` fixture creates a new `AsyncEngine` with `NullPool` pointing at the shared test database but `truncate_all` uses a separate psycopg2 connection with `autocommit=True`, creating a potential race if any connection holds locks during truncation.
- Hardcoded default credentials (`pioneer_password`) appear in `.env.example`, `ci.yml`, and `_test_config.py` fallback logic; while `.env.example` is intentionally illustrative, the CI workflow embeds these same weak credentials directly in the workflow file.
- `psycopg2-binary` is added to `requirements.txt` (used only for test helpers) but will be installed in the production Docker image, adding unnecessary weight and a potential security surface; test dependencies should be separated.

backend/alembic/versions/20260512_000002_drop_guild_id_child_tables.py
```diff
@@ -328,7 +328,7 @@ def downgrade() -> None:
     op.execute(
```
jmelloy: The `downgrade()` function now uses `SERIAL PRIMARY KEY` for the old tables, but these are supposed to recreate the pre-migration SQLite-style schema. This means `downgrade()` is now broken — it will fail on PostgreSQL if ever run, and the semantic intent (restoring the previous state) is lost. The old `AUTOINCREMENT` syntax was SQLite-specific, but the correct PostgreSQL equivalent for a plain serial PK in a downgrade path should at minimum be documented or the downgrade marked as non-reversible.

backend/tests/helpers.py
```diff
@@ -2,101 +2,144 @@
...
+    conn.autocommit = True
+    cur = conn.cursor(cursor_factory=psycopg2.extras.RealDictCursor)
     try:
-        command.upgrade(cfg, "head")
+        cur.execute("SELECT tablename FROM pg_tables WHERE schemaname = 'public'")
+        tables = [row["tablename"] for row in cur.fetchall()]
+        if tables:
+            cur.execute(
+                "TRUNCATE TABLE {} RESTART IDENTITY CASCADE".format(
+                    ", ".join(f'"{t}"' for t in tables)
```
jmelloy, line 53: Using `TRUNCATE ... RESTART IDENTITY CASCADE` with `autocommit=True` will fail if any other connection holds a lock on these tables (e.g., the NullPool engine from the test fixture hasn't fully released connections). Consider adding explicit connection disposal before truncation, or use `DELETE FROM` which is less lock-sensitive.

backend/tests/test_agent_state_ws_persistence.py
```diff
@@ -56,47 +54,41 @@ async def _stubbed_reset_connection_state() -> None:
 
     mp.setattr(main_module, "reset_connection_state", _stubbed_reset_connection_state)
     mp.setenv("DATABASE_URL", db_url)
-    mp.setenv("DB_PATH", db_path)
 
     with TestClient(main_module.app) as c:
-        yield c, db_path
+        yield c, db_url
```
jmelloy, line 57: This fixture calls `create_db(TEST_DATABASE_URL)` on every test function invocation, which re-runs all Alembic migrations each time. Since the session-scoped `_setup_schema` fixture in `conftest.py` already handles this, this call is redundant and will slow down the test suite. Remove the `create_db` call and rely on the session fixture.

.github/workflows/ci.yml
```diff
@@ -32,6 +32,24 @@ jobs:
     defaults:
       run:
         working-directory: backend
+    services:
+      postgres:
+        image: postgres:16
+        env:
+          POSTGRES_DB: pioneer_test
+          POSTGRES_USER: pioneer
```
jmelloy, line 40: The `TEST_DATABASE_URL` contains the password in plaintext in the workflow file. While this is a test-only database, it sets a poor precedent. Consider using a GitHub Actions secret (e.g., `${{ secrets.TEST_DB_PASSWORD }}`) even for test credentials to establish consistent security practices.

backend/tests/conftest.py
```diff
@@ -20,39 +20,46 @@
...
+def client(monkeypatch, _setup_schema):
+    """Fresh test DB (tables truncated) + TestClient for each test."""
+    from helpers import truncate_all
 
-    # Create DB tables synchronously before the TestClient starts.
-    # This avoids any asyncio.to_thread calls during startup
-    # which can deadlock under pytest's anyio event loop.
-    os.environ["DATABASE_URL"] = db_url
-    _create_db(db_path)
+    truncate_all(TEST_DATABASE_URL)
```
jmelloy, line 49: The `_setup_schema` fixture has `autouse=True` with session scope, but individual test modules (e.g., `test_agent_state_ws_persistence.py`) also call `create_db` independently. This creates a redundant and potentially conflicting double-migration path. The autouse fixture should be the single authority for schema setup.

jmelloy (commented):
- Migration is architecturally sound: SQLite-specific constructs (AUTOINCREMENT, sqlite_insert, aiosqlite, check_same_thread, render_as_batch) are all correctly replaced with PostgreSQL equivalents
- Security concern: default credentials in `.env.example` (`pioneer_password`) may be copied verbatim to production; the example should use obviously-placeholder values
- Test isolation is weakened: the `client` fixture in `conftest.py` shares a single PostgreSQL database across all tests using TRUNCATE, but the module-scoped fixture in `test_agent_state_ws_persistence.py` was downgraded to function-scoped without a clear rationale, and both call `create_db` redundantly (migrations run twice)
- The `SERIAL` type used in raw SQL migration strings is a PostgreSQL legacy alias; modern PostgreSQL (10+) prefers `GENERATED ALWAYS AS IDENTITY`, though `SERIAL` still works correctly
- `psycopg2-binary` in `requirements.txt` is only needed for test helpers (`truncate_all`, `raw_conn`) but is installed in the production image, adding unnecessary weight and a known production-use caveat from the psycopg2 authors

backend/tests/helpers.py
```diff
@@ -2,101 +2,144 @@
...
+                )
+            )
     finally:
-        engine.dispose()
+        cur.close()
+        conn.close()
+
```
jmelloy, line 60: The URL replacement `replace('postgresql+asyncpg://', 'postgresql://')` is fragile: if the scheme is ever `postgresql+psycopg2://` or uses `postgres://` (common on Heroku/Railway), the replace is a no-op and psycopg2.connect will receive an asyncpg URL it cannot parse. Use `urllib.parse` or `make_url` from SQLAlchemy to swap the driver component reliably.

backend/tests/test_agent_state_ws_persistence.py
```diff
@@ -31,18 +30,17 @@
 
 import database as database_module  # noqa: E402
 import main as main_module  # noqa: E402
+from _test_config import TEST_DATABASE_URL  # noqa: E402
 from helpers import create_db as _create_db  # noqa: E402
+from helpers import raw_conn, truncate_all
 from starlette.testclient import TestClient  # noqa: E402
```
jmelloy, line 36: This local fixture calls `create_db(TEST_DATABASE_URL)` directly, duplicating the session-scoped `_setup_schema` fixture in `conftest.py`. If both run in the same pytest session the migrations will be applied twice (which is idempotent but wasteful and confusing). Remove the `create_db` call here and rely on the session-scoped fixture instead.

backend/tests/conftest.py
```diff
@@ -20,39 +20,46 @@
...
+    db_url = TEST_DATABASE_URL
 
     new_engine = create_async_engine(db_url, echo=False, poolclass=NullPool)
     new_session = async_sessionmaker(new_engine, expire_on_commit=False, class_=AsyncSession)
```
jmelloy, line 53: Using `truncate_all` with `RESTART IDENTITY CASCADE` between every test means all tests share the same physical PostgreSQL database. A test that leaves a transaction open (e.g. due to a crash) will block the next test's truncation. Consider wrapping each test in a savepoint/transaction that is rolled back instead, or at minimum add a statement timeout to the truncation connection.

.env.example
```diff
@@ -1,5 +1,10 @@
 # Copy to .env and fill in. Used by docker-compose.
 
+# --- PostgreSQL (managed by docker compose) ---
+POSTGRES_DB=pioneer_square
```
jmelloy, line 4: The default password `pioneer_password` is weak and descriptive enough to be guessed. `.env.example` files are frequently copied as-is; use a clearly-placeholder value like `CHANGE_ME` or `<your-secure-password>` to reduce the risk of accidental production deployment with a known credential.

backend/requirements.txt
```diff
@@ -2,7 +2,8 @@ cryptography
 fastapi
 uvicorn[standard]
 websockets
-aiosqlite
+asyncpg
```
jmelloy, line 5: `psycopg2-binary` is only used by test helpers (`truncate_all`, `raw_conn`). Installing it in the production Docker image adds ~10 MB and the psycopg2 project explicitly warns against using the binary package in production. Move it to a `requirements-test.txt` (or a `[test]` extras group) and install it only in test/CI environments.

backend/tests/conftest.py
```diff
@@ -20,39 +20,46 @@
...
+    create_db(TEST_DATABASE_URL)
 
 
 @pytest.fixture()
-def client(tmp_path, monkeypatch):
-    """TestClient backed by a fresh temporary SQLite database."""
-    db_path = str(tmp_path / "test.db")
-    db_url = f"sqlite+aiosqlite:///{db_path}"
+def client(monkeypatch, _setup_schema):
+    """Fresh test DB (tables truncated) + TestClient for each test."""
```
jmelloy, line 46: `truncate_all` runs with `autocommit=True` against the shared `TEST_DATABASE_URL` — if tests ever run in parallel (e.g. via `pytest-xdist`) this will silently truncate data mid-test for other workers. Consider per-test schemas or transaction-level rollback isolation instead.

jmelloy (commented):
- Migration is architecturally sound: Docker Compose health-check dependency, asyncpg engine, and dialect-specific upserts are all correctly updated
- **Critical test isolation issue**: `test_agent_state_ws_persistence.py` uses `scope='function'` but calls `_create_db` every invocation, and the shared `TEST_DATABASE_URL` in `conftest.py` means concurrent or sequential tests can stomp on each other via `truncate_all` without proper transaction isolation
- **Security**: Hardcoded default credentials (`pioneer_password`) in `.env.example` are fine as examples, but the CI workflow hardcodes the same password in plain YAML; secrets should use GitHub Actions secrets even for test DBs
- The downgrade migrations in `20260512_000002` still use `SERIAL PRIMARY KEY` syntax for the rollback tables, which is correct for PostgreSQL but inconsistent with using `BIGSERIAL` or `IDENTITY` — minor but worth standardizing
- `psycopg2-binary` is listed in `requirements.txt` alongside `asyncpg`; it is only used in test helpers, so it should be in a separate `requirements-test.txt` to avoid shipping it in the production image

backend/tests/helpers.py
```diff
@@ -2,101 +2,144 @@
...
+    import psycopg2
+    import psycopg2.extras
+
+    sync_url = db_url.replace("postgresql+asyncpg://", "postgresql://")
+    conn = psycopg2.connect(sync_url)
+    conn.autocommit = True
+    cur = conn.cursor(cursor_factory=psycopg2.extras.RealDictCursor)
     try:
-        command.upgrade(cfg, "head")
+        cur.execute("SELECT tablename FROM pg_tables WHERE schemaname = 'public'")
```
jmelloy: The URL replacement `db_url.replace('postgresql+asyncpg://', 'postgresql://')` is fragile and will silently produce a broken URL if the scheme is ever `postgresql+psycopg2://` or similar. Use `urllib.parse` or `sqlalchemy.engine.make_url` to swap the driver component reliably.

.github/workflows/ci.yml
```diff
@@ -32,6 +32,24 @@ jobs:
...
+          POSTGRES_PASSWORD: pioneer_password
+        options: >-
+          --health-cmd pg_isready
+          --health-interval 5s
```
jmelloy, line 44: The Postgres password `pioneer_password` is hardcoded in the workflow YAML. Even for ephemeral CI databases this should use a GitHub Actions secret (e.g. `${{ secrets.TEST_POSTGRES_PASSWORD }}`) to avoid normalising plaintext credentials in source-controlled workflow files.

backend/tests/test_agent_state_ws_persistence.py
```diff
@@ -31,18 +30,17 @@
...
-    db_path = str(tmp_path / "test.db")
-    db_url = f"sqlite+aiosqlite:///{db_path}"
-
-    os.environ["DATABASE_URL"] = db_url
-    _create_db(db_path)
+@pytest.fixture(scope="function")
+def client():
+    _create_db(TEST_DATABASE_URL)
+    truncate_all(TEST_DATABASE_URL)
+    db_url = TEST_DATABASE_URL
```
jmelloy, line 42: `_create_db(TEST_DATABASE_URL)` is called on every `function`-scoped fixture invocation. If migrations have already been applied this is wasteful and potentially error-prone; the schema creation should stay in the `session`-scoped `_setup_schema` fixture in `conftest.py` and only `truncate_all` should be called here.

backend/requirements.txt
```diff
@@ -2,7 +2,8 @@ cryptography
...
+psycopg2-binary
```
jmelloy, line 6: `psycopg2-binary` is a runtime dependency only needed by test helpers (`helpers.py`). Including it in the main `requirements.txt` means it is installed in the production Docker image unnecessarily; move it to a `requirements-test.txt` or similar.

backend/tests/helpers.py
```diff
@@ -2,101 +2,144 @@
...
+            cur.execute(
+                "TRUNCATE TABLE {} RESTART IDENTITY CASCADE".format(
+                    ", ".join(f'"{t}"' for t in tables)
+                )
+            )
     finally:
-        engine.dispose()
+        cur.close()
+        conn.close()
+
```
jmelloy, line 59: The `truncate_all` function uses `conn.autocommit = True` and executes TRUNCATE directly, but if the cursor factory or connection setup raises, `conn` may not be closed. Consider wrapping the connection itself in a try/finally or using a context manager to ensure cleanup on error.

backend/tests/test_agent_state_ws_persistence.py
```diff
@@ -31,18 +30,17 @@
 
 import database as database_module  # noqa: E402
 import main as main_module  # noqa: E402
+from _test_config import TEST_DATABASE_URL  # noqa: E402
 from helpers import create_db as _create_db  # noqa: E402
+from helpers import raw_conn, truncate_all
 from starlette.testclient import TestClient  # noqa: E402
```
jmelloy, line 35: `create_db` (which runs all Alembic migrations) is called on every test invocation because the fixture scope was changed from `module` to `function`. Migrations are idempotent but this is unnecessary overhead; the schema setup should remain session- or module-scoped while only `truncate_all` runs per function.

backend/requirements.txt
jmelloy, line 6: `psycopg2-binary` is only used by test helpers (`helpers.py`) for direct DB access; including it in the production `requirements.txt` adds an unnecessary C-extension dependency to the runtime image. Move it to a `requirements-test.txt` or `requirements-dev.txt`.

.github/workflows/ci.yml
jmelloy, line 44: The `TEST_DATABASE_URL` contains the plaintext password `pioneer_password` hardcoded in the workflow file. Even for a test database this is poor practice and will appear in public CI logs; use a GitHub Actions secret or at minimum an environment variable referenced from a secret.

backend/tests/conftest.py
```diff
@@ -20,39 +20,46 @@
...
+
+    create_db(TEST_DATABASE_URL)
 
 
 @pytest.fixture()
-def client(tmp_path, monkeypatch):
-    """TestClient backed by a fresh temporary SQLite database."""
-    db_path = str(tmp_path / "test.db")
-    db_url = f"sqlite+aiosqlite:///{db_path}"
+def client(monkeypatch, _setup_schema):
```
jmelloy, line 45: `truncate_all` truncates all tables including the `alembic_version` table, which will cause the next `create_db` / `alembic upgrade head` call to re-run all migrations from scratch rather than recognising the schema is current. Add the alembic version table to an exclusion list in `truncate_all`.

jmelloy (commented):
- Migration is technically sound: SQLite-specific constructs (`aiosqlite`, `sqlite_insert`, `render_as_batch`, `check_same_thread`) are all removed and replaced with PostgreSQL equivalents
- **Security concern**: Default credentials in `.env.example` (`pioneer_password`) are weak and likely to be copied verbatim into production; the CI workflow also hardcodes the same password in plain text
- **Test isolation regression**: The `client` fixture in `conftest.py` uses `scope='function'` but shares a single database via `truncate_all`, meaning parallel test execution would cause races; the module-scoped fixture in `test_agent_state_ws_persistence.py` was changed to `scope='function'` which calls `create_db` redundantly on every test
- The `downgrade()` function in migration `20260512_000002` uses `SERIAL PRIMARY KEY` for the rollback tables, which is incorrect — `SERIAL` is PostgreSQL syntax but the downgrade path should restore the original SQLite-style tables; since SQLite is being dropped this is a minor dead-code issue, but it signals the downgrade path is broken
- `psycopg2-binary` is added to `requirements.txt` as a runtime dependency solely for test helpers; it should live in a test-only requirements file to avoid bloating the production image

backend/tests/helpers.py
```diff
@@ -2,101 +2,144 @@
...
     token = "test-session-" + secrets.token_hex(8)
     now = datetime.now(UTC).isoformat()
-    with sqlite3.connect(db_path) as conn:
-        conn.execute(
-            "INSERT OR IGNORE INTO github_tokens "
-            "(github_user_id, github_username, access_token, token_type, scope, "
-            "created_at, updated_at) VALUES (?, ?, ?, ?, ?, ?, ?)",
+    with raw_conn(db_url) as (conn, cur):
+        cur.execute(
+            "INSERT INTO github_tokens "
```
jmelloy, line 94: If the `ON CONFLICT ... DO UPDATE` branch is taken (guild already exists), `RETURNING id` still returns the row so `guild_pk` will be set correctly. However if there is an unexpected error and `row` is `None`, the function silently skips inserting the owner member without raising, which can cause confusing test failures downstream. Consider asserting `guild_pk is not None` here.

backend/tests/test_agent_state_ws_persistence.py
jmelloy, line 36: This fixture calls `_create_db` (which runs all Alembic migrations) on every test function invocation. Since the session-scoped `_setup_schema` fixture in `conftest.py` already handles this, the `_create_db` call here is redundant and adds latency; the `truncate_all` call is sufficient for per-test isolation.

.github/workflows/ci.yml
jmelloy, line 44: The `TEST_DATABASE_URL` contains the plaintext password `pioneer_password` directly in the workflow file. Even for a test/non-production credential this sets a poor precedent. Consider using a GitHub Actions secret (e.g. `${{ secrets.TEST_DB_PASSWORD }}`) or at minimum a randomly generated ephemeral credential.

backend/tests/helpers.py
jmelloy, line 53: The `truncate_all` function uses string formatting to build the `TRUNCATE` statement (`", ".join(...)`). While table names come from `pg_tables` (not user input) this pattern is worth noting. More importantly, this will also truncate Alembic's `alembic_version` table, which means subsequent `alembic upgrade head` calls within the same session may re-run migrations unexpectedly. Consider excluding `alembic_version` from truncation.

backend/alembic/versions/20260512_000002_drop_guild_id_child_tables.py
jmelloy: The `downgrade()` function uses `SERIAL PRIMARY KEY` in raw SQL for creating the rollback tables. `SERIAL` is a PostgreSQL shorthand that implicitly creates a sequence, but when recreating via raw DDL in a migration the sequence naming may differ from what SQLAlchemy/Alembic expects. Using `op.create_table()` with `sa.Column('id', sa.Integer(), autoincrement=True, primary_key=True)` would be more robust and portable.

jmelloy (commented):
- Migration scripts use raw SQL with SQLite-specific `SERIAL PRIMARY KEY` syntax in downgrade paths, which is technically valid PostgreSQL but inconsistent with using SQLAlchemy's `op` helpers elsewhere and leaves the downgrade path untested
- Test isolation relies on `TRUNCATE ... CASCADE` between tests rather than per-test transactions/rollbacks, which is slower and could mask ordering dependencies between tests
- The `_test_config.py` module raises `RuntimeError` at import time if env vars are missing, which gives poor pytest error messages and blocks collection of unrelated tests
- Hard-coded credentials (`pioneer_password`) appear in `.env.example`, CI workflow, and comments — while example credentials are expected, the CI workflow embeds them directly rather than using GitHub secrets
- The `insert_guild` helper in `helpers.py` silently swallows failures when `guild_pk` is `None` after the upsert's `RETURNING id`, potentially leaving tests in a bad state without any error

PR #414, opened 2026-05-20
fix(locking): clear task lock when agent goes idle; add stale-lock watchdog

jmelloy:
## Summary

Fixes the bug where a task remained stuck in `working` state after its agent transitioned to `idle` or any other terminal state, because the lock was never released.

### What changed

**New `Lock` model + `locks` table** (`backend/models.py`, `backend/lock_service.py`)
- Standalone key-value lock table with per-key TTL (default 30 min) — locks expire automatically even if the explicit release path is skipped.
- `LockService` class: `acquire(key, owner, ttl)`, `release(key)`, `is_locked(key)`, `cleanup_expired()`.
- Alembic migrations: `20260520_000001_add_task_locking`, `20260520_000002_add_locks_table`, and a merge migration `20260520_000003_merge_locking_and_messages` to resolve the dual-branch conflict with the `add_role_and_meta_to_messages` migration.

**Lock release on all terminal agent states** (`backend/ws_handlers.py`)
- `handle_agent_state` now checks the agent's `current_task_id` when the new state is any non-working value (`idle`, `done`, `error`, `cancelled`, `timeout`) and calls `LockService.release(f"task:{task_id}")`, transitioning the task out of `working`.
- `handle_task_complete`, `handle_task_followup_done` also release the lock explicitly at their natural completion points.

**Stale-lock watchdog** (`backend/main.py` liveness loop)
- Added to the existing liveness-check periodic task.
- Finds tasks in `working` state whose lock has been held longer than `WORKER_OFFLINE_AFTER_SECONDS` with no actively-running agent and moves them to `awaiting-review`, releasing the lock.
- `LockService.cleanup_expired()` also runs each tick to purge expired lock rows.

### Tests added

- `backend/tests/test_lock_service.py` (177 lines, 9 tests): acquire, double-acquire, release, idempotent release, `is_locked`, TTL expiry, `cleanup_expired`, concurrent acquire.
- `backend/tests/test_foreman.py`: `test_stale_lock_overridden` — expired lock is evicted on the next `send_followup` call.
- `backend/tests/test_agent_state_ws_persistence.py`: coverage for lock release triggered by agent-state transitions.

## Test plan

- [ ] `cd backend && python -m pytest tests/test_lock_service.py tests/test_foreman.py -x -q` — all 118 tests pass
- [ ] Reproduce original bug: assign a task, have the worker go idle without sending `task-complete` — task should now move to `awaiting-review` within `WORKER_OFFLINE_AFTER_SECONDS`
- [ ] Verify stale-lock watchdog log line appears in backend logs when a stuck task is detected


backend/lock_service.py
```diff
@@ -0,0 +1,71 @@
...
+than the previous 1-hour stale-cutoff while still generous enough to cover
+normal follow-up round-trips.
+
+Lock keys use namespaced strings: ``task:<task_id>``, ``worker:<worker_id>``, etc.
+"""
+
+from __future__ import annotations
+
+from datetime import UTC, datetime, timedelta
+
```
jmelloy, line 21: `sqlite_insert` is imported from `sqlalchemy.dialects.sqlite`, making `LockService` silently SQLite-only. If this service is ever used against PostgreSQL or another DB, `on_conflict_do_nothing()` will raise an import/dialect error. Consider using `sqlalchemy.dialects.sqlite` explicitly documented as a constraint, or abstract the upsert behind a dialect-aware helper.

jmelloy (commented):
- The lock service correctly moves from inline task-column locking to a dedicated `locks` table with TTL-based expiry, which is a sound architectural improvement.
- **Critical issue**: `LockService` uses `sqlite_insert` with `on_conflict_do_nothing`, making it SQLite-specific and non-portable; the service should either declare its database dependency explicitly or use dialect-agnostic upsert patterns.
- The stale-task watchdog in `main.py` uses `literal("task:").op("||")(Task.id)` for string concatenation in a correlated subquery, which is fragile and SQLite-specific; a Python-side formatted key or a proper SQLAlchemy expression is safer.
- The `handle_agent_state` fix correctly reads `current_task_id` before clearing it, but `"cancelled"` and `"done"` are absent from `_LOCK_RELEASE_AGENT_STATES`, meaning tasks completed via normal flow through ws_handlers won't trigger the agent-state lock release path (though other code paths cover it).
- Test coverage for `LockService` is thorough and the migration chain (including the merge migration) is handled correctly.

backend/main.py
```diff
@@ -142,6 +143,71 @@ async def _sweep_stale_workers_once() -> int:
...
+                        # that we can assume it's not a brand-new dispatch.
+                        select(Lock.key)
+                        .where(
+                            Lock.key == literal("task:").op("||")(Task.id),
+                            Lock.acquired_at < cutoff_lock,
+                        )
+                        .exists(),
+                        # No agent is actively running this task right now.
+                        ~select(Agent.id)
+                        .where(
```
jmelloy, line 184: `literal("task:").op("||")(Task.id)` relies on SQLite's `||` string-concatenation operator inside a correlated subquery. This will silently produce wrong results or errors on any other database. Use `func.concat("task:", Task.id)` or build the key comparison differently (e.g. a parameterised subquery per task_id in Python) for portability and clarity.

backend/ws_handlers.py
```diff
@@ -339,6 +340,9 @@ async def handle_join(ctx: WSContext, data: dict) -> None:
     await broadcast(ctx.guild_id, broadcast_payload)
 
 
+_LOCK_RELEASE_AGENT_STATES = frozenset({"idle", "offline", "error", "cancelled", "timeout"})
```
jmelloy: `_LOCK_RELEASE_AGENT_STATES` omits `"done"` and `"cancelled"`. While those states are handled in other code paths (`handle_task_complete`, `handle_task_followup_done`, `cancel_task_endpoint`), if an agent sends an `agent-state` message with `state="done"` directly (bypassing those handlers) the lock will not be released. At minimum, add a comment explaining the intentional exclusion, or add the states to be defensive.

backend/lock_service.py
```diff
@@ -0,0 +1,71 @@
...
+        """Try to acquire *key*. Returns True if the lock was granted, False if already held."""
+        now = datetime.now(UTC)
+        now_iso = now.isoformat()
+        expires_at = (now + timedelta(seconds=ttl_seconds)).isoformat()
+
+        # Evict any expired lock for this key before attempting the insert.
+        await self._db.execute(
+            delete(Lock).where(
+                Lock.key == key,
+                Lock.expires_at.isnot(None),
```
jmelloy, line 56: The expired-lock eviction `DELETE` and the subsequent `INSERT … ON CONFLICT DO NOTHING` are two separate statements with no transaction isolation between them. A concurrent caller could re-insert after the delete but before this session's insert, causing both to get `rowcount == 1`. Consider wrapping both statements in a savepoint or relying on the DB-level primary-key constraint alone (just attempt the insert and let conflict detection handle expiry separately).

backend/ws_handlers.py
```diff
@@ -356,11 +360,34 @@ async def handle_agent_state(ctx: WSContext, data: dict) -> None:
...
         update_vals["current_task_id"] = None
+
+    # When the agent transitions to a non-working state, release the task lock
+    # so the task doesn't stay stuck in "working" indefinitely.  We must read
+    # current_task_id from the DB *before* the update clears it.
+    task_id_to_release: str | None = None
+    if state in _LOCK_RELEASE_AGENT_STATES and agent_id:
+        row = await ctx.db.execute(
+            select(Agent.current_task_id).where(
+                Agent.id == agent_id, Agent.guild_pk == ctx.guild_pk
```
jmelloy, line 379: This unconditionally moves the task to `awaiting-review` for any agent state in `_LOCK_RELEASE_AGENT_STATES` (including `done`, `cancelled`, `error`). If the task was already set to a terminal state (e.g., `done`) by `handle_task_complete` or `handle_task_update` before the agent sends its idle signal, this update will overwrite it — the `.where(Task.state == 'working')` guard is missing here, unlike the watchdog code in `main.py`.

backend/lock_service.py
```diff
@@ -0,0 +1,78 @@
...
+
+    async def acquire(self, key: str, owner: str, ttl_seconds: int = DEFAULT_TTL_SECONDS) -> bool:
+        """Try to acquire *key*. Returns True if the lock was granted, False if already held."""
+        now = datetime.now(UTC)
+        now_iso = now.isoformat()
+        expires_at = (now + timedelta(seconds=ttl_seconds)).isoformat()
+
+        # Evict any expired lock for this key before attempting the insert.
+        await self._db.execute(
+            delete(Lock).where(
```
jmelloy, line 54: Comparing ISO-8601 datetime strings lexicographically works only for UTC timestamps without timezone offset characters. The `datetime.isoformat()` output includes `+00:00` for timezone-aware datetimes in Python 3.11+, which sorts correctly, but any inconsistency (e.g., a Z-suffix vs +00:00) would silently break expiry checks. Consider storing Unix timestamps as integers or using a `REAL` column for deterministic comparison.

backend/ws_handlers.py
```diff
@@ -356,11 +360,34 @@ async def handle_agent_state(ctx: WSContext, data: dict) -> None:
         update_vals["current_task_id"] = data.get("taskId")
     elif state in ("idle", "offline"):
         update_vals["current_task_id"] = None
+
+    # When the agent transitions to a non-working state, release the task lock
```
jmelloy, line 372: `_LOCK_RELEASE_AGENT_STATES` includes `'done'` and `'cancelled'`, but `handle_agent_state` is a generic WebSocket message handler — agents typically send `task-followup-done` or `task-update` for task completion, not an agent-state of `done`. Including these states risks premature lock release if an agent legitimately transitions through these states mid-task on a multi-step workflow. Restricting to `{'idle', 'offline', 'error', 'timeout'}` would be safer.

backend/lock_service.py
```diff
@@ -0,0 +1,78 @@
...
+                Lock.key == key,
+                Lock.expires_at.isnot(None),
+                Lock.expires_at < now_iso,
+            )
+        )
+
+        # Use a savepoint so a primary-key conflict rolls back only this
+        # attempt without aborting the enclosing transaction.
+        try:
+            async with self._db.begin_nested():
```
jmelloy, line 67: The `begin_nested()` savepoint approach for detecting duplicate key insertion is correct for SQLite, but `IntegrityError` is a broad catch — it will also swallow constraint violations unrelated to the primary key (e.g., a `NOT NULL` violation on `acquired_at`). Consider checking `e.orig` or narrowing the exception handling to avoid masking real bugs.

backend/main.py
```diff
@@ -142,6 +143,71 @@ async def _sweep_stale_workers_once() -> int:
...
+                            Agent.current_task_id == Task.id,
+                            Agent.state.in_(("working", "thinking", "busy")),
+                        )
+                        .exists(),
```
jmelloy, line 188: The `func.concat('task:', Task.id)` correlated subquery may not work correctly on SQLite, which does not have a `concat` function — it uses the `||` operator instead. This would silently return no results, making the watchdog ineffective on SQLite deployments. Use `'task:' || Task.id` or a literal expression instead.

jmelloy (commented):
- Good architectural improvement: moving from in-row lock columns to a dedicated `locks` table with TTL expiry is a cleaner design that correctly handles the idle-agent bug.
- **Critical bug in `handle_agent_state`**: unconditionally moving a task to `awaiting-review` when any agent goes idle/done will interfere with legitimate `task-followup-done` and `task-update` flows that set tasks to terminal states (`done`, `cancelled`, etc.) — the task state transition should guard against overwriting terminal states.
- **String comparison for ISO timestamps is fragile**: the entire locking mechanism relies on lexicographic comparison of ISO-8601 strings (e.g., `Lock.expires_at < now_iso`), which works for UTC-only dates but breaks silently if any timezone offset is ever introduced; using a proper `DateTime` column type would be safer.
- The watchdog logic in `_sweep_stale_workers_once` runs inside its own `AsyncSessionLocal` context but `stale_task_ids` is collected and logged outside it — this is fine, but the two-session structure adds unnecessary complexity and a TOCTOU window.
- Test coverage is solid: unit tests for `LockService`, integration tests for the watchdog, and the agent-idle scenario are all present and well-structured.

PR #422, opened 2026-05-20
Partial unique index on locks(key) for active rows

### LINKED ISSUE #420: Add partial unique index on locks(key) for non-expired rows
## Problem

The `locks` table currently allows multiple active (non-expired) rows with the same `key`, which means two workers can hold the same lock simultaneously — leading to the "stale lock" bugs we've been seeing.

## Proposed Fix

Add a partial unique index so only one non-expired row per `key` can exist at a time:

```sql
CREATE UNIQUE INDEX locks_key_active_unique
    ON locks (key)
    WHERE expired_at IS NULL OR expired_at > (now() - interval '1 hour');
```

The `now() - 1 hour` (or similar) window means recently-expired locks are still visible for debugging/auditing but can't block new lock acquisition.

## Tasks
- [ ] Add an Alembic migration that creates the partial unique index
- [ ] Update the lock-acquisition upsert/insert logic to handle `UniqueViolation` gracefully (i.e. treat it as "lock already held")
- [ ] Add a test: two concurrent lock attempts on the same key → only one succeeds

jmelloy:
## Summary

- **Migration** (`20260520_000004`): drops and recreates the `locks` table with an auto-increment `id` PK instead of `key` as PK, then adds `locks_key_active_unique` — a partial unique index on `key` where `expires_at IS NULL OR expires_at > (now() - interval '1 hour')`. This lets expired locks remain in the table while the database enforces that only one *active* lock per key exists.
- **`lock_service.LockService.acquire`**: replaced `pg_insert().on_conflict_do_nothing()` with a plain `insert()` inside a `begin_nested()` savepoint, catching `IntegrityError` and returning `False`. The savepoint ensures the expired-lock eviction DELETE is not rolled back if the insert fails.
- **`models.Lock`**: added `id` auto-increment PK; `key` is now a regular (indexed) column.
- **Test** (`test_unique_index_rejects_second_active_lock`): acquires the same key in two separate sessions — verifies first returns `True` and second returns `False`, exercising the DB-level partial index constraint.

## Test plan

- [ ] `cd backend && python -m pytest tests/test_lock_service.py -v` — all existing tests plus the new one pass
- [ ] `alembic upgrade head` applies cleanly against an empty schema
- [ ] `alembic downgrade -1` restores the original PK-based locks table

Closes #420

### DIFF
--- a/backend/alembic/versions/20260520_000004_partial_unique_index_on_locks_key.py
+++ b/backend/alembic/versions/20260520_000004_partial_unique_index_on_locks_key.py
@@ -1,0 +1,53 @@
+"""partial_unique_index_on_locks_key
+
+Recreate the locks table with an auto-increment id PK (dropping key as PK)
+and add a partial unique index so only one non-expired lock per key can exist
+at a time.  Expired locks (expires_at < now - 1 hour) fall outside the index
+predicate and are therefore not covered by the uniqueness constraint.
+
+Revision ID: 20260520_000004_partial_unique_index_on_locks_key
+Revises: 20260520_000003_merge_locking_and_messages
+Create Date: 2026-05-20 00:00:04.000000
+
+"""
+
+from collections.abc import Sequence
+
+import sqlalchemy as sa
+from alembic import op
+
+revision: str = "20260520_000004_partial_unique_index_on_locks_key"
+down_revision: str | Sequence[str] | None = "20260520_000003_merge_locking_and_messages"
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+
+def upgrade() -> None:
+    op.drop_table("locks")
+    op.create_table(
+        "locks",
+        sa.Column("id", sa.Integer(), primary_key=True, autoincrement=True),
+        sa.Column("key", sa.Text(), nullable=False),
+        sa.Column("owner", sa.Text(), nullable=True),
+        sa.Column("acquired_at", sa.Text(), nullable=False),
+        sa.Column("expires_at", sa.Text(), nullable=True),
+    )
+    op.execute(
+        """
+        CREATE UNIQUE INDEX locks_key_active_unique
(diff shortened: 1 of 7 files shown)


jmelloy (commented):
{"summary": "- The partial unique index predicate `expires_at > (now() - interval '1 hour')` is overly permissive: it keeps locks \"active\" for up to 1 hour after their TTL expires, meaning the eviction DELETE and the index predicate are not aligned — a lock could be deleted by the eviction logic but still be covered by the index (or vice versa).\n- Storing `acquired_at` and `expires_at` as `Text` rather than `TIMESTAMPTZ` is a pre-existing issue not corrected here; the partial index relies on comparing `expires_at` to `now()` as if it were a timestamp, but text comparison semantics in PostgreSQL can be unreliable and this will silently misbehave if ISO strings include timezone offsets or vary in format.\n- The `begin_nested()` savepoint approach is correct for preserving the eviction DELETE on insert failure, but there is no handling for other `IntegrityError` causes (e.g., NOT NULL violations, FK violations), which would be silently swallowed and returned as `False` instead of propagating.\n- The test only exercises in-process sequential sessions and does not truly verify concurrent contention at the DB level; the existing `test_concurrent_acquire` already covers race conditions better, so the new test's value is limited.\n- The downgrade migration drops the entire `locks` table without preserving existing data, which is a destructive operation that should be clearly documented or handled with a safer migration strategy.", "comments": [{"path": "backend/alembic/versions/20260520_000004_partial_unique_index_on_locks_key.py", "line": 38, "side": "RIGHT", "body": "The index predicate `expires_at > (now() - interval '1 hour')` gives expired locks a 1-hour grace window where they remain 'active' per the index, even though the eviction DELETE above removes locks where `expires_at < now()`. These two conditions are misaligned: a lock expiring at T is evicted at T but stays covered by the index until T+1h, which could block new acquisitions for up to an hour after evicti

PR #423, opened 2026-05-20
Migrate to SQLModel + rename guild_pk → guild_id

### LINKED ISSUE #421: Migrate SQLAlchemy models to SQLModel + clean up guild_pk/guild_id naming
## Overview

Two related cleanups bundled together:

### 1. SQLAlchemy → SQLModel migration

Replace the current `DeclarativeBase` / `Column`-style SQLAlchemy models with [SQLModel](https://sqlmodel.tiangolo.com/) models that unify SQLAlchemy ORM and Pydantic validation in a single class hierarchy.

**What to do:**
- Replace `Base = DeclarativeBase()` (or equivalent) with `class Model(SQLModel, table=True)` definitions
- Convert all `Column(...)` field declarations to `Field(...)` equivalents
- Keep Alembic for migrations — SQLModel is still SQLAlchemy under the hood, so `autogenerate` should continue to work
- Ensure relationships (`Mapped[...]`, `relationship(...)`) are expressed using SQLModel conventions
- Retain all existing validators/properties; move pure-Pydantic schemas to non-table `SQLModel` classes (replacing any standalone `BaseModel` response schemas where it makes sense)

### 2. `guild_pk` / `guild_id` foreign-key naming cleanup

Historically the guild primary key was exposed as both `guild_pk` (the internal PK integer) and `guild_id` (the human-readable string id). Clean this up so naming is consistent:
- Audit every model, migration, and query that references `guild_pk` or `guild_id`
- Settle on one canonical name (prefer `guild_id` for the FK column that holds the string identifier, or `guild_pk` for a pure surrogate integer — document the choice in a comment)
- Write an Alembic migration to rename the column(s) if any DB-level rename is needed
- Update all call-sites (queries, API route handlers, serialisers, tests)

## Acceptance Criteria
- [ ] All models are `SQLModel` subclasses; no bare `DeclarativeBase` / `Column` imports remain in model files
- [ ] `guild_pk` / `guild_id` usage is consistent and documented
- [ ] Alembic migrations cover any column renames
- [ ] All existing tests pass; add tests for any renamed fields
- [ ] No regressions in API response shapes

jmelloy:
## Summary

- Replaces all `DeclarativeBase`/`Column(...)` model definitions with `SQLModel` table classes using `Field(...)`
- Renames the `guild_pk` FK column to `guild_id` across all child tables via Alembic migration `20260520_000004`
- Fixes all `Model.guild_pk` ORM field accesses → `Model.guild_id` throughout routes, foreman, ws_handlers, and agent_runner
- Fixes all model instantiation kwargs (`guild_pk=value` → `guild_id=value`)
- Updates raw SQL strings in foreman context queries to use the renamed column
- Fixes raw SQL INSERT column names in all 10 affected test files

## Test plan

- [ ] Models import cleanly (`python -c "from models import *"` → OK)
- [ ] `ruff check . --fix && ruff format .` passes with zero remaining errors
- [ ] All existing backend tests pass against a PostgreSQL test DB (`pytest backend/`)
- [ ] `alembic upgrade head` runs without errors on a fresh DB
- [ ] No `DeclarativeBase` or bare `Column(...)` imports remain in model files
- [ ] Zero `guild_pk` SQL column references remain (only Python local variable names)

Closes #421

### DIFF
--- a/backend/alembic/versions/20260520_000004_rename_guild_pk_to_guild_id.py
+++ b/backend/alembic/versions/20260520_000004_rename_guild_pk_to_guild_id.py
@@ -16,6 +16,7 @@ Create Date: 2026-05-20 00:00:04.000000
 
 from collections.abc import Sequence
 
+import sqlalchemy as sa
 from alembic import op
 
 revision: str = "20260520_000004_rename_guild_pk_to_guild_id"
@@ -43,22 +44,63 @@ _UNIQUE_FK_TABLES = [
 ]
 
 
+def _drop_fk(table: str, col: str) -> None:
+    """Drop the FK on *col* from *table*, robust to auto-generated constraint names.
+
+    Migration 20260512_000002 recreated tables as ``{table}_new`` then renamed
+    them.  PostgreSQL keeps the original auto-generated constraint name
+    (e.g. ``agents_new_guild_pk_fkey``) after a table rename, so we cannot
+    assume the conventional ``{table}_{col}_fkey`` pattern.  We query
+    ``pg_constraint`` at migration time to find and drop whatever name exists.
+    """
+    op.execute(
+        sa.text(f"""
+        DO $$ DECLARE cname TEXT; BEGIN
+            SELECT c.conname INTO cname
+              FROM pg_constraint c
+              JOIN pg_class t ON t.oid = c.conrelid
+             WHERE t.relname = '{table}'
+               AND c.contype = 'f'
+               AND c.conname LIKE '%{col}_fkey';
+            IF cname IS NOT NULL THEN
+                EXECUTE format('ALTER TABLE {table} DROP CONSTRAINT %I', cname);
+            END IF;
+        END $$
+        """)
+    )
+
+
+def _drop_unique(table: str, col: str) -> None:
+    """Drop the UNIQUE constraint on *col* from *table*, robust to auto-generated names."""
+    op.execute(
+        sa.text(f"""
+        DO $$ DECLARE cname TEXT; BEGIN
+            SELECT c.conname INTO cname
+              FROM pg_constraint c
+              JOIN pg_class t ON t.oid = c.conrelid
+             WHERE t.relname = '{table}'
+               AND c.contype = 'u'
+               AND c.conname LIKE '%{col}_key';
+            IF cname IS NOT NULL THEN
+                EXECUTE format('ALTER TABLE {table} DROP CONSTRAINT %I', cname);
+            END IF;
+        END $$
+        """)
+    )
+
+
 def upgrade() -> None:
     for table in _SIMPLE_FK_TABLES:
-        op.drop_constraint(f"{table}_guild_pk_fkey", table, type_="foreignkey")
+        _drop_fk(table, "guild_pk")
         op.alter_column(table, "guild_pk", new_column_name="guild_id")
-        op.create_foreign_key(
-            f"{table}_guild_id_fkey", table, "guilds", ["guild_id"], ["id"]
-        )
+        op.create_foreign_key(f"{table}_guild_id_fkey", table, "guilds", ["guild_id"], ["id"])
 
     for table in _UNIQUE_FK_TABLES:
-        op.drop_constraint(f"{table}_guild_pk_fkey", table, type_="foreignkey")
-        op.drop_constraint(f"{table}_guild_pk_key", table, type_="unique")
+        _drop_fk(table, "guild_pk")
+        _drop_unique(table, "guild_pk")
         op.alter_column(table, "guild_pk", new_column_name="guild_id")
         op.create_unique_constraint(f"{table}_guild_id_key", table, ["guild_id"])
-        op.create_foreign_key(
-            f"{table}_guild_id_fkey", table, "guilds", ["guild_id"], ["id"]
-        )
+        op.create_foreign_key(f"{table}_guild_id_fkey", table, "guilds", ["guild_id"], ["id"])
 
 
 def downgrade() -> None:
@@ -67,13 +109,9 @@ def downgrade() -> None:
         op.drop_constraint(f"{table}_guild_id_key", table, type_="unique")
         op.alter_column(table, "guild_id", new_column_name="guild_pk")
         op.create_unique_constraint(f"{table}_guild_pk_key", table, ["guild_pk"])
-        op.create_foreign_key(
-            f"{table}_guild_pk_fkey", table, "guilds", ["guild_pk"], ["id"]
-        )
+        op.create_foreign_key(f"{table}_guild_pk_fkey", table, "guilds", ["guild_pk"], ["id"])
 
     for table in reversed(_SIMPLE_FK_TABLES):
         op.drop_constraint(f"{table}_guild_id_fkey", table, type_="foreignkey")
         op.alter_column(table, "guild_id", new_column_name="guild_pk")
-        op.create_foreign_key(
-            f"{table}_guild_pk_fkey", table, "guilds", ["guild_pk"], ["id"]
-        )
+        op.create_foreign_key(f"{table}_guild_pk_fkey", table, "guilds", ["guild_pk"], ["id"])
--- a/backend/alembic/versions/20260520_000004_rename_guild_pk_to_guild_id.py
+++ b/backend/alembic/versions/20260520_000004_rename_guild_pk_to_guild_id.py
@@ -16,7 +16,6 @@ Create Date: 2026-05-20 00:00:04.000000
 
 from collections.abc import Sequence
 
-import sqlalchemy as sa
 from alembic import op
 
 revision: str = "20260520_000004_rename_guild_pk_to_guild_id"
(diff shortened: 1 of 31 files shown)


jmelloy (commented):
## Code Review: Migrate to SQLModel + rename guild_pk → guild_id

### Overview

This PR does two independent things atomically: (1) replaces all `DeclarativeBase`/`Column(...)` model definitions with SQLModel `Field(...)` equivalents, and (2) renames the `guild_pk` integer FK column to `guild_id` across all nine child tables. The rename is motivated by the transitional name becoming permanent post-migration #20260512_000002 when the original TEXT `guild_id` column was dropped.

---

### ✅ What's done well

- **Complete coverage**: All 9 child tables are in the migration; all ORM field accesses updated across `main.py`, `ws_handlers.py`, `foreman/`, `routes/`, tests, and `agent_runner.py`. No stranded `Model.guild_pk` attribute accesses remain.
- **Migration correctness**: The `ALTER TABLE ... RENAME COLUMN` approach is safe, atomic, and correct. The downgrade path reverses the tables in reverse order (good practice though not strictly necessary for renames).
- **Disambiguation in multi-join queries (`main.py`)**: The `.label("guild_slug")` technique correctly handles queries that select from both `Worker.guild_id` (int FK) and `Guild.guild_id` (str slug) in the same result row. Without it, the broadcast/log calls at lines 403–408 would silently use the integer instead of the string guild slug.
- **Test updates**: All 10 affected test helpers using raw SQL have been updated to `guild_id`. The API assertion in `test_worker_api.py` (`workers[0]["guild_id"]`) confirms the serialized response key changed too.
- **`alembic/env.py`**: Switching to `SQLModel.metadata` is correct. The side-effect import `import models  # noqa: F401` is properly documented.

---

### 🟡 Issues to address before merge

#### 1. `server_default` silently dropped — DB-level defaults are lost for new installs

Every `server_default="..."` in the old models is replaced with a Python-level `default=...` in SQLModel. **The migration only renames columns; it does not `ALTER COLUMN ... SET DEFAULT ...`.** This means:

- On existing databases the old `DEFAULT` constraints from earlier migrations remain, so behaviour is unchanged.
- On a **fresh `alembic upgrade head`** from scratch, the generated DDL from SQLModel's metadata will not include column-level `DEFAULT` clauses (SQLModel `Field(default=...)` sets the value in Python, not as a SQL `DEFAULT`). Raw SQL inserts or third-party tooling that omits these columns will get `NULL` instead of `'idle'`/`'pending'`/`'execute'`/etc.

Affected fields: `Agent.type`, `Agent.state`, `Worker.repos`, `Worker.state`, `Task.tool`, `Task.state`, `Task.phase`, `GuildMember.role`, `ForemanTurn.is_tool_response`, `GithubToken.token_type`.

**Recommendation**: Either add `sa_column_kwargs={"server_default": "..."}` in `Field(...)` for critical columns, or add a follow-up migration that reinstates `DEFAULT` constraints to keep a fresh install consistent with an upgraded one.

#### 2. `Task.phase` nullability widened without a migration

```python
# Before
phase = Column(Text, server_default="execute")   # effectively NOT NULL via server_default

# After  
phase: str | None = Field(default="execute")     # explicitly nullable in Python
```

The original column had `server_default` but no explicit `nullable=False`, so it was already nullable at the DB level. However, the model now _declares_ it as `str | None`, which is a semantic broadening. Any code that previously assumed `task.phase` was never `None` could break if a legacy NULL row is encountered. Minor, but worth a comment.

#### 3. Breaking REST API change not called out in the PR description

`test_worker_api.py` line 1696:
```python
-    assert isinstance(workers[0]["guild_pk"], int)
+    assert isinstance(workers[0]["guild_id"], int)
```

The `GET /guilds/{id}/workers` response now returns `guild_id` instead of `guild_pk`. Any frontend or external client that reads `worker.guild_pk` will silently get `undefined`. This should be explicitly called out as a breaking change in the PR description so it's easy to find in git history.

#### 4. `ctx.guild_pk` attribute on `WSContext` still uses the old name

In `ws_handlers.py`, all the new lines correctly use `Model.guild_id == ctx.guild_pk`. But the `WSContext` attribute itself is still named `guild_pk` (a local Python attribute, not a DB column). This is not a bug — the integer value is correct — but it creates a subtle naming mismatch: the model field is `guild_id`, the context attribute is `guild_pk`. Worth renaming in a follow-up for consistency.

---

### 🔵 Minor nits

- **`Guild.__table_args__` index**: The partial unique index `uq_guilds_guild_id_active` is now declared in the model, which is good for documentation. Ensure the index name matches what was created in earlier migrations to avoid Alembic autogenerate trying to recreate it.
- **Local variable names**: Many local variables are still named `guild_pk` / `guild_pk_val` (e.g. in `foreman/runner.py`, `routes/foreman.py`). These hold the integer PK of the guild row. While functionally fine, they now disagree with the model field name. Renaming them to `guild_fk` or `guild_int_pk` in a follow-up would remove the last ambiguity.
- **SQLModel global metadata**: `SQLModel.metadata` is a process-global object. If any future dependency also uses SQLModel tables, they'd appear in Alembic's autogenerate output. Not an immediate concern but worth noting in a docstring or comment in `env.py`.

---

### Verdict

**Approve with recommendations.** The core logic is correct and the coverage is thorough. The `server_default` issue (#1) is the most impactful for new installs — everything else is cosmetic or future-proofing. Tests should pass on an existing upgraded DB; the risk is a fresh DB behaving differently from an upgraded one for raw SQL inserts that rely on column defaults.

backend/alembic/versions/20260520_000004_rename_guild_pk_to_guild_id.py
```diff
@@ -0,0 +1,47 @@
...
+"""
+
+from collections.abc import Sequence
+
+import sqlalchemy as sa
+from alembic import op
+
+revision: str = "20260520_000004_rename_guild_pk_to_guild_id"
+down_revision: str | Sequence[str] | None = "20260520_000006_merge_nullable_worker_and_locks_fix"
+branch_labels: str | Sequence[str] | None = None
```
jmelloy, line 24: Critical bug: `down_revision` is set to `20260520_000006_merge_nullable_worker_and_locks_fix`, which is a *later* revision. This creates an invalid migration chain (a revision depending on one that doesn't exist yet in the upgrade path). It should point to the actual predecessor revision (e.g., `20260520_000003_merge_locking_and_messages` as stated in the docstring). This will cause `alembic upgrade head` to fail.

backend/models.py
```diff
@@ -20,262 +16,277 @@ def live_tasks_filter(now: str | None = None):
...
-    activity = Column(Text, nullable=True)
+    id: str = Field(primary_key=True)
+    # guild_id is the integer FK to guilds.id (renamed from guild_pk).
+    guild_id: int = Field(foreign_key="guilds.id")
+    worker_id: str | None = Field(default=None, foreign_key="workers.id")
+    name: str
+    type: str = Field(default="worker")
+    state: str = Field(default="idle")
+    activity: str | None = None
     # Task this agent is currently executing (NULL when idle/offline). Set
```
jmelloy, line 60: SQLModel `Field(default=...)` sets a Python-side default only, not a `server_default` in the database schema. The original `Column(..., server_default='worker')` for `type`, `state`, `repos`, etc. ensured the DB itself provided defaults for raw INSERTs. Consider using `Field(default='worker', sa_column_kwargs={'server_default': 'worker'})` or `sa_column=Column(Text, server_default='worker')` for columns that need DB-level defaults to preserve schema compatibility.

backend/alembic/versions/20260520_000004_rename_guild_pk_to_guild_id.py
```diff
@@ -0,0 +1,47 @@
...
+    "tasks",
+    "guild_members",
+    "claude_credentials",
+    "guild_keys",
+    "foreman_turns",
+    "github_events",
+]
+
+
+def upgrade() -> None:
```
jmelloy, line 92: The rename uses raw `ALTER TABLE ... RENAME COLUMN` via `op.execute(sa.text(...))` rather than `op.alter_column(..., new_column_name=...)`. While functionally equivalent for PostgreSQL, `op.alter_column` is the Alembic-idiomatic approach and handles dialect differences. Also, this does not drop and recreate foreign key constraints, which on some DB configurations may cause issues — verify FK constraint names are not affected by the rename on the target PostgreSQL version.

backend/main.py
```diff
@@ -97,8 +97,8 @@ async def _sweep_stale_workers_once() -> int:
         # and handled via the agent cascade below.
         zombie_workers = (
             await db.execute(
-                select(Worker.id, Worker.guild_pk, Guild.guild_id)
-                .join(Guild, Guild.id == Worker.guild_pk)
+                select(Worker.id, Worker.guild_id, Guild.guild_id.label("guild_slug"))
```
jmelloy, line 100: The select now includes `Worker.guild_id` (integer FK) and `Guild.guild_id.label('guild_slug')` (text slug). However, `row.guild_id` used later in `stale_worker_keys.add((row.worker_id, row.guild_id))` refers to the integer FK — ensure downstream consumers of `stale_worker_keys` (the `Worker` update WHERE clause) correctly use this integer value, which they do. This is correct but the dual use of `guild_id` for two different semantic types in the same query is a latent readability hazard.

jmelloy (commented):
- The migration file has a **critical dependency inversion bug**: `down_revision` points to `20260520_000006_merge_nullable_worker_and_locks_fix` which is a *later* revision number, creating a circular/invalid migration chain that will break `alembic upgrade head`.
- The `main.py` sweep query correctly introduces `guild_slug` labels to disambiguate `Guild.guild_id` (the text slug) from `Worker/Agent.guild_id` (the integer FK), but logging now references `row.guild_slug` — this is correct and the fix is sound.
- SQLModel migration is generally clean; `server_default` values from SQLAlchemy `Column` definitions are replaced with Python-side `default=` in `Field(...)`, which means the database-level defaults are lost for columns like `type`, `state`, `tool`, `phase`, and `is_tool_response` — existing rows inserted via raw SQL or other tools won't benefit from DB defaults.
- The `Guild` model moves its partial unique index into `__table_args__` using `Index(...)` which is correct, but this index was presumably already created by an earlier migration — Alembic autogenerate may try to re-create it unless the migration history accounts for it.
- Raw SQL in `_fetch_online_workers` is correctly updated to use `guild_id` parameter binding, and all ORM filter sites appear consistently renamed.

jmelloy (commented):
## Code Review: SQLModel migration + guild_pk → guild_id rename

### 1. Migration chain correctness ✅

`_000004_rename_guild_pk_to_guild_id` correctly sets `down_revision = "20260520_000003_merge_locking_and_messages"`, branching from `_000003` in parallel with the already-merged `_000004_nullable_task_worker_id` and `_000004_partial_unique_index_on_locks_key`.

`_000007_merge_rename_with_prior_head` correctly closes the fork:
```python
down_revision = (
    "20260520_000006_merge_nullable_worker_and_locks_fix",
    "20260520_000004_rename_guild_pk_to_guild_id",
)
```
Downgrade ops in `_000004` are complete and symmetric (reversed table order, proper constraint name reversal). ✅

**⚠️ Hardcoded constraint names**: the migration assumes `{table}_guild_pk_fkey` / `{table}_guild_pk_key` for every table. If any constraint was created with a non-standard name (e.g., by Alembic autogenerate at an intermediate schema version), `op.drop_constraint` will raise at upgrade time. A comment showing where constraint names were verified (e.g., from `\d+ agents` output) would add confidence.

---

### 2. SQLModel Field(...) vs bare Column(...) ⚠️

The translation is thorough. One critical regression:

**`Lock.acquired_at` / `expires_at` lose `timezone=True`**

Migration `20260520_000005_fix_locks_timestamp_types_and_index_predicate` created these as `DateTime(timezone=True)` (`TIMESTAMP WITH TIME ZONE` in PostgreSQL). SQLModel infers bare `datetime` as `DateTime()` **without** timezone, i.e., `TIMESTAMP WITHOUT TIME ZONE`. The DB schema and the model are now inconsistent — running `alembic revision --autogenerate` on any DB that has been through `_000005` will generate a spurious `ALTER COLUMN` migration.

Fix:
```python
from sqlalchemy import Column, DateTime
from sqlmodel import Field

class Lock(SQLModel, table=True):
    ...
    acquired_at: datetime = Field(sa_column=Column(DateTime(timezone=True), nullable=False))
    expires_at: datetime | None = Field(default=None, sa_column=Column(DateTime(timezone=True), nullable=True))
```

**`server_default` quoting is correct**: `sa_column_kwargs={"server_default": "'value'"}" (SQL string literals) for strings, and `"0"` for integers — both correct for PostgreSQL. ✅

**`Guild.__table_args__` partial index**: already created by `20260512_000000_guilds_integer_pk_soft_delete`. Adding it to `__table_args__` is purely documentation — no migration drift. ✅

**`sqlmodel` not version-pinned** in `requirements.txt`: SQLModel has had API changes around `sa_column_kwargs` behavior across minor versions. Should be pinned (e.g., `sqlmodel>=0.0.21,<1.0`) to avoid silent breakage on fresh installs.

---

### 3. guild_id naming consistency ⚠️

The rename across all 29 files (routes, foreman, ws_handlers, auth, webhooks, 11 test files) is comprehensive. Raw SQL strings in `foreman/runner.py` and `routes/foreman.py` are correctly updated. The `main.py` `Guild.guild_id.label("guild_slug")` disambiguator is a good defensive change. ✅

**One incomplete rename — `WSContext.guild_pk`** (`ws_handlers.py:195`):
```python
class WSContext:
    guild_pk: int | None = None   # ← still the old attribute name
```
The attribute is used throughout as `guild_id=ctx.guild_pk`, `Agent.guild_id == ctx.guild_pk`, etc. The code is correct (both sides are integers), but the attribute name is now semantically inverted: `ctx.guild_pk` holds the value for the field named `guild_id`. This is the only surviving `guild_pk` in application code that should be renamed to `ctx.guild_id`.

**Local variable naming** (`guild_pk`, `guild_pk_val`) throughout routes/foreman: the code is correct — `get_guild_pk()` returns an integer which is then used against the integer FK field `Model.guild_id`. But readers now encounter `Agent.guild_id == guild_pk`, where the right-hand side name suggests the old column name. Consider renaming local variables to `guild_row_id` or `guild_int_id` in a follow-up to align with the new field name.

---

### 4. Remaining regressions

- `Lock` datetime type drift (see §2) — most actionable issue here
- `WSContext.guild_pk` attribute (see §3) — low-risk but creates future confusion
- Constraint name assumptions in migration (see §1) — safe on unmodified DBs
- `task_logs`, `user_sessions`, `github_tokens` correctly absent from the migration (none have a `guild_pk` FK). ✅
- All 11 test files updated including `ON CONFLICT (guild_id, user_id)` clauses. ✅
- `test_worker_api.py` assertion `isinstance(workers[0]["guild_id"], int)` correctly updated from `guild_pk`. ✅

---

### Summary

| Area | Status |
|---|---|
| Migration chain `down_revision` ordering | ✅ Correct |
| Merge migration closes independent fork | ✅ Correct |
| SQLModel `Field()` for strings/ints | ✅ Correct |
| `Lock.acquired_at/expires_at` `timezone=True` | ❌ Schema drift — needs `sa_column=Column(DateTime(timezone=True))` |
| `guild_id` rename across all routes/tests | ✅ Comprehensive |
| `WSContext.guild_pk` attribute | ⚠️ Missed rename |
| `sqlmodel` version pin | ⚠️ Missing |
| FK constraint name assumptions | ⚠️ Undocumented assumption |

The PR is well-executed. The `Lock` timezone issue is the only change that will silently corrupt `autogenerate` output; everything else is cosmetic or defensive hygiene.

PR #435, opened 2026-05-21
Worker join/leave notifies foreman

### LINKED ISSUE #434: Notify foreman when a worker joins or leaves
## Summary

When a worker agent comes online or goes offline, the foreman should receive a notification so it can react accordingly (reassign queued work, log the event, alert on unexpected disconnects, etc.).

## Requirements

- When a worker transitions to **online** (registers / reconnects), send the foreman a message indicating the worker ID, its repos, and agent_count.
- When a worker transitions to **offline** (graceful shutdown or unexpected disconnect), send the foreman a message indicating the worker ID and reason (if known).
- The message should arrive on the foreman's standard input channel (same path as `[github-event]` and `[periodic-check]` messages) so the foreman can handle it in its normal turn loop.
- The message format should follow the existing `[...]` prefix convention, e.g. `[worker-online]` / `[worker-offline]`.

## Acceptance Criteria

- [ ] Worker emits a `[worker-online]` event to the foreman when it successfully registers.
- [ ] Worker emits a `[worker-offline]` event to the foreman on graceful shutdown.
- [ ] Server-side disconnect (unexpected drop) also triggers a `[worker-offline]` message to the foreman.
- [ ] Foreman documentation / prompt updated to describe how to handle these events.

jmelloy:
## Summary

- **`[worker-online]`** — `handle_worker_register` in `ws_handlers.py` now calls `_trigger_foreman` after updating the worker's repo list. Emits `[worker-online] worker_id=<id> repos=<comma-separated> agent_count=<n>` so the foreman can assign queued tasks to the newly available worker.
- **`[worker-offline] reason=shutdown`** — `handle_worker_disconnect` in `ws_handlers.py` calls `_trigger_foreman` after broadcasting the offline state, emitting `[worker-offline] worker_id=<id> reason=shutdown` for clean shutdowns.
- **`[worker-offline] reason=disconnect`** — the `finally` block in `routes/websocket.py` calls `ws_handlers._trigger_foreman` for each stale worker ID after marking agents offline, emitting `[worker-offline] worker_id=<id> reason=disconnect` for abrupt drops.
- **Foreman system prompt** (`foreman/prompt.py`) — new "Reacting to worker lifecycle events" section tells the foreman to check pending tasks on `worker-online`, ignore clean shutdowns unless tasks are affected, and escalate unexpected `disconnect` events to the human.

Both call sites use the existing `_trigger_foreman` helper, which routes to an external foreman if one is connected and falls back to the embedded loop—no new infrastructure required.

## Test plan

- [ ] `test_worker_online_notifies_foreman` — join + worker-register → `[worker-online]` with correct `worker_id`, `repos`, and `agent_count`
- [ ] `test_worker_graceful_offline_notifies_foreman` — worker-disconnect → `[worker-offline] reason=shutdown`
- [ ] `test_abrupt_disconnect_notifies_foreman` — close WS without worker-disconnect → `[worker-offline] reason=disconnect`
- [ ] All existing disconnect and liveness tests continue to pass (foreman trigger is patched in new tests; existing tests are unaffected because `spawn(run_foreman_ai(...))` schedules a task that completes after the test WS closes)

Closes #434

### DIFF
--- a/backend/routes/websocket.py
+++ b/backend/routes/websocket.py
@@ -204,6 +204,13 @@ async def websocket_endpoint(websocket: WebSocket, guild_id: str):
                             "state": "offline",
                         },
                     )
+                for wid in stale_worker_ids:
+                    await ws_handlers._trigger_foreman(
+                        guild_id,
+                        "worker-offline",
+                        f"[worker-offline] worker_id={wid} reason=disconnect",
+                        task_name=f"foreman.worker-offline:{wid}",
+                    )
             finally:
                 # If a cancellation was delivered inside a handler (e.g.
                 # handle_worker_disconnect) the underlying asyncpg connection
--- a/backend/routes/websocket.py
+++ b/backend/routes/websocket.py
@@ -204,12 +204,19 @@ async def websocket_endpoint(websocket: WebSocket, guild_id: str):
                             "state": "offline",
                         },
                     )
-                for wid in stale_worker_ids:
+                # Batch all stale-worker notifications into a single foreman
+                # trigger so a simultaneous mass-disconnect doesn't stampede
+                # the foreman with N concurrent embedded-foreman tasks.
+                if stale_worker_ids:
+                    offline_lines = "\n".join(
+                        f"[worker-offline] worker_id={wid} reason=disconnect"
+                        for wid in sorted(stale_worker_ids)
+                    )
                     await ws_handlers._trigger_foreman(
                         guild_id,
                         "worker-offline",
-                        f"[worker-offline] worker_id={wid} reason=disconnect",
-                        task_name=f"foreman.worker-offline:{wid}",
+                        offline_lines,
+                        task_name="foreman.worker-offline:disconnect-batch",
                     )
             finally:
                 # If a cancellation was delivered inside a handler (e.g.
--- a/backend/tests/test_worker_foreman_notify.py
+++ b/backend/tests/test_worker_foreman_notify.py
@@ -99,10 +99,10 @@ def test_worker_online_notifies_foreman(client):
                     "repos": ["org/repo1", "org/repo2"],
                 }
             )
-            # handler runs synchronously; give the event loop a tick so the
-            # awaited fake_trigger coroutine completes before we exit
-            ws.send_json({"type": "ping"})
-            ws.receive_json()  # pong
+            # No explicit sync needed: _trigger_foreman is directly awaited
+            # inside handle_worker_register, so triggered is populated before
+            # the handler returns. The WS context exit (close frame) ensures
+            # the server processes worker-register before we reach assertions.
 
     online = [(e, m) for e, m in triggered if e == "worker-online"]
     assert online, f"Expected worker-online trigger, got: {triggered}"
--- a/backend/tests/test_worker_foreman_notify.py
+++ b/backend/tests/test_worker_foreman_notify.py
@@ -1,0 +1,180 @@
+"""Tests that worker join/leave events trigger foreman notifications.
+
+Verifies that:
+- handle_worker_register triggers [worker-online] to the foreman
+- handle_worker_disconnect triggers [worker-offline] reason=shutdown
+- abrupt WebSocket disconnect triggers [worker-offline] reason=disconnect
+"""
+
+from __future__ import annotations
+
+import os
+import sys
+from unittest.mock import patch
+
+import pytest
+from sqlalchemy.ext.asyncio import AsyncSession, async_sessionmaker, create_async_engine
+from sqlalchemy.pool import NullPool
+from starlette.testclient import TestClient
+
+sys.path.insert(0, os.path.join(os.path.dirname(__file__), ".."))
+sys.path.insert(0, os.path.dirname(__file__))
+
+import database as database_module  # noqa: E402
+import main as main_module  # noqa: E402
+import ws_handlers  # noqa: E402
+from _test_config import TEST_DATABASE_URL  # noqa: E402
+from helpers import insert_guild, insert_worker  # noqa: E402
+
+
+@pytest.fixture(scope="module")
+def client(_setup_schema):
+    """Module-scoped TestClient shared by all notify tests."""
+    db_url = TEST_DATABASE_URL
+    new_engine = create_async_engine(db_url, echo=False, poolclass=NullPool)
+    new_session = async_sessionmaker(new_engine, expire_on_commit=False, class_=AsyncSession)
+    mp = pytest.MonkeyPatch()
+    mp.setattr(database_module, "AsyncSessionLocal", new_session)
+    mp.setattr(main_module, "AsyncSessionLocal", new_session)
+
+    async def _stubbed_reset_connection_state() -> None:
+        pass
+
+    mp.setattr(main_module, "reset_connection_state", _stubbed_reset_connection_state)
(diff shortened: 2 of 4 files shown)


backend/ws_handlers.py
```diff
@@ -540,6 +540,14 @@ async def handle_worker_register(ctx: WSContext, data: dict) -> None:
         .values(**update_vals)
     )
     await ctx.db.commit()
+    repos_str = ",".join(repos) if repos else ""
```
jmelloy, line 543: `ctx.joined_agents` counts all agents currently tracked on this WebSocket context, which may span multiple workers or miss agents registered in previous sessions. Consider querying the DB for agents belonging specifically to `worker_id` to get an accurate and consistent `agent_count`.

backend/ws_handlers.py
```diff
@@ -564,6 +572,13 @@ async def handle_worker_disconnect(ctx: WSContext, data: dict) -> None:
             {"type": "agent-state", "agentId": agent_id, "state": "offline"},
         )
     reset_foreman_poll(ctx.guild_id)
```
jmelloy: The `reset_foreman_poll` call just above this block may cancel or supersede the `_trigger_foreman` call if `reset_foreman_poll` schedules a competing foreman task. Ensure the two calls are compatible and that the offline trigger is not dropped by the task deduplication logic inside `_trigger_foreman`.

backend/routes/websocket.py
```diff
@@ -204,6 +204,13 @@ async def websocket_endpoint(websocket: WebSocket, guild_id: str):
                             "state": "offline",
                         },
                     )
+                for wid in stale_worker_ids:
```
jmelloy: The `for wid in stale_worker_ids` loop fires one foreman task per stale worker, which could cause a burst of concurrent foreman invocations if many workers disconnect simultaneously. Consider batching these into a single notification or adding a short stagger to avoid stampeding the foreman.

jmelloy (commented):
- The feature correctly wires up worker join/leave events to foreman notifications across three code paths (register, graceful disconnect, abrupt disconnect), with matching test coverage for each path.
- A subtle ordering issue exists in `handle_worker_register`: `agent_count` is read from `ctx.joined_agents` which reflects the WebSocket connection pool at handler time, but this may include agents from other workers or not yet reflect the current worker's agents, making the count potentially misleading.
- The `for wid in stale_worker_ids` loop in `websocket.py` is placed *before* the `finally:` block comment, meaning it runs inside the `finally` block; however the indentation in the diff suggests it is actually in the `except`/cleanup block — this placement needs verification to ensure notifications fire even when exceptions occur during cleanup.
- The `repos` field in the foreman message uses comma-separation with no quoting or escaping, so repo names containing commas would silently corrupt the message format; while unlikely for repo slugs, this is worth a defensive note.
- Test isolation relies on module-scoped `client` fixture with separate guild IDs per test, which is acceptable, but the `insert_worker` with `state='online'` pre-seeding may conflict with the handler's own state update logic if the DB constraints are strict.

backend/tests/test_worker_foreman_notify.py
```diff
@@ -0,0 +1,180 @@
...
+            )
+            # handler runs synchronously; give the event loop a tick so the
+            # awaited fake_trigger coroutine completes before we exit
+            ws.send_json({"type": "ping"})
+            ws.receive_json()  # pong
+
+    online = [(e, m) for e, m in triggered if e == "worker-online"]
+    assert online, f"Expected worker-online trigger, got: {triggered}"
+    _event, msg = online[0]
+    assert f"worker_id={worker_id}" in msg, msg
```
jmelloy, line 110: The comment says 'handler runs synchronously; give the event loop a tick' but the mechanism used is sending a ping and waiting for a pong. This is fragile — if the ping/pong round-trip completes before the async `fake_trigger` coroutine is scheduled, the assertion may occasionally fail. Consider an explicit `asyncio.sleep` or awaiting the task directly instead.

jmelloy (commented):
## Code Review: Worker join/leave notifies foreman

### What the PR does

Wires up `_trigger_foreman` at three points in the WebSocket layer so the foreman AI is notified when workers come and go:

1. **`handle_worker_register`** (`ws_handlers.py`) — emits `[worker-online] worker_id=<id> repos=<comma-list> agent_count=<n>` after committing the repos update. Agent count is read directly from the DB (not from in-memory `joined_agents`) to handle multi-worker connections and stale session artefacts correctly.
2. **`handle_worker_disconnect`** (`ws_handlers.py`) — emits `[worker-offline] worker_id=<id> reason=shutdown` after broadcasting the agent-state-offline messages.
3. **Stale-worker cleanup in `websocket.py` `finally`** — batches all simultaneous abrupt disconnects into one `_trigger_foreman` call, preventing an N-worker mass-disconnect from spawning N concurrent embedded foreman tasks.

A new "Reacting to worker lifecycle events" section in `foreman/prompt.py` tells the foreman what each event means and how to respond.

---

### Correctness

**Agent count timing is sound.** Querying the DB after `commit()` means the count reflects the committed state. The typical call sequence is `join` → `receive ack` → `worker-register`, so the join's agent row is already persisted before the count runs. Confirmed by the test, which waits for the `agent-joined` broadcast before sending `worker-register`.

**Batching is correct.** `sorted(stale_worker_ids)` gives deterministic output order; the comment explains the stampede concern clearly.

**`task_name` uniqueness.** Online uses `foreman.worker-online:{worker_id}`, offline uses `foreman.worker-offline:{worker_id}` or the shared `disconnect-batch` name. If two `worker-offline` messages could race for the same worker (e.g., a graceful disconnect racing with a stale-worker cleanup), a `spawn` task name collision would silently drop the second. Worth auditing whether both paths can fire for the same worker concurrently — though in practice the graceful path marks the worker offline before the stale-detection loop runs, so the race window is narrow.

---

### Minor issues

1. **Multi-line batch not described in the prompt.** The `disconnect` batch sends multiple `[worker-offline]` lines in one message, but the prompt describes only single-line examples. An LLM processing the message could stop at the first line. Consider either noting in the prompt that the body may contain multiple lines, or splitting the batch into one trigger per worker (if the stampede concern is acceptable at normal scale).

2. **Empty `repos` representation.** The prompt says `repos=<repos>` without specifying the empty case. The code sends `repos=` (empty string). A small clarification in the prompt (e.g., "repos may be empty") would remove ambiguity for the foreman.

3. **Test coverage gap for the batch path.** `test_abrupt_disconnect_notifies_foreman` inserts a pre-seeded online worker but never sends `worker-register` on the WS, so it tests the stale-cleanup path without verifying that a multi-worker batch produces one call with multiple lines (vs. two separate calls). A second assertion on `len(triggered)` or on the message containing both `worker_id`s would tighten this.

---

### Code quality

- The DB-count approach over `joined_agents` is the right call and the comment explains why.
- All three call sites reuse `_trigger_foreman` cleanly — no new infrastructure.
- `sorted()` on the set is a nice defensive touch.
- Tests are well-structured with a shared `_make_trigger_spy()` factory.

---

### Verdict

The implementation is correct and the architecture is clean. The issues above are minor: (1) is a prompt-quality concern that could affect foreman behaviour under mass disconnect, (2) is documentation, (3) is a test coverage gap. None block the feature. **Approving with the suggestion to address the multi-line prompt gap before the foreman is expected to handle large-scale simultaneous disconnects.**

PR #441, opened 2026-05-21
fix: prevent WS close-race InterfaceError in test_worker_online_notifies_foreman

jmelloy:
## Summary

- `test_worker_online_notifies_foreman` was failing in CI with `sqlalchemy.exc.InterfaceError: connection is closed` (1 failing test, 438 passing)
- Root cause: the test sent `worker-register` then immediately closed the WebSocket. The close frame could race `handle_worker_register`'s `await db.commit()`, delivering a `CancelledError` that corrupted the asyncpg connection. The `finally` block then attempted a `SELECT agents.worker_id` on the closed connection and raised `InterfaceError`.
- Fix 1 (test): add a `ping`/`pong` sync after `worker-register` so the server finishes the handler before the WebSocket closes.
- Fix 2 (production code): wrap the stale-agent DB operations in the `finally` block's teardown in `try/except` so a broken connection yields a warning log rather than an unhandled exception. Also move `agent_owners.pop()` before the DB try-block so ownership is always released even if DB ops fail.

## Test plan

- [ ] `pytest backend/tests/test_worker_foreman_notify.py -v` — all 3 notify tests pass
- [ ] `pytest backend/tests/ -v` — full suite passes (requires postgres-test container)
- [ ] CI should show green on this PR

### DIFF
--- a/backend/routes/websocket.py
+++ b/backend/routes/websocket.py
@@ -162,39 +162,55 @@ async def websocket_endpoint(websocket: WebSocket, guild_id: str):
                     stale_agents = [
                         aid for aid in joined_agents if agent_owners.get(aid) is websocket
                     ]
+                    # Release ownership first so a racing join/reconnect sees a
+                    # clean slate even if the DB operations below fail.
+                    for agent_id in stale_agents:
+                        agent_owners.pop(agent_id, None)
                     # Look up the worker_ids for all stale agents *before* we
                     # modify their rows, so we can update the Worker table with
                     # the correct IDs (worker IDs are "w-…"; agent IDs are
                     # "a-…" — using agent_id as a Worker.id filter never
                     # matched and left zombie workers stuck in "online").
                     stale_worker_ids: set[str] = set()
-                    if stale_agents:
-                        wid_rows = (
-                            await db.execute(
-                                select(Agent.worker_id).where(
-                                    Agent.id.in_(stale_agents),
-                                    Agent.guild_id == _guild_pk,
-                                    Agent.worker_id.isnot(None),
+                    try:
+                        if stale_agents:
+                            wid_rows = (
+                                await db.execute(
+                                    select(Agent.worker_id).where(
+                                        Agent.id.in_(stale_agents),
+                                        Agent.guild_id == _guild_pk,
+                                        Agent.worker_id.isnot(None),
+                                    )
                                 )
+                            ).all()
+                            stale_worker_ids = {r[0] for r in wid_rows}
+                        for agent_id in stale_agents:
+                            await db.execute(
+                                update(Agent)
+                                .where(Agent.id == agent_id, Agent.guild_id == _guild_pk)
+                                .values(state="offline", activity=None, current_task_id=None)
                             )
-                        ).all()
-                        stale_worker_ids = {r[0] for r in wid_rows}
-                    for agent_id in stale_agents:
-                        agent_owners.pop(agent_id, None)
-                        await db.execute(
-                            update(Agent)
-                            .where(Agent.id == agent_id, Agent.guild_id == _guild_pk)
-                            .values(state="offline", activity=None, current_task_id=None)
+                        # Mirror into workers table using the real worker IDs.
+                        for wid in stale_worker_ids:
+                            await db.execute(
+                                update(Worker)
+                                .where(Worker.id == wid, Worker.guild_id == _guild_pk)
+                                .values(state="offline")
+                            )
+                        if stale_agents:
+                            await db.commit()
+                    except Exception:
+                        # The DB connection may have been closed by a
+                        # CancelledError delivered during a handler (e.g.
+                        # when a close frame races a db.commit() inside the
+                        # message loop).  Log and continue so the broadcast
+                        # and foreman-trigger below still run.
+                        logger.warning(
+                            "WS teardown: DB error for guild %s — stale agents/workers"
+                            " may remain online in DB until sweeper runs",
+                            guild_id,
+                            exc_info=True,
                         )
-                    # Mirror into workers table using the real worker IDs.
-                    for wid in stale_worker_ids:
-                        await db.execute(
-                            update(Worker)
-                            .where(Worker.id == wid, Worker.guild_id == _guild_pk)
-                            .values(state="offline")
-                        )
-                    if stale_agents:
-                        await db.commit()
                 for agent_id in stale_agents:
                     await broadcast(
                         guild_id,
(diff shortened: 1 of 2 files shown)


backend/routes/websocket.py
```diff
@@ -162,39 +162,55 @@ async def websocket_endpoint(websocket: WebSocket, guild_id: str):
...
-                    # Mirror into workers table using the real worker IDs.
-                    for wid in stale_worker_ids:
-                        await db.execute(
-                            update(Worker)
-                            .where(Worker.id == wid, Worker.guild_id == _guild_pk)
-                            .values(state="offline")
+                        # Mirror into workers table using the real worker IDs.
+                        for wid in stale_worker_ids:
+                            await db.execute(
+                                update(Worker)
```
jmelloy, line 196: Catching bare `Exception` here is overly broad — it will silently swallow programming errors (e.g., `AttributeError`, `NameError`) alongside the intended `InterfaceError`/`CancelledError`. Consider catching only `(sqlalchemy.exc.InterfaceError, asyncio.CancelledError, Exception)` with a tighter guard, or at minimum re-raise anything that isn't an asyncpg/SQLAlchemy connection error so real bugs aren't hidden.

backend/routes/websocket.py
```diff
@@ -162,39 +162,55 @@ async def websocket_endpoint(websocket: WebSocket, guild_id: str):
...
+                                        Agent.id.in_(stale_agents),
+                                        Agent.guild_id == _guild_pk,
+                                        Agent.worker_id.isnot(None),
+                                    )
                                 )
+                            ).all()
+                            stale_worker_ids = {r[0] for r in wid_rows}
+                        for agent_id in stale_agents:
+                            await db.execute(
+                                update(Agent)
```
jmelloy, line 189: If `stale_agents` is non-empty but the DB block raises early (e.g., on the `SELECT worker_id` query), the per-agent `UPDATE` and `db.commit()` are both skipped, leaving agents stuck as 'online' in the DB. The log message acknowledges this, but it may be worth attempting a `db.rollback()` in the except block to ensure the connection is in a clean state for any subsequent use.

backend/tests/test_worker_foreman_notify.py
```diff
@@ -99,10 +99,14 @@ def test_worker_online_notifies_foreman(client):
...
             )
-            # No explicit sync needed: _trigger_foreman is directly awaited
-            # inside handle_worker_register, so triggered is populated before
-            # the handler returns. The WS context exit (close frame) ensures
-            # the server processes worker-register before we reach assertions.
+            # Sync: ping/pong ensures the server has fully processed
+            # worker-register (including db.commit()) before we close the
+            # WebSocket.  Without this, the close frame can arrive while
+            # handle_worker_register is still awaiting inside db.commit(),
+            # which delivers a CancelledError that corrupts the asyncpg
```
jmelloy, line 106: The ping/pong sync relies on the server responding to `{"type": "ping"}` with a pong before the `worker-register` handler has fully committed — this works because the event loop is single-threaded and processes messages in order, but it would be worth a brief comment clarifying that assumption (i.e., the server processes messages sequentially, so a pong reply guarantees the preceding handler completed).

jmelloy (commented):
- The test fix (ping/pong sync) correctly addresses the race condition and is a straightforward, low-risk change that makes the test deterministic.
- The production-code fix swallows all exceptions in the teardown path with a bare `except Exception`, which could silently hide unrelated bugs (e.g., programming errors, ORM misconfiguration) beyond the intended asyncpg connection-closed scenario.
- Moving `agent_owners.pop()` before the DB block is a good correctness improvement — ownership is always released regardless of DB failures, preventing stuck ownership state.
- The `except Exception` block logs but continues, meaning the broadcast and foreman-trigger below still run with potentially stale in-memory state; this is documented in the log message but could cause confusing behavior in production.
- No security concerns introduced; the change is scoped to connection teardown and test infrastructure.

backend/routes/websocket.py
```diff
@@ -162,39 +162,55 @@ async def websocket_endpoint(websocket: WebSocket, guild_id: str):
...
+                                update(Worker)
+                                .where(Worker.id == wid, Worker.guild_id == _guild_pk)
+                                .values(state="offline")
+                            )
+                        if stale_agents:
+                            await db.commit()
+                    except Exception:
+                        # The DB connection may have been closed by a
+                        # CancelledError delivered during a handler (e.g.
+                        # when a close frame races a db.commit() inside the
```
jmelloy, line 205: Catching bare `Exception` here is overly broad and could swallow unexpected bugs (e.g. a programming error like `NameError` or `AttributeError`). Consider catching only connection-related errors such as `sqlalchemy.exc.InterfaceError` and `asyncio.CancelledError` so that genuine bugs still propagate and are visible.

backend/routes/websocket.py
jmelloy, line 196: The `stale_worker_ids` set is initialised to empty before the try-block, which is correct. However, if the SELECT query raises before populating it, the subsequent Worker `update` loop (line ~202) will silently be a no-op, potentially leaving zombie workers online. This is noted in the warning message, but it's worth confirming the sweeper actually handles this case in practice.

backend/tests/test_worker_foreman_notify.py
```diff
@@ -99,10 +99,14 @@ def test_worker_online_notifies_foreman(client):
...
+            # connection and causes InterfaceError in the finally block.
+            ws.send_json({"type": "ping"})
```
jmelloy, line 108: If the server does not respond with a pong (e.g. the ping handler is missing or broken), `ws.receive_json()` will block until the test framework's socket timeout, causing a slow and confusing failure. Consider adding an assertion on the received message (`assert resp['type'] == 'pong'`) so a missing pong gives a clear diagnostic rather than a timeout.

jmelloy (commented):
- The test fix (ping/pong sync) is the correct and minimal approach to eliminate the close-race; it directly ensures `handle_worker_register` completes before the WebSocket tears down.
- Moving `agent_owners.pop()` before the DB try-block is a sound improvement — ownership is always released regardless of DB failures, preventing phantom ownership.
- The broad `except Exception` catch in the teardown is a reasonable defensive measure, but silently swallowing errors (including unexpected bugs unrelated to connection closure) could hide real issues; a tighter exception type (e.g. `asyncpg.InterfaceError`, `sqlalchemy.exc.InterfaceError`, or `asyncio.CancelledError`) would be safer.
- The `stale_agents` variable is referenced in the `for agent_id in stale_agents` broadcast loop below the try/except block; if the list was never populated (exception before assignment), this would raise `UnboundLocalError`, but since `stale_agents` is assigned before the try-block this is fine — however, the code's readability would benefit from a brief comment.
- Overall the fix is well-reasoned and the two-pronged approach (test synchronization + defensive production teardown) is appropriate belt-and-suspenders engineering.

PR #442, opened 2026-05-22
feat(foreman): add Bedrock support and sync standalone foreman with main patterns

jmelloy:
## Summary

- **Bedrock support** (`runner.py`, `config.py`, `pyproject.toml`): The standalone foreman was missing the `AsyncAnthropicBedrock` path that the embedded foreman gained in #430. Added `FOREMAN_PROVIDER` / `AWS_DEFAULT_REGION` env-var reading, `AsyncAnthropicBedrock` client construction, and `provider` / `aws_region` fields to `Config` with full TOML/env/override layering. The `anthropic[bedrock]` extra is now declared in `pyproject.toml`.

- **Prompt parity** (`prompt.py`): The `[worker-online]` / `[worker-offline]` worker lifecycle events section was added to the embedded foreman in PR #435 but never ported to the standalone copy. Synced it in.

- **Logging parity** (`runner.py`): Each API round now logs `cache_read` and `cache_write` token counts (matching the embedded runner), plus `system_chars`, `state_chars`, and `extra_context_chars` on the first `run_foreman_ai` log line.

- **Docs** (`pioneer-foreman.toml.example`): Added a commented Bedrock section documenting `provider` and `aws_region` options.

- **requirements.txt**: Updated stale "Phase 3 will replace this" comment now that Phase 3 is shipped; clarified the file is for the legacy `main.py` entry point only.

## Test plan

- [ ] `pioneer-foreman` starts with default config (provider=anthropic) — existing behaviour unchanged
- [ ] `FOREMAN_PROVIDER=bedrock AWS_DEFAULT_REGION=us-east-1` creates `AsyncAnthropicBedrock`; confirmed by the new startup log line
- [ ] `provider = "bedrock"` in `pioneer-foreman.toml` also triggers Bedrock path
- [ ] Worker join/leave events routed to the standalone foreman now trigger the correct lifecycle instructions from the updated system prompt
- [ ] `pip install -e foreman/` pulls in `anthropic[bedrock]` (boto3 included)

### DIFF
--- a/foreman/pioneer_foreman/runner.py
+++ b/foreman/pioneer_foreman/runner.py
@@ -9,6 +9,7 @@ from __future__ import annotations
 import copy
 import json
 import logging
+import os
 from collections.abc import Awaitable, Callable
 from datetime import UTC, datetime
 from typing import TYPE_CHECKING
@@ -35,17 +36,34 @@ _HUMAN_TURN_WINDOW = 5
 _TERMINAL_STATES = frozenset({"done", "failed", "cancelled"})
 _24H_SECS = 86_400
 
+# Set FOREMAN_PROVIDER=bedrock to use Amazon Bedrock instead of the Anthropic API.
+# Requires: pip install "anthropic[bedrock]"  +  AWS credentials in env / IAM role.
+# On Bedrock, model IDs use the "anthropic." prefix, e.g.:
+#   anthropic.claude-sonnet-4-5   (Sonnet 4.x on Bedrock)
+#   anthropic.claude-opus-4-5     (Opus 4.x on Bedrock)
+# Check your Bedrock console for exact IDs available in your region.
+FOREMAN_PROVIDER = os.environ.get("FOREMAN_PROVIDER", "anthropic").lower()
+_BEDROCK_REGION = os.environ.get("AWS_DEFAULT_REGION", "us-east-1")
+
 # Module-level Anthropic client (reused across calls)
-_anthropic_client: _anthropic.AsyncAnthropic | None = None
+_anthropic_client: _anthropic.AsyncAnthropic | _anthropic.AsyncAnthropicBedrock | None = None
 
 
-def _get_anthropic_client(api_key: str | None = None) -> _anthropic.AsyncAnthropic:
+def _get_anthropic_client(
+    config: Config,
+) -> _anthropic.AsyncAnthropic | _anthropic.AsyncAnthropicBedrock:
     global _anthropic_client
     if _anthropic_client is None:
-        kwargs: dict = {}
-        if api_key:
-            kwargs["api_key"] = api_key
-        _anthropic_client = _anthropic.AsyncAnthropic(**kwargs)
+        provider = (getattr(config, "provider", None) or FOREMAN_PROVIDER).lower()
+        region = getattr(config, "aws_region", None) or _BEDROCK_REGION
+        if provider == "bedrock":
+            _anthropic_client = _anthropic.AsyncAnthropicBedrock(aws_region=region)
+            logger.info("Foreman using Amazon Bedrock (region=%s, model=%s)", region, config.model)
+        else:
+            kwargs: dict = {}
+            if getattr(config, "api_key", None):
+                kwargs["api_key"] = config.api_key
+            _anthropic_client = _anthropic.AsyncAnthropic(**kwargs)
     return _anthropic_client
 
 
@@ -356,10 +374,14 @@ async def run_foreman_ai(
         )
 
         logger.info(
-            "guild=%s run_foreman_ai: workers=%d tasks_in_context=%d",
+            "guild=%s run_foreman_ai: workers=%d tasks_in_context=%d "
+            "system_chars=%d state_chars=%d extra_context_chars=%d",
             guild_id,
             len(worker_rows),
             len(summarized_tasks),
+            len(system_blocks[0]["text"]),
+            len(state_preamble),
+            len(extra_context),
         )
 
         await _save_turn(guild_id, user_id, "system", audit_system, http=http)
@@ -368,7 +390,7 @@ async def run_foreman_ai(
 
         _inject_state_preamble(messages, state_preamble)
 
-        client = _get_anthropic_client(config.api_key)
+        client = _get_anthropic_client(config)
 
         text_parts = []
         for round_num in range(config.max_rounds):
@@ -392,11 +414,15 @@ async def run_foreman_ai(
             _input_tokens = getattr(usage, "input_tokens", 0) or 0
             _output_tokens = getattr(usage, "output_tokens", 0) or 0
             logger.info(
-                "guild=%s round %d: stop_reason=%s input=%d output=%d",
+                "guild=%s round %d: stop_reason=%s content_blocks=%d "
+                "input=%d cache_read=%d cache_write=%d output=%d",
                 guild_id,
                 round_num,
                 resp.stop_reason,
+                len(resp.content),
                 _input_tokens,
+                getattr(usage, "cache_read_input_tokens", 0) or 0,
+                getattr(usage, "cache_creation_input_tokens", 0) or 0,
                 _output_tokens,
             )
 
@@ -499,6 +525,15 @@ async def run_foreman_ai(
             wrap_usage = wrap_resp.usage
             _wrap_input = getattr(wrap_usage, "input_tokens", 0) or 0
             _wrap_output = getattr(wrap_usage, "output_tokens", 0) or 0
+            logger.info(
+                "guild=%s wrap-up: stop_reason=%s input=%d cache_read=%d cache_write=%d output=%d",
+                guild_id,
+                wrap_resp.stop_reason,
+                _wrap_input,
+                getattr(wrap_usage, "cache_read_input_tokens", 0) or 0,
+                getattr(wrap_usage, "cache_creation_input_tokens", 0) or 0,
+                _wrap_output,
+            )
             wrap_turn_id = await _save_turn(
                 guild_id, user_id, "assistant", wrap_resp.content, http=http
             )
--- a/foreman/pyproject.toml
+++ b/foreman/pyproject.toml
@@ -10,7 +10,7 @@ requires-python = ">=3.11"
 dependencies = [
     "websockets>=12.0",
     "httpx>=0.27",
-    "anthropic>=0.49",
+    "anthropic[bedrock]>=0.49",
     "anyio>=4.0",
 ]
 
--- a/foreman/pioneer-foreman.toml.example
+++ b/foreman/pioneer-foreman.toml.example
@@ -17,6 +17,14 @@ api_key       = ""                     # falls back to ANTHROPIC_API_KEY env var
 max_rounds    = 10
 history_limit = 40
 
+# AI provider: "anthropic" (default) or "bedrock" (Amazon Bedrock).
+# When using Bedrock: set provider = "bedrock", ensure AWS credentials are
+# available (env vars, ~/.aws/credentials, or IAM role), and update model to
+# the Bedrock model ID (e.g. "anthropic.claude-sonnet-4-5").
+# Requires: pip install "anthropic[bedrock]"
+# provider   = "anthropic"
+# aws_region = "us-east-1"             # falls back to AWS_DEFAULT_REGION env var
+
 [poll]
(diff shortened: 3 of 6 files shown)


foreman/pioneer_foreman/runner.py
```diff
@@ -35,17 +36,34 @@
...
-_anthropic_client: _anthropic.AsyncAnthropic | None = None
+_anthropic_client: _anthropic.AsyncAnthropic | _anthropic.AsyncAnthropicBedrock | None = None
 
 
-def _get_anthropic_client(api_key: str | None = None) -> _anthropic.AsyncAnthropic:
+def _get_anthropic_client(
+    config: Config,
+) -> _anthropic.AsyncAnthropic | _anthropic.AsyncAnthropicBedrock:
     global _anthropic_client
     if _anthropic_client is None:
```
jmelloy, line 56: The module-level `FOREMAN_PROVIDER` / `_BEDROCK_REGION` constants are read once at import time, but `_anthropic_client` is cached for the lifetime of the process after the first `_get_anthropic_client` call. If two different `Config` objects are passed (e.g., in tests or a future multi-guild scenario where configs differ), the second call silently returns the client built for the first config. Consider keying the cache on `(provider, region)` or simply constructing the client inside `run_foreman_ai` and storing it on the Config/context object rather than as a module global.

foreman/pioneer_foreman/runner.py
```diff
@@ -35,17 +36,34 @@
...
-        if api_key:
-            kwargs["api_key"] = api_key
-        _anthropic_client = _anthropic.AsyncAnthropic(**kwargs)
+        provider = (getattr(config, "provider", None) or FOREMAN_PROVIDER).lower()
+        region = getattr(config, "aws_region", None) or _BEDROCK_REGION
+        if provider == "bedrock":
+            _anthropic_client = _anthropic.AsyncAnthropicBedrock(aws_region=region)
+            logger.info("Foreman using Amazon Bedrock (region=%s, model=%s)", region, config.model)
+        else:
+            kwargs: dict = {}
```
jmelloy, line 63: `getattr(config, 'provider', None)` is unnecessarily defensive — `Config` is a dataclass with `provider` declared as a required field with a default. Using plain `config.provider` is cleaner and will surface any accidental attribute removal at development time rather than silently falling back to the env-var path.

foreman/pyproject.toml
```diff
@@ -10,7 +10,7 @@ requires-python = ">=3.11"
 dependencies = [
     "websockets>=12.0",
     "httpx>=0.27",
-    "anthropic>=0.49",
+    "anthropic[bedrock]>=0.49",
```
jmelloy, line 13: Making `anthropic[bedrock]` a hard dependency forces `boto3`, `botocore`, and related AWS libraries (~50 MB+) onto every installation even when Bedrock is never used. Consider keeping `anthropic>=0.49` as the base dependency and documenting `pip install 'anthropic[bedrock]'` as an optional step in the example config, matching the pattern already used in the example TOML comments.

jmelloy (commented):
- **Module-level globals create a provider-switching bug**: `FOREMAN_PROVIDER` and `_BEDROCK_REGION` are read once at import time from env vars, but `_anthropic_client` is cached after the first call — if `config.provider` differs from the env var on a second call with a different config, the cached client from the first call is silently reused regardless.
- **`getattr` defensive coding masks real bugs**: Multiple `getattr(config, 'provider', None)` and `getattr(config, 'api_key', None)` calls are used on a typed `Config` dataclass where these fields are always present; this pattern hides typos and attribute renames at the cost of clarity.
- **`pyproject.toml` change makes `bedrock` a hard dependency**: Changing `anthropic>=0.49` to `anthropic[bedrock]>=0.49` forces `boto3` and related AWS SDKs onto all users even when Bedrock is never used, significantly increasing install size and potential supply-chain surface.
- **Prompt parity and logging changes are clean**: The worker lifecycle event documentation sync and cache token logging additions are straightforward, correct, and match the described intent well.
- **`requirements.txt` comment update is accurate and helpful** but the file still pins `anthropic[bedrock]>=0.40.0` while `pyproject.toml` now also pins bedrock — these two dependency files could diverge in version requirements over time.

foreman/pioneer_foreman/runner.py
```diff
@@ -356,10 +374,14 @@ async def run_foreman_ai(
         )
 
         logger.info(
-            "guild=%s run_foreman_ai: workers=%d tasks_in_context=%d",
+            "guild=%s run_foreman_ai: workers=%d tasks_in_context=%d "
+            "system_chars=%d state_chars=%d extra_context_chars=%d",
```
jmelloy, line 378: `system_blocks[0]["text"]` will raise `IndexError` if `system_blocks` is empty, and `KeyError` if the first block is not a text block (e.g., a cache-control block). Add a guard such as `next((b['text'] for b in system_blocks if b.get('type') == 'text'), '')` to avoid a crash on unexpected prompt shapes.

foreman/pioneer_foreman/runner.py
```diff
@@ -35,17 +36,34 @@
...
+def _get_anthropic_client(
+    config: Config,
+) -> _anthropic.AsyncAnthropic | _anthropic.AsyncAnthropicBedrock:
     global _anthropic_client
     if _anthropic_client is None:
-        kwargs: dict = {}
-        if api_key:
-            kwargs["api_key"] = api_key
-        _anthropic_client = _anthropic.AsyncAnthropic(**kwargs)
+        provider = (getattr(config, "provider", None) or FOREMAN_PROVIDER).lower()
```
jmelloy, line 57: The module-level `FOREMAN_PROVIDER` / `_BEDROCK_REGION` constants are read at import time from env vars, but the actual client construction already prefers `config.provider` / `config.aws_region` via `getattr`. These module-level constants are only a fallback and are not documented as such, making the precedence rules easy to misread. Consider removing the module-level constants entirely and relying solely on the `Config` object (with `Config.load()` already reading the same env vars), or at minimum add a comment clarifying they are last-resort fallbacks.

foreman/pioneer_foreman/runner.py
```diff
@@ -35,17 +36,34 @@
...
+        region = getattr(config, "aws_region", None) or _BEDROCK_REGION
+        if provider == "bedrock":
+            _anthropic_client = _anthropic.AsyncAnthropicBedrock(aws_region=region)
+            logger.info("Foreman using Amazon Bedrock (region=%s, model=%s)", region, config.model)
+        else:
```
jmelloy, line 62: The `_anthropic_client` singleton is initialised from the first `Config` passed to `_get_anthropic_client`. A second call with a different `config.provider` or `config.aws_region` silently reuses the wrong client. This is unlikely today but will cause silent misconfiguration in tests or if the function is ever called with different configs. Consider keying the cache on `(provider, region)` or constructing the client inline and letting callers cache it.

foreman/pioneer_foreman/runner.py
```diff
@@ -356,10 +374,14 @@ async def run_foreman_ai(
...
 
         logger.info(
-            "guild=%s run_foreman_ai: workers=%d tasks_in_context=%d",
+            "guild=%s run_foreman_ai: workers=%d tasks_in_context=%d "
+            "system_chars=%d state_chars=%d extra_context_chars=%d",
             guild_id,
             len(worker_rows),
             len(summarized_tasks),
+            len(system_blocks[0]["text"]),
+            len(state_preamble),
```
jmelloy, line 383: `system_blocks[0]["text"]` will raise an `IndexError` if `system_blocks` is empty for any reason. Add a guard such as `len(system_blocks[0].get('text', ''))` with a fallback, or assert/log a warning when the list is unexpectedly empty.

foreman/pyproject.toml
jmelloy, line 13: Changing the default dependency from `anthropic>=0.49` to `anthropic[bedrock]>=0.49` forces every user to install `boto3` and `botocore` regardless of whether they use Bedrock. This adds ~30 MB of transitive dependencies. Prefer an optional/extras dependency: `[project.optional-dependencies]` with a `bedrock` extra, and keep the base dependency as plain `anthropic>=0.49`.

jmelloy (commented):
- Module-level `FOREMAN_PROVIDER` and `_BEDROCK_REGION` constants are read at import time, but the `_get_anthropic_client` function also accepts a `Config` object — the config values correctly override the module-level ones, yet the module-level constants serve no purpose beyond the cached client and add confusion.
- The `_anthropic_client` singleton is module-level and is initialized once from whichever `Config` is passed first; if two different configs (e.g., different regions or providers) are used in the same process the second call silently reuses the first client — this is a latent correctness bug.
- `system_blocks[0]["text"]` is accessed without a bounds check; if `system_blocks` is empty this will raise an `IndexError` at runtime.
- The `pyproject.toml` change makes `anthropic[bedrock]` a hard dependency for all users, forcing `boto3`/`botocore` installation even for users who never use Bedrock — this should remain optional.
- Prompt and logging parity changes are clean and well-structured; the documented Bedrock example config is clear and accurate.

jmelloy (commented):
## Review of PR #442: Add Bedrock support and sync standalone foreman

### Summary
The prompt parity, logging additions, and docs changes are clean and well-structured. Three issues need addressing before merge.

---

### Critical: module-level client cache doesn't respect `config` values

`_anthropic_client` is a singleton initialized on first call. If `_get_anthropic_client(config)` is called twice with different `config.provider` or `config.aws_region` values (e.g., in tests, or a future multi-guild scenario), the second call silently returns the client built for the first config:

```python
# First call: config.provider = "bedrock", region = "us-east-1"  → creates Bedrock client
# Second call: config.provider = "anthropic"  → returns the Bedrock client!
```

The `getattr(config, "provider", None) or FOREMAN_PROVIDER` fallback reads the module-level env constant, but once `_anthropic_client` is set that branch never runs again. **Fix**: key the cache on `(provider, region, api_key)` using a dict:

```python
_anthropic_clients: dict[tuple, ...] = {}

def _get_anthropic_client(config: Config) -> ...:
    key = (config.provider.lower(), config.aws_region, config.api_key or "")
    if key not in _anthropic_clients:
        if config.provider.lower() == "bedrock":
            _anthropic_clients[key] = _anthropic.AsyncAnthropicBedrock(aws_region=config.aws_region)
        else:
            kwargs = {"api_key": config.api_key} if config.api_key else {}
            _anthropic_clients[key] = _anthropic.AsyncAnthropic(**kwargs)
    return _anthropic_clients[key]
```

This also lets you remove the redundant module-level `FOREMAN_PROVIDER` / `_BEDROCK_REGION` constants — `Config.load()` already reads the same env vars, so the runner can just use `config.provider` / `config.aws_region` directly.

---

### Critical: `pyproject.toml` makes `bedrock` a hard dependency

```toml
-    "anthropic>=0.49",
+    "anthropic[bedrock]>=0.49",
```

This forces `boto3`, `botocore`, and related AWS packages (~30–50 MB) onto every installation even when Bedrock is never used. The example config already documents the optional install step correctly — match that in `pyproject.toml`:

```toml
# base dependency (unchanged)
"anthropic>=0.49",

# in [project.optional-dependencies]
bedrock = ["anthropic[bedrock]>=0.49"]
```

Then users who want Bedrock run `pip install "pioneer-foreman[bedrock]"`.

---

### Moderate: `system_blocks[0]["text"]` is unguarded

```python
len(system_blocks[0]["text"]),
```

Raises `IndexError` if `system_blocks` is empty, and `KeyError` if the first block isn't a text block. Use a safe accessor:

```python
next((len(b["text"]) for b in system_blocks if b.get("type") == "text"), 0)
```

---

### Minor: `getattr` defensive coding on a typed dataclass

```python
provider = (getattr(config, "provider", None) or FOREMAN_PROVIDER).lower()
region = getattr(config, "aws_region", None) or _BEDROCK_REGION
```

`Config` is a dataclass with `provider` and `aws_region` declared with defaults — both attributes are always present. Using `getattr(..., None)` hides typos and attribute renames silently. Plain `config.provider` and `config.aws_region` are cleaner and fail loudly if something goes wrong.

---

### What's good
- Prompt parity (`[worker-online]` / `[worker-offline]` lifecycle section) is correct and matches the embedded foreman exactly.
- Cache token logging (`cache_read_input_tokens`, `cache_creation_input_tokens`) is the right pattern and matches the embedded runner.
- `pioneer-foreman.toml.example` Bedrock documentation is accurate and helpful.
- `requirements.txt` comment cleanup is accurate — Phase 3 shipped.

PR #444, opened 2026-05-22
Audit & fix: stop workers opening PRs from review results

jmelloy:
## Summary

- **Root cause (PR #443):** The foreman assigned a "Review PR #442" task to a worker via `assign_task`. Workers always end by committing and opening a PR, so the worker committed its review findings to files and opened PR #443 — wrong output.
- Three independent layers of guardrails added to prevent recurrence.

### `worker/pioneer_worker/worker.py`
Tasks with `phase="review"` no longer receive the `git add -A && git commit && gh pr create` footer. Instead, they receive explicit instructions to post findings via `gh pr review` or the GitHub Reviews API, with a hard "do NOT push commits or open a PR" warning.

### Foreman system prompts (3 copies: `backend/foreman/`, `foreman/pioneer_foreman/`, `pioneer-foreman/pioneer_foreman/`)
- **Multi-step flows:** the `review` bullet now says to use `review_pr_internal`/`review_pr`, not `assign_task`, and includes an explicit NEVER line.
- **PR review tracking:** new CRITICAL block explains that review output goes to GitHub PR review comments only — never to a new PR — and warns that assigning a worker produces the wrong result.
- `pioneer-foreman/` also gains the full PR review tracking section it was missing.

### Tool descriptions (3 copies of `tools.py`)
- **`assign_task`:** new WARNING that it must not be used for PR code reviews.
- **`review_pr` / `review_pr_internal`:** both now say they are the CORRECT path and post findings to the original PR, never open a new one.

## Test plan
- [ ] Verify `phase="review"` tasks in `worker.py` do not include `gh pr create` in their injected description
- [ ] Verify all 3 foreman prompts contain the CRITICAL review guardrail
- [ ] Verify `assign_task` description in all 3 tools.py carries the WARNING
- [ ] Confirm `ruff check` passes (clean)


worker/pioneer_worker/worker.py
```diff
@@ -1722,13 +1722,35 @@ def _on_proc(proc: claude_runner.ClaudeProcess) -> None:
...
                     pr_title = (task.get("name") or desc)[:72].replace('"', "'")
                     closes = f" Closes #{task['issue_number']}" if task.get("issue_number") else ""
-                    current_desc = (
-                        f"{desc}\n\n"
-                        f"After completing your changes, commit, push, and open a PR:\n"
-                        f'  git add -A && git commit -m "<concise commit message>"\n'
-                        f"  git push origin {branch}\n"
-                        f'  gh pr create --title "{pr_title}" --body "<summary of changes>{closes}"\n'
-                    )
+                    if task.get("phase") == "review":
```
jmelloy: The guard is a case-sensitive exact match on `'review'`. If the foreman ever passes `phase='Review'` or `phase='REVIEW'`, the worker will silently fall through to the commit/PR footer. Consider `(task.get('phase') or '').lower() == 'review'` to make this robust against capitalisation differences.

worker/pioneer_worker/worker.py
```diff
@@ -1722,13 +1722,35 @@ def _on_proc(proc: claude_runner.ClaudeProcess) -> None:
...
+                        # Review-phase tasks must post findings as GitHub PR review
+                        # comments — NEVER by committing files and opening a new PR.
+                        current_desc = (
```
jmelloy, line 1739: The review-phase prompt tells the worker to call `gh pr review <PR-URL>` but the task description may not contain a PR URL — it might only have a PR number. The inline API example also hard-codes `OWNER/REPO` placeholders. Consider injecting the actual repository and PR number from the task metadata so the worker has a concrete, actionable command rather than a template it must fill in.

backend/foreman/tools.py
```diff
@@ -98,9 +98,13 @@ def _resolve_finalize_deleted_at(inp: dict) -> tuple[str, str | None]:
         "name": "assign_task",
         "description": (
             "Queue a coding task for a worker agent. The worker creates a git worktree, "
-            "runs the chosen coding agent on the description, then pushes the branch. "
+            "runs the chosen coding agent on the description, then pushes the branch and "
```
jmelloy: The WARNING added to the `assign_task` description is purely advisory text for the LLM; there is no runtime guard that rejects or re-routes a review-phase task submitted via `assign_task`. If the foreman ignores the warning, the bug will recur silently. Consider adding a server-side check that logs or raises when `assign_task` is called with `phase='review'`.

pioneer-foreman/pioneer_foreman/prompt.py
```diff
@@ -34,7 +34,31 @@
 For complex work use phases:
 1. **plan** — create_task(phase='plan'), assign a worker to produce an outline/spec
 2. **execute** — assign workers to implement
-3. **review** — assign a worker to verify correctness, run tests, check the PR
+3. **review** — use review_pr_internal (or review_pr) to post a GitHub PR review; do NOT assign a worker to "review a PR" — that causes the worker to open a new PR with its findings instead of posting review comments
```
jmelloy, line 37: The 'PR review tracking' block (create_task → review_pr_internal → finalize_task pattern) was added only to the `pioneer-foreman` copy, while the other two foreman copies (`backend/foreman/prompt.py` and `foreman/pioneer_foreman/prompt.py`) already had this block further down in the file. Verify that neither of the other two copies is missing this new CRITICAL block, and consider consolidating the three copies to avoid future drift.

jmelloy (commented):
- The fix correctly targets the root cause: review-phase worker tasks were receiving git commit/PR instructions, causing spurious PRs. The new branch in `worker.py` strips those instructions and replaces them with `gh pr review` guidance.
- Three layers of defence (worker prompt, tool descriptions, foreman system prompts) provide good redundancy, though all three layers ultimately rely on the LLM reading and following text instructions — there is no hard enforcement.
- The three foreman directories (`backend/foreman/`, `foreman/pioneer_foreman/`, `pioneer-foreman/pioneer_foreman/`) appear to be near-identical copies; the `pioneer-foreman` copy received extra content (the PR review tracking block) that was already present in the other two, which is good, but the triplication itself is a maintenance risk.
- The worker-side guard only fires when `task.get('phase') == 'review'`; if the foreman omits the phase field or uses a different string (e.g. `'Review'`), the worker will still append the commit/PR footer — a small but real gap.
- No automated test is added to assert the correct prompt is generated for `phase='review'` tasks, leaving the fix unverifiable by CI.

jmelloy (commented):
## Overall Assessment

Good multi-layer approach. The system prompt changes, tool description warnings, and runtime guard all address the root cause correctly. However, there is a **critical bug** that makes the primary worker.py guard completely ineffective: the `phase` field is never populated in the task dict when a task arrives via WebSocket (which is the normal assignment path).

---

## Critical Bug: `phase` is missing from the WS task dict

**File:** `worker/pioneer_worker/worker.py`, `task-assigned` handler (around the `task_queue.put` call)

The dict built from the incoming `task-assigned` WebSocket message does **not** include `"phase"`:

```python
await self.task_queue.put(
    {
        "id": task_id,
        ...
        "tool": msg.get("tool", "claude"),
        "issue_number": msg.get("issueNumber"),
        "issue_repo": msg.get("issueRepo"),
        "repos": msg.get("repos") or [],
        # ← "phase" is missing here
    }
)
```

But the backend **does** send `phase` in the `task-assigned` message (visible in `backend/foreman/tools.py` around the `broadcast(...)` call, which includes `"phase": phase`).

Because `phase` is absent from the task dict, `task.get("phase")` always returns `None` in `_task_runner`, which means the entire review-path guard introduced in this PR (`if (task.get("phase") or "").lower() == "review":`) will never fire for any task assigned via WebSocket — which is the primary assignment path.

The idle-puller path (`_fetch_pending_tasks`) returns raw DB rows via `row_to_dict`, which **do** include `phase`, so that path would work. But the WS path is where new tasks land, making the guard effectively dead code for the common case.

**Fix:** add `"phase": msg.get("phase")` to the `task_queue.put(...)` dict in the `task-assigned` handler. The same omission exists in the `task-followup` handler dict — worth fixing there too for consistency.

---

## Minor: Runtime guard exists in only one of three `tools.py` copies

The `logger.warning(...)` guard that fires when `assign_task` is called with `phase='review'` was added to `backend/foreman/tools.py` but **not** to `foreman/pioneer_foreman/tools.py` or `pioneer-foreman/pioneer_foreman/tools.py`.

The standalone-foreman paths (which use those two tools.py copies) would log nothing if a foreman incorrectly calls `assign_task(phase='review')`. The same guard should be in all three for consistent defense-in-depth.

---

## Minor: Fallback placeholder is not actionable

In `worker.py`, when `_pr_repo` or `_pr_num` is falsy the injected instructions include:

```
gh pr review <PR-URL> --comment --body '...'
gh api repos/OWNER/REPO/pulls/NUMBER/reviews ...
```

A worker receiving this literal placeholder will run `gh pr review <PR-URL>` which will fail loudly. Consider logging a warning at the point where the placeholder is used so the problem is visible in the task output, or making the injected text tell the worker to report an error rather than attempt an unparseable command.

---

## What's working well

- System prompt changes across all three foreman copies are clear and direct; the NEVER/CRITICAL repetition is appropriate given the severity.
- `pioneer-foreman/pioneer_foreman/prompt.py` correctly gains the full "## PR review tracking" section that was missing there.
- Tool description updates accurately set the foreman's expectations for `assign_task`, `review_pr`, and `review_pr_internal`.
- Three-layer approach (prompt + tool descriptions + runtime guard) is the right philosophy.
- No ruff/lint issues visible in the diff.

The PR is one line away from being correct — just add `"phase": msg.get("phase")` to the WS task dict. Request changes for that fix before merging.

backend/foreman/tools.py
```diff
@@ -1108,6 +1116,14 @@ async def _exec_one_tool(guild_id: str, tu, user_id: str | None = None) -> dict:
                 wid = inp["worker_id"]
                 desc = inp.get("description", "")
                 phase = inp.get("phase", "execute")
+                if (phase or "").lower() == "review":
+                    logger.warning(
+                        "assign_task called with phase='review' for worker %s — "
+                        "review tasks must use review_pr_internal or review_pr, not assign_task. "
+                        "The worker will commit findings and open a new PR instead of posting "
```
jmelloy, line 613: This warning log fires but execution continues unimpeded — the foreman's misuse of `assign_task` for a review still proceeds. Consider returning an error response (or at least appending a strong error message to the task description) so the foreman receives clear feedback that it used the wrong tool, rather than silently logging and proceeding.

worker/pioneer_worker/worker.py
```diff
@@ -1722,13 +1722,43 @@ def _on_proc(proc: claude_runner.ClaudeProcess) -> None:
...
+                            "  - Post your findings directly as a GitHub PR review using the gh CLI or API.\n"
+                            "  - NEVER create a new branch, commit review findings to files, or open a new PR.\n"
+                            "  - The review is complete when you post it, for example:\n"
+                            f"      gh pr review {_pr_ref} --comment --body '<your findings>'\n"
+                            "    or for APPROVE/REQUEST_CHANGES:\n"
+                            f"      gh pr review {_pr_ref} --approve --body '<comment>'\n"
+                            f"      gh pr review {_pr_ref} --request-changes --body '<what needs fixing>'\n"
+                            "    or via the API for inline comments:\n"
+                            f"      gh api {_api_ref}/reviews \\\n"
+                            "        -f body='...' -f event='COMMENT|APPROVE|REQUEST_CHANGES' \\\n"
```
jmelloy, line 1752: When `_pr_repo` or `_pr_num` is falsy the worker receives placeholder strings like `<PR-URL>` in the `gh pr review` command. The worker will either skip posting the review or produce a confusing error. A warning log and/or a clearer fallback instruction (e.g., "Look for the PR URL in the task description and post your review there") would be safer than silent placeholder substitution.

pioneer-foreman/pioneer_foreman/prompt.py
```diff
@@ -34,7 +34,31 @@
...
+
+NEVER assign a worker to review a PR and expect it to post comments. Workers that receive review instructions will try to commit findings and open a new PR — the wrong output. Always use review_pr_internal or review_pr for PR code reviews.
+
+## PR review tracking
+review_pr_internal and review_pr always run as tracked tasks — every review must have \
+a sidebar entry so the human can see what was reviewed and what was found.
+
```
jmelloy, line 44: The `## PR review tracking` block (steps 1–3 with `create_task` / `review_pr_internal` / `finalize_task`) was added only to `pioneer-foreman/pioneer_foreman/prompt.py` and is absent from the other two copies. If this content is intentional and correct it should be applied to `backend/foreman/prompt.py` and `foreman/pioneer_foreman/prompt.py` as well, otherwise the three forks will behave inconsistently.

worker/pioneer_worker/worker.py
```diff
@@ -1722,13 +1722,43 @@ def _on_proc(proc: claude_runner.ClaudeProcess) -> None:
...
+                        else:
+                            _pr_ref = "<PR-URL>"
+                            _api_ref = "repos/OWNER/REPO/pulls/NUMBER"
+                        current_desc = (
+                            f"{desc}\n\n"
+                            "IMPORTANT — this is a review-phase task.\n"
+                            "If you are reviewing a pull request:\n"
+                            "  - Post your findings directly as a GitHub PR review using the gh CLI or API.\n"
+                            "  - NEVER create a new branch, commit review findings to files, or open a new PR.\n"
+                            "  - The review is complete when you post it, for example:\n"
```
jmelloy, line 1745: The review-phase branch builds a worktree and sets up a branch (`branch` variable used in the else path) before reaching this point. For a pure review task no git branch is needed at all; consider short-circuiting the worktree setup for `phase='review'` so no stale branch/worktree is left behind after the review completes.

jmelloy (commented):
- The core fix in `worker.py` correctly gates the commit/PR footer on `phase != 'review'` and substitutes `gh pr review` instructions instead, directly addressing the root cause of PR #443.
- The warning log added in `tools.py` (`_exec_one_tool`) is informational only — it does not block or reject the misuse; a foreman that repeatedly makes this mistake will still proceed and the worker will still misbehave if the review task lacks `issue_repo`/`issue_number`.
- Triplication of prompt/tool changes across `backend/foreman/`, `foreman/pioneer_foreman/`, and `pioneer-foreman/pioneer_foreman/` suggests a structural duplication problem in the codebase; these three copies will likely drift again, as `pioneer-foreman/pioneer_foreman/prompt.py` already received extra content (`## PR review tracking` block) that the other two did not.
- The fallback path in `worker.py` when `_pr_repo` or `_pr_num` is missing produces placeholder strings (`<PR-URL>`, `repos/OWNER/REPO/pulls/NUMBER`) which would cause `gh` commands to fail silently or with a confusing error; there is no warning or early-exit for this case.
- Overall the layered-guardrails approach is pragmatic and the changes are low-risk, but the lack of a hard enforcement path (e.g., returning an error from `assign_task` when `phase='review'`) means the fix relies entirely on prompt-steering, which is inherently fragile.

backend/foreman/tools.py
jmelloy, line 613: This logs a warning but still proceeds to queue the task. Consider returning an error dict here (e.g. `{'error': 'assign_task with phase=review is not supported; use review_pr_internal instead'}`) so the foreman LLM gets immediate corrective feedback rather than silently proceeding to create a bad PR.

worker/tests/test_review_phase_guard.py
```diff
@@ -0,0 +1,202 @@
...
+        captured_desc.append(desc)
+        return True, "end_turn", "done"
+
+    with (
+        patch("pioneer_worker.worker.git_ops.ensure_repo", return_value="/tmp/fake-repo"),
+        patch("pioneer_worker.worker.git_ops.create_worktree", return_value=True),
+        patch("pioneer_worker.worker.github_pr.push_branch", return_value=True),
+        patch("pioneer_worker.worker.github_pr.find_existing_pr", return_value=None),
+        patch("pioneer_worker.worker.claude_runner.run_claude_auto", side_effect=fake_run_claude),
+        patch.dict(os.environ, {}),
```
jmelloy, line 185: If `fake_run_claude` is never invoked (e.g. due to an early-exit code path or a changed call site), `captured_desc` will be empty and the `assert captured_desc` guard raises, masking the real issue. Prefer initialising `captured_desc` to a sentinel and asserting it was populated with a clearer message, or assert `len(captured_desc) == 1` explicitly.

worker/pioneer_worker/worker.py
```diff
@@ -1722,13 +1725,43 @@ def _on_proc(proc: claude_runner.ClaudeProcess) -> None:
...
+                    if (task.get("phase") or "").lower() == "review":
+                        # Review-phase tasks must post findings as GitHub PR review
+                        # comments — NEVER by committing files and opening a new PR.
+                        _pr_repo = task.get("issue_repo") or ""
+                        _pr_num = task.get("issue_number")
+                        if _pr_repo and _pr_num:
+                            _pr_ref = f"https://github.com/{_pr_repo}/pull/{_pr_num}"
+                            _api_ref = f"repos/{_pr_repo}/pulls/{_pr_num}"
+                        else:
+                            _pr_ref = "<PR-URL>"
```
jmelloy, line 1737: When `issue_repo` or `issue_number` are absent, `_pr_ref` falls back to the literal string `'<PR-URL>'`, which will be emitted verbatim to the agent. It would be safer to also log a warning here so operators know the review task was dispatched without a PR reference, since the agent won't know which PR to post to.

worker/tests/test_review_phase_guard.py
```diff
@@ -0,0 +1,202 @@
...
+    """When phase='review', _execute_task must build a description that forbids
+    opening a new PR, not the usual 'commit/push/gh pr create' instructions.
+
+    We verify this by inspecting the description passed to the runner, which is
+    captured in the emit log before runner launch.
+    """
+    import os
+    import tempfile
+    from unittest.mock import patch
+
```
jmelloy, line 155: The `task-redirect` handler re-queues the task as a follow-up internally; this test asserts `phase='review'` is in the queue but does not verify the constructed follow-up message structure matches what `task-followup` handler expects. A more integrated assertion (e.g. checking the full dict shape) would prevent future handler divergence.

jmelloy (commented):
- The core fix correctly threads `phase` through all three WS message handlers (`task-assigned`, `task-followup`, `task-redirect`) so the review guard in `_execute_task` can fire — this was the actual bug.
- The three-layer defence (prompt text, tool description warning, runtime logger warning + worker description injection) is a sound strategy for an LLM-driven system where any single layer can be ignored.
- The runtime guard in `backend/foreman/tools.py` only logs a warning but does **not** block or redirect the call — a determined or confused LLM will still proceed and create the wrong PR.
- The test for `_execute_task` (`test_review_phase_injects_no_pr_instructions`) patches `claude_runner.run_claude_auto` but the real code path calls `_run_claude` / `run_claude_auto` conditionally; if the call site changed the test could silently never invoke the fake, making `captured_desc` empty and the assertion vacuously pass only via the guard `assert captured_desc`.
- There is no test covering the `backend/foreman/tools.py` runtime warning path, leaving the logger call untested.

PR #447, opened 2026-05-22
Fix: clear agent state when task parks awaiting-review

### LINKED ISSUE #446: Bug: agent state not cleared when task parks in awaiting-review
## Problem

When a worker task completes and parks in `awaiting-review`, the associated agent's `agent_state` is not being reset. The agent shows as `working` (or similar active state) in the task record even after the worker has logged "Done" and pushed the branch.

Example: task `t-14i21x` finished with "✓ Done in 59 turns" and the branch was pushed, but `get_task_status` still returns `agent_state: "working"`.

## Expected behaviour

When a task transitions to `awaiting-review`, the agent record attached to the task should be cleared or set to an idle/done state so the foreman can accurately read task health and worker availability.

## Impact

- Foreman periodic-checks misread healthy tasks as potentially stalled
- Worker agent slot may appear occupied, affecting scheduling decisions
- Hard to distinguish genuinely stuck agents from cleanly finished ones

## Steps to reproduce

1. Assign a followup task to a worker
2. Worker completes, pushes branch, logs "Done"
3. Call `get_task_status` — `agent.agent_state` still shows `"working"`

## Likely location

Look at the task state-transition code that sets `state = awaiting-review` — it probably doesn't update the agent record at the same time.

jmelloy:
## Summary

- `handle_task_complete` and `handle_task_followup_done` now immediately set the agent to `idle` (clearing `current_task_id` and `activity`) in the same DB transaction that moves the task to `awaiting-review`
- A synthesized `agent-state: idle` broadcast is emitted right after the commit so the frontend updates without waiting for the worker's duplicate message
- Previously, the agent stayed in `working` state until the worker's own `agent-state: idle` message arrived; the foreman's `get_task_status` call (triggered immediately after) would see stale `agent_state: "working"` on an already-parked task

## Root cause

The worker message sequence is:
1. `task-update` (state=awaiting-review) → task parked
2. `task-complete` / `task-followup-done` → foreman triggered
3. `agent-state: idle` → agent cleared

The foreman fires at step 2 and calls `get_task_status` before step 3 completes, hitting the race window.

## Test plan

- [ ] `test_task_complete_clears_agent_state` — verifies agent row is idle and `current_task_id=None` before the worker's `agent-state:idle` arrives
- [ ] `test_task_followup_done_clears_agent_state` — same for the follow-up path
- [ ] Existing `test_agent_state_ws_persistence.py` tests continue to pass (idempotent: the subsequent `agent-state:idle` from the worker is a no-op)

Closes #446

### DIFF
--- a/backend/tests/test_agent_state_ws_persistence.py
+++ b/backend/tests/test_agent_state_ws_persistence.py
@@ -327,6 +327,127 @@ def test_guild_get_returns_current_task_id(client):
             assert agents_by_id[agent_id]["activity"] == "editing"
 
 
+def test_task_complete_clears_agent_state(client):
+    """task-complete with agentId must immediately clear the agent to idle.
+
+    The worker sends agent-state:idle only after task-complete, so there is a
+    window where the foreman can call get_task_status and see agent_state
+    'working' on an already-parked awaiting-review task. Fixes #446.
+    """
+    test_client, db_url = client
+    guild_id, worker_id, task_id, agent_id = "gas-6", "w-gas6", "t-gas6", "a-gas6"
+    _insert_guild_worker_task(db_url, guild_id=guild_id, worker_id=worker_id, task_id=task_id)
+
+    with test_client.websocket_connect(f"/ws/{guild_id}") as ws_worker:
+        _join_ws(ws_worker, agent_id, worker_id)
+        with test_client.websocket_connect(f"/ws/{guild_id}") as ws_obs:
+            _join_ws(ws_obs, "a-obs6")
+            ws_worker.receive_json()  # drain obs's agent-joined broadcast
+
+            # Set agent to 'working' on the task.
+            ws_worker.send_json(
+                {
+                    "type": "agent-state",
+                    "workerId": worker_id,
+                    "agentId": agent_id,
+                    "taskId": task_id,
+                    "state": "working",
+                }
+            )
+            ws_obs.receive_json()  # drain agent-state broadcast
+
+            # Worker finishes and sends task-complete (agent-state:idle not sent yet).
+            ws_worker.send_json(
+                {
+                    "type": "task-complete",
+                    "workerId": worker_id,
+                    "agentId": agent_id,
+                    "taskId": task_id,
+                    "branch": "feature/fix",
+                    "description": "done",
+                    "prUrl": "",
+                    "sessionId": "",
+                    "lastText": "",
+                }
+            )
+
+            # Handler emits agent-state:idle before the task-complete broadcast.
+            msg = ws_obs.receive_json()
+            assert msg["type"] == "agent-state"
+            assert msg["agentId"] == agent_id
+            assert msg["state"] == "idle"
+            assert msg["taskId"] is None
+
+            # DB is already committed at this point.
+            with _sync_session(db_url) as session:
+                row = session.execute(
+                    select(Agent.state, Agent.current_task_id).where(Agent.id == agent_id)
+                ).first()
+
+    assert row.state == "idle"
+    assert row.current_task_id is None
+
+
+def test_task_followup_done_clears_agent_state(client):
+    """task-followup-done with agentId must immediately clear the agent to idle.
+
+    Same race as task-complete: the foreman is triggered right after, so the
+    agent row must be idle before the get_task_status call can arrive. #446.
+    """
+    test_client, db_url = client
+    guild_id, worker_id, task_id, agent_id = "gas-7", "w-gas7", "t-gas7", "a-gas7"
+    _insert_guild_worker_task(db_url, guild_id=guild_id, worker_id=worker_id, task_id=task_id)
+
+    with test_client.websocket_connect(f"/ws/{guild_id}") as ws_worker:
+        _join_ws(ws_worker, agent_id, worker_id)
+        with test_client.websocket_connect(f"/ws/{guild_id}") as ws_obs:
+            _join_ws(ws_obs, "a-obs7")
+            ws_worker.receive_json()  # drain obs's agent-joined broadcast
+
+            # Set agent to 'working' on the task.
+            ws_worker.send_json(
+                {
+                    "type": "agent-state",
+                    "workerId": worker_id,
+                    "agentId": agent_id,
+                    "taskId": task_id,
+                    "state": "working",
+                }
+            )
+            ws_obs.receive_json()  # drain agent-state broadcast
+
+            # Worker finishes a follow-up iteration (agent-state:idle not sent yet).
+            ws_worker.send_json(
+                {
+                    "type": "task-followup-done",
+                    "workerId": worker_id,
+                    "agentId": agent_id,
+                    "taskId": task_id,
+                    "success": True,
+                    "stopReason": "end_turn",
+                    "branch": "feature/fix",
+                    "sessionId": "",
+                    "prUrl": "",
+                }
+            )
+
+            # Handler emits agent-state:idle before the task-followup-done broadcast.
+            msg = ws_obs.receive_json()
+            assert msg["type"] == "agent-state"
+            assert msg["agentId"] == agent_id
+            assert msg["state"] == "idle"
+            assert msg["taskId"] is None
+
+            # DB is already committed at this point.
+            with _sync_session(db_url) as session:
+                row = session.execute(
+                    select(Agent.state, Agent.current_task_id).where(Agent.id == agent_id)
+                ).first()
+
+    assert row.state == "idle"
+    assert row.current_task_id is None
+
+
 def test_agent_idle_releases_task_lock(client):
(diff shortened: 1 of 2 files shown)


backend/ws_handlers.py
```diff
@@ -701,8 +725,28 @@ async def handle_task_followup_done(ctx: WSContext, data: dict) -> None:
                 .where(Task.id == task_id, Task.state.in_(_TERMINAL_STATES))
```
jmelloy, line 735: The `if agent_id_msg:` UPDATE block is placed inside `if task_id:` in `handle_task_complete` but in `handle_task_followup_done` the agent update comes *after* the `if task_id:` block closes (only the PR-update logic is inside it). This means the agent UPDATE and `LockService.release` + `commit` only happen when `task_id` is truthy anyway, but the structural inconsistency is confusing. Consider placing the agent update inside the `if task_id:` block for clarity and to make the transactional grouping explicit.

backend/ws_handlers.py
```diff
@@ -648,8 +649,30 @@ async def handle_task_complete(ctx: WSContext, data: dict) -> None:
             .where(Task.id == task_id, Task.state == "working")
             .values(state="awaiting-review")
         )
+        # Clear the agent that ran this task so get_task_status doesn't return
+        # agent_state:"working" after the task has parked in awaiting-review.
+        # The worker will also send agent-state:idle shortly after, but the
+        # foreman may call get_task_status before that message arrives.
+        if agent_id_msg:
```
jmelloy, line 656: The Agent UPDATE uses `Agent.guild_id == ctx.guild_pk` as a safety scope, but `agentId` is taken directly from the untrusted client message (`data.get('agentId', '')`). A malicious worker could supply a different agent's ID that happens to belong to the same guild and force-idle it. Consider verifying that `agent_id_msg` matches the agent actually associated with `worker_id_msg` (e.g. via a JOIN or prior lookup) before issuing the UPDATE.

backend/tests/test_agent_state_ws_persistence.py
```diff
@@ -327,6 +327,127 @@ def test_guild_get_returns_current_task_id(client):
...
+                    "workerId": worker_id,
+                    "agentId": agent_id,
+                    "taskId": task_id,
+                    "branch": "feature/fix",
+                    "description": "done",
+                    "prUrl": "",
+                    "sessionId": "",
+                    "lastText": "",
+                }
+            )
```
jmelloy, line 372: The test asserts that the very first message received by `ws_obs` after `task-complete` is the synthesized `agent-state: idle`. This tightly couples the test to broadcast ordering, which is an internal implementation detail. If the broadcast order changes (e.g. the original `task-complete` is emitted first), the test will fail even though the DB state is still correct. Consider making the assertion order-independent by draining a few messages and asserting one of them is the idle state.

backend/ws_handlers.py
```diff
@@ -648,8 +649,30 @@ async def handle_task_complete(ctx: WSContext, data: dict) -> None:
...
+                    "agentId": agent_id_msg,
+                    "workerId": worker_id_msg,
+                    "state": "idle",
+                    "taskId": None,
+                    "activity": None,
+                },
+            )
     await broadcast(ctx.guild_id, data, exclude=ctx.websocket)
     if task_id:
         spawn(
```
jmelloy, line 688: If `broadcast` raises an exception after the DB commit, the agent is idle in the DB but the synthesized `agent-state: idle` websocket message is never delivered. This is pre-existing infrastructure risk, but since this PR adds a critical ordering guarantee (frontend must see idle before foreman acts), consider logging a warning or swallowing the exception gracefully so the commit is not silently lost.

jmelloy (commented):
- The fix correctly addresses the race condition by clearing agent state within the same DB transaction that parks the task, ensuring `get_task_status` never sees a stale `working` state on an `awaiting-review` task.
- The synthesized `agent-state: idle` broadcast is emitted after commit but before the original message broadcast, which is the right ordering — however this means observers may receive two `agent-state: idle` broadcasts (one synthesized here, one from the worker's own subsequent `agent-state` message), which should be harmless but is worth noting.
- The agent-clearing UPDATE in `handle_task_followup_done` is placed outside the `if task_id:` block's transaction scope — if `task_id` is falsy the agent update and broadcast are still skipped (due to the `if agent_id_msg:` guard), but the structural asymmetry with `handle_task_complete` is a readability/maintenance concern.
- Tests are well-structured and cover both handlers, but they assert message ordering (agent-state idle arrives *before* task-complete broadcast) which is an implementation detail that could make tests brittle if broadcast ordering changes.
- No security concerns identified; the `guild_pk` filter on the Agent UPDATE correctly scopes the update to the right guild.

backend/ws_handlers.py
```diff
@@ -648,8 +649,40 @@ async def handle_task_complete(ctx: WSContext, data: dict) -> None:
...
             .values(state="awaiting-review")
         )
+        # Clear the agent that ran this task so get_task_status doesn't return
+        # agent_state:"working" after the task has parked in awaiting-review.
+        # The worker will also send agent-state:idle shortly after, but the
+        # foreman may call get_task_status before that message arrives.
+        if agent_id_msg:
+            await ctx.db.execute(
+                update(Agent)
+                .where(
```
jmelloy, line 659: The `Agent.worker_id == worker_id_msg` guard is correct for security, but if `worker_id_msg` is an empty string (e.g. a misconfigured client omits `workerId`), this WHERE clause silently matches nothing and the agent is never cleared. Consider adding an explicit guard: `if agent_id_msg and worker_id_msg:` to ensure both fields are present before attempting the update.

backend/ws_handlers.py
```diff
@@ -648,8 +649,40 @@ async def handle_task_complete(ctx: WSContext, data: dict) -> None:
...
+                    Agent.id == agent_id_msg,
+                    Agent.guild_id == ctx.guild_pk,
+                    Agent.worker_id == worker_id_msg,
+                )
```
jmelloy, line 663: The agent-clearing UPDATE in `handle_task_complete` runs before `LockService.release`, but it is inside the `if task_id:` block only implicitly — actually it is outside that block. If `task_id` is falsy the earlier `update(Task)` is skipped but this agent update will still execute because the `if agent_id_msg:` check is independent. Verify this is intentional; if `task_id` is absent the task was never moved to `awaiting-review`, so clearing the agent prematurely would be incorrect.

backend/tests/test_agent_state_ws_persistence.py
```diff
@@ -327,6 +327,137 @@ def test_guild_get_returns_current_task_id(client):
...
+                    "type": "task-complete",
+                    "workerId": worker_id,
+                    "agentId": agent_id,
+                    "taskId": task_id,
+                    "branch": "feature/fix",
+                    "description": "done",
+                    "prUrl": "",
+                    "sessionId": "",
+                    "lastText": "",
+                }
```
jmelloy, line 371: The test hard-codes `received = [ws_obs.receive_json(), ws_obs.receive_json()]` and expects exactly 2 messages. If `handle_task_complete` or broadcast logic ever emits a third message (e.g. task-update broadcast), the test will deadlock waiting for a third receive or miss the idle message. Consider draining with a timeout loop or using `receive_json(timeout=1)` for robustness.

backend/ws_handlers.py
```diff
@@ -648,8 +649,40 @@ async def handle_task_complete(ctx: WSContext, data: dict) -> None:
...
+                .values(state="idle", current_task_id=None, activity=None)
+            )
         await LockService(ctx.db).release(f"task:{task_id}")
         await ctx.db.commit()
+        if agent_id_msg:
```
jmelloy, line 668: The broadcast exception is silently swallowed with only a warning log. Since the DB has already committed at this point, a broadcast failure means the frontend will show stale state until the worker's own `agent-state:idle` arrives. This is acceptable, but the comment should note that the worker's subsequent message is the fallback recovery path, so this degraded-mode behaviour is intentional.

jmelloy (commented):
- The fix correctly addresses the race condition by clearing agent state within the same DB transaction that parks the task, preventing `get_task_status` from seeing stale `working` state.
- The synthesized `agent-state:idle` broadcast is emitted after commit but before the original `data` broadcast and foreman spawn, which is the right ordering.
- A potential double-update issue exists: the worker will subsequently send its own `agent-state:idle` message, which will trigger `handle_agent_state` and update the DB again — this is harmless but redundant and could cause a no-op conflict if agent ownership has changed.
- The test coverage is good and correctly validates both DB state and broadcast order, though the tests assume exactly 2 messages will be broadcast after task-complete, which could be fragile if other broadcasts are added.
- There is meaningful code duplication between `handle_task_complete` and `handle_task_followup_done` for the agent-clearing and broadcast logic that should be extracted into a shared helper.

PR #454, opened 2026-05-26
Fix codex_runner readline buffer overflow (#453)

### LINKED ISSUE #453: Fix codex_runner.py asyncio readline buffer overflow (ValueError: Separator is not found)
## Problem

`codex_runner.py` crashes with:

```
ValueError: Separator is not found, and chunk exceed the limit
```

This happens in `run_codex_auto` at the `async for raw in proc.stdout` loop when Codex emits a line longer than asyncio's default 64 KB `StreamReader` limit.

Traceback:
```
File "codex_runner.py", line 106, in run_codex_auto
    async for raw in proc.stdout:
...
File "asyncio/streams.py", line 571, in readline
    raise ValueError(e.args[0])
ValueError: Separator is not found, and chunk exceed the limit
```

## Fix

When creating the subprocess, pass a custom `asyncio.StreamReader` with a raised `_limit` (e.g. 8 MB) via `limit=` on `asyncio.create_subprocess_exec`, **or** switch the read loop from `async for raw in proc.stdout` to `await proc.stdout.read(n)` chunks that don't rely on newline delimiters.

The simplest one-liner fix is to pass `limit=8*1024*1024` to `asyncio.create_subprocess_exec(...)`.

jmelloy:
## Summary

- Added `limit=8 * 1024 * 1024` to the `asyncio.create_subprocess_exec` call in `run_codex_auto` (`codex_runner.py`)
- This raises the `StreamReader` buffer from asyncio's default 64 KB to 8 MB, matching the limit already used in `claude_runner.py`
- Without this fix, any stdout line emitted by Codex longer than 64 KB causes `ValueError: Separator is not found, and chunk exceed the limit`

Closes #453

### DIFF
--- a/worker/pioneer_worker/codex_runner.py
+++ b/worker/pioneer_worker/codex_runner.py
@@ -90,6 +90,7 @@ async def run_codex_auto(
             stdout=asyncio.subprocess.PIPE,
             stderr=asyncio.subprocess.PIPE,
             env=env,
+            limit=8 * 1024 * 1024,  # raise readline limit to 8 MB
         )
         os.close(slave_fd)
         slave_fd = None


worker/pioneer_worker/codex_runner.py
```diff
@@ -90,6 +90,7 @@ async def run_codex_auto(
             stdout=asyncio.subprocess.PIPE,
             stderr=asyncio.subprocess.PIPE,
             env=env,
+            limit=8 * 1024 * 1024,  # raise readline limit to 8 MB
```
jmelloy, line 93: 8 MB is still a hard ceiling. If Codex output ever produces a single line larger than 8 MB the same `ValueError` will be raised again. Consider either handling `LimitOverrunError` explicitly (e.g., reading in chunks with `read(n)` as a fallback) or documenting why 8 MB is a safe upper bound for this workload.

jmelloy, line 93: It would be worth extracting this constant (and the identical one in `claude_runner.py`) into a shared config or module-level constant (e.g., `READLINE_LIMIT = 8 * 1024 * 1024`) so both runners stay in sync automatically and the value only needs to change in one place.

jmelloy (commented):
- The fix correctly aligns `codex_runner.py` with `claude_runner.py` by raising the `StreamReader` buffer limit from asyncio's default 64 KB to 8 MB, directly addressing the `ValueError` on long stdout lines.
- The change is minimal, targeted, and low-risk; it touches only the subprocess creation call without altering any surrounding logic.
- An 8 MB hard cap still exists: if Codex ever emits a single line exceeding 8 MB the same `ValueError` will recur, so the fix trades one threshold for a higher one rather than eliminating the class of problem.
- No tests are included or referenced; a regression test that pipes a >64 KB line through the runner would prevent this from silently regressing.
- The comment is clear and the magic number is expressed as a readable arithmetic expression, which is good practice.

PR #457, opened 2026-05-26
Remove auto-PR on task completion from worker

### LINKED ISSUE #455: Remove automatic PR creation from worker task completion flow
## Problem

The worker agent automatically opens a PR when a task completes, but this is incorrect for non-PR tasks (e.g. review tasks, ephemeral checks, status polls). This causes phantom PR noise and confusion.

## Requested Change

Remove (or make conditional) the code in the worker that automatically opens a pull request at the end of a task. A PR should only be opened when the task explicitly involves pushing code to a new branch that warrants review — not for every task completion.

jmelloy:
## Summary

- Added `branch_has_new_commits()` to `github_pr.py` — runs `git rev-list --count origin/main..HEAD` (falling back to `origin/master`) to check whether the branch actually has new commits relative to the base
- Gated `open_pr()` in `worker.py` on this check: a PR is only opened when `push_ok=True` **and** the branch has commits ahead of the base branch
- Added 3 tests for `branch_has_new_commits` covering: has commits, no commits, and fallback when base ref can't be resolved
- Updated `test_review_phase_guard.py` to patch `branch_has_new_commits` (consistent with existing push/find patches)

Review tasks, status polls, and other ephemeral tasks that don't commit code will no longer produce phantom PRs. Normal code tasks that commit and push are unaffected — the fallback to `True` also ensures PRs still open if git ref resolution fails.

Closes #455


worker/tests/test_open_pr.py
```diff
@@ -129,3 +129,44 @@ async def test_open_pr_non_422_http_error_propagates():
 
     assert exc_info.value.code == 500
     mock_find.assert_not_called()
+
+
```
jmelloy: `_fake_run_git_count` is an `async def`, so calling `await _fake_run_git_count(3)` inside `with patch(...)` is valid but confusing — you're awaiting the coroutine to get the inner `_impl` function. A plain (non-async) factory function returning `_impl` would be clearer and less surprising to future readers.

worker/pioneer_worker/github_pr.py
```diff
@@ -94,6 +94,30 @@ async def push_branch(
...
+    Checks against origin/main then origin/master. Falls back to True (safe
+    default — open the PR) if neither ref can be resolved or git fails.
+    """
+    try:
+        for base_ref in ("origin/main", "origin/master"):
+            rc, out, _ = await git_ops.run_git(
+                ["rev-list", "--count", f"{base_ref}..HEAD"],
+                cwd=worktree_path,
+            )
+            if rc == 0:
```
jmelloy: The emit call for "no new commits" fires before the `return count > 0` statement, meaning it always emits when count == 0 regardless of which base ref matched. This is correct for count==0, but note the message is emitted even when the function will return `False` — that's fine. However, if `count > 0`, no emit fires at all (no positive confirmation log), which may make debugging harder.

worker/pioneer_worker/github_pr.py
```diff
@@ -94,6 +94,30 @@ async def push_branch(
...
 
 
+async def branch_has_new_commits(*, worktree_path: str, emit: EmitFn) -> bool:
+    """Return True if the current branch has commits not present on the base branch.
+
+    Checks against origin/main then origin/master. Falls back to True (safe
+    default — open the PR) if neither ref can be resolved or git fails.
+    """
+    try:
+        for base_ref in ("origin/main", "origin/master"):
```
jmelloy: The `except Exception` silently swallows all errors including unexpected ones (e.g., `ValueError` from `int()` if git outputs non-numeric text). Consider catching more specific exceptions or at minimum logging the exception details with `logger.warning(..., exc_info=True)` to aid debugging.

jmelloy (commented):
- The core logic is sound: gating `open_pr()` on `branch_has_new_commits()` correctly prevents phantom PRs for tasks that don't commit code.
- A subtle bug exists in the loop: when `origin/main` resolves successfully with 0 commits, the function correctly returns `False`, but the emit message is logged *before* the return, yet if `origin/master` is the one that returns 0 the same path is taken — this is fine. However, the loop will never try `origin/master` if `origin/main` succeeds (returns on first success regardless of count), which is the intended behavior but worth noting.
- The fallback to `True` on any exception or unresolved ref is a safe default that preserves existing behavior for edge cases, though it means a misconfigured repo could still produce unwanted PRs.
- The `await _fake_run_git_count(count)` pattern in tests is unusual and unnecessary — `_fake_run_git_count` is already an `async def` returning a coroutine, so `await`-ing it returns the inner `_impl` function correctly, but this is confusing and fragile; a synchronous factory or `AsyncMock` would be clearer.
- Test coverage is adequate for the new function but there is no integration-level test verifying the full `worker.py` guard path (i.e., confirming `open_pr` is *not* called when `branch_has_new_commits` returns `False`).

worker/pioneer_worker/github_pr.py
```diff
@@ -94,6 +94,30 @@ async def push_branch(
...
+            rc, out, _ = await git_ops.run_git(
+                ["rev-list", "--count", f"{base_ref}..HEAD"],
+                cwd=worktree_path,
+            )
```
jmelloy: When `rc == 0` and `count == 0`, the emit fires correctly but then `return count > 0` returns `False` immediately without trying `origin/master`. This means if a repo only has `origin/master` (no `origin/main`), the function would only skip the PR if `origin/main` happens to resolve — otherwise it falls through to the `True` fallback. The intent seems to be: return on first successful ref resolution, which is correct, but a comment here would clarify that zero-commit detection only applies to whichever base ref resolves first.

worker/tests/test_open_pr.py
```diff
@@ -129,3 +129,44 @@ async def test_open_pr_non_422_http_error_propagates():
...
+async def _fake_run_git_count(count: int):
+    async def _impl(args, cwd=None):
+        if args[:2] == ["rev-list", "--count"]:
```
jmelloy: `_fake_run_git_count` is an async function that must be awaited (`await _fake_run_git_count(3)`) before being passed to `patch()`. This works, but it's an unusual pattern — a plain synchronous factory `def _fake_run_git_count(count)` returning the inner async function would be simpler and less surprising.

jmelloy (commented):
- The core logic is correct and well-motivated: gating PR creation on actual commits prevents phantom PRs from non-code tasks
- A subtle bug exists in the `branch_has_new_commits` loop: the emit message for zero commits is printed but then `return count > 0` exits on the *first* ref that resolves successfully, meaning `origin/master` is never tried if `origin/main` returns 0 commits — this is probably intentional but the early-exit on zero is not clearly documented
- The fallback-to-`True` on git failure is a safe default and is well-documented, avoiding regression for broken git environments
- The `_fake_run_git_count` helper is an async factory that must be awaited before use in `patch()`, which is a subtle but valid pattern; however it could be simplified to a plain sync factory returning an async function
- Test coverage is adequate (3 new tests + guard patch update), but no test covers the behavior when `origin/main` returns rc!=0 and `origin/master` returns commits, verifying the fallback iteration actually works

worker/tests/test_open_pr.py
```diff
@@ -129,3 +129,44 @@ async def test_open_pr_non_422_http_error_propagates():
...
+@pytest.mark.asyncio
+async def test_branch_has_new_commits_fallback_when_base_missing():
+    """If neither origin/main nor origin/master can be resolved, default to True."""
+    emit = AsyncMock()
+
+    async def _always_fail(args, cwd=None):
+        return (1, "", "unknown revision")
+
+    with patch("pioneer_worker.git_ops.run_git", _always_fail):
+        result = await github_pr.branch_has_new_commits(worktree_path="/tmp/wt", emit=emit)
```
jmelloy: There is no test for the partial-fallback case: `origin/main` returns `rc != 0` (unresolvable) and `origin/master` returns a non-zero count. This is the most important branch of the loop to validate, since it exercises the actual iteration rather than early-exit or full-failure paths.

worker/pioneer_worker/github_pr.py
```diff
@@ -94,6 +94,30 @@ async def push_branch(
...
+            if rc == 0:
+                count = int(out.strip() or "0")
+                if count == 0:
```
jmelloy: Once `origin/main` resolves successfully (rc == 0), the function always returns on this iteration — `origin/master` is never consulted. This means if the repo uses `origin/master` as its base but `origin/main` also happens to exist (e.g. a stale remote ref), the count will be measured against the wrong base. Consider only breaking out of the loop when a ref is confirmed to actually be the repo's default branch, or document clearly that `origin/main` always takes precedence when both refs exist.

worker/tests/test_open_pr.py
jmelloy: `_fake_run_git_count` is an `async def` that is `await`ed before being passed to `patch`, which means it's called once at test-setup time to produce the inner `_impl` coroutine — but `patch` receives the already-resolved coroutine object, not a callable. This pattern works by accident here because the inner `_impl` is a closure returned from the outer await, but it's fragile and misleading. Simplify to a plain `def` factory or define `_impl` directly at the test level.

worker/pioneer_worker/github_pr.py
jmelloy: The emit call noting "no new commits" is only triggered for `origin/main` (the first successful ref). If `origin/main` resolves but has 0 commits ahead, the message is emitted and `False` is returned correctly. But if `origin/main` fails and `origin/master` resolves with 0 commits, the same message is never emitted — the silent `False` return could make debugging harder. Move the emit outside the condition or ensure it fires for any successful base ref.

jmelloy (commented):
- The core logic in `branch_has_new_commits` has a subtle bug: the emit message for "no new commits" is only printed when `origin/main` resolves with count 0, but the function still returns `False` — if `origin/main` resolves with count > 0 it correctly returns `True`, but it never falls through to check `origin/master` at all once `origin/main` succeeds, which is actually correct behavior. However, the emit message is printed and then the function returns on the same `origin/main` iteration regardless of count, so `origin/master` is never checked even when `origin/main` resolves successfully with 0 commits — this is intentional but worth clarifying.
- The `_fake_run_git_count` helper is an async function that returns a coroutine which is then `await`ed before being passed to `patch`, which is unnecessary complexity and could confuse readers; a regular function or direct `AsyncMock` would be cleaner.
- The fallback to `True` on git failure is a safe default for the stated goal (don't block legitimate PRs), and the two-ref fallback loop (`origin/main` → `origin/master`) is reasonable for supporting both repo conventions.
- The `test_review_phase_guard.py` patch correctly uses `return_value=False` rather than a coroutine, which matches how `AsyncMock` or plain mock would resolve for an `await`ed call — this is correct since `patch` on an async function auto-wraps it.
- Overall the change is small, well-targeted, and the test coverage addresses the main paths; the primary concern is a subtle correctness issue in the loop logic around when `origin/master` is ever actually tried.

worker/pioneer_worker/github_pr.py
```diff
@@ -94,6 +94,42 @@ async def push_branch(
...
+            if rc != 0:
+                continue  # ref absent, try next candidate
+            resolved_any = True
+            count = int(out.strip() or "0")
+            if count > 0:
+                await emit(
+                    f"[worker] Branch has {count} new commit(s) relative to {base_ref} — opening PR"
+                )
+                return True
+        if resolved_any:
```
jmelloy: The `logger.warning` + `return True` at the bottom is only reached when `resolved_any is False` after the loop — it is NOT inside the `except` block. This is correct, but the visual proximity to the `except` block makes it easy to mistake it for dead code. Consider adding an explicit `# no refs resolved` comment or restructuring slightly (e.g., `return True  # no refs resolved — safe default`) to make the control flow obvious.

worker/pioneer_worker/github_pr.py
```diff
@@ -94,6 +94,42 @@ async def push_branch(
...
+    """Return True if the current branch has commits not present on the base branch.
+
+    Tries origin/main first, then origin/master. Skips refs that fail to resolve.
+    Returns True immediately when a resolved ref shows new commits. After exhausting
+    all candidates, returns False if at least one resolved (no new commits found), or
+    True (safe default — open the PR) if no refs resolved at all.
+    """
+    try:
+        resolved_any = False
+        for base_ref in ("origin/main", "origin/master"):
```
jmelloy: When `origin/main` successfully resolves to 0 commits, `resolved_any` is set to `True` and the loop continues to check `origin/master`. If `origin/master` then reports commits, the function returns `True`. This is arguably incorrect: the branch is truly 0 commits ahead of `origin/main` (the authoritative base), so even if it has commits ahead of `origin/master` (diverged history), a PR against main would be empty. Consider breaking after the first successfully-resolved ref rather than treating all refs as independent candidates.

worker/tests/test_open_pr.py
```diff
@@ -129,3 +129,85 @@ async def test_open_pr_non_422_http_error_propagates():
...
+@pytest.mark.asyncio
+async def test_branch_has_new_commits_fallback_to_master():
+    """When origin/main is absent, fall back to origin/master with commits."""
+    emit = AsyncMock()
+
+    async def _main_fails_master_has_commits(args, cwd=None):
+        if args[:2] == ["rev-list", "--count"]:
+            base_ref = args[2].split("..")[0]
+            if base_ref == "origin/main":
+                return (1, "", "unknown revision")
```
jmelloy: The `test_branch_has_new_commits_main_zero_master_has_commits` test asserts `result is True` because `origin/master` has 3 commits, but as noted above this scenario — where `origin/main` resolves cleanly to 0 and `origin/master` has diverged commits — will produce a PR with no net changes against main. This test encodes the current (arguably buggy) behavior; if the logic is intentional, add a comment explaining why both refs are checked independently.

jmelloy (commented):
- The core logic in `branch_has_new_commits` has a subtle bug: when `origin/main` resolves to 0 commits, the loop continues to check `origin/master`, but if `origin/master` also resolves to 0 commits the function correctly returns `False`. However, the PR description claims the fallback ensures PRs still open if git ref resolution fails, which is correct via the `resolved_any` guard.
- The `except Exception` block returns `True` before the `logger.warning` / final `return True` at the bottom is reachable from the non-exception path — the unreachable final `return True` is actually the fallback for `not resolved_any`, but it sits outside the `try` block in an awkward position that can confuse readers into thinking it's dead code.
- The new guard in `worker.py` introduces an extra async git subprocess call on the hot path for every task completion that previously only called `push_branch`. For tasks that never push, `push_ok` will be `False` and the short-circuit `and` will skip the call — this is fine.
- Test coverage is solid: five scenarios cover commits present, no commits, both refs missing, main-missing/master-has, and main-zero/master-has. The `test_review_phase_guard.py` update correctly patches the new function to return `False`, preventing phantom PRs in that test.
- The fallback-to-`True` on git error is a reasonable safe default (prefer opening an unnecessary PR over silently dropping a legitimate one), but this design decision is not documented in the function signature or a comment near the call site in `worker.py`.

jmelloy (commented):
## Code Review: Remove auto-PR on task completion from worker

### Overview

This PR solves a real annoyance: review tasks and other ephemeral work that don't produce commits were opening phantom GitHub PRs. The fix adds a lightweight `branch_has_new_commits()` guard that runs `git rev-list --count` before `open_pr()` is called. Clean, targeted change.

---

### What Works Well

- **Safe default**: all failure paths (unresolvable refs, git errors, `int()` parse failure) fall back to `True`, so the guard never silently swallows a legitimate PR. Exactly right for a system where a missed PR is worse than a phantom one.
- **Correct placement**: the guard is on the `elif push_ok` branch, so the fast-path for pre-existing PRs (`existing_pr`) is untouched.
- **Test coverage**: 5 tests (PR description says 3, but 5 were added) cover the main paths — commits present, no commits, all refs absent, main-absent/master-has-commits, and the tricky main-zero/master-has-commits case.
- **`int(out.strip() or "0")`**: handles the empty-string edge case where git returns `rc=0` but empty stdout.

---

### Issues / Suggestions

**1. Minor: `int()` can still raise on unexpected git output**

`int(out.strip() or "0")` only guards empty string. If `git rev-list --count` returns `rc=0` but writes non-numeric text, `int()` raises `ValueError`. This is caught by the outer `except Exception`, so it's safe, but the behavior is a silent fallback to `True` rather than a deliberate parse error. Consider:

```python
try:
    count = int(out.strip() or "0")
except ValueError:
    logger.warning("branch_has_new_commits: unexpected rev-list output %r", out)
    return True
```

**2. Minor: `test_review_phase_guard.py` patch mocks async function correctly (Python 3.8+)**

```python
patch("pioneer_worker.worker.github_pr.branch_has_new_commits", return_value=False)
```

This works because Python 3.8+ auto-promotes to `AsyncMock` when patching a coroutine. Worth a brief comment so reviewers don't question it, but functionally fine.

**3. Suggestion: add a test for the PR-open path in `test_review_phase_guard.py`**

The existing test patches `branch_has_new_commits` to `False`, so the PR-open code path inside that test suite is never exercised. An additional test with `return_value=True` would verify the end-to-end guard integration. Not blocking, but increases confidence.

**4. Nit: docstring says "Skips refs that fail to resolve" but the behavior on ref failure is `continue`, not quite "skip"**

The phrasing is fine, just slightly loose. "Skips" implies deliberate intent, which is accurate, so this is purely optional.

**5. Edge case: repos without `origin` remote**

If the worktree has no `origin` configured, both `git rev-list` calls return `rc != 0`, `resolved_any` stays `False`, and the function returns `True` (opens PR). This is the right call, and it matches the documented fallback, but it isn't tested. Low priority.

---

### Summary

Solid, focused fix. The fallback-to-`True` contract is the key design decision and it's the right one. The 5 tests are well-structured. Items 1 and 3 above are the only ones worth acting on before merge; the rest are nits.

**Verdict: Approve with minor suggestions — ready to merge after addressing #1 (optional but clean) and #3 (low-lift confidence boost).**

jmelloy (commented):
## Code Review: Remove auto-PR on task completion

**Overall**: Solid, well-scoped fix. The safe-default-to-`True` design (open a PR when refs can't be resolved) is the right call — it preserves existing behavior for normal code tasks and only suppresses PRs when there's clear evidence of no new commits. The integration change in `worker.py` is minimal and correct.

---

### Issues

**1. Multi-line docstring violates project conventions** (`github_pr.py`)

The project's `CLAUDE.md` specifies: *"Never write multi-paragraph docstrings or multi-line comment blocks — one short line max."* The 7-line docstring on `branch_has_new_commits` should be trimmed to a single line.

**2. Confusing control-flow after the `try` block** (`github_pr.py`)

The "no refs resolved" case exits the `try` block by falling through (not via `return` or `raise`), so the `logger.warning` + `return True` after the `except` clause silently double as that handler. This reads like dead code at a glance. Cleaner to keep it inside the `try`:

```python
    if resolved_any:
        await emit("[worker] Branch has no new commits relative to base — skipping PR")
        return False
    # no refs resolved
    logger.warning("branch_has_new_commits: could not resolve base ref, assuming new commits exist")
    return True
except Exception:
    logger.warning("branch_has_new_commits: git error, assuming new commits exist", exc_info=True)
    return True
```

**3. PR summary says "3 tests" but diff adds 5** — minor doc nit, but `test_branch_has_new_commits_fallback_to_master` and `test_branch_has_new_commits_main_zero_master_has_commits` are also new tests.

---

### Minor notes

- **Missing test for the `except Exception` path**: None of the new tests exercise the branch where `run_git` raises an exception (vs returning a non-zero exit code). Low priority given the safe-default behaviour, but a quick `side_effect=RuntimeError` test would close the gap.

- **Both `origin/main` (0 commits) and `origin/master` (N commits) coexist**: The function will return `True` based on `master`, even if `main` is the canonical base. This is intentional per the safe-default philosophy but could open a phantom PR for a repo where `master` is a stale legacy branch. Worth noting in a comment if this is a known trade-off.

- **`test_review_phase_guard.py` patch** uses `return_value=False` on an async function — Python 3.8+ `patch` auto-detects coroutines and creates an `AsyncMock`, so this works correctly. No change needed.

---

### Summary

The logic is correct and the test coverage handles the key scenarios. The two actionable items are trimming the docstring (project convention) and restructuring the post-`try` control flow to make the "no refs resolved" case explicit inside the `try` block. Everything else is minor.

PR #458, opened 2026-05-26
Update tab title to PS / Guild Name - Guild ID

### LINKED ISSUE #456: Update browser tab title to show guild context when inside a guild
## Feature Request

When viewing the top-level app the tab title should remain `Pioneer Square`, but when a user navigates into a specific guild the title should update to:

```
PS / {Guild Name} - {Guild ID}
```

For example: `PS / My Guild - g-abc123`

## Acceptance Criteria
- Top-level / home page: tab title = `Pioneer Square`
- Inside a guild view: tab title = `PS / {guild name} - {guild id}`
- Title should update reactively as the user navigates between guilds

jmelloy:
## Summary

- Updated the reactive title watcher in `AppView.vue` to use the new format `PS / {Guild Name} - {Guild ID}` (e.g. `PS / My Guild - g-abc123`) when inside a guild
- Falls back to `Pioneer Square` on home / when no guild is active
- The watcher now observes the full `currentGuild` object (instead of just `name`) so both name and ID are reactive

Closes #456

### DIFF
--- a/frontend/src/views/AppView.vue
+++ b/frontend/src/views/AppView.vue
@@ -165,7 +165,7 @@ watch(
       console.warn('Guild is missing name or id', { name, id })
     }
     const label = name && id ? `${name} / ${id}` : name || id
-    document.title = `PS / ${label}`
+    document.title = `Pioneer Square / ${label}`
   },
   { immediate: true },
 )
--- a/frontend/src/views/AppView.vue
+++ b/frontend/src/views/AppView.vue
@@ -154,12 +154,18 @@ watch(
 )
 
 watch(
-  () => guildStore.currentGuild,
-  (guild) => {
-    document.title =
-      guild?.name && guild?.id
-        ? `PS / ${guild.name} - ${guild.id}`
-        : 'Pioneer Square'
+  () => [guildStore.currentGuild?.name, guildStore.currentGuild?.id],
+  ([name, id]) => {
+    const guild = guildStore.currentGuild
+    if (!guild) {
+      document.title = 'Pioneer Square'
+      return
+    }
+    if (!name || !id) {
+      console.warn('Guild is missing name or id', { name, id })
+    }
+    const label = name && id ? `${name} / ${id}` : name || id
+    document.title = `PS / ${label}`
   },
   { immediate: true },
 )
--- a/frontend/src/views/AppView.vue
+++ b/frontend/src/views/AppView.vue
@@ -154,9 +154,12 @@ watch(
 )
 
 watch(
-  () => guildStore.currentGuild?.name,
-  (name) => {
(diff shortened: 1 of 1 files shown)


frontend/src/views/AppView.vue
```diff
@@ -154,9 +154,12 @@ watch(
 )
 
 watch(
-  () => guildStore.currentGuild?.name,
-  (name) => {
-    document.title = name ? `${name} — Pioneer Square` : 'Pioneer Square'
+  () => guildStore.currentGuild,
+  (guild) => {
+    document.title =
+      guild?.name && guild?.id
```
jmelloy: Watching the entire `currentGuild` object will fire the watcher on any property change, not just `name` or `id`. Consider using a computed getter `() => [guildStore.currentGuild?.name, guildStore.currentGuild?.id]` to limit reactivity to only the fields that affect the title.

frontend/src/views/AppView.vue
```diff
@@ -154,9 +154,12 @@ watch(
...
+        ? `PS / ${guild.name} - ${guild.id}`
```
jmelloy: If a guild exists but is missing either `name` or `id` (e.g., partially loaded state), the title silently falls back to `'Pioneer Square'` with no indication of the problem. Consider logging a warning or using whichever fields are available to avoid a misleading fallback.

frontend/src/views/AppView.vue
```diff
@@ -154,9 +154,12 @@ watch(
...
+        : 'Pioneer Square'
```
jmelloy: The separator style is inconsistent: a forward-slash (`/`) is used between `PS` and the guild name, but a hyphen (`-`) is used between the guild name and ID. Standardise on one separator style (e.g., all ` — ` em-dashes, or all ` / `) for a cleaner appearance.

jmelloy (commented):
- The tab title format changes from `{Guild Name} — Pioneer Square` to `PS / {Guild Name} - {Guild ID}`, which is a meaningful UX change but the separator inconsistency (` / ` vs ` - `) may look odd and the `PS` abbreviation may be unclear to users unfamiliar with the product.
- Watching the full `currentGuild` object instead of just `name` is correct for reactivity, but it may trigger unnecessary re-renders if the guild object reference changes frequently (e.g., on every store update) even when name and ID haven't changed.
- The fallback title `'Pioneer Square'` is inconsistent with the new abbreviated format; if the intent is brevity, the fallback could also use `'PS'`, or vice versa the full name could be used in both cases.
- The condition `guild?.name && guild?.id` silently falls back to `'Pioneer Square'` if only one field is missing (e.g., a guild with a name but no ID), which may hide data integrity issues rather than surface them.
- No tests or snapshots appear to have been updated alongside this change; given that the title format is user-visible and potentially tested, this should be verified.

PR #461, opened 2026-05-26
Fix: agent stays active on frontend after task completes

### LINKED ISSUE #459: Agent shown as active on frontend after review/small tasks complete
## Bug

When a review task or small/ephemeral task finishes, the worker agent is still displayed as **active** in the frontend. It should transition back to idle/available once the task is done.

## Steps to Reproduce
1. Assign a review task or short-lived task to a worker
2. Task completes
3. Frontend still shows the agent as active rather than idle

## Expected Behaviour
Agent status updates to idle/available as soon as the task is marked complete.

## Notes
This likely relates to the frontend not receiving or processing the task-complete event properly for short-lived tasks, or the agent state not being cleared when the task finishes without a PR/branch to push.

jmelloy:
## Summary

- The `agents` store in the frontend previously relied entirely on the backend's `agent-state: idle` WebSocket frame to clear an agent's active indicator after a task finished
- For short-lived or review-phase tasks, the foreman can finalize the task so quickly that the gap between `task-complete` and the worker's `agent-state: idle` broadcast left the sidebar showing the agent as working indefinitely
- Added handlers for `task-complete` and `task-followup-done` in `agents.ts` that immediately set the agent to idle — the worker is guaranteed to return to its idle pool after sending these messages, so the frontend can safely reflect that without waiting for the follow-up `agent-state` frame
- Falls back to matching by `taskId` when `agentId` is absent from the message

## Test plan

- [x] 5 new unit tests in `agents.spec.ts` covering `task-complete`/`task-followup-done` with `agentId`, with `taskId` fallback, and the no-op case
- [x] All 84 frontend tests pass
- [x] TypeScript type check passes

Closes #459

### DIFF
--- a/frontend/src/stores/__tests__/agents.spec.ts
+++ b/frontend/src/stores/__tests__/agents.spec.ts
@@ -215,6 +215,41 @@ describe('useAgentsStore', () => {
 
       expect(store.agents[0].state).toBe('working')
     })
+
+    it('handles a subsequent agent-state:idle gracefully after task-complete already cleared the agent (double-update)', () => {
+      const store = useAgentsStore()
+      store.registerAgent({ agentId: 'a-1', workerId: 'w-x', state: 'working', taskId: 't-5' })
+
+      // First frame: task-complete sets agent to idle and clears taskId
+      store.handleWebSocketMessage({ type: 'task-complete', taskId: 't-5', agentId: 'a-1' })
+      expect(store.agents[0].state).toBe('idle')
+      expect(store.agents[0].taskId).toBeNull()
+
+      // Second frame: trailing agent-state:idle arrives; taskId is already null
+      // so the taskId-based lookup would be a no-op — but the agentId lookup
+      // must still find the agent and leave it idle without throwing.
+      store.handleWebSocketMessage({ type: 'agent-state', agentId: 'a-1', state: 'idle' })
+      expect(store.agents[0].state).toBe('idle')
+      expect(store.agents[0].taskId).toBeNull()
+    })
+
+    it('falls through to taskId lookup when agentId is present but not registered (deregistration race)', () => {
+      const store = useAgentsStore()
+      // Agent is registered with a taskId but its agentId will not match the
+      // one in the task-complete message (simulates a race where the ws frame
+      // carries a stale agentId while the slot reconnected under a new id).
+      store.registerAgent({ agentId: 'a-new', workerId: 'w-x', state: 'working', taskId: 't-6' })
+
+      store.handleWebSocketMessage({
+        type: 'task-complete',
+        taskId: 't-6',
+        agentId: 'a-old', // not registered
+      })
+
+      // agentId lookup misses; taskId fallback must find a-new and reset it
+      expect(store.agents[0].state).toBe('idle')
+      expect(store.agents[0].taskId).toBeNull()
+    })
   })
 
   describe('concurrent slots on the same worker', () => {
--- a/frontend/src/stores/__tests__/agents.spec.ts
+++ b/frontend/src/stores/__tests__/agents.spec.ts
@@ -149,6 +149,74 @@ describe('useAgentsStore', () => {
     })
   })
 
+  describe('task-complete and task-followup-done reset agent to idle', () => {
+    it('resets agent to idle on task-complete when agentId is present', () => {
+      const store = useAgentsStore()
+      store.registerAgent({ agentId: 'a-1', workerId: 'w-x', state: 'working', taskId: 't-1' })
+
+      store.handleWebSocketMessage({
+        type: 'task-complete',
+        taskId: 't-1',
+        agentId: 'a-1',
+      })
+
+      expect(store.agents[0].state).toBe('idle')
+      expect(store.agents[0].taskId).toBeNull()
+    })
+
+    it('resets agent to idle on task-complete by taskId fallback when agentId is absent', () => {
+      const store = useAgentsStore()
+      store.registerAgent({ agentId: 'a-1', workerId: 'w-x', state: 'working', taskId: 't-2' })
+
+      store.handleWebSocketMessage({
+        type: 'task-complete',
+        taskId: 't-2',
+      })
+
+      expect(store.agents[0].state).toBe('idle')
+      expect(store.agents[0].taskId).toBeNull()
+    })
+
+    it('resets agent to idle on task-followup-done when agentId is present', () => {
+      const store = useAgentsStore()
+      store.registerAgent({ agentId: 'a-1', workerId: 'w-x', state: 'working', taskId: 't-3' })
+
+      store.handleWebSocketMessage({
+        type: 'task-followup-done',
+        taskId: 't-3',
+        agentId: 'a-1',
+      })
+
+      expect(store.agents[0].state).toBe('idle')
+      expect(store.agents[0].taskId).toBeNull()
+    })
+
+    it('resets agent to idle on task-followup-done by taskId fallback when agentId is absent', () => {
+      const store = useAgentsStore()
+      store.registerAgent({ agentId: 'a-1', workerId: 'w-x', state: 'working', taskId: 't-4' })
+
+      store.handleWebSocketMessage({
+        type: 'task-followup-done',
+        taskId: 't-4',
+      })
+
+      expect(store.agents[0].state).toBe('idle')
+      expect(store.agents[0].taskId).toBeNull()
+    })
+
+    it('is a no-op for task-complete when neither agentId nor matching taskId exists', () => {
+      const store = useAgentsStore()
+      store.registerAgent({ agentId: 'a-1', workerId: 'w-x', state: 'working', taskId: 't-99' })
+
+      store.handleWebSocketMessage({
+        type: 'task-complete',
+        taskId: 't-unknown',
+      })
+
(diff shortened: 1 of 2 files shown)


frontend/src/stores/agents.ts
```diff
@@ -279,6 +279,24 @@ export const useAgentsStore = defineStore('agents', () => {
         data.activity ?? undefined,
         data.taskId ?? undefined,
       )
```
jmelloy, line 291: The two `else if` branches for `task-complete` and `task-followup-done` are identical. Extract into a small inline helper (e.g., `resetAgentForTask(data)`) to avoid duplication and make future changes easier to maintain consistently.

jmelloy (commented):
- The fix correctly addresses the race condition where `agent-state: idle` arrives too late for short-lived tasks, proactively resetting state on `task-complete` and `task-followup-done`.
- Code duplication exists between the two new `else if` branches — the identical lookup/update logic for `task-complete` and `task-followup-done` could be extracted into a shared helper.
- A potential double-update risk exists: when the backend does eventually send `agent-state: idle` after `task-complete`, the store will process it again; this is likely harmless but worth noting as an idempotency assumption.
- Test coverage is solid with 5 new cases covering agentId-present, taskId-fallback, and no-op scenarios for both message types.
- There is no test verifying that a subsequent `agent-state: idle` frame after `task-complete` doesn't corrupt state (e.g., if `taskId` was already cleared to null, the fallback lookup would silently no-op).

frontend/src/stores/agents.ts
```diff
@@ -279,6 +279,24 @@ export const useAgentsStore = defineStore('agents', () => {
...
+    } else if (data.type === 'task-complete') {
+      // The worker returns to its idle pool immediately after sending task-complete.
+      // Reset the agent proactively so the sidebar doesn't lag behind waiting for
+      // the separate agent-state:idle frame (which may be delayed for short tasks).
```
jmelloy, line 295: When `agentId` is present, the `taskId` fallback is skipped entirely. If the backend ever sends an `agentId` that doesn't match any registered agent (e.g., a race during deregistration), the reset is silently dropped with no fallback to `taskId`. Consider falling through to the `taskId` lookup when `updateAgentState` returns a falsy/not-found result.

frontend/src/stores/agents.ts
```diff
@@ -279,6 +279,24 @@ export const useAgentsStore = defineStore('agents', () => {
...
+      if (data.agentId) {
+        updateAgentState(data.agentId, 'idle', null, null)
+      } else {
```
jmelloy: Once `task-complete` fires and `taskId` is set to `null` on the agent, the subsequent `agent-state: idle` frame from the backend will attempt a `taskId`-based lookup that will silently no-op if it also lacks `agentId`. Ensure `updateAgentState` is robust to being called with a already-idle agent, and consider adding a test for this double-update scenario.

frontend/src/stores/agents.ts
```diff
@@ -100,15 +100,25 @@ export const useAgentsStore = defineStore('agents', () => {
...
-  ) {
+  ): boolean {
     const agent = agents.value.find((a) => a.id === agentId)
-    if (!agent) return
+    if (!agent) return false
     agent.state = state
     if (activity !== undefined) agent.activity = activity
     if (taskId !== undefined) agent.taskId = taskId
     // Idle/offline agents are by definition not working a task; clear the
     // link so a stale taskId from a prior run can't latch the agent to a row.
```
jmelloy, line 110: If `data.agentId` is present but unregistered, the code falls through to the `taskId` lookup. However, if the `task-complete` message arrives *after* `agent-state: idle` already cleared `taskId` to `null`, the `taskId` lookup (`a.taskId === data.taskId`) will never match because `taskId` is already `null`. This means a genuine deregistration race where `agentId` is stale AND `taskId` was already cleared will silently no-op. Consider logging a warning in that case to aid debugging.

jmelloy (commented):
- The fix correctly addresses a real race condition where short-lived tasks complete before the `agent-state: idle` WebSocket frame arrives, causing the sidebar to show stale working state indefinitely.
- The fallback from `agentId` to `taskId` lookup is a reasonable resilience measure, but the 'deregistration race' test case (stale agentId in message) reveals an implicit assumption that may not always hold — if both `agentId` is stale and `taskId` has already been cleared, the message is silently dropped.
- Changing `updateAgentState` to return a boolean is a reasonable minimal change, but it slightly muddies the function's responsibility by making it double as a lookup predicate; a dedicated `findAgentById` helper would be cleaner.
- Test coverage is thorough: the double-update idempotency test and the deregistration-race test are particularly good regression guards.
- No security concerns; the change is purely frontend state management with no authentication or data-exposure implications.

frontend/src/stores/agents.ts
```diff
@@ -279,6 +289,14 @@ export const useAgentsStore = defineStore('agents', () => {
         data.activity ?? undefined,
         data.taskId ?? undefined,
       )
```
jmelloy, line 291: The two `else if` branches for `task-complete` and `task-followup-done` call identical logic. Consider extracting the shared behavior into a single branch: `else if (data.type === 'task-complete' || data.type === 'task-followup-done')` to reduce duplication and make it easier to add future completion-type events.

frontend/src/stores/__tests__/agents.spec.ts
```diff
@@ -149,6 +149,109 @@ describe('useAgentsStore', () => {
...
+        agentId: 'a-1',
+      })
+
+      expect(store.agents[0].state).toBe('idle')
+      expect(store.agents[0].taskId).toBeNull()
+    })
+
+    it('resets agent to idle on task-followup-done by taskId fallback when agentId is absent', () => {
+      const store = useAgentsStore()
+      store.registerAgent({ agentId: 'a-1', workerId: 'w-x', state: 'working', taskId: 't-4' })
```
jmelloy, line 196: The 'deregistration race' test comment says the frame carries a 'stale agentId while the slot reconnected under a new id', but the actual scenario being tested is just that `agentId` doesn't match any registered agent. These are meaningfully different scenarios — consider adding a separate test where `agentId` is stale *and* `taskId` is already null (i.e., `agent-state: idle` arrived first), to confirm the expected no-op behavior is intentional and documented.

jmelloy (commented):
## Code Review

### Overview

Fixes a real race condition: for short-lived or review-phase tasks, the foreman can finalize before the worker's `agent-state: idle` broadcast, leaving the sidebar permanently showing the agent as working. The fix proactively resets agent state in the frontend on `task-complete` and `task-followup-done`, with a fallback from `agentId` to `taskId` lookup for resilience.

---

### Correctness

**Approve with one concern.**

The core fix is sound. Workers are guaranteed to return to idle after sending either message, so the frontend can safely reflect that without waiting for the follow-up `agent-state` frame.

However, there is a subtle UI flicker case worth noting: when a task has **multiple followup rounds**, the sequence is:

```
task-followup-done  → agent set to idle (this PR)
task-followup       → worker sets agent back to working
task-followup-done  → agent set to idle again
```

The agent will briefly flash idle between rounds. For tasks that zip through multiple followups quickly, this will produce a working→idle→working flicker in the sidebar. It's a minor visual glitch but worth a comment since the PR description frames the fix as "guaranteed" to be correct — that guarantee holds for single-round tasks; multi-round tasks trade the "stuck working" bug for a transient flicker.

---

### Code Quality

- **`resetAgentForTask` helper is clean**: the deduplication between the two identical `else if` branches into a single helper is the right call. No redundancy.

- **`updateAgentState` now returns `boolean`**: minor API responsibility creep — the function now doubles as a lookup predicate. A dedicated `findAgentById(id)` would keep responsibilities cleaner. Not a blocker, but worth revisiting if `updateAgentState` gains more callers that use its return value for control flow.

- **Silent-drop edge case**: if `agentId` is stale (not registered) *and* `taskId` has already been cleared (e.g., a duplicate `task-complete` arrives after the first already reset the slot), `resetAgentForTask` silently no-ops. This is the correct behavior — the agent is already idle — but the comment in the code only mentions the deregistration-race scenario, not the duplicate-message scenario. A small comment addition would make this explicit.

---

### Tests

The test suite is excellent — 7 tests, not 5 as stated in the PR description. Both the double-update idempotency test and the deregistration-race test are strong regression guards that go beyond the minimum required. Specifically:

- `handles a subsequent agent-state:idle gracefully...` — good: confirms idempotency
- `falls through to taskId lookup when agentId is present but not registered` — good: covers the stale-agentId race

**One gap**: there is no test for the multi-round followup flicker scenario described above (agent goes `working → idle → working` when `task-followup` arrives after `task-followup-done`). Not blocking, but would prevent silent regressions if the handler logic changes.

---

### Summary

| Area | Assessment |
|------|-----------|
| Correctness | ✅ Fixes the reported bug; minor flicker risk on multi-round followups |
| Code quality | ✅ Clean helper extraction; `boolean` return on `updateAgentState` is slight overloading |
| Tests | ✅ Thorough; one missing case (multi-round followup flicker) |
| Security | ✅ No concerns — purely frontend state management |
| Breaking changes | ✅ None — no WS protocol changes, purely additive frontend logic |

**Verdict: Approve with minor suggestions.** The flicker on multi-round followups is worth a follow-up issue or a comment in the code, but it does not block merging — it is a visual degradation, not a regression from the current bug.

PR #467, opened 2026-05-26
Convert date fields in models.py from str to datetime

### LINKED ISSUE #466: Convert date fields in SQLModel models.py from str to proper date/datetime types
All date/datetime fields in the SQLModel `models.py` are currently typed as `str`. These should be converted to proper `datetime` / `date` types so that SQLModel/SQLAlchemy handles serialisation, validation, and DB storage correctly.

## What to do
1. Locate `models.py` (likely under `backend/` or `app/`).
2. Find every field typed as `str` that represents a date or datetime value (e.g. `created_at`, `finished_at`, `deleted_at`, `updated_at`, etc.).
3. Change the type annotations from `str` to `datetime` (or `date` where only a date is needed), using `Optional[datetime]` for nullable fields.
4. Add/update any `default_factory` or `default` values accordingly (e.g. `default_factory=datetime.utcnow` → `default_factory=lambda: datetime.now(timezone.utc)`).
5. Ensure any Alembic migration or schema is updated if needed.
6. Check that API serialisers/response models still work correctly after the type change (Pydantic will serialise `datetime` to ISO-8601 strings automatically).
7. Run existing tests to confirm nothing is broken.

jmelloy:
## Summary

- Replace all `str`-typed date/time columns in `backend/models.py` with `datetime` using `sa_column=Column(DateTime(timezone=True), ...)` — 15 models, 22 columns total
- Add Alembic migration `20260526_000000_convert_datetime_columns` that ALTERs all 22 columns from `Text` to `TIMESTAMPTZ` via `postgresql_using=column::timestamptz` (with full downgrade support)
- Update every caller across 15 application files: DB writes use `datetime.now(UTC)` directly; WebSocket broadcast payloads call `.isoformat()` explicitly (since `json.dumps` can't handle `datetime`); HTTP responses rely on FastAPI's `jsonable_encoder`
- Update 9 test files: helpers and assertions now use `datetime` objects instead of ISO strings for DB operations

## Key design choices

- **DB writes**: raw `datetime` — no `.isoformat()` needed, psycopg2 and asyncpg accept native Python types for `TIMESTAMPTZ` columns
- **WebSocket broadcasts**: explicit `.isoformat()` retained — `json.dumps` does not handle `datetime` objects
- **HTTP responses**: no change — FastAPI's `jsonable_encoder` converts `datetime` → ISO-8601 automatically
- **`live_tasks_filter()`**: parameter changed from `str | None` to `datetime | None`; comparison against the `Task.deleted_at` DateTime column works correctly via SQLAlchemy

## Test plan

- [ ] `docker compose up -d postgres-test && cd backend && python -m pytest` — all 119 tests should pass
- [ ] `ruff check . && ruff format .` — passes (verified locally, 0 issues)
- [ ] Alembic upgrade: `alembic upgrade head` against an existing DB applies migration cleanly
- [ ] Alembic downgrade: `alembic downgrade -1` restores Text columns

Closes #466

### DIFF
--- a/backend/alembic/versions/20260526_000000_convert_datetime_columns.py
+++ b/backend/alembic/versions/20260526_000000_convert_datetime_columns.py
@@ -46,6 +46,11 @@ _COLUMNS = [
 
 
 def upgrade() -> None:
+    # IMPORTANT: postgresql_using casts existing text values directly to
+    # timestamptz. This succeeds only when every stored value is a valid
+    # ISO-8601 timestamp string. If any row contains a non-parseable value
+    # the ALTER TABLE will fail mid-migration. Verify clean data beforehand:
+    #   SELECT table, column, value FROM <table> WHERE <column> !~ '^\d{4}-\d{2}-\d{2}'
     for table, column, nullable in _COLUMNS:
         op.alter_column(
             table,
@@ -53,7 +58,7 @@ def upgrade() -> None:
             existing_type=sa.Text(),
             type_=sa.DateTime(timezone=True),
             existing_nullable=nullable,
-            postgresql_using=f"{column}::timestamptz",
+            postgresql_using=f"{column}::timestamptz",  # safe for ISO-8601 text strings
         )
 
 
--- a/backend/alembic/versions/20260526_000000_convert_datetime_columns.py
+++ b/backend/alembic/versions/20260526_000000_convert_datetime_columns.py
@@ -1,0 +1,69 @@
+"""Convert all datetime columns from Text to DateTime(timezone=True).
+
+Revision ID: 20260526_000000_convert_datetime_columns
+Revises: 20260520_000007_merge_rename_with_prior_head
+Create Date: 2026-05-26
+
+"""
+
+from __future__ import annotations
+
+from collections.abc import Sequence
+
+import sqlalchemy as sa
+from alembic import op
+
+revision: str = "20260526_000000_convert_datetime_columns"
+down_revision: str | Sequence[str] | None = "20260520_000007_merge_rename_with_prior_head"
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+# (table, column, nullable)
+_COLUMNS = [
+    ("guilds", "created_at", False),
+    ("guilds", "deleted_at", True),
+    ("agents", "joined_at", False),
+    ("agents", "last_seen", True),
+    ("messages", "created_at", False),
+    ("workers", "created_at", False),
+    ("workers", "last_seen", True),
+    ("tasks", "created_at", False),
+    ("tasks", "finished_at", True),
+    ("tasks", "deleted_at", True),
+    ("github_tokens", "created_at", False),
+    ("github_tokens", "updated_at", False),
+    ("user_sessions", "created_at", False),
+    ("users", "created_at", False),
+    ("users", "updated_at", False),
+    ("guild_members", "created_at", False),
+    ("task_logs", "timestamp", False),
+    ("claude_credentials", "updated_at", False),
+    ("guild_keys", "created_at", False),
+    ("foreman_turns", "created_at", False),
+    ("github_events", "created_at", False),
+    ("task_events", "created_at", False),
+]
+
+
+def upgrade() -> None:
+    for table, column, nullable in _COLUMNS:
+        op.alter_column(
+            table,
+            column,
+            existing_type=sa.Text(),
+            type_=sa.DateTime(timezone=True),
+            existing_nullable=nullable,
+            postgresql_using=f"{column}::timestamptz",
+        )
+
+
+def downgrade() -> None:
+    for table, column, nullable in reversed(_COLUMNS):
+        op.alter_column(
+            table,
+            column,
+            existing_type=sa.DateTime(timezone=True),
+            type_=sa.Text(),
+            existing_nullable=nullable,
+            postgresql_using=f"{column}::text",
+        )
--- a/backend/foreman/tools.py
+++ b/backend/foreman/tools.py
@@ -45,10 +45,10 @@ FOREMAN_MODEL = os.environ.get("FOREMAN_MODEL", "claude-sonnet-4-6")
 DEFAULT_FINALIZE_TTL_SECONDS = 3 * 24 * 60 * 60  # 3 days
 
 
-def _resolve_finalize_deleted_at(inp: dict) -> tuple[str, str | None]:
+def _resolve_finalize_deleted_at(inp: dict) -> tuple[datetime | None, str | None]:
     """Compute the soft-delete instant for a finalize_task tool call.
 
-    Returns ``(deleted_at_iso, error)`` — error is non-None when the inputs
+    Returns ``(deleted_at, error)`` — error is non-None when the inputs
     were malformed. Honours an explicit ``deleted_at`` first, then
     ``expires_in_seconds``, otherwise falls back to the default 3-day window.
     """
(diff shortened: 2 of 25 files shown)


backend/foreman/tools.py
```diff
@@ -908,11 +907,13 @@ async def _exec_one_tool(guild_id: str, tu, user_id: str | None = None) -> dict:
                                 "type": "task-update",
                                 "taskId": task_id,
                                 "state": "done",
```
jmelloy, line 909: `deleted_at` comes from `_resolve_finalize_deleted_at` which returns `(None, error_str)` on failure. The code calls `deleted_at.isoformat()` here without checking if `deleted_at` is None — if the error-handling branch above failed to return early, this would raise `AttributeError`. Verify that all error paths from `_resolve_finalize_deleted_at` are guarded before these `.isoformat()` calls.

backend/models.py
```diff
@@ -29,97 +29,73 @@ class Guild(SQLModel, table=True):
```
jmelloy, line 29: Numerous explanatory comments documenting non-obvious business logic (soft-delete semantics for `deleted_at`, the purpose of `webhook_secret`, the `auth_token` legacy-worker note, the `TaskEvent` docstring rationale) have been silently removed. These should be preserved or moved to a separate doc, as they capture important design intent.

backend/alembic/versions/20260526_000000_convert_datetime_columns.py
```diff
@@ -0,0 +1,69 @@
...
+def upgrade() -> None:
+    for table, column, nullable in _COLUMNS:
+        op.alter_column(
+            table,
+            column,
+            existing_type=sa.Text(),
+            type_=sa.DateTime(timezone=True),
+            existing_nullable=nullable,
+            postgresql_using=f"{column}::timestamptz",
+        )
```
jmelloy, line 62: The `postgresql_using` cast `column::timestamptz` will fail at runtime if any existing Text value is not a valid ISO-8601/RFC-3339 timestamp. There is no pre-migration data validation step or advisory note. Consider adding a comment or a separate data-check step to surface bad rows before the ALTER fails mid-migration on production data.

jmelloy (commented):
- The migration correctly converts all 22 columns from Text to TIMESTAMPTZ using `postgresql_using` casting, with full downgrade support — this is well-structured.
- The layered boundary pattern is correctly applied: raw `datetime` objects for DB writes, `.isoformat()` for WebSocket/JSON payloads, and reliance on FastAPI's `jsonable_encoder` for HTTP responses.
- One potential bug exists in `foreman/tools.py`: `deleted_at` from `_resolve_finalize_deleted_at` is used in a broadcast payload with `.isoformat()`, but if the error path returns `None`, calling `.isoformat()` on it would raise an `AttributeError` — the error check must be verified as exhaustive.
- Several inline comments documenting important business logic (webhook secret purpose, soft-delete semantics, bearer token auth requirements, etc.) were removed from `models.py` without replacement, reducing maintainability.
- The `_COLUMNS` list in the migration is not validated against the actual model definitions, creating a risk of drift if models change before the migration runs.

backend/routes/tasks.py
```diff
@@ -292,11 +288,11 @@ async def finalize_task_endpoint(
             "type": "task-update",
             "taskId": task_id,
             "state": "done",
-            "finishedAt": finished_at,
-            "deletedAt": deleted_at,
+            "finishedAt": finished_at.isoformat(),
```
jmelloy, line 291: The HTTP response `return {"status": "finalized", "taskId": task_id, "deletedAt": deleted_at.isoformat()}` manually calls `.isoformat()` while the PR description states HTTP responses rely on FastAPI's `jsonable_encoder`. This inconsistency is harmless but breaks the stated design convention — either always use `jsonable_encoder` or always use `.isoformat()` explicitly.

jmelloy (commented):
## Code Review: Convert date fields from str to datetime

### Overview

This PR correctly upgrades 22 datetime columns across 12 tables from `Text` storage with manual `.isoformat()` calls to proper `DateTime(timezone=True)` (TIMESTAMPTZ) columns. The approach is systematic, the migration is complete, and the layer separation is well-thought-out (raw `datetime` for DB ops, explicit `.isoformat()` for WebSocket payloads, relying on FastAPI's `jsonable_encoder` for HTTP responses).

---

### ✅ What works well

- **DRY migration**: The loop over `_COLUMNS` in the Alembic migration avoids 22 nearly-identical `alter_column` calls — clean pattern.
- **Correct `postgresql_using` cast**: `postgresql_using=f"{column}::timestamptz"` is the right approach for in-place type coercion; the `env.py` confirms this is a PostgreSQL-only project (PL/pgSQL in `_ENSURE_VERSION_TABLE_SQL`), so no SQLite concerns.
- **Return type fix for `_resolve_finalize_deleted_at`**: Changing the error-case return from `("", error)` to `(None, error)` is semantically cleaner; tests updated consistently.
- **`live_tasks_filter`**: Comparing a `datetime` column directly with a `datetime` object is correct SQLAlchemy usage and avoids the subtle string-sort vs. timestamp-sort bugs that the old approach risked.

---

### 🐛 Bug / Data-integrity concern

**Silent datetime parse failure in `ws_handlers.py:handle_task_update`** (lines 1372–1383 of the diff):

```python
try:
    parsed = datetime.fromisoformat(str(raw).replace("Z", "+00:00"))
    ...
    update_values["finished_at"] = parsed
except (ValueError, TypeError):
    pass  # ← silently drops the field update
```

If a worker sends a malformed `finishedAt`, the field is silently skipped — no log, no error, no signal to the caller. A task will appear to never finish from the DB's perspective. This is a behavior change from before (previously the raw string was stored as-is). At minimum this should log a warning:

```python
except (ValueError, TypeError) as exc:
    logger.warning("Ignoring invalid finishedAt %r: %s", raw, exc)
```

---

### ⚠️ Inconsistency: HTTP response serialization

Most routes now explicitly call `.isoformat()` before putting datetimes into response dicts (e.g., `routes/tasks.py:767`, `routes/foreman.py:668`). But `routes/foreman.py:671` does not:

```python
return {"task_id": task_id, "created_at": created_at}  # created_at is datetime
```

FastAPI's `jsonable_encoder` handles this transparently, so it won't break — but it's inconsistent with the rest of the PR's explicit pattern. Either call `.isoformat()` everywhere, or trust the encoder everywhere; mixing signals confuses future readers about which approach is intentional.

---

### 📝 Comment over-removal in `models.py`

The PR strips field-level comments that explained WHY (invariants, non-obvious design choices) — not just WHAT. A few worth restoring:

| Removed comment | Why it matters |
|---|---|
| `ForemanTurn.is_tool_response` — *"1 if this 'user' turn carries tool_results; 0 otherwise"* | The int-as-bool pattern is non-obvious; the comment clarified intent |
| `TaskEvent` docstring was shortened | Original explained the lock-serialization mechanic and the re-trigger flow on `task-followup-done`. The new one-liner loses that context entirely. |
| `Task.deleted_at` — *"NULL = live; once now() > deleted_at, list/get queries hide the row"* | Explains the soft-delete semantic; derivable from code but aids navigation |

It's fine to remove comments that just describe what the field name says; these went beyond that.

---

### 🔎 Minor notes

- **`test_finalize_endpoint_default_sets_three_day_window`**: The assertion at line 1218 round-trips through `isoformat()` then `fromisoformat()` to compare DB value to HTTP response. This is correct but could be simplified now that `_read_task()` returns a native `datetime`.
- **Migration nullable flag ordering**: The `_COLUMNS` list in the migration passes `nullable` as the third tuple element, but it maps to `existing_nullable` in `op.alter_column`. The naming is clear in context, but a named-tuple or dataclass would make this more self-documenting for a list this long.
- **Downgrade uses `reversed(_COLUMNS)`**: Correct for FK-ordered teardown; no issue, just worth noting it's intentional.

---

### Summary

The core change is sound and the approach is correct. Two actionable items before merging:

1. **Fix silent parse failure** in `ws_handlers.py:handle_task_update` — add a warning log.
2. **Consistency pass** on HTTP response serialization (either always use `.isoformat()` or always rely on FastAPI's encoder, don't mix).

The comment cleanup is a judgment call; the ones flagged above are worth restoring, but not blocking.

PR #476, opened 2026-05-27
User settings: view & clear Claude/Codex tokens (#475)

### LINKED ISSUE #475: User settings: view and clear Claude/Codex API tokens
## Summary

Users should be able to view (masked) and clear their stored Claude and Codex API tokens from the user settings UI.

### Backend
- Add API endpoints (or extend existing user settings endpoints) to:
  - `GET /api/users/me/tokens` — return which tokens are set (masked, e.g. `sk-ant-****1234`), not the raw values
  - `DELETE /api/users/me/tokens/{provider}` — clear a specific token (`claude` or `codex`)

### Frontend (User Settings page)
- In the user settings UI, add a "API Tokens" section showing:
  - Claude token: masked value (or "Not set") + a **Clear** button
  - Codex token: masked value (or "Not set") + a **Clear** button
- Clearing a token should call the DELETE endpoint and update the UI

### Security
- Tokens must never be returned in plaintext — always mask all but the last 4 characters
- Clearing should zero/remove the token from the database

## Acceptance Criteria
- [ ] User can see whether their Claude and Codex tokens are set (masked display)
- [ ] User can clear either token from the settings page
- [ ] No raw token values are ever returned by the API

jmelloy:
## Summary

- **DB migration** (`20260527_000000`): adds `claude_api_key` and `openai_api_key` nullable `TEXT` columns to the `users` table.
- **`GET /api/users/me/tokens`**: returns which tokens the caller has set, masked (`sk-ant-****1234` style) — raw values are never returned.
- **`DELETE /api/users/me/tokens/{provider}`**: clears (`NULL`s) the stored token for `claude` or `codex`; returns `400` for unknown providers.
- **`GitHubConfigModal.vue`**: renamed to "User Settings", adds an **API Tokens** section with a row per provider showing the masked value or "Not set", and a **Clear** button (disabled when not set or while clearing).

## Security

- `_mask_token()` helper in `auth.py` shows at most the first 7 chars + last 4; the raw key never leaves the server.
- Both new endpoints are gated behind `require_user` (login token required).

## Test plan

- [ ] Log in, open the user settings modal — API Tokens section loads showing "Not set" for both rows.
- [ ] Manually set a value in the DB (`UPDATE users SET claude_api_key = 'sk-ant-api03-test1234' WHERE id = ...`), reload modal — shows `sk-ant-****1234`.
- [ ] Click **Clear** — token row reverts to "Not set"; DB column is NULL.
- [ ] `GET /api/users/me/tokens` without auth returns 401.
- [ ] `DELETE /api/users/me/tokens/github` returns 400.

Closes #475

### DIFF
--- a/backend/models.py
+++ b/backend/models.py
@@ -192,6 +192,10 @@ class User(SQLModel, table=True):
     avatar_url: str | None = None
     created_at: datetime = Field(sa_column=Column(DateTime(timezone=True), nullable=False))
     updated_at: datetime = Field(sa_column=Column(DateTime(timezone=True), nullable=False))
+    # Per-user API keys set from the user-settings UI; stored in plaintext,
+    # never returned raw — always masked at the API layer.
+    claude_api_key: str | None = None
+    openai_api_key: str | None = None
 
 
 class GuildMember(SQLModel, table=True):
--- a/backend/routes/auth.py
+++ b/backend/routes/auth.py
@@ -39,6 +39,20 @@ from utils import generate_guild_id
 router = APIRouter()
 
 
+def _mask_token(token: str | None) -> str | None:
+    """Return a masked representation of *token*, or None when not set.
+
+    Shows the first 7 characters (the recognizable prefix) and the last 4,
+    with ``****`` in between — e.g. ``sk-ant-****1234``.
+    """
+    if not token:
+        return None
+    if len(token) <= 8:
+        return "****"
+    prefix_len = min(7, len(token) - 4)
+    return token[:prefix_len] + "****" + token[-4:]
+
+
 class CodeExchangeRequest(BaseModel):
     code: str
     state: str
@@ -333,6 +347,51 @@ async def guest_login():
         "gh_name": "Dev Guest",
         "gh_avatar": "",
     }
+
+
+@router.get("/api/users/me/tokens")
+async def get_my_tokens(github_user_id: str = Depends(require_user)):
+    """Return which API tokens the current user has configured, masked.
+
+    Never returns raw token values — each present token is shown as e.g.
+    ``sk-ant-****1234``.
+    """
+    db = await get_db()
+    try:
+        u_res = await db.execute(select(User).where(User.id == github_user_id))
+        user = u_res.scalar_one_or_none()
+        if not user:
+            raise HTTPException(status_code=404, detail="User not found")
+        return {
+            "claude": _mask_token(user.claude_api_key),
+            "codex": _mask_token(user.openai_api_key),
+        }
+    finally:
+        await db.close()
+
+
+@router.delete("/api/users/me/tokens/{provider}")
+async def clear_my_token(provider: str, github_user_id: str = Depends(require_user)):
+    """Clear the stored API token for *provider* (``claude`` or ``codex``).
+
+    Sets the column to NULL in the database; the raw value is never returned.
+    """
+    if provider not in ("claude", "codex"):
+        raise HTTPException(status_code=400, detail="provider must be 'claude' or 'codex'")
+    db = await get_db()
+    try:
+        u_res = await db.execute(select(User).where(User.id == github_user_id))
+        user = u_res.scalar_one_or_none()
+        if not user:
+            raise HTTPException(status_code=404, detail="User not found")
+        if provider == "claude":
+            user.claude_api_key = None
+        else:
+            user.openai_api_key = None
+        await db.commit()
+    finally:
+        await db.close()
+    return {"ok": True}
 
 
 @router.delete("/auth/logout")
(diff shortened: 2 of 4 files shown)


backend/routes/auth.py
```diff
@@ -39,6 +39,20 @@
 router = APIRouter()
 
 
+def _mask_token(token: str | None) -> str | None:
+    """Return a masked representation of *token*, or None when not set.
+
+    Shows the first 7 characters (the recognizable prefix) and the last 4,
+    with ``****`` in between — e.g. ``sk-ant-****1234``.
```
jmelloy, line 46: `prefix_len = min(7, len(token) - 4)`: for a 9-character token this gives `prefix_len = 5`, so the output is `token[:5] + '****' + token[-4:]` — which exposes all 9 characters. Consider returning `'****'` for any token shorter than, say, 12 characters to avoid leaking short keys entirely.

backend/routes/auth.py
```diff
@@ -335,6 +349,51 @@ async def guest_login():
     }
 
 
+@router.get("/api/users/me/tokens")
+async def get_my_tokens(github_user_id: str = Depends(require_user)):
+    """Return which API tokens the current user has configured, masked.
+
+    Never returns raw token values — each present token is shown as e.g.
+    ``sk-ant-****1234``.
+    """
```
jmelloy, line 358: Both endpoints call `await get_db()` directly and close the session in a `finally` block. If `get_db` is designed as a FastAPI dependency (generator), calling it this way bypasses the dependency lifecycle and may leak connections or skip rollback logic. Use `Depends(get_db)` as a parameter instead, consistent with the rest of the codebase.

backend/routes/auth.py
```diff
@@ -335,6 +349,51 @@ async def guest_login():
...
+
+@router.delete("/api/users/me/tokens/{provider}")
+async def clear_my_token(provider: str, github_user_id: str = Depends(require_user)):
+    """Clear the stored API token for *provider* (``claude`` or ``codex``).
+
+    Sets the column to NULL in the database; the raw value is never returned.
+    """
+    if provider not in ("claude", "codex"):
+        raise HTTPException(status_code=400, detail="provider must be 'claude' or 'codex'")
+    db = await get_db()
```
jmelloy, line 381: If the `db.commit()` raises (e.g. a transient DB error), the `finally` block closes the session without rolling back, potentially leaving the session in a dirty state. Add an explicit `await db.rollback()` on exception, or rely on the session context manager pattern.

frontend/src/components/GitHubConfigModal.vue
```diff
@@ -24,10 +59,45 @@
...
+const tokens = ref<{ claude: string | null; codex: string | null }>({
+  claude: null,
+  codex: null,
+})
+const clearing = ref<'claude' | 'codex' | null>(null)
+const tokenError = ref<string | null>(null)
+
+async function loadTokens() {
+  try {
+    const data = await api<{ claude: string | null; codex: string | null }>(
```
jmelloy, line 78: `loadTokens` silently swallows all errors with an empty `catch` block. An expired login token or 5xx error will leave the UI showing 'Not set' with no indication that the load actually failed. At minimum set `tokenError` so the user knows the state could not be fetched.

backend/models.py
```diff
@@ -192,6 +192,10 @@ class User(SQLModel, table=True):
     avatar_url: str | None = None
     created_at: datetime = Field(sa_column=Column(DateTime(timezone=True), nullable=False))
     updated_at: datetime = Field(sa_column=Column(DateTime(timezone=True), nullable=False))
+    # Per-user API keys set from the user-settings UI; stored in plaintext,
+    # never returned raw — always masked at the API layer.
```
jmelloy, line 196: Storing API keys as plaintext `Text` columns means a DB dump or read replica immediately exposes all user credentials. Consider encrypting at rest using SQLAlchemy's `EncryptedType` or a secrets-management service, and track this as a required security hardening item before shipping to production.

jmelloy (commented):
- **Security concern**: API keys are stored in plaintext in the database with no encryption at rest; the PR description acknowledges this but it warrants explicit tracking as a follow-up issue.
- **Missing `SET` endpoint**: The PR adds view and clear functionality but there is no corresponding `PUT`/`POST` endpoint to actually store tokens, making the feature only half-implemented and the UI's 'Clear' button useless until tokens can be set.
- **`_mask_token` edge case**: When `len(token) == 9`, `prefix_len = min(7, 5) = 5`, so the masked output reveals 9 of 9 characters (5 prefix + 4 suffix), effectively showing the full key for short tokens.
- **DB session lifecycle**: Both new endpoints use a module-level `get_db()` call that could conflict with the pattern used elsewhere in the codebase; if `get_db` is a dependency-injection helper, calling it directly and manually closing may bypass connection-pool bookkeeping.
- **Silent failure on token load**: `loadTokens` swallows all errors silently, meaning auth failures (e.g. expired session) or server errors are invisible to the user; at minimum a non-fatal error indicator should be surfaced.

PR #479, opened 2026-05-27
Fix: chat messages not showing in real-time (#477)

### LINKED ISSUE #477: Chat: incoming messages not displayed in real-time — requires page refresh
## Bug

Foreign (non-self) messages sent in a chat thread are not appearing in the UI in real-time. The user must manually refresh the page to see new messages.

## Expected behaviour

New messages from other participants should appear in the chat immediately via WebSocket push (or polling), with no page refresh required.

## Likely causes

- The WebSocket connection for the chat channel is not being established or maintained correctly.
- Incoming message events are not being handled / dispatched to update UI state.
- The chat component may not be subscribing to the correct channel/topic.
- A React/Svelte/Vue state update is missing so the component doesn't re-render when new data arrives.

## Steps to reproduce

1. Open two browser tabs / sessions as different users.
2. Send a message from one session.
3. Observe that the other session does not show the message until the page is refreshed.

## Acceptance criteria

- Incoming messages appear in real-time without a page refresh.
- The WebSocket (or equivalent) connection is stable and reconnects on drop.

jmelloy:
## Summary

- `computed(() => guildStore.messages)` in `ChatTab.vue` only tracks changes to the ref's **identity** — it is not invalidated when the underlying array is mutated in-place via `.push()`
- The WebSocket handler in `guild.ts` was calling `messages.value.push(chatMsg)`, which mutated the reactive array but never replaced the ref's value
- Because the outer computed was never invalidated, the `watch(messages, …)` scroll handler never fired, and the `groupedMessages` computed's dependency chain was broken — the chat pane silently received other users' messages but the UI never re-rendered
- Changing to `messages.value = [...messages.value, chatMsg]` replaces the array reference, triggers the ref's setter, and propagates the update through the full computed/watch dependency chain so incoming messages render immediately

## Test plan

- [ ] Confirm all 86 existing unit tests pass (`npm test`)
- [ ] Confirm TypeScript type-check is clean (`npm run type-check`)
- [ ] Open two browser tabs connected to the same guild; verify a message sent in one tab appears immediately in the other without a refresh
- [ ] Verify own sent messages continue to appear normally (echo path unchanged)
- [ ] Verify system messages (task-assigned, task-complete, etc.) still render correctly

Closes #477

### DIFF
--- a/frontend/src/stores/guild.ts
+++ b/frontend/src/stores/guild.ts
@@ -119,7 +119,7 @@ export const useGuildStore = defineStore('guild', () => {
         if (data.type === 'chat') {
           const chatMsg = data as ChatMessage
           if ((chatMsg.from || chatMsg.from_agent) !== 'github') {
-            messages.value.push(chatMsg)
+            messages.value = [...messages.value, chatMsg]
           }
         }
         if (data.type === 'guild-updated') {


frontend/src/stores/guild.ts
```diff
@@ -119,7 +119,7 @@ export const useGuildStore = defineStore('guild', () => {
         if (data.type === 'chat') {
           const chatMsg = data as ChatMessage
           if ((chatMsg.from || chatMsg.from_agent) !== 'github') {
-            messages.value.push(chatMsg)
+            messages.value = [...messages.value, chatMsg]
```
jmelloy, line 122: Spreading the entire array on every message (`[...messages.value, chatMsg]`) is O(n) in both time and memory and will become a performance bottleneck in long-running chat sessions. Consider using Vue's `triggerRef` after `.push()` — e.g., `messages.value.push(chatMsg); triggerRef(messages)` — which forces dependency invalidation without copying the array, and is the idiomatic Vue 3 fix for this pattern.

jmelloy (commented):
- The fix correctly addresses the Vue reactivity issue: replacing the array reference instead of mutating in-place ensures computed properties and watchers are properly invalidated and re-executed.
- However, the chosen solution (`[...messages.value, chatMsg]`) creates a full shallow copy of the entire messages array on every incoming message, which will degrade performance as chat history grows — a more idiomatic Vue 3 approach exists.
- The root cause description in the PR is partially inaccurate: Vue 3's `ref` wrapping an array *does* track mutations like `.push()` via its Proxy, but the real issue is more likely in how `computed(() => guildStore.messages)` in the consuming component accesses the store — returning the ref itself rather than its `.value` breaks the dependency tracking chain.
- No new tests were added to cover the real-time message rendering behavior, and the test plan checklist appears incomplete (cut off mid-sentence).
- The change is minimal and low-risk in terms of correctness, but the performance trade-off and missing test coverage are concerns worth addressing.

PR #480, opened 2026-05-27
Fix chat messages reactivity performance (#479 follow-up)

jmelloy:
## Summary

- Replace `messages.value = [...messages.value, chatMsg]` with `messages.value.push(chatMsg)` in `frontend/src/stores/guild.ts`
- `messages` is already a `ref<ChatMessage[]>([])` — Vue 3 tracks mutations on `.value` arrays, so `.push()` triggers reactivity correctly without copying the entire array
- Existing messages still load on mount via direct assignment in `joinGuild()` (unchanged)

## Why

The spread pattern from PR #479 created an O(n) full array copy on every incoming WebSocket chat message. In a long session with hundreds of messages, this compounds with every new message.

## Test plan

- [ ] Existing messages load correctly when joining a guild
- [ ] New incoming WebSocket `chat` messages appear in real-time
- [ ] Sending your own message still appears immediately

### DIFF
--- a/frontend/src/stores/guild.ts
+++ b/frontend/src/stores/guild.ts
@@ -119,7 +119,7 @@ export const useGuildStore = defineStore('guild', () => {
         if (data.type === 'chat') {
           const chatMsg = data as ChatMessage
           if ((chatMsg.from || chatMsg.from_agent) !== 'github') {
-            messages.value = [...messages.value, chatMsg]
+            messages.value.push(chatMsg)
           }
         }
         if (data.type === 'guild-updated') {


frontend/src/stores/guild.ts
```diff
@@ -119,7 +119,7 @@ export const useGuildStore = defineStore('guild', () => {
         if (data.type === 'chat') {
           const chatMsg = data as ChatMessage
           if ((chatMsg.from || chatMsg.from_agent) !== 'github') {
-            messages.value = [...messages.value, chatMsg]
+            messages.value.push(chatMsg)
```
jmelloy, line 122: `.push()` is correct here for Vue 3 reactive refs, but consider whether a maximum message count / ring-buffer cap is needed. Without one, the array grows unboundedly for long-lived WebSocket sessions, which is the same memory-growth problem the O(n) copy was hiding.

jmelloy (commented):
- The change from spread assignment to `.push()` is correct for Vue 3 reactivity — `ref<T[]>` wraps the array in a reactive proxy, so mutations like `.push()` are tracked and trigger updates.
- The performance motivation is valid: the previous O(n) array copy on every WebSocket message would degrade in long chat sessions, and `.push()` is O(1) amortized.
- No functional regression is expected since `joinGuild()` still uses direct assignment for bulk load, and only the incremental-append path is changed.
- The diff is minimal and focused, reducing review risk; however, no automated tests are added or referenced to guard against future regression.
- The PR description's test plan items are all unchecked, suggesting no explicit verification was recorded before submission.

PR #485, opened 2026-05-27
Notify foreman on max-turns (#484)

### LINKED ISSUE #484: Worker doesn't notify foreman when Claude hits max turns
## Problem

When a Claude agent subprocess hits max turns (`error_max_turns`), the worker auto-commits whatever partial state it has and pushes the branch, but the foreman is not notified that the task ended abnormally. This means:

- The task parks in `awaiting-review` with no indication it was truncated
- The foreman has no signal to send a follow-up continuation
- Humans only notice if they happen to check logs or the periodic-check catches it

## Expected behaviour

When `error_max_turns` is raised by the Claude subprocess, the worker should:
1. Emit a structured event/notification to the foreman (e.g. a `task-max-turns` or `task-error` message) so the foreman knows the agent ran out of turns mid-task
2. Optionally include the last N log lines so the foreman has context on where it stopped
3. The task state should reflect the truncation (e.g. a flag or a distinct terminal state vs a clean finish)

## Steps to reproduce

Assign a task that requires many tool calls. When Claude hits its turn limit the worker silently parks the task.

## Notes

- Seen in task `t-tvxkcl` — logs show `✗ error_max_turns:` at the end, worker committed and pushed but foreman was not told
- The fix should be in the worker's Claude subprocess result handler, wherever `stop_reason == "max_turns"` or `error_max_turns` is caught

jmelloy:
## Summary

- When Claude hits its `max_turns` limit, the subprocess exits cleanly (code 0) with `stop_reason="max_turns"`, so the worker correctly sends a `task-complete` message — but `handle_task_complete` was constructing the foreman trigger message as if the task finished normally, with no indication of truncation.
- The foreman would receive the generic "parked in awaiting-review" message and have no signal to issue a follow-up continuation.
- Same gap existed in `handle_task_followup_done` for follow-up runs that hit the turn limit.

## Changes

**`backend/ws_handlers.py`**
- `handle_task_complete`: reads `stopReason` from the WS payload; when it's `"max_turns"`, sends a distinct `[task-complete/max-turns]` trigger message to the foreman that explains the truncation, includes the last 200 chars of Claude's output, and explicitly prompts the foreman to call `send_followup` with a continuation prompt.
- `handle_task_followup_done`: same fix for the follow-up path (`[followup-done/max-turns]`).
- Normal completions are unchanged.

**`backend/tests/test_worker_foreman_notify.py`**
- Added `test_task_complete_max_turns_notifies_foreman`: uses the existing `_trigger_foreman` spy pattern to verify that a `task-complete` with `stopReason=max_turns` triggers a message containing `max-turns`, the task ID, `send_followup` guidance, and the `lastText` snippet.

## Test plan

- [ ] Run `python -m pytest backend/tests/test_worker_foreman_notify.py` (requires `postgres-test` container at localhost:5433)
- [ ] Manually trigger a task that hits the turn limit and verify the foreman chat shows a truncation notice and issues a follow-up

Closes #484

### DIFF
--- a/backend/tests/test_worker_foreman_notify.py
+++ b/backend/tests/test_worker_foreman_notify.py
@@ -4,6 +4,7 @@ Verifies that:
 - handle_worker_register triggers [worker-online] to the foreman
 - handle_worker_disconnect triggers [worker-offline] reason=shutdown
 - abrupt WebSocket disconnect triggers [worker-offline] reason=disconnect
+- task-complete with stopReason=max_turns triggers a max-turns foreman message
 """
 
 from __future__ import annotations
@@ -24,7 +25,7 @@ import database as database_module  # noqa: E402
 import main as main_module  # noqa: E402
 import ws_handlers  # noqa: E402
 from _test_config import TEST_DATABASE_URL  # noqa: E402
-from helpers import insert_guild, insert_worker  # noqa: E402
+from helpers import insert_guild, insert_task, insert_worker  # noqa: E402
 
 
 @pytest.fixture(scope="module")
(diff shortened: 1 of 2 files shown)


backend/tests/test_worker_foreman_notify.py
```diff
@@ -182,3 +183,73 @@ def test_abrupt_disconnect_notifies_foreman(client):
...
+                        "agentId": "a-nfy004-obs",
+                        "agentName": "Observer",
+                        "agentType": "browser",
+                    }
+                )
+                ws_obs.receive_json()  # agent-joined for observer
+                ws_worker.receive_json()  # observer's agent-joined broadcast to worker
+
+                ws_worker.send_json(
+                    {
```
jmelloy, line 228: This assertion is a no-op. `assert <expr>, True` passes the boolean `True` as the failure message, not as the expected value — the assertion will always pass regardless of whether `max_turns` appears in the string. The intent was likely `assert "max_turns" not in ..., "message"` with a real condition, but as written it provides zero test coverage for this invariant.

backend/ws_handlers.py
```diff
@@ -694,6 +710,8 @@ async def handle_task_complete(ctx: WSContext, data: dict) -> None:
 async def handle_task_followup_done(ctx: WSContext, data: dict) -> None:
     task_id = data.get("taskId")
```
jmelloy, line 711: The variable name `last_text_fud` is an unexplained abbreviation. Given that `handle_task_complete` uses `last_text` for the same field, consistency and readability would be improved by using `last_text` here as well, avoiding the need to mentally decode the suffix.

backend/ws_handlers.py
```diff
@@ -643,6 +643,7 @@ async def handle_task_complete(ctx: WSContext, data: dict) -> None:
     branch = data.get("branch", "")
```
jmelloy, line 643: There is no logging or validation for unexpected `stopReason` values. If a worker sends an unrecognised reason (e.g., `"error"`, `"timeout"`, or a future Claude stop reason), it silently falls through to the normal completion message. Consider logging a warning for unrecognised values so they are observable in production.

backend/ws_handlers.py
```diff
@@ -736,7 +754,15 @@ async def handle_task_followup_done(ctx: WSContext, data: dict) -> None:
```
jmelloy, line 754: The diff shows `human_msg` being constructed in the `max_turns` branch but the code where it is passed to `_trigger_foreman` is not visible in the diff. Please verify the full function ensures `human_msg` is always forwarded to `_trigger_foreman` and that the non-max-turns, non-queued path also correctly triggers the foreman (the prior else-chain must be complete).

jmelloy (commented):
- The PR correctly addresses the gap where `max_turns` completions were indistinguishable from normal completions in foreman notifications, adding a distinct `[task-complete/max-turns]` trigger with actionable guidance.
- The default value for `stop_reason` is `"success"` which is fine, but no validation or logging occurs for unexpected `stopReason` values, meaning typos or new stop reasons would silently fall through to the normal path.
- The variable name `last_text_fud` in `handle_task_followup_done` is an awkward abbreviation that reduces readability, especially when compared to `last_text` used in the sibling function.
- The test assertion on line 228 (`assert "max_turns" not in msg.split("max-turns")[0].replace("max_turns", ""), True`) is a no-op — asserting a value against `True` using a positional argument always passes regardless of the expression's truthiness; this check provides zero coverage.
- The `handle_task_followup_done` max-turns branch builds `human_msg` but the diff is cut off before showing where it is passed to `_trigger_foreman`, making it impossible to verify the full handler logic is correct; the prior `queued_payloads` branch also appears to fall through without a final else, which could mean no trigger fires for a normal followup-done without queued payloads.

PR #487, opened 2026-05-27
Audit worker max-turns handling & follow-up queue bug

jmelloy:
## Summary

- Adds `docs/max-turns-audit.md` investigating why follow-up tasks are silently dropped after `max_turns` exits (task `t-tvxkcl`, ~16:14 UTC 2026-05-27)
- Identifies two independent root causes and evaluates PR #485 for regression risk

## Root causes found

**Bug 1a — Lock not released on `error` state** (`backend/ws_handlers.py:40,631`):
`_TERMINAL_STATES = ("done", "failed", "cancelled")` — `"error"` is absent. When a follow-up task exits with non-zero code, `handle_task_update` sets `state="error"` but never calls `LockService.release()`. All subsequent `send_followup` calls queue as `pending-followup` (which only drains on `task-followup-done`, a message that's never sent for failed tasks). Follow-ups are permanently blocked with no visible error.

**Bug 1b — No max-turns signal to foreman** (`backend/ws_handlers.py:675-691`):
When max-turns exits with code 0, `handle_task_complete` fires with a generic "default: leave open" message. The foreman has no directive to call `send_followup` and silently does nothing. Even if `send_followup` is called manually, `_select_followup_worker` returns `None` if the worker is still in `working` state (pre-idle transition), causing all calls to fail with no state change and no log entries — matching the t-tvxkcl symptom.

## PR #485 assessment

PR #485 (`claude/notify-foreman-on-max-turns-484-t-ty0s`) correctly fixes Bug 1b by adding a max-turns-specific foreman message. The `stop_reason == "max_turns"` check matches the value documented in `claude_runner.py` and used in tests. It does **not** fix Bug 1a (lock not released on error). No commit on `main` regressed this behavior — both issues predate recent activity.

## Recommended fixes

1. **P0** — Add `"error"` to the lock-release condition in `handle_task_update` (one-liner fix)
2. **P1** — Merge PR #485
3. **P2** — Harden `stop_reason` checks against `"error_max_turns"` variant

### DIFF
--- a/docs/max-turns-audit.md
+++ b/docs/max-turns-audit.md
@@ -1,0 +1,253 @@
+# Max-Turns Audit: Follow-up Queue Bug
+
+**Date:** 2026-05-27  
+**Task under investigation:** `t-tvxkcl` — `error_max_turns` at ~16:14 UTC, three subsequent `send_followup` calls silently dropped  
+**Branch investigated:** `origin/claude/notify-foreman-on-max-turns-484-t-ty0s` (PR #485)
+
+---
+
+## 1. Root Cause: Why Follow-ups Don't Run After Max-Turns
+
+There are two independent mechanisms that can cause follow-ups to be silently dropped after max-turns. Depending on whether Claude exits with code 0 or non-zero, one or both apply.
+
+### 1a. Lock not released when task transitions to `error` state
+
+**File:** `backend/ws_handlers.py:40,631`
+
+```python
+# line 40
+_TERMINAL_STATES = ("done", "failed", "cancelled")  # "error" is missing
+
+# line 629-633 — handle_task_update
+if update_values:
+    await ctx.db.execute(update(Task).where(Task.id == task_id).values(**update_values))
+    if update_values.get("state") in _TERMINAL_STATES:          # "error" never matches
+        await LockService(ctx.db).release(f"task:{task_id}")    # lock stays held
+    await ctx.db.commit()
+```
+
+When `success=False` — exit code non-zero, which can happen on any error including some
+versions of Claude CLI reporting max-turns — the worker sends:
+
+```python
+# worker/pioneer_worker/worker.py:1902-1908
+await self._task_update(
+    task_id,
+    agent=agent,
+    type="needs-input" if stop_reason == "needs_input" else "error",
+    state="error",          # <-- sends state="error" via handle_task_update
+    **msg,
+)
+```
+
+`handle_task_update` sets the task to `error` state but **never releases the follow-up
+lock** because `"error"` is absent from `_TERMINAL_STATES`. From this point:
+
+1. `send_followup` calls `LockService.acquire(f"task:{task_id}")` → fails (lock held).
+2. Each call is stored as a `pending-followup` `TaskEvent` row.
+3. The `pending-followup` queue is only drained in `handle_task_followup_done`
+   (`ws_handlers.py:724-735`), which is triggered by `task-followup-done` — a message
+   the worker only sends on **success**.
+4. The queue therefore never drains. All follow-ups are permanently blocked.
+
+**Verdict:** This is a definite code bug. Any task that exits with a non-zero code (error,
+interrupted, or some CLI versions of max-turns) leaves the lock held forever and silently
+swallows all subsequent `send_followup` tool calls.
+
+### 1b. Foreman receives no max-turns signal — follow-up loop never starts
+
+**File:** `backend/ws_handlers.py:675-691`
+
+When max-turns exits with code 0 (`success=True`, the normal case per `claude_runner.py`
+docstring and tests), the task correctly lands in `awaiting-review` with the lock released.
+However, the foreman is triggered with a generic "task complete" message:
+
+```python
+# ws_handlers.py:675-691 (current main)
+await _trigger_foreman(
+    ctx.guild_id,
+    "task-complete",
+    f"[task-complete] Worker {worker_id_msg} finished task {task_id}: "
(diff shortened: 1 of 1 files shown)


docs/max-turns-audit.md
```diff
@@ -0,0 +1,253 @@
...
+         ...
+```
+
+**Assessment of PR #485:**
+
+- Correctly fixes Bug 1b for the normal case (`exit_code=0`, `stop_reason="max_turns"`).
+- The `"max_turns"` string matches the value emitted by `claude_runner.py` (confirmed by
+  docstring at line 282 and test at `worker/tests/test_claude_runner.py:270`).
+- Moving the `max_turns` check **before** `queued_payloads` in `handle_task_followup_done`
+  is correct — a max-turns exit should tell the foreman to resume, regardless of whether
```
jmelloy, line 164: Option 2 is cleaner but still leaves the `pending-followup` queue in a permanently drained state for tasks that hit the `error` path before this fix — accumulated `pending-followup` rows won't be replayed. The recommendation should include a one-time migration or drain step for existing stuck tasks, and ideally the drain should be tied to lock release rather than the `task-followup-done` message type.

docs/max-turns-audit.md
```diff
@@ -0,0 +1,253 @@
...
+| Time | Event |
+|------|-------|
+| ~16:14 UTC | Claude subprocess hits max-turns; exits with code 0; `stop_reason="max_turns"` |
+| +0s | Worker sends `task-complete` (`stopReason="max_turns"`, `state="awaiting-review"`) |
+| +0s | `handle_task_complete` sets task to `awaiting-review`, releases lock |
+| +0s | Foreman triggered with generic "task complete, default: leave open" message |
+| +Xs | Foreman calls `send_followup` (3 times, likely based on prior context) |
+| +Xs | `_select_followup_worker` returns `None` — no idle agents at that moment |
+| +Xs | All 3 calls return "No idle worker available"; no state change; no log entries |
+| now | Task stuck in `awaiting-review`; no follow-up ever dispatched |
```
jmelloy, line 102: The latent risk note about `"error_max_turns"` vs `"max_turns"` is valuable, but the audit dismisses it as a 'display artifact' without verifying the raw WebSocket payload. The recommendation in Fix C is correct regardless — the string comparison should be confirmed against the actual `stopReason` field in the wire message, not inferred from terminal rendering.

docs/max-turns-audit.md
```diff
@@ -0,0 +1,253 @@
...
++    stop_reason = data.get("stopReason", "success")
++    last_text_fud = data.get("lastText", "")
+     ...
++    if stop_reason == "max_turns":
++        human_msg = (
++            f"[followup-done/max-turns] ... hit Claude's max-turns limit ..."
++            "Call send_followup with a continuation prompt to resume ..."
++        )
++    elif queued_payloads:
+         ...
```
jmelloy, line 155: Option 1 (`new_state == 'error'`) introduces a string literal that can silently drift if the state value changes. Option 2 is preferable but `_LOCK_RELEASE_STATES` should be defined close to `_TERMINAL_STATES` with a comment explaining why `error` is deliberately excluded from the terminal set — otherwise a future refactor may inadvertently merge them and cause behavioral changes.

docs/max-turns-audit.md
```diff
@@ -0,0 +1,253 @@
...
+1. `send_followup` calls `LockService.acquire(f"task:{task_id}")` → fails (lock held).
+2. Each call is stored as a `pending-followup` `TaskEvent` row.
+3. The `pending-followup` queue is only drained in `handle_task_followup_done`
+   (`ws_handlers.py:724-735`), which is triggered by `task-followup-done` — a message
+   the worker only sends on **success**.
+4. The queue therefore never drains. All follow-ups are permanently blocked.
+
+**Verdict:** This is a definite code bug. Any task that exits with a non-zero code (error,
+interrupted, or some CLI versions of max-turns) leaves the lock held forever and silently
+swallows all subsequent `send_followup` tool calls.
```
jmelloy, line 55: The timeline notes that `send_followup` returned 'No idle worker available' with no state change and no log entries, but doesn't confirm whether the calls arrived *before* or *after* the worker transitioned back to idle. If the race window is narrow, a simple retry with backoff in `send_followup` (or an async re-queue on idle-worker reconnection) would close both Bug 1b and this timing issue without requiring foreman LLM intervention.

jmelloy (commented):
- The audit document is thorough and well-structured, correctly identifying two independent root causes (lock not released on `error` state, and no max-turns directive to foreman) that together explain the silent follow-up drop observed in task t-tvxkcl.
- Bug 1a fix (Option 2) is the stronger recommendation — introducing `_LOCK_RELEASE_STATES` makes the intent explicit and avoids a subtle logic branch, but neither option handles the case where a task transitions to `error` *without* going through `handle_task_update` (e.g., direct DB writes or other code paths).
- The `pending-followup` drain mechanism is a latent footgun beyond just the `error` state: any state not in `_TERMINAL_STATES` that terminates a task will leave the queue permanently blocked, suggesting the drain should be triggered on lock release rather than on a specific message type.
- The Fix C recommendation (`_MAX_TURNS_REASONS` frozenset) is sound defensive practice, but the audit should also recommend a test covering the `error_max_turns` subtype to prevent silent regression.
- Priority ordering (P0 Fix A → P1 PR #485 → P2 Fix C) is correct; however, Fix A and Fix B together are insufficient unless the `pending-followup` drain is also reviewed, since queued follow-ups accumulated before the fix will never be replayed.

PR #488, opened 2026-05-27
Fix foreman chat messages not showing in UI (#486)

### LINKED ISSUE #486: Foreman chat messages not appearing in UI
## Problem

Messages sent by the Foreman (operator) in the chat panel are not showing up in the UI. This persists after PR #480 (chat reactivity `.push()` fix) was merged.

## Symptoms

- Foreman sends a chat message
- The message does not appear in the chat UI
- Worker/user messages may or may not be affected

## Investigation needed

1. Check how foreman chat messages are emitted from the backend — look for the event type/channel used for foreman vs worker messages.
2. Check the frontend WebSocket handler to see if foreman message events are handled/filtered differently from other message types.
3. Check whether the message is stored in the DB but not pushed to the WebSocket, or emitted over WS but not rendered (e.g. wrong `role` / `sender` field causing the component to skip it).
4. Check the chat Vue component for any conditional render logic that might exclude foreman messages.

jmelloy:
## Root cause

`run_foreman_ai` in `backend/foreman/runner.py` passed raw `datetime` objects as `"createdAt"` values inside the dicts handed to `broadcast()`. FastAPI's `WebSocket.send_json` uses the standard `json.dumps` with no custom encoder, so it raises `TypeError: Object of type datetime is not JSON serializable`.

The `broadcast()` helper catches **all** exceptions from `send_json` and marks the offending WebSocket as dead, removing it from the connections pool. This meant:
1. The very first foreman message failed to serialize.
2. The client's WebSocket was evicted from `connections[guild_id]`.
3. Every subsequent broadcast in the same foreman run had no recipients — silently dropped.

## Fix

Called `.isoformat()` on the three `_now` datetime objects that were passed directly into broadcast payloads (text-block, tool-use, and tool-result events). The safety-cap path on the same line range was already correct (its `_now` was assigned as an ISO string).

## Testing

Verified with `ruff check` — no issues. The change is a one-liner per call site with no logic changes.

Closes #486

### DIFF
--- a/backend/foreman/runner.py
+++ b/backend/foreman/runner.py
@@ -469,7 +469,7 @@ async def run_foreman_ai(
                             "from": "foreman",
                             "to": "user",
                             "content": b.text.strip(),
-                            "createdAt": _now,
+                            "createdAt": _now.isoformat(),
                         },
                     )
 
@@ -490,7 +490,7 @@ async def run_foreman_ai(
                         "toolName": tu.name,
                         "toolInput": dict(tu.input) if tu.input else {},
                         "toolId": tu.id,
-                        "createdAt": _now,
+                        "createdAt": _now.isoformat(),
                     },
                 )
 
@@ -520,7 +520,7 @@ async def run_foreman_ai(
                         "toolId": result.get("tool_use_id"),
                         "toolOutput": result.get("content", ""),
                         "isError": result.get("is_error", False),
-                        "createdAt": _now,
+                        "createdAt": _now.isoformat(),
                     },
                 )
 


backend/foreman/runner.py
```diff
@@ -469,7 +469,7 @@ async def run_foreman_ai(
                             "from": "foreman",
                             "to": "user",
                             "content": b.text.strip(),
-                            "createdAt": _now,
+                            "createdAt": _now.isoformat(),
```
jmelloy, line 472: Consider defining `_now` as `datetime.utcnow().isoformat()` (or equivalent) at the point of assignment rather than calling `.isoformat()` at each use site. This would be more DRY and eliminate the risk of future callers forgetting the conversion, especially since there are at least 3 broadcast sites using the same variable.

backend/foreman/runner.py
```diff
@@ -490,7 +490,7 @@ async def run_foreman_ai(
                         "toolName": tu.name,
                         "toolInput": dict(tu.input) if tu.input else {},
                         "toolId": tu.id,
-                        "createdAt": _now,
+                        "createdAt": _now.isoformat(),
```
jmelloy, line 493: The fix is correct here, but note that `_now` is reassigned at multiple points in the loop. If `_now` were typed as `str` from the start (assigned as `.isoformat()` on creation), the type checker would catch any future raw-datetime broadcast attempts at development time rather than at runtime.

backend/foreman/runner.py
```diff
@@ -520,7 +520,7 @@ async def run_foreman_ai(
                         "toolId": result.get("tool_use_id"),
                         "toolOutput": result.get("content", ""),
                         "isError": result.get("is_error", False),
-                        "createdAt": _now,
+                        "createdAt": _now.isoformat(),
```
jmelloy, line 523: This hunk is correct. As a follow-up, the `broadcast()` helper's exception handler that evicts WebSockets on *any* exception (including serialization errors) should be tightened — serialization failures should arguably raise rather than silently drop the connection, so bugs like this are immediately visible in logs.

jmelloy (commented):
- The fix correctly identifies and resolves the root cause: raw `datetime` objects being passed to `broadcast()` which uses `json.dumps` without a custom encoder, causing `TypeError` and silent WebSocket eviction.
- The three-line change is minimal, targeted, and consistent — all affected `_now` usages in broadcast payloads are updated to `.isoformat()`.
- A deeper concern remains: the `broadcast()` helper silently swallows serialization errors and evicts the WebSocket, which is an overly aggressive error-handling strategy that can mask bugs. This PR doesn't address that design issue, but it's worth noting.
- No tests are included in the diff despite the PR description referencing verification; automated regression coverage for this serialization path would prevent recurrence.
- It would be worth auditing other `broadcast()` call sites in the codebase for similar datetime or non-serializable object issues, as the pattern may exist elsewhere.

PR #491, opened 2026-05-27
Tighten broadcast() + datetime-aware WS JSON (#490)

### LINKED ISSUE #489: This hunk is correct. As a follow-up, the `broadcast()` helper's exception handler that evicts WebSockets on *any* exception (including serialization errors) should be tightened — serialization failures should arguably raise rather than silently drop the connection, so bugs like this are immediately visible in logs.
This hunk is correct. As a follow-up, the `broadcast()` helper's exception handler that evicts WebSockets on *any* exception (including serialization errors) should be tightened — serialization failures should arguably raise rather than silently drop the connection, so bugs like this are immediately visible in logs.

_Originally posted by @jmelloy in https://github.com/jmelloy/pioneer-square/pull/488#discussion_r3312556205_

### LINKED ISSUE #490: Tighten broadcast() exception handling & add datetime-aware JSON encoding for WebSocket
## Problem

Two related issues in the WebSocket broadcast path:

### 1. `broadcast()` swallows serialization errors silently
The `broadcast()` helper catches all exceptions and evicts the WebSocket on any error — including `json.dumps()` serialization failures (e.g. unserializable types like `datetime`). This means a bug in the payload is invisible: the connection is silently dropped and no error surfaces in the logs. Serialization errors should **raise** immediately so they are visible and loud, not swallowed.

**Fix:** Split the exception handler in `broadcast()` into two cases:
- **Serialization errors** (raised before any send): re-raise so the bug is immediately visible in logs and doesn't silently evict a healthy WebSocket connection.
- **Send errors** (network/connection failures after serialization): keep the existing eviction behavior — the client really is gone.

### 2. `json.dumps()` is not datetime-aware
Any `datetime` object in a WS payload causes an unhandled `TypeError` in `json.dumps()`, which then hits the broad exception handler above and silently kills the connection.

**Fix:** Pass a custom `default` encoder to every `json.dumps()` call on the WS path (or wrap in a shared `ws_json_dumps()` helper) that serializes `datetime` / `date` objects to ISO-8601 strings and `UUID` to `str`. Example:

```python
def _json_default(obj):
    if isinstance(obj, (datetime, date)):
        return obj.isoformat()
    if isinstance(obj, UUID):
        return str(obj)
    raise TypeError(f"Object of type {type(obj)} is not JSON serializable")

def ws_dumps(payload: Any) -> str:
    return json.dumps(payload, default=_json_default)
```

## Acceptance criteria
- [ ] `broadcast()` re-raises on serialization failure instead of evicting the WebSocket
- [ ] All `json.dumps()` calls on the WS broadcast path use a datetime/UUID-aware encoder
- [ ] A `datetime` value in a WS payload serializes cleanly to an ISO-8601 string
- [ ] Connection eviction still happens on genuine send failures (network errors)
- [ ] Closes #489

jmelloy:
## Summary

- Adds `ws_dumps()` helper in `backend/events.py` with a custom `json.dumps` encoder that serialises `datetime`/`date` values to ISO-8601 and `UUID` to `str` — preventing silent failures when broadcast payloads contain these types.
- Splits `broadcast()` into two distinct steps:
  - **Serialisation** (`ws_dumps()`): `TypeError` propagates immediately — the connection is healthy, so the bug should be loud and visible, not swallowed.
  - **Send** (`ws.send_text(data)`): network/connection errors still evict the dead WebSocket as before.
- Switches from `ws.send_json()` to `ws.send_text(ws_dumps())` so the single encoder is used consistently for every outbound message.
- Adds `logger.warning()` on send failures so evictions are visible in logs.
- Adds unit tests for `ws_dumps` covering datetime, date, UUID, plain dict, and unhandled type.

## Test plan

- [x] Broadcast a payload containing a `datetime` field — verify it serialises to an ISO-8601 string on the client side with no error
- [x] Broadcast a payload containing an unserializable type — verify `TypeError` is raised immediately (not silently swallowed)
- [x] Disconnect a client while the backend is running — verify the dead WebSocket is still evicted on the next send attempt

Closes #490
Closes #489

### DIFF
--- a/backend/events.py
+++ b/backend/events.py
@@ -12,7 +12,7 @@ from typing import Any
 from uuid import UUID
 
 from database import get_db
-from fastapi import WebSocket
+from fastapi import WebSocket, WebSocketDisconnect
 from models import Agent, TaskLog
 from sqlalchemy import select
 
@@ -97,8 +97,11 @@ async def broadcast(guild_id: str, message: dict, exclude: WebSocket | None = No
             continue
         try:
             await ws.send_text(data)
-        except Exception as exc:
+        except (WebSocketDisconnect, RuntimeError, OSError) as exc:
             logger.warning("broadcast: send failed for %s guild=%s: %s", ws, guild_id, exc)
+            dead.append(ws)
+        except Exception as exc:
+            logger.warning("Unexpected error evicting WebSocket: %s", exc, exc_info=True)
             dead.append(ws)
     for ws in dead:
         connections[guild_id].remove(ws)
--- a/backend/events.py
+++ b/backend/events.py
@@ -5,8 +5,11 @@ broadcast/emit_terminal_line without circular dependencies.
 """
 
 import asyncio
+import json
 import logging
-from datetime import UTC, datetime
+from datetime import UTC, date, datetime
+from typing import Any
+from uuid import UUID
 
 from database import get_db
 from fastapi import WebSocket
@@ -14,6 +17,20 @@ from models import Agent, TaskLog
 from sqlalchemy import select
 
 logger = logging.getLogger(__name__)
+
+
+def _json_default(obj: Any) -> Any:
+    if isinstance(obj, (datetime, date)):
+        return obj.isoformat()
+    if isinstance(obj, UUID):
+        return str(obj)
+    raise TypeError(f"Object of type {type(obj)} is not JSON serializable")
+
+
+def ws_dumps(payload: Any) -> str:
+    """Serialize payload to a JSON string with datetime/UUID support."""
+    return json.dumps(payload, default=_json_default)
+
 
 # In-memory WebSocket connections: guild_id -> list of WebSocket
 connections: dict[str, list[WebSocket]] = {}
@@ -71,12 +88,15 @@ async def broadcast(guild_id: str, message: dict, exclude: WebSocket | None = No
         return
     if guild_id not in connections:
         return
+    # Serialize once before iterating — let TypeError propagate immediately so
+    # serialization bugs are visible rather than silently evicting connections.
+    data = ws_dumps(message)
     dead = []
     for ws in connections[guild_id]:
         if ws is exclude:
             continue
         try:
-            await ws.send_json(message)
+            await ws.send_text(data)
         except Exception:
             dead.append(ws)
     for ws in dead:
--- a/backend/events.py
+++ b/backend/events.py
@@ -97,7 +97,8 @@ async def broadcast(guild_id: str, message: dict, exclude: WebSocket | None = No
             continue
         try:
             await ws.send_text(data)
-        except Exception:
+        except Exception as exc:
+            logger.warning("broadcast: send failed for %s guild=%s: %s", ws, guild_id, exc)
             dead.append(ws)
     for ws in dead:
         connections[guild_id].remove(ws)
--- a/backend/tests/test_ws_dumps.py
+++ b/backend/tests/test_ws_dumps.py
@@ -5,7 +5,7 @@ from __future__ import annotations
 import json
 import os
 import sys
-from datetime import date, datetime, timezone
+from datetime import UTC, date, datetime, timezone
 from uuid import UUID
 
 import pytest
@@ -16,7 +16,7 @@ from events import ws_dumps  # noqa: E402
 
 
 def test_datetime_serializes_to_iso8601():
-    dt = datetime(2024, 3, 15, 10, 30, 0, tzinfo=timezone.utc)
+    dt = datetime(2024, 3, 15, 10, 30, 0, tzinfo=UTC)
     result = json.loads(ws_dumps({"ts": dt}))
     assert result["ts"] == "2024-03-15T10:30:00+00:00"
 
--- a/backend/tests/test_ws_dumps.py
+++ b/backend/tests/test_ws_dumps.py
@@ -1,0 +1,47 @@
+"""Unit tests for ws_dumps in events.py."""
+
+from __future__ import annotations
+
+import json
+import os
+import sys
+from datetime import date, datetime, timezone
+from uuid import UUID
+
+import pytest
+
+sys.path.insert(0, os.path.join(os.path.dirname(__file__), ".."))
+
+from events import ws_dumps  # noqa: E402
+
+
+def test_datetime_serializes_to_iso8601():
+    dt = datetime(2024, 3, 15, 10, 30, 0, tzinfo=timezone.utc)
+    result = json.loads(ws_dumps({"ts": dt}))
+    assert result["ts"] == "2024-03-15T10:30:00+00:00"
+
+
+def test_date_serializes_to_iso8601():
+    d = date(2024, 3, 15)
+    result = json.loads(ws_dumps({"d": d}))
+    assert result["d"] == "2024-03-15"
+
+
+def test_uuid_serializes_to_string():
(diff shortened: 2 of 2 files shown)


jmelloy (commented):
{"summary": "- Serialising once before the loop is a good correctness and performance improvement — avoids redundant serialisation per connection and ensures a `TypeError` is loud rather than silently swallowed by the `except Exception` block.\n- `ws_dumps` is a sensible, reusable abstraction, but it is not tested directly and `_json_default` has no coverage for subclasses of `datetime`/`date` that may bypass `isinstance` checks in edge cases (unlikely, but worth noting).\n- The `except Exception` catch on `send_text` is intentionally broad to evict dead connections, but it will also swallow transient errors (e.g. `asyncio.CancelledError` in Python <3.8 or other unexpected exceptions) — consider catching a narrower set such as `(WebSocketDisconnect, RuntimeError, OSError)`.\n- `ws_dumps` is defined as a module-level public symbol but there are no corresponding exports or `__all__` entries, which is fine for a small module but could become confusing as the file grows.\n- The PR description's test plan is incomplete (the third checklist item is cut off), which may leave a regression path untested around concurrent disconnects during broadcast."], "comments": [{"path": "backend/events.py", "line": 97, "side": "RIGHT", "body": "Catching bare `Exception` here also silences `asyncio.CancelledError` (Python ≤3.7) and other non-network errors, making debugging harder. Prefer a narrower set like `except (RuntimeError, OSError)` (Starlette raises `RuntimeError` for a closed WebSocket) so only genuine connection failures trigger eviction."}, {"path": "backend/events.py", "line": 21, "side": "RIGHT", "body": "`isinstance(obj, (datetime, date))` correctly catches `datetime` first because `datetime` is a subclass of `date`, but the order matters — if this were reversed (checking `date` first) a `datetime` would be serialised via `date.isoformat()` and lose the time component. Add a comment or reorder with `datetime` explicitly first to document this intent and prevent a future re

backend/events.py
```diff
@@ -15,6 +18,20 @@
...
+
+def _json_default(obj: Any) -> Any:
+    if isinstance(obj, (datetime, date)):
+        return obj.isoformat()
+    if isinstance(obj, UUID):
+        return str(obj)
+    raise TypeError(f"Object of type {type(obj)} is not JSON serializable")
+
+
+def ws_dumps(payload: Any) -> str:
```
jmelloy, line 30: `ws_dumps` is module-level public API but has no unit tests in this PR. Given the explicit test-plan items are all unchecked, please add at least one pytest parametrize covering `datetime`, `date`, `UUID`, and an unserializable type before merging.

backend/events.py
```diff
@@ -15,6 +18,20 @@
...
 logger = logging.getLogger(__name__)
 
+
+def _json_default(obj: Any) -> Any:
+    if isinstance(obj, (datetime, date)):
+        return obj.isoformat()
+    if isinstance(obj, UUID):
+        return str(obj)
+    raise TypeError(f"Object of type {type(obj)} is not JSON serializable")
+
```
jmelloy, line 28: The `datetime`/`date` isinstance check will return a plain ISO string (e.g. `"2024-01-15T10:30:00"`) with no timezone suffix when a naive `datetime` is passed. Consider asserting or warning when a timezone-unaware `datetime` is encountered, since clients may misinterpret naive datetimes as local time.

backend/events.py
```diff
@@ -71,12 +88,15 @@ async def broadcast(guild_id: str, message: dict, exclude: WebSocket | None = No
         return
     if guild_id not in connections:
         return
+    # Serialize once before iterating — let TypeError propagate immediately so
+    # serialization bugs are visible rather than silently evicting connections.
+    data = ws_dumps(message)
     dead = []
     for ws in connections[guild_id]:
         if ws is exclude:
             continue
```
jmelloy, line 97: `except Exception` catches everything including `asyncio.CancelledError` (Python < 3.8 only, but also `BaseException` subclasses in some Starlette versions). More critically, a `RuntimeError: Cannot call send after a WebSocket is disconnected` from Starlette is silently evicted rather than logged. Consider catching `Exception` but adding a `logger.debug` for unexpected error types that are not WebSocket disconnect errors.

jmelloy (commented):
- Good separation of serialisation from network errors; pre-serialising once before the loop is a clear correctness improvement.
- `ws_dumps` is public but lacks any test coverage in the PR, and the test-plan items are all unchecked checkboxes.
- The `except Exception` swallow in the send loop is intentionally kept broad, which is fine for dead-socket eviction, but `RuntimeError` from Starlette (e.g. sending on a closed websocket before disconnect is detected) will be silently swallowed alongside genuine network errors — this was pre-existing but worth noting.
- `_json_default` correctly raises `TypeError` for unknown types, matching the contract of `json.dumps`; no issues there.
- Minor nit: `ws_dumps` is exported at module level but never added to `__all__` (if one exists) and the naming convention (`ws_dumps`) slightly mixes domains — consider `serialize_ws_payload` or similar for clarity.

backend/events.py
```diff
@@ -15,6 +18,20 @@
 
 logger = logging.getLogger(__name__)
 
+
+def _json_default(obj: Any) -> Any:
+    if isinstance(obj, (datetime, date)):
+        return obj.isoformat()
+    if isinstance(obj, UUID):
+        return str(obj)
+    raise TypeError(f"Object of type {type(obj)} is not JSON serializable")
```
jmelloy, line 27: `ws_dumps` is not exported or listed in `__all__`, yet the docstring and PR description treat it as a public API. If it is intended for use in other modules (e.g. to pre-serialise payloads before calling `broadcast`), consider either exporting it explicitly or renaming it with a leading underscore to match `_json_default` and signal it is internal.

jmelloy (commented):
- Serialising once before iterating is a clear correctness and performance improvement; the design decision to let `TypeError` propagate immediately (rather than swallow it in the send-error handler) is sound and well-motivated.
- `ws_dumps` is a thin, focused helper and the custom `_json_default` covers the most common non-serialisable types (`datetime`, `date`, `UUID`), though it does not handle types like `Decimal`, `Enum`, or `bytes` that also appear frequently in FastAPI/SQLAlchemy payloads.
- The `except Exception` catch on the send path is intentionally broad; while acceptable for network teardown, it silently discards non-network errors (e.g. a bug introduced in a `WebSocket` subclass), which could hide regressions.
- The PR description test plan is incomplete (the third bullet is cut off), but the logic changes are self-contained and low-risk.
- No security concerns identified; the change does not alter authentication, authorisation, or input-validation logic.

backend/events.py
```diff
@@ -15,6 +18,20 @@
 
 logger = logging.getLogger(__name__)
 
+
```
jmelloy, line 21: The `_json_default` handler covers `datetime`, `date`, and `UUID` but not `decimal.Decimal` (common in SQLAlchemy models) or `enum.Enum` (common in Pydantic schemas). Omitting these means the first `broadcast` call with such a value will still raise `TypeError`; it may be worth adding them now while the encoder is being established.

backend/events.py
jmelloy, line 97: `except Exception` catches every error from `ws.send_text`, including programming errors unrelated to the connection state (e.g. passing a non-`str` argument). Consider narrowing to `except (WebSocketDisconnect, RuntimeError, OSError)` or at minimum logging the exception at WARNING level so silent evictions are observable in production logs.

PR #494, opened 2026-05-27
Consolidate pioneer-foreman/ and foreman/ directories (#492)

### LINKED ISSUE #492: Consolidate pioneer-foreman/ and foreman/ directories
There are now two directories — `pioneer-foreman/` and `foreman/` — that appear to serve the same purpose. The backend currently imports from one or both of them, creating confusion and potential drift.

## Work needed
1. Audit both directories: list all modules/files in each, identify overlaps and unique content.
2. Determine which directory the backend actually imports from (grep for `from foreman`, `from pioneer_foreman`, `import foreman`, etc.).
3. Migrate any code that is still only in the "losing" directory into the canonical one.
4. Update all imports in `backend/` (and anywhere else) to use only the canonical directory.
5. Delete the redundant directory entirely.
6. Verify tests pass after the consolidation.

Closes this issue once the duplicate is gone and imports are clean.

jmelloy:
## Summary

- **Deletes `pioneer-foreman/`** entirely — this was an older Phase 3 standalone foreman implementation (no JWT auth, no Bedrock support, tool execution implemented inline rather than delegated to the backend)
- **Keeps `foreman/`** as the single canonical standalone foreman directory — it has JWT auth via `ForemanHTTPClient`, Bedrock provider support, and clean tool delegation to the backend's `/foreman/exec_tool` endpoint
- **Removes the `pioneer-foreman` service from `docker-compose.yml`** — only the `foreman` service (profile `foreman`) remains
- **Simplifies `backend/tests/test_tools_dedup.py`** — removes the now-gone `test_pioneer_foreman_uses_canonical_tools` test and its `sys.path` manipulation; remaining tests continue to verify `foreman/pioneer_foreman/tools.py` uses the canonical tool list from `backend.foreman_core`

## Why `foreman/` wins

| Feature | `foreman/` | `pioneer-foreman/` |
|---|---|---|
| JWT auth (`PIONEER_FOREMAN_KEY`) | ✅ `JWTTokenManager` | ❌ static token only |
| Bedrock / provider config | ✅ | ❌ |
| Tool execution | Delegates to `POST /foreman/exec_tool` (single source of truth) | Reimplements all tools inline (~1 200 lines) |
| Prompt/utils | Re-exports from `backend.foreman_core` | Duplicates inline |
| Config richness | `backend_key`, `provider`, `aws_region`, `max_rounds`, `history_limit` | Subset only |

## Test plan

- [x] `python -c "import pioneer_foreman.tools as t; from backend.foreman_core.tools_schema import FOREMAN_TOOLS as c; assert t.FOREMAN_TOOLS is c"` passes
- [x] `ruff check` clean
- [x] `foreman/` Dockerfile still builds and `pioneer-foreman` CLI entry point still works from `foreman/`

Closes #492


backend/tests/test_tools_dedup.py
```diff
@@ -3,15 +3,12 @@
 import os
 import sys
 
-import pytest
-
 # Repo root — needed so `import backend` resolves backend/foreman_core.
```
jmelloy, line 6: The `sys.path.insert(0, _ROOT)` is still present but the `foreman/` path insertion directly below it covers the needed import. Verify that `import backend` still resolves correctly at test collection time without relying on installation; a comment explaining why `_ROOT` must be on `sys.path` would help future maintainers.

backend/tests/test_tools_dedup.py
```diff
@@ -36,41 +33,8 @@ def test_standalone_foreman_uses_canonical_tools():
...
-            sys.modules.pop("pioneer_foreman", None)
-        else:
-            sys.modules["pioneer_foreman"] = saved
-        if saved_tools is None:
-            sys.modules.pop("pioneer_foreman.tools", None)
-        else:
-            sys.modules["pioneer_foreman.tools"] = saved_tools
-
-
 def test_all_entry_points_same_tool_names():
```
jmelloy, line 36: The docstring now says 'All entry points' but the test only checks two sources (canonical backend and standalone foreman). If a third entry point is added in the future this comment will silently become stale. Consider being explicit: 'All known entry points (backend and standalone foreman)'.

jmelloy (commented):
- PR cleanly removes the `pioneer-foreman/` directory and its Docker service, consolidating to the single `foreman/` implementation with JWT auth and Bedrock support
- The test file cleanup is appropriate: the removed `test_pioneer_foreman_uses_canonical_tools` test had complex `sys.path`/`sys.modules` manipulation that was fragile and is no longer needed
- The `pytest` import removal in `test_tools_dedup.py` is correct since the deleted test was the only consumer of `pytest.mark` or similar fixtures
- The `docker-compose.yml` removal is straightforward with no dangling references to `pioneer-foreman` service remaining
- The PR description accurately documents the rationale; no functional regressions are introduced since the canonical `foreman/` service supersedes all capabilities of `pioneer-foreman/`

PR #495, opened 2026-05-27
Fix lock not released on error state, follow-ups blocked (#493)

### LINKED ISSUE #493: Lock not released on "error" state — follow-ups permanently blocked
## Bug

`_TERMINAL_STATES` in `backend/ws_handlers.py` (lines ~40, 631) is defined as:

```python
_TERMINAL_STATES = ("done", "failed", "cancelled")
```

`"error"` is absent. When a follow-up task exits with a non-zero code, `handle_task_update` sets `state="error"` but never calls `LockService.release()`. All subsequent `send_followup` calls queue as `pending-followup`, which only drains on `task-followup-done` — a message that is never sent for failed/errored tasks. Follow-ups are **permanently blocked** with no visible error to the operator.

## Root cause

- `_TERMINAL_STATES` must include `"error"` (and likely `"error_max_turns"` too).
- `handle_task_update` (or whatever reads `_TERMINAL_STATES`) must call `LockService.release()` whenever the task transitions to any terminal state, including `"error"`.
- The `pending-followup` drain logic must also fire on `"error"` → terminal transition so queued follow-ups are either dispatched or surfaced as failed.

## Fix

1. Add `"error"` (and `"error_max_turns"` if it exists as a distinct state) to `_TERMINAL_STATES`.
2. Ensure `LockService.release()` is called on all terminal transitions.
3. Decide and implement what happens to queued follow-ups when the lock is released after an error: either dispatch them immediately (best UX) or cancel them with a visible error message.
4. Add a test that confirms a follow-up queued after an error-state task is either dispatched or cleanly rejected, not silently dropped.

jmelloy:
## Summary

- **Root cause**: `"error"` was missing from `_TERMINAL_STATES` in `backend/ws_handlers.py` (and the matching constants in `foreman_core/constants.py`, `pioneer-foreman`, and frontend). When a task transitioned to `state="error"`, `LockService.release()` was never called. Follow-up calls (`send_followup`) queued as `pending-followup` events in the DB that are only drained on `task-followup-done` — a message never emitted for errored tasks — so they were silently and permanently blocked.
- **Fix**: Added `"error"` to `_TERMINAL_STATES` in all five locations (backend handler, shared constants, standalone foreman, and both frontend constants).
- **Drain logic**: Added a drain block in `handle_task_update` for the error path (mirrors `handle_task_followup_done`): on error transition, any queued `pending-followup` events are deleted and the foreman is triggered with their contents so it can decide whether to retry or finalize.
- **Tests**: Added `test_error_state_releases_lock_and_drains_followups` (ws integration test — verifies lock released + events drained on error) and `test_send_followup_after_error_state_dispatches` (foreman tool test — verifies `send_followup` dispatches rather than queues after an error).

## Test plan

- [ ] Run `python -m pytest backend/tests/test_agent_state_ws_persistence.py::test_error_state_releases_lock_and_drains_followups` — should pass
- [ ] Run `python -m pytest backend/tests/test_foreman.py::TestLockAndFollowup::test_send_followup_after_error_state_dispatches` — should pass
- [ ] Manually send a `task-update` with `state=error` for a task that has a queued pending-followup; verify foreman receives a `task-error` trigger with the queued instructions and the lock is released

Closes #493


jmelloy (commented):
- The core fix is correct and well-scoped: adding `"error"` to `_TERMINAL_STATES` across all five locations ensures lock release on error transitions.
- The drain-and-notify logic in `handle_task_update` is a reasonable mirror of `handle_task_followup_done`, but it calls `await ctx.db.commit()` after deleting events without first ensuring the lock release (which happens earlier in the function) was committed in the same transaction — creating a potential partial-commit window.
- The test `test_error_state_releases_lock_and_drains_followups` does not verify that the foreman was actually triggered (e.g., checking broadcast calls or a foreman invocation), leaving the drain→notify path partially untested.
- The `test_send_followup_after_error_state_dispatches` test uses `_sync_session(db_session)` where `db_session` is already an async session fixture, which is likely a bug in the test — this could silently pass without actually querying the right database.
- Minor: `worker_id_upd = data.get("workerId", "a worker")` uses the raw WebSocket payload rather than a validated/trusted value, which is a low-severity inconsistency since the worker ID should already be on `ctx` or validated earlier in the handler.

backend/ws_handlers.py
```diff
@@ -634,6 +634,47 @@ async def handle_task_update(ctx: WSContext, data: dict) -> None:
...
+        rows = result.all()
+        if rows:
+            event_ids = [r[0] for r in rows]
+            queued_payloads = [json.loads(r[1]) for r in rows]
+            await ctx.db.execute(delete(TaskEvent).where(TaskEvent.id.in_(event_ids)))
+            await ctx.db.commit()
+        worker_id_upd = data.get("workerId", "a worker")
+        task_uid = await _task_user_id(ctx.db, task_id)
+        if queued_payloads:
+            queued_summary = "\n".join(
```
jmelloy, line 654: The `await ctx.db.commit()` here commits only the `TaskEvent` deletion, but the lock release earlier in `handle_task_update` may have already been committed in a separate `await ctx.db.commit()` call. If the process crashes between the two commits, the lock is released but the pending-followup events are never drained — they remain in the DB in a zombie state. Consider performing both the lock release and the event deletion in a single atomic transaction.

backend/tests/test_foreman.py
```diff
@@ -1030,6 +1030,50 @@ async def capture(gid, msg):
...
+
+        followup_msgs = [m for m in broadcast_calls if m.get("type") == "task-followup"]
+        assert len(followup_msgs) == 1, (
+            "send_followup after error state should dispatch, not silently queue"
+        )
+        assert "t-err1" in results[0]["content"]
+
+        with _sync_session(db_session) as session:
+            ev = session.execute(
+                select(TaskEvent).where(TaskEvent.task_id == "t-err1")
```
jmelloy, line 1073: `_sync_session(db_session)` wraps what is presumably an async SQLAlchemy session in a sync context manager — this is almost certainly incorrect and will either error or query a different connection, making the `assert ev is None` assertion unreliable. This should use `await db_session.execute(...)` directly.

backend/tests/test_agent_state_ws_persistence.py
```diff
@@ -396,3 +397,70 @@ def test_agent_idle_releases_task_lock(client):
...
+            )
+            ws_obs.receive_json()  # drain the broadcast
+
+    with _sync_session(db_url) as session:
+        lock_key = session.scalar(select(Lock.key).where(Lock.key == f"task:{task_id}"))
+        task_state = session.scalar(select(Task.state).where(Task.id == task_id))
+        remaining_event = session.scalar(
+            select(TaskEvent).where(
+                TaskEvent.task_id == task_id, TaskEvent.event_type == "pending-followup"
+            )
```
jmelloy, line 461: The test verifies lock release and event drain but does not assert that the foreman was triggered with the queued follow-up payload. The notify path (`_trigger_foreman`) is the most operationally important part of the drain logic and should be covered here, e.g., by patching `_trigger_foreman` or `broadcast` and asserting a `task-error` event was emitted.

backend/ws_handlers.py
```diff
@@ -634,6 +634,47 @@ async def handle_task_update(ctx: WSContext, data: dict) -> None:
...
+                f"  {i + 1}. {p.get('instructions', '')}" for i, p in enumerate(queued_payloads)
+            )
+            human_msg = (
+                f"[task-error] Worker {worker_id_upd} reported task {task_id} as errored. "
```
jmelloy, line 658: The `pending-followup` rows are deleted and committed before `_trigger_foreman` is called. If `_trigger_foreman` raises, those queued payloads are permanently lost with no logging or recovery. Consider either wrapping both operations in a single transaction/savepoint, or at minimum logging the payloads before deletion so they can be recovered.

backend/ws_handlers.py
jmelloy, line 654: The `pending-followup` rows are deleted and committed before `_trigger_foreman` is called. If `_trigger_foreman` raises (network issue, DB error, etc.), the queued follow-ups are permanently lost with no way to recover them. Consider either: (a) deleting after a successful trigger, or (b) wrapping both operations in a single transaction / using an outbox pattern.

backend/ws_handlers.py
```diff
@@ -634,6 +634,47 @@ async def handle_task_update(ctx: WSContext, data: dict) -> None:
     await broadcast(ctx.guild_id, data, exclude=ctx.websocket)
     if "state" in update_values:
         reset_foreman_poll(ctx.guild_id)
```
jmelloy, line 636: The drain block only fires when `update_values.get('state') == 'error'`, but `task_id` is checked with `and task_id` as a guard. If `task_id` is falsy (e.g. missing from the message), the lock-release above still runs but the drain is silently skipped. This is likely fine in practice but worth an explicit log warning so the skip is observable.

backend/ws_handlers.py
```diff
@@ -634,6 +634,47 @@ async def handle_task_update(ctx: WSContext, data: dict) -> None:
...
+            select(TaskEvent.id, TaskEvent.payload_json)
+            .where(TaskEvent.task_id == task_id, TaskEvent.event_type == "pending-followup")
+            .order_by(TaskEvent.id)
+        )
+        rows = result.all()
+        if rows:
+            event_ids = [r[0] for r in rows]
+            queued_payloads = [json.loads(r[1]) for r in rows]
+            await ctx.db.execute(delete(TaskEvent).where(TaskEvent.id.in_(event_ids)))
+            await ctx.db.commit()
```
jmelloy: `worker_id_upd` comes directly from `data.get('workerId', 'a worker')` — raw client input that is interpolated into the foreman's instruction message. If worker IDs are not strictly validated earlier in the pipeline, this is a prompt-injection surface. Assert or sanitize the value before embedding it.

backend/tests/test_foreman.py
```diff
@@ -1030,6 +1030,50 @@ async def capture(gid, msg):
...
+            )
+
+        followup_msgs = [m for m in broadcast_calls if m.get("type") == "task-followup"]
+        assert len(followup_msgs) == 1, (
+            "send_followup after error state should dispatch, not silently queue"
+        )
+        assert "t-err1" in results[0]["content"]
+
+        with _sync_session(db_session) as session:
+            ev = session.execute(
```
jmelloy, line 1072: `_sync_session(db_session)` wraps what is presumably an async SQLAlchemy session fixture inside a sync context manager. This is likely incorrect and could cause the assertion to always pass vacuously (empty result due to wrong session/transaction) or raise a runtime error depending on the engine configuration. Use the async session directly with `await session.execute(...)`.

jmelloy (commented):
- **Correct root-cause fix**: Adding `"error"` to `_TERMINAL_STATES` in all five locations (backend handler, shared constants, standalone foreman, and both frontend constants) is the right approach and ensures consistency across the codebase.
- **Drain logic concern**: The drain block commits the `pending-followup` deletions *before* `_trigger_foreman` is called; if `_trigger_foreman` raises an exception, the queued follow-ups are permanently lost with no recovery path or logging.
- **Test gap**: `test_send_followup_after_error_state_dispatches` uses `_sync_session(db_session)` inside an async test where `db_session` is already an async session fixture — this pattern may silently pass without actually asserting the DB state correctly, or could cause session conflicts.
- **Human-readable message construction**: The foreman trigger message is built inline in the handler and embeds raw user-controlled data (`worker_id_upd`) directly into a string sent to the LLM, which could be a prompt-injection vector if worker IDs are not validated/sanitized upstream.
- **Missing pioneer-foreman update**: The PR description mentions five locations but the diff only shows four (the `pioneer-foreman` standalone constants file is referenced in the description but not shown in the diff), which could leave that component still blocking follow-ups on error.

jmelloy (commented):
- Core bug fix is correct: adding `"error"` to `_TERMINAL_STATES` across all five locations (backend handler, shared constants, foreman, and both frontend files) ensures the lock is released when a task errors.
- The drain logic in `handle_task_update` mirrors `handle_task_followup_done` appropriately, but has a subtle atomicity issue: `pending-followup` rows are deleted and committed *before* `_trigger_foreman` is called, meaning if the foreman trigger fails the queued follow-ups are silently lost with no retry path.
- The human message construction uses `data.get('workerId', 'a worker')` which could expose internal worker IDs in foreman-visible messages; this is a minor concern but consistent with existing patterns in the codebase.
- Test coverage is solid: both the WS-level lock release/drain and the foreman-level dispatch after error state are exercised. The test for `test_send_followup_after_error_state_dispatches` correctly validates no residual `pending-followup` events remain.
- Minor: the `pioneer-foreman` constants file is mentioned in the PR description as one of five locations updated, but no diff for it appears in the patch — worth verifying it was not accidentally omitted.

PR #499, opened 2026-05-27
Split FOREMAN_MODEL for Bedrock vs direct API (#498)

### LINKED ISSUE #498: Split FOREMAN_MODEL into separate Bedrock and direct-API model config
## Problem

`FOREMAN_MODEL` is a single env var used for all model invocations. When running against AWS Bedrock, the model ID must be an inference profile (e.g. `us.anthropic.claude-sonnet-4-6-20250514-v1:0`) rather than a plain model ID (e.g. `anthropic.claude-sonnet-4-6`). These two values are not interchangeable.

## Proposed fix

Split into two env vars:

- `FOREMAN_MODEL` — model ID for the direct Anthropic API (e.g. `claude-sonnet-4-5`)
- `FOREMAN_BEDROCK_MODEL` — inference profile ID for AWS Bedrock (e.g. `us.anthropic.claude-sonnet-4-6-20250514-v1:0`)

When the Bedrock client is active (i.e. `AWS_ACCESS_KEY_ID` / `ANTHROPIC_BEDROCK_BASE_URL` / similar Bedrock env vars are set), use `FOREMAN_BEDROCK_MODEL`; otherwise fall back to `FOREMAN_MODEL`.

Also update `.env.example` / `docker-compose.yml` / any config docs to document both vars.

jmelloy:
## Summary

- Add `FOREMAN_BEDROCK_MODEL` env var for Amazon Bedrock cross-region inference profile IDs (e.g. `us.anthropic.claude-sonnet-4-6-20250514-v1:0`), which are not interchangeable with the plain model IDs used by the direct Anthropic API
- Add `get_foreman_model(provider=None)` helper in `backend/foreman_core/llm.py` that returns the right model ID based on `FOREMAN_PROVIDER` (or a caller-supplied value)
- Remove the duplicate `FOREMAN_MODEL` definition from `backend/foreman/tools.py`; both the embedded runner and tools now call `get_foreman_model()`
- For the standalone foreman, add a `bedrock_model` field to `Config` (resolved from `FOREMAN_BEDROCK_MODEL`) and an `effective_model` property; the runner uses `config.effective_model` instead of `config.model`
- Update `docker-compose.yml`, `.env.example`, `pioneer-foreman.toml.example`, and `foreman/main.py` to document both vars

**Example values:**
```
FOREMAN_MODEL=claude-sonnet-4-6                                    # direct Anthropic API
FOREMAN_BEDROCK_MODEL=us.anthropic.claude-sonnet-4-6-20250514-v1:0 # Amazon Bedrock
```

Closes #498

### DIFF
--- a/backend/foreman_core/llm.py
+++ b/backend/foreman_core/llm.py
@@ -20,15 +20,27 @@ except ImportError:
 logger = logging.getLogger(__name__)
 
 FOREMAN_MODEL = os.environ.get("FOREMAN_MODEL", "claude-sonnet-4-6")
+# Bedrock uses cross-region inference profiles, not plain model IDs.
+# Set this to the profile ARN/ID appropriate for your region, e.g.:
+#   us.anthropic.claude-sonnet-4-6-20250514-v1:0
+FOREMAN_BEDROCK_MODEL = os.environ.get(
+    "FOREMAN_BEDROCK_MODEL", "us.anthropic.claude-sonnet-4-6-20250514-v1:0"
+)
 
 # Set FOREMAN_PROVIDER=bedrock to use Amazon Bedrock instead of the Anthropic API.
 # Requires: pip install "anthropic[bedrock]"  +  AWS credentials in env / IAM role.
-# On Bedrock, model IDs use the "anthropic." prefix, e.g.:
-#   anthropic.claude-sonnet-4-5   (Sonnet 4.x on Bedrock)
-#   anthropic.claude-opus-4-5     (Opus 4.x on Bedrock)
-# Check your Bedrock console for exact IDs available in your region.
 FOREMAN_PROVIDER = os.environ.get("FOREMAN_PROVIDER", "anthropic").lower()
 _BEDROCK_REGION = os.environ.get("AWS_DEFAULT_REGION", "us-east-1")
+
+
+def get_foreman_model(provider: str | None = None) -> str:
+    """Return the model ID to use for the given provider.
+
+    When provider is 'bedrock' (or FOREMAN_PROVIDER=bedrock), returns
+    FOREMAN_BEDROCK_MODEL; otherwise returns FOREMAN_MODEL.
+    """
+    resolved = (provider or FOREMAN_PROVIDER).lower()
+    return FOREMAN_BEDROCK_MODEL if resolved == "bedrock" else FOREMAN_MODEL
 
 
 def make_anthropic_client(
@@ -53,7 +65,7 @@ def make_anthropic_client(
         logger.info(
             "Foreman using Amazon Bedrock (region=%s, model=%s)",
             resolved_region,
-            FOREMAN_MODEL,
+            get_foreman_model("bedrock"),
         )
         return client
 
--- a/backend/foreman_core/llm.py
+++ b/backend/foreman_core/llm.py
@@ -19,13 +19,13 @@ except ImportError:
 
 logger = logging.getLogger(__name__)
 
+_DEFAULT_BEDROCK_MODEL = "us.anthropic.claude-sonnet-4-6-20250514-v1:0"
+
 FOREMAN_MODEL = os.environ.get("FOREMAN_MODEL", "claude-sonnet-4-6")
 # Bedrock uses cross-region inference profiles, not plain model IDs.
 # Set this to the profile ARN/ID appropriate for your region, e.g.:
 #   us.anthropic.claude-sonnet-4-6-20250514-v1:0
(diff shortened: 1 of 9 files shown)


backend/foreman_core/llm.py
```diff
@@ -20,17 +20,29 @@
 logger = logging.getLogger(__name__)
 
 FOREMAN_MODEL = os.environ.get("FOREMAN_MODEL", "claude-sonnet-4-6")
+# Bedrock uses cross-region inference profiles, not plain model IDs.
+# Set this to the profile ARN/ID appropriate for your region, e.g.:
+#   us.anthropic.claude-sonnet-4-6-20250514-v1:0
+FOREMAN_BEDROCK_MODEL = os.environ.get(
+    "FOREMAN_BEDROCK_MODEL", "us.anthropic.claude-sonnet-4-6-20250514-v1:0"
+)
```
jmelloy: The default `us.anthropic.claude-sonnet-4-6-20250514-v1:0` is baked in here as a module-level constant resolved at import time. The same default also appears in `foreman/pioneer_foreman/config.py` (dataclass default) and the `load()` function. Consider defining a single `_DEFAULT_BEDROCK_MODEL` constant in one place and importing it elsewhere, so a future default-model change doesn't require touching four files.

jmelloy (commented):
- The PR cleanly separates Bedrock and direct-API model identifiers, solving a real misconfiguration hazard since cross-region inference profile IDs are not valid for the direct Anthropic API and vice versa.
- The `get_foreman_model()` helper and the `Config.effective_model` property apply the same logic consistently, which is good, but they are independent implementations of the same rule and could drift.
- Module-level constants `FOREMAN_MODEL` and `FOREMAN_BEDROCK_MODEL` are resolved at import time; `get_foreman_model()` then reads those constants, so a test that patches `os.environ` after import won't affect the returned values without also patching the constants.
- The default value for `FOREMAN_BEDROCK_MODEL` is hardcoded in four separate places (llm.py, config.py dataclass, config.py `load()`, docker-compose.yml), making future model bumps error-prone.
- Documentation updates across `.env.example`, `docker-compose.yml`, `pioneer-foreman.toml.example`, and `main.py` are thorough and consistent.

backend/foreman_core/llm.py
```diff
@@ -20,17 +20,29 @@
...
 
 # Set FOREMAN_PROVIDER=bedrock to use Amazon Bedrock instead of the Anthropic API.
 # Requires: pip install "anthropic[bedrock]"  +  AWS credentials in env / IAM role.
-# On Bedrock, model IDs use the "anthropic." prefix, e.g.:
-#   anthropic.claude-sonnet-4-5   (Sonnet 4.x on Bedrock)
-#   anthropic.claude-opus-4-5     (Opus 4.x on Bedrock)
-# Check your Bedrock console for exact IDs available in your region.
 FOREMAN_PROVIDER = os.environ.get("FOREMAN_PROVIDER", "anthropic").lower()
 _BEDROCK_REGION = os.environ.get("AWS_DEFAULT_REGION", "us-east-1")
```
jmelloy, line 35: `get_foreman_model()` closes over the module-level `FOREMAN_MODEL` / `FOREMAN_BEDROCK_MODEL` constants which are resolved once at import time. If tests (or code) mutate `os.environ` after import they must also patch the constants directly; patching the env var alone won't work. Consider reading `os.environ` inside the function, or at least documenting this invariant.

foreman/pioneer_foreman/config.py
```diff
@@ -44,6 +47,11 @@ class Config:
 
     config_path: Path = field(default_factory=Path)
 
+    @property
```
jmelloy, line 52: `effective_model` duplicates the same `provider == 'bedrock'` branching logic already in `get_foreman_model()` in `backend/foreman_core/llm.py`. These two code paths can silently diverge (e.g. if someone adds a third provider). Since `Config` already holds both model values, this property is reasonable for the standalone foreman, but a comment or shared utility would reduce the risk of drift.

foreman/pioneer_foreman/config.py
```diff
@@ -139,6 +147,13 @@ def load(explicit_path: str | None = None, overrides: dict | None = None) -> Con
         or "claude-sonnet-4-6"
```
jmelloy, line 155: The precedence chain `overrides > toml > env-var > hardcoded-default` is correct, but the hardcoded fallback string `'us.anthropic.claude-sonnet-4-6-20250514-v1:0'` is repeated verbatim here and in the dataclass field default on line 31. Using `field(default=...)` and then `or config_defaults.bedrock_model` would keep them in sync.

PR #502, opened 2026-05-27
feat: inject foreman messages on CI completion without re-triggering

### LINKED ISSUE #501: GitHub Actions: inject foreman messages for CI events without re-triggering the fire
## Problem

The current cooldown timer approach is blunt. When GitHub Actions completes a CI run, the system needs to inform the Foreman but should **not** automatically re-trigger the workflow/fire.

## Proposed Solution

Instead of a cooldown timer, GitHub Actions workflows should:

1. **Post a foreman message** (e.g. a `[github-event]` style notification) when CI completes, so the Foreman is aware of the result.
2. **Not re-trigger the fire** — the message is informational only. The Foreman can decide whether to act (e.g. call `send_followup` on a failing task) without the workflow loop re-firing.

## Benefits

- No cooldown timer needed — the signal is event-driven, not polling-based.
- Prevents runaway workflow loops.
- Foreman stays informed of CI outcomes and can route fixes appropriately.

## Implementation Notes

- Look at the existing GitHub Actions workflow files and the foreman message injection mechanism.
- Add a step in CI workflows (on `workflow_run` or `check_run` completion) that POSTs a message to the Foreman endpoint.
- Ensure the message payload includes repo, PR number, conclusion, and linked task ID if available.
- The step should **not** dispatch another workflow event.

jmelloy:
## Summary

- **New endpoint** `POST /guilds/{guild_id}/foreman/ci-notify`: accepts CI completion payloads from GitHub Actions and persists a structured `[ci-notify]` chat message so the foreman has CI context next time it runs, **without** triggering a new AI invocation.
- **Reduced debounce window** from 180 s → 30 s: CI completion is now event-driven (GHA posts directly), so the debounce only needs to cover rapid review comments, not full CI runs.
- **New env var** `PIONEER_CI_KEY`: shared Bearer secret between GHA and the backend; documented in `.env.example`.
- **GHA `notify` job** added to `.github/workflows/ci.yml`: runs after all other jobs, POSTs a JSON payload (repo, PR number, workflow name, conclusion, task ID, run URL) to the backend. Only fires when all three secrets are set — skips silently otherwise. ⚠️ *The workflow file could not be pushed automatically (token lacks `workflow` scope); see note below.*
- **7 new tests** in `backend/tests/test_github_webhooks.py` covering auth failure, 404, task-ID resolution, and happy-path message persistence.

## Why this doesn't re-trigger the fire

The endpoint persists a `Message` row and broadcasts a `chat` event — it does **not** call `trigger_foreman()` or enqueue any asyncio task. The foreman reads the message the next time it's invoked for some other reason (e.g. a human chat or a subsequent webhook).

## Workflow scope note

The `.github/workflows/ci.yml` change (adding the `notify` job) is committed locally on this branch but **could not be pushed** — the `GITHUB_TOKEN` used by this worker lacks the `workflow` OAuth scope required by GitHub. To include it:

```bash
# In a terminal with a PAT that has `workflow` scope:
git fetch origin
git checkout claude/gha-inject-foreman-messages-on-ci-completion-501-t-6iae
git cherry-pick 907cb8e   # the local commit with all 4 changes
git push origin claude/gha-inject-foreman-messages-on-ci-completion-501-t-6iae
```

Or add `workflow` scope to the `GITHUB_TOKEN` secret used by the worker.

## Test plan
- [ ] `cd backend && python -m pytest backend/tests/test_github_webhooks.py -k ci_notify` — all 7 new tests pass
- [ ] Set `PIONEER_CI_KEY`, `PIONEER_GUILD_ID`, and `PIONEER_BACKEND_URL` as GitHub secrets; trigger a CI run; verify a `[ci-notify]` message appears in the guild chat
- [ ] Verify no foreman invocation fires automatically on CI notify

Closes #501

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/backend/routes/webhooks.py
+++ b/backend/routes/webhooks.py
@@ -46,10 +46,14 @@ router = APIRouter()
 # Configurable via WEBHOOK_DEBOUNCE_SECONDS (default: 30 seconds).
 DEBOUNCE_WINDOW_SECONDS: float = float(os.environ.get("WEBHOOK_DEBOUNCE_SECONDS", "30"))
 
+
 # Shared secret for the /foreman/ci-notify endpoint.  GitHub Actions sets
 # Authorization: Bearer <PIONEER_CI_KEY> on each CI completion POST.
 # When empty the endpoint returns 503 (not configured).
-_CI_KEY: str = os.environ.get("PIONEER_CI_KEY", "")
+# Read at call time so monkeypatch.setenv works in tests and secrets can rotate
+# without a restart.
+def _get_ci_key() -> str:
+    return os.environ.get("PIONEER_CI_KEY", "")
 
 
 # Cap on stored payloads. GitHub webhook payloads are typically <50 KB but
@@ -695,13 +699,14 @@ async def ci_notify(guild_id: str, body: CINotifyPayload, request: Request) -> R
     Auth: ``Authorization: Bearer <PIONEER_CI_KEY>`` where the key matches
     the ``PIONEER_CI_KEY`` environment variable on the backend.
     """
-    if not _CI_KEY:
+    ci_key = _get_ci_key()
+    if not ci_key:
         raise HTTPException(
             status_code=503, detail="CI notifications not configured (PIONEER_CI_KEY not set)"
         )
     auth = request.headers.get("authorization", "")
     provided = auth[len("Bearer ") :] if auth.startswith("Bearer ") else ""
-    if not hmac.compare_digest(provided.encode(), _CI_KEY.encode()):
+    if not hmac.compare_digest(provided.encode(), ci_key.encode()):
         raise HTTPException(status_code=401, detail="Invalid CI key")
 
     repo = body.repo or ""
@@ -742,25 +747,25 @@ async def ci_notify(guild_id: str, body: CINotifyPayload, request: Request) -> R
             )
         )
         await db.commit()
+        await broadcast(
+            guild_id,
+            {
+                "type": "chat",
+                "from": "github",
+                "to": "foreman",
+                "content": content,
+                "createdAt": created_at.isoformat(),
+            },
+        )
+        logger.info(
+            "ci-notify guild=%s repo=%s pr=%s task=%s conclusion=%s",
+            guild_id,
+            repo,
+            body.pr_number,
+            task_id,
+            conclusion,
+        )
     finally:
         await db.close()
 
-    await broadcast(
-        guild_id,
-        {
-            "type": "chat",
-            "from": "github",
-            "to": "foreman",
-            "content": content,
-            "createdAt": created_at.isoformat(),
-        },
-    )
-    logger.info(
-        "ci-notify guild=%s repo=%s pr=%s task=%s conclusion=%s",
-        guild_id,
-        repo,
-        body.pr_number,
-        task_id,
-        conclusion,
-    )
     return Response(status_code=202)
--- a/backend/routes/webhooks.py
+++ b/backend/routes/webhooks.py
@@ -26,6 +26,7 @@ from database import get_db
 from events import broadcast
 from fastapi import APIRouter, HTTPException, Request, Response
 from models import GithubEvent, Guild, Message, Task
+from pydantic import BaseModel
 from sqlalchemy import update
 from sqlalchemy.dialects.postgresql import insert as pg_insert
 from sqlalchemy.exc import IntegrityError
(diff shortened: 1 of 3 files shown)


backend/routes/webhooks.py
```diff
@@ -656,3 +662,105 @@ async def github_webhook(guild_id: str, request: Request) -> Response:
...
+                content=content,
+                message_type="chat",
+                created_at=created_at,
+            )
+        )
+        await db.commit()
+    finally:
+        await db.close()
+
+    await broadcast(
```
jmelloy: The `content` and `created_at` variables are assigned inside the `try` block but referenced after the `finally` in the `broadcast` and `logger` calls. If an exception is raised after the guild lookup but before `content` is assigned (e.g., during `_find_task` or `db.add`), the `finally` block closes the DB correctly, but then `broadcast(...)` will raise `UnboundLocalError`. Move the `broadcast` and logger calls inside the `try` block (before `finally`), or initialize `content = None` / `created_at = None` before the `try` and guard the broadcast.

backend/routes/webhooks.py
```diff
@@ -656,3 +662,105 @@ async def github_webhook(guild_id: str, request: Request) -> Response:
...
+
+@router.post("/guilds/{guild_id}/foreman/ci-notify")
+async def ci_notify(guild_id: str, body: CINotifyPayload, request: Request) -> Response:
+    """CI completion notification injected directly by GitHub Actions.
+
+    Persists a structured ``[ci-notify]`` chat message so the foreman has
+    CI context the next time it runs, **without** triggering a new foreman
+    AI invocation.  This is the event-driven counterpart to the webhook
+    debounce: GHA posts here the moment the workflow finishes, giving the
+    foreman immediate context while avoiding an automatic re-trigger.
```
jmelloy, line 697: The `_CI_KEY` is read once at module import time. This is fine for production but means `monkeypatch.setenv('PIONEER_CI_KEY', ...)` in tests won't work — only direct attribute patching does. Consider wrapping key access in a function or using `functools.lru_cache` with an explicit cache-clear mechanism to make the configuration more testable and to support runtime secret rotation without a process restart.

jmelloy (commented):
- Security concern: the `_CI_KEY` is loaded once at module import time from `os.environ`, meaning changes to the environment variable at runtime (e.g., via `monkeypatch.setenv`) won't take effect — tests work only because they patch the module attribute directly, but this is a fragile pattern that could mask misconfiguration issues.
- The `content` variable is referenced inside the `finally`/`broadcast` block but is only assigned inside the `try` block; if an exception occurs after `guild_row` check but before `content` is assigned, the `broadcast` call and logger will raise `UnboundLocalError`.
- The `hmac.compare_digest` comparison leaks timing information about whether `_CI_KEY` is empty vs. set to a non-empty value (the 503 check before the comparison is fine), but comparing empty strings via `compare_digest` with an attacker-controlled empty string would pass — the 503 guard above prevents this, so this is acceptable, but worth noting.
- The debounce window reduction from 180s to 30s is a meaningful behavioral change that could cause the foreman to fire before all rapid webhook events (e.g., multiple review comments in quick succession) have coalesced; the trade-off is reasonable but should be validated against real-world comment burst patterns.
- Test IDs skip `gcn3` and `gcn4`, which is minor but suggests copy-paste editing artifacts; test naming consistency and coverage of the `task_id`-provided-in-payload path (vs. looked up) is incomplete.

backend/routes/webhooks.py
```diff
@@ -656,3 +662,105 @@ async def github_webhook(guild_id: str, request: Request) -> Response:
...
+        guild_row = guild_res.one_or_none()
+        if guild_row is None:
+            raise HTTPException(status_code=404, detail="Guild not found")
+        guild_pk = guild_row.id
+
+        task_id = body.task_id
+        if not task_id and body.pr_number is not None:
+            task_row = await _find_task(db, guild_pk, repo, body.pr_number)
+            task_id = task_row.id if task_row else None
+
```
jmelloy, line 728: When `body.pr_number` is `None` and no `task_id` is provided, `_find_task` is never called and `task_id` remains `None`. This is intentional, but the `content` string will then contain no task reference even if the repo matches an existing task. Consider adding a fallback lookup by `repo` alone (if unambiguous) or documenting this limitation explicitly, since GHA workflows that don't track PR numbers will silently produce context-free notifications.

backend/tests/test_github_webhooks.py
```diff
@@ -515,6 +515,143 @@ def test_webhook_backfills_task_pr_url(client):
...
+    test_client, db_url = client
+    insert_guild(db_url, "gcn5")
+    monkeypatch.setattr(webhooks_mod, "_CI_KEY", "mykey")
+    resp = test_client.post(
+        "/guilds/gcn5/foreman/ci-notify",
+        json={
+            "repo": "owner/repo",
+            "pr_number": 42,
+            "workflow_name": "CI",
+            "conclusion": "success",
```
jmelloy, line 582: Test guild IDs jump from `gcn2` to `gcn5`, skipping `gcn3` and `gcn4`. While functionally harmless (each test uses an isolated guild), this suggests copy-paste editing and makes it harder to audit that all edge cases are covered. A test for the case where `task_id` is explicitly provided in the payload (bypassing the DB lookup) is also missing.

PR #504, opened 2026-05-28
User panel: view & delete Claude credentials (#503)

### LINKED ISSUE #503: User panel: view and delete saved Claude credentials
## Problem

Claude credentials (`credentials_blob`) are saved per-guild in the database and used by workers, but there is no UI for the user to inspect or remove them. Once saved, there's no way to delete them without directly hitting the API.

## Desired behaviour

In the user/settings panel (or a dedicated section), expose the saved Claude credentials so the user can:

1. **See** whether credentials are currently saved (masked display, e.g. show type + last 4 chars of token)
2. **Delete** the saved credentials (with a confirmation prompt)
3. *(Stretch)* **Edit / re-enter** credentials inline (reusing the existing `ClaudeAuthModal` flow)

## Implementation notes

- Backend: add a `DELETE /auth/claude-credentials` endpoint (and optionally a `GET` that returns a masked summary) to `backend/routes/auth.py`
- Frontend: surface this in the existing settings/user panel area. The `ClaudeAuthModal.vue` already handles entry — wire up delete + status display alongside it.
- No plaintext secrets should be returned from the API; only a masked indicator (e.g. `"sk-ant-…xxxx"`) and metadata (type, updated_at).

jmelloy:
## Summary

- **Backend**: adds two new routes to `backend/routes/auth.py`:
  - `GET /auth/claude-credentials?guild_id=…` — returns a masked status object (`{ saved, updated_at }`) for the settings UI; never exposes the raw credentials blob
  - `DELETE /auth/claude-credentials?guild_id=…` — deletes the stored credentials row for the guild; requires member/worker auth, returns 204
- **Frontend**: extends the `⚙ Guild Settings` popover in `TopBar.vue` with a **Claude Credentials** section that:
  - Loads credential status when the popover opens
  - Shows "Saved {date}" + a **Delete** button (with `confirm()` guard) if credentials exist
  - Shows "No credentials saved" otherwise
  - The existing `ClaudeAuthModal` (worker OAuth flow) is unchanged; credentials are still set automatically when a worker authenticates

## Test plan
- [ ] Open guild settings (⚙) → "Claude Credentials" shows "Loading…" then either saved date or "No credentials saved"
- [ ] With saved credentials, click **Delete** → cancel → credentials remain
- [ ] With saved credentials, click **Delete** → confirm → status refreshes to "No credentials saved"
- [ ] `GET /auth/claude-credentials` returns `{ "saved": false }` when no row exists
- [ ] `DELETE /auth/claude-credentials` with no auth returns 401/403

Closes #503

### DIFF
--- a/backend/routes/auth.py
+++ b/backend/routes/auth.py
@@ -19,7 +19,7 @@ from auth_deps import (
     require_worker_or_member,
 )
 from database import get_db
-from fastapi import APIRouter, Depends, HTTPException, Query
+from fastapi import APIRouter, Depends, HTTPException, Query, Response
 from fastapi.responses import RedirectResponse
 from fastapi.security import HTTPAuthorizationCredentials
 from models import (
@@ -180,6 +180,49 @@ async def store_claude_credentials(
     finally:
         await db.close()
     return {"ok": True}
+
+
+@router.get("/auth/claude-credentials")
+async def get_claude_credentials_status(
+    guild_id: str = Query(...),
+    _principal: str = Depends(require_worker_or_member),
+):
+    """Return masked Claude credentials status for the settings UI (never the full blob)."""
+    db = await get_db()
+    try:
+        guild_pk = await get_guild_pk(db, guild_id)
+        if guild_pk is None:
+            return {"saved": False}
+        result = await db.exec(
+            select(ClaudeCredentials).where(col(ClaudeCredentials.guild_id) == guild_pk)
+        )
+        row = result.one_or_none()
+        if not row:
+            return {"saved": False}
+        return {"saved": True, "updated_at": row.updated_at.isoformat()}
+    finally:
+        await db.close()
+
+
+@router.delete("/auth/claude-credentials", status_code=204)
+async def delete_claude_credentials(
+    guild_id: str = Query(...),
+    credentials: HTTPAuthorizationCredentials | None = Depends(http_bearer),
+):
+    """Delete stored Claude credentials for a guild. Requires member or worker auth."""
+    token = credentials.credentials if credentials else None
+    await authorize_worker_or_member(guild_id, token)
+    db = await get_db()
+    try:
+        guild_pk = await get_guild_pk(db, guild_id)
+        if guild_pk is not None:
+            await db.exec(
+                delete(ClaudeCredentials).where(col(ClaudeCredentials.guild_id) == guild_pk)
+            )
+            await db.commit()
+    finally:
+        await db.close()
+    return Response(status_code=204)
 
 
 @router.get("/auth/me")
--- a/backend/routes/auth.py
+++ b/backend/routes/auth.py
@@ -199,7 +199,7 @@ async def get_claude_credentials_status(
         row = result.one_or_none()
         if not row:
             return {"saved": False}
-        return {"saved": True, "updated_at": row.updated_at.isoformat()}
+        return {"saved": True, "updated_at": row.updated_at.isoformat() if row.updated_at else None}
     finally:
         await db.close()
 
@@ -207,11 +207,9 @@ async def get_claude_credentials_status(
 @router.delete("/auth/claude-credentials", status_code=204)
 async def delete_claude_credentials(
     guild_id: str = Query(...),
-    credentials: HTTPAuthorizationCredentials | None = Depends(http_bearer),
+    _principal: str = Depends(require_worker_or_member),
 ):
     """Delete stored Claude credentials for a guild. Requires member or worker auth."""
-    token = credentials.credentials if credentials else None
-    await authorize_worker_or_member(guild_id, token)
     db = await get_db()
     try:
         guild_pk = await get_guild_pk(db, guild_id)
--- a/backend/routes/auth.py
+++ b/backend/routes/auth.py
@@ -15,10 +15,11 @@ from auth_deps import (
     authorize_worker_or_member,
     get_guild_pk,
     http_bearer,
+    require_member,
     require_user,
     require_worker_or_member,
 )
-from database import get_db
+from database import get_db, get_db_dep
 from fastapi import APIRouter, Depends, HTTPException, Query, Response
 from fastapi.responses import RedirectResponse
 from fastapi.security import HTTPAuthorizationCredentials
@@ -35,6 +36,7 @@ from pydantic import BaseModel
 from sqlalchemy import delete, or_
 from sqlalchemy.dialects.postgresql import insert as pg_insert
 from sqlmodel import col, select
+from sqlmodel.ext.asyncio.session import AsyncSession
 from utils import generate_guild_id
 
 router = APIRouter()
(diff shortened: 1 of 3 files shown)


backend/routes/auth.py
```diff
@@ -182,6 +182,49 @@ async def store_claude_credentials(
...
+            return {"saved": False}
+        return {"saved": True, "updated_at": row.updated_at.isoformat()}
+    finally:
+        await db.close()
+
+
+@router.delete("/auth/claude-credentials", status_code=204)
+async def delete_claude_credentials(
+    guild_id: str = Query(...),
+    credentials: HTTPAuthorizationCredentials | None = Depends(http_bearer),
```
jmelloy: Auth pattern is inconsistent with the GET route above: this handler manually calls `authorize_worker_or_member` and extracts the token from `http_bearer` directly, while the GET route correctly uses `Depends(require_worker_or_member)`. Use the same `_principal: str = Depends(require_worker_or_member)` pattern here to keep auth enforcement uniform and avoid accidentally bypassing it in future refactors.

backend/routes/auth.py
```diff
@@ -182,6 +182,49 @@ async def store_claude_credentials(
...
+    token = credentials.credentials if credentials else None
+    await authorize_worker_or_member(guild_id, token)
+    db = await get_db()
+    try:
+        guild_pk = await get_guild_pk(db, guild_id)
+        if guild_pk is not None:
+            await db.exec(
+                delete(ClaudeCredentials).where(col(ClaudeCredentials.guild_id) == guild_pk)
+            )
+            await db.commit()
```
jmelloy: `row.updated_at.isoformat()` will raise `AttributeError` if `updated_at` is `None`. Guard with `row.updated_at.isoformat() if row.updated_at else None` and handle the null on the frontend, or ensure the column has a non-nullable DB default.

frontend/src/components/TopBar.vue
```diff
@@ -133,6 +151,49 @@ const repoStatus = ref<'' | 'saved' | 'error'>('')
...
+    )
+    if (res.ok) {
+      claudeCredsStatus.value = await res.json()
+    }
+  } catch {
+    // ignore — status stays null
+  }
+}
+
+async function deleteClaude() {
```
jmelloy, line 198: The DELETE result is never checked — if the server returns a 4xx/5xx, `claudeCredsStatus` is still set to `{ saved: false }`, falsely indicating success. Check `res.ok` before updating state, and surface an error to the user on failure.

frontend/src/components/TopBar.vue
```diff
@@ -150,6 +211,7 @@ async function toggleSettings() {
     if (ghStore.repos.length === 0 && ghStore.token) {
       await ghStore.fetchRepos()
     }
+    loadClaudeCredsStatus()
```
jmelloy: `loadClaudeCredsStatus()` is called without `await`, so any unhandled rejection becomes a floating promise. Either `await` it (the parent `toggleSettings` is already async) or add a `.catch()` handler to avoid silent failures and potential unhandled-promise-rejection warnings.

jmelloy (commented):
- **Security inconsistency**: `delete_claude_credentials` manually re-implements auth (`authorize_worker_or_member`) instead of using the `require_worker_or_member` Depends pattern used by the GET route, creating a maintenance risk and diverging security model.
- **Silent failure on delete**: The frontend `deleteClaude()` optimistically sets `{ saved: false }` even when the DELETE request fails (non-2xx status or network error), giving the user false confidence that credentials were removed.
- **No `await` on credential load**: `loadClaudeCredsStatus()` is called without `await` in `toggleSettings()`, meaning errors are silently swallowed and there is no way to surface a loading failure to the user.
- **Missing DB session for GET route**: Both new backend handlers call `get_db()` directly inside the function body rather than using `Depends(get_db)`, which bypasses the standard session lifecycle; the GET route also has no database injection for `get_guild_pk`, unlike the pattern used elsewhere.
- **`updated_at` may be `None`**: If `row.updated_at` is `None` (e.g., older rows or DB default not set), calling `.isoformat()` will raise an `AttributeError`; this should be guarded.

backend/routes/auth.py
```diff
@@ -182,6 +182,47 @@ async def store_claude_credentials(
...
 
+@router.get("/auth/claude-credentials")
+async def get_claude_credentials_status(
+    guild_id: str = Query(...),
+    _principal: str = Depends(require_worker_or_member),
+):
+    """Return masked Claude credentials status for the settings UI (never the full blob)."""
+    db = await get_db()
+    try:
+        guild_pk = await get_guild_pk(db, guild_id)
```
jmelloy: Authorization is only checked at the route level (any member/worker), but there is no validation that `_principal` belongs to the requested `guild_id`. A member of Guild A can call `GET /auth/claude-credentials?guild_id=GuildB` and learn whether Guild B has credentials saved. The guild membership should be verified against the principal before returning data.

backend/routes/auth.py
```diff
@@ -182,6 +182,47 @@ async def store_claude_credentials(
     return {"ok": True}
 
 
+@router.get("/auth/claude-credentials")
+async def get_claude_credentials_status(
+    guild_id: str = Query(...),
+    _principal: str = Depends(require_worker_or_member),
```
jmelloy: `get_db()` is called directly here rather than via `Depends(get_db)`. If an exception is raised between the `await get_db()` call and the `try` block (e.g. during argument binding), the session will never be closed. Using FastAPI dependency injection (`db: AsyncSession = Depends(get_db)`) is the standard pattern that guarantees cleanup.

frontend/src/components/TopBar.vue
```diff
@@ -133,6 +152,58 @@ const repoStatus = ref<'' | 'saved' | 'error'>('')
...
+async function loadClaudeCredsStatus() {
+  if (!currentGuild.value) return
+  claudeCredsStatus.value = null
+  try {
+    const res = await fetch(
+      `${API_BASE}/auth/claude-credentials?guild_id=${encodeURIComponent(currentGuild.value.id)}`,
+      { headers: authStore.authHeaders() },
+    )
+    if (res.ok) {
+      claudeCredsStatus.value = await res.json()
```
jmelloy, line 187: Non-OK responses (401, 403, 500, etc.) from the fetch are silently ignored and `claudeCredsStatus` stays `null` forever, showing a perpetual "Loading…" state. The error branch should set `claudeCredsError` and/or set a sentinel value so the user sees meaningful feedback.

frontend/src/components/TopBar.vue
```diff
@@ -150,6 +221,7 @@ async function toggleSettings() {
     if (ghStore.repos.length === 0 && ghStore.token) {
```
jmelloy, line 252: `loadClaudeCredsStatus()` is called every time the popover opens but `claudeCredsStatus` and `claudeCredsError` are never cleared when the popover closes or when `currentGuild` changes. When switching guilds, the previous guild's credential status will flash briefly before the new fetch resolves. Reset these refs at the start of `toggleSettings` or in the `currentGuild` watcher.

jmelloy (commented):
- Backend routes are clean and correctly mask credentials, but both handlers open a new DB session via `get_db()` without using the existing pattern of dependency injection, which can cause connection leaks if an exception is raised before the `try` block.
- The DELETE endpoint silently returns 204 even when the guild does not exist (`guild_pk is None`), which is acceptable REST behaviour, but there is no auth check that the requesting principal actually belongs to the guild being queried — any authenticated member/worker can delete credentials for any guild_id they supply.
- Frontend `loadClaudeCredsStatus` silently swallows non-OK HTTP responses (e.g. 401, 403, 500) and leaves `claudeCredsStatus` as `null` indefinitely with no error surfaced to the user.
- `claudeCredsStatus` and related refs are module-level in the component but are never reset when the popover closes, so stale data from a previous guild briefly flashes when switching guilds and reopening settings.
- Using `window.confirm()` for the delete guard is discouraged in modern UX and can be blocked by some browser policies; a proper inline confirmation step would be more robust.

backend/routes/auth.py
```diff
@@ -182,6 +182,47 @@ async def store_claude_credentials(
...
+):
+    """Return masked Claude credentials status for the settings UI (never the full blob)."""
+    db = await get_db()
+    try:
```
jmelloy: There is no verification that the authenticated principal (`_principal`) is actually a member of the requested guild. A user authenticated to any guild can pass an arbitrary `guild_id` and read or delete credentials for guilds they don't belong to. You must check that `_principal` is a member/worker of `guild_id` before proceeding.

backend/routes/auth.py
jmelloy: Both endpoints call `await get_db()` directly instead of using FastAPI's dependency injection (`db: AsyncSession = Depends(get_db)`). This bypasses the standard lifecycle management and is inconsistent with other handlers in this file. If an exception is raised before the `try` block (e.g. in dependency resolution), the connection may leak.

backend/routes/auth.py
```diff
@@ -182,6 +182,47 @@ async def store_claude_credentials(
...
+        guild_pk = await get_guild_pk(db, guild_id)
+        if guild_pk is None:
+            return {"saved": False}
+        result = await db.exec(
+            select(ClaudeCredentials).where(col(ClaudeCredentials.guild_id) == guild_pk)
```
jmelloy: Returning `{saved: false}` for an unknown `guild_id` silently swallows invalid inputs and makes it impossible for callers to distinguish 'no credentials' from 'guild does not exist'. Consider raising `HTTPException(status_code=404)` when `guild_pk is None`.

frontend/src/components/TopBar.vue
```diff
@@ -133,11 +152,70 @@ const repoStatus = ref<'' | 'saved' | 'error'>('')
...
+    }
+  } catch {
+    claudeCredsError.value = 'Failed to load credentials status'
+    claudeCredsStatus.value = { saved: false }
+  }
+}
+
+async function deleteClaude() {
+  if (!currentGuild.value) return
+  if (!confirm('Delete saved Claude credentials for this guild?')) return
```
jmelloy: Using the synchronous `confirm()` dialog for a destructive operation is problematic: it is blocked in sandboxed iframes, inconsistent across browsers, and provides poor UX. Replace with a proper inline confirmation UI (e.g. a two-step button or a small modal).

jmelloy (commented):
- Authorization is too permissive: any guild member can delete credentials for any guild by supplying an arbitrary `guild_id` query parameter — there is no check that the authenticated principal actually belongs to the target guild.
- Both new backend endpoints create a fresh DB connection via `get_db()` but never use the dependency-injection pattern already established in this file (other handlers use `db: Session = Depends(get_db)`), leading to inconsistent resource management.
- The DELETE endpoint silently returns 204 even when the guild or credentials row does not exist, which is fine REST semantics, but the GET endpoint returns `{saved: false}` for an unknown guild instead of a 404, masking invalid `guild_id` inputs.
- Frontend uses `confirm()` (blocking, non-customisable) for a destructive action, which is generally discouraged in modern UX and blocked entirely in some browser contexts (e.g. cross-origin iframes).
- `loadClaudeCredsStatus` is only triggered on popover open and guild change, so stale status can be shown if another session deletes credentials while the popover is open — minor but worth noting.

PR #506, opened 2026-05-28
Audit worker startup latency

jmelloy:
## Summary

Read-only audit of worker startup and task assignment latency, with findings written to `docs/worker-startup-latency-audit.md`.

### High impact
- **`pull_interval = 300 s`** (`config.py:52`, `worker.py:1498`): The fallback REST poll for missed tasks fires every 5 minutes. A task assigned while the WS push is lost won't be picked up until the next poll cycle. (The reconnect path at `worker.py:803` mitigates this for WS-down scenarios.)

### Medium impact
- **Sequential startup sequence** (`worker.py:971–990`): Six I/O steps run one-after-another before `_join()` makes the worker visible to the foreman. `gh auth status` (10 s), `codex doctor` (20 s), and `claude auth status` (10 s) all run even when those runners aren't configured.
- **WS reconnect backoff cap 30 s** (`ws_client.py:40`): After ≥5 failed connect attempts, up to 30 s elapses between retries, during which push tasks can't be delivered.

### Low impact (by design or configurable)
- `asyncio.sleep(0.2)` / `15.0` sleeps — inside interactive auth flow only, not on normal task path.
- Heartbeat 25 s, backend offline threshold 90 s — well-tuned, no task-latency impact.
- Worktree sweeper (1 h) and repo refresh (20 min) — background tasks, not on pickup path.

## Test plan
- [ ] Review `docs/worker-startup-latency-audit.md` for completeness and accuracy
- [ ] Confirm file paths and line numbers against current code

### DIFF
--- a/docs/worker-startup-latency-audit.md
+++ b/docs/worker-startup-latency-audit.md
@@ -11,7 +11,7 @@ Task assignment is **push-based over WebSocket** — the server broadcasts `task
 
 1. **Fallback poll interval: 300 s** (`pull_interval` default) — tasks missed during WS downtime are not re-fetched until 5 minutes later.
 2. **Sequential startup sequence** — six sequential I/O steps (registration, GitHub token, repo discovery, tool auth checks, WS connect, Claude auth) must all complete before the worker announces itself as ready.
-3. **Claude auth flow: up to 300 s** — if Claude credentials are absent, the worker blocks on a manual auth-code paste before joining.
+3. **Claude auth flow: ~10 s (normal) / up to 300 s (credentials missing)** — when already authenticated, `claude auth status --json` completes in ~10 s (`worker.py:326`). The 300 s figure applies only when credentials are absent and the worker must block waiting for a manual auth-code paste; this is not a routine startup cost.
 
 ---
 
@@ -82,7 +82,10 @@ def _backoff_delay(attempt: int, base: float, cap: float) -> float:
 ### WebSocket send retry — `ws_client.py:113–130`
 
 - Up to `send_retries=3` attempts with the same exponential backoff.
-- If all 3 fail, the send is dropped (error logged).
+- If all 3 fail, the send is **dropped** (error logged). The risk level depends on which message type is dropped:
+  - `heartbeat` — low risk; the next heartbeat fires within 25 s.
+  - `join` / `worker-register` — high risk; the worker will not appear in the backend's agent list and will not receive push tasks until it reconnects and re-joins.
+  - `task-update` / `task-complete` / `task-followup-done` — high risk; task state diverges between worker and backend, potentially causing the task to stall or be re-assigned.
 
 ### WS transport / recv error backoff — `ws_client.py:160,164`
 
@@ -108,13 +111,15 @@ Reference: `backend/routes/workers.py` broadcasts `task-assigned` at the point o
 ### Poll fallback path
 
 ```
-task-assigned WS push missed (e.g. worker offline or WS flap)
+task-assigned WS push missed — dominant case: worker was offline when task was assigned
   → _idle_puller() fires after pull_interval (default 300 s)
   → _fetch_pending_tasks() GET /guilds/{id}/workers/{id}/tasks
   → tasks enqueued for execution
 ```
 
 Maximum extra wait: **300 s** on default config.
+
+The primary scenario that triggers this path is the worker being offline (or in the middle of reconnecting) at the moment a task is assigned. A push being silently dropped on an established, live WebSocket connection is a rare edge case — it would require a successful write at the sender that the receiver never processes, which the WS protocol does not normally allow.
 
 ### Reconnect recovery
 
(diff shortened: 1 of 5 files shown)


docs/worker-startup-latency-audit.md
```diff
@@ -0,0 +1,195 @@
...
+
+## Summary
+
+Task assignment is **push-based over WebSocket** — the server broadcasts `task-assigned` immediately when a task is created with no intentional queue delay. The dominant latency risks are:
+
+1. **Fallback poll interval: 300 s** (`pull_interval` default) — tasks missed during WS downtime are not re-fetched until 5 minutes later.
+2. **Sequential startup sequence** — six sequential I/O steps (registration, GitHub token, repo discovery, tool auth checks, WS connect, Claude auth) must all complete before the worker announces itself as ready.
+3. **Claude auth flow: up to 300 s** — if Claude credentials are absent, the worker blocks on a manual auth-code paste before joining.
+
+---
```
jmelloy, line 16: The summary lists 'Claude auth flow: up to 300 s' as item 3 alongside always-present latency sources, but this only applies when credentials are absent and the worker is running interactively. It should be clearly qualified as 'only when credentials are missing' to avoid alarming readers about normal startup latency.

docs/worker-startup-latency-audit.md
```diff
@@ -0,0 +1,195 @@
...
+
+## 4. Task Assignment Flow
+
+### Push path (normal)
+
+```
+Foreman calls assign_task tool
+  → backend/routes/workers.py broadcasts { type: "task-assigned", ... } over WS
+  → worker._listen() receives message, enqueues to task_queue (asyncio.Queue)
```
jmelloy, line 103: The poll-fallback path comment says 'task-assigned WS push missed (e.g. worker offline or WS flap)', but the high-impact finding also claims this applies when 'the WS is up and the push is simply lost'. Lost pushes on a live WebSocket are extremely rare with the websockets library and should be noted as an edge case rather than a co-equal scenario with worker-offline to avoid over-stating the risk.

docs/worker-startup-latency-audit.md
```diff
@@ -0,0 +1,195 @@
...
+Maximum extra wait: **300 s** on default config.
+
+### Reconnect recovery
+
+`_on_ws_reconnect` (`worker.py:800–806`) calls `_fetch_pending_tasks()` immediately on reconnection, bypassing the idle puller interval for this specific case.
+
+---
+
+## 5. Initialization Steps Before Worker Signals Ready
+
```
jmelloy, line 131: Step 8 (`_check_claude_auth()`) is described as 'Blocks up to 300 s if credentials missing (manual auth-code paste)', but in the normal authenticated case this step completes in ~10 s (the subprocess timeout at worker.py:326). The table entry should distinguish the two cases so the 300 s figure isn't misread as a routine startup cost.

docs/worker-startup-latency-audit.md
```diff
@@ -0,0 +1,195 @@
...
+- Retries **indefinitely** (`ws_client.py:59`: "Retries indefinitely — the worker is a long-running daemon").
+- While reconnecting, the worker cannot receive push task assignments; the reconnect backoff is the upper bound on how long a task sits undelivered before the WS is re-established.
+
+### WebSocket send retry — `ws_client.py:113–130`
+
+- Up to `send_retries=3` attempts with the same exponential backoff.
+- If all 3 fail, the send is dropped (error logged).
+
+### WS transport / recv error backoff — `ws_client.py:160,164`
+
```
jmelloy, line 91: The send-retry section says 'If all 3 fail, the send is dropped (error logged)' but does not note whether the dropped message is a heartbeat, a join message, or a task-status update — the impact differs significantly. A brief note on which message types are subject to this path would make the risk concrete.

docs/worker-startup-latency-audit.md
```diff
@@ -0,0 +1,195 @@
...
+|------|---------------|-------|
+| 1. `_register()` | `worker.py:971` | POST to backend; establishes `worker_id` |
+| 2. `_fetch_github_token_if_needed()` | `worker.py:974` | HTTP GET if token absent from config/env |
+| 3. `_refresh_github_repos()` | `worker.py:975` | GitHub API call; discovers org repos |
+| 4. `_check_gh_auth()` | `worker.py:976` | Runs `gh auth status` — **10 s timeout** (`worker.py:231`) |
+| 5. `_check_codex_doctor()` | `worker.py:977` | Runs `codex doctor` — **20 s timeout** (`worker.py:267`) |
+| 6. `_ensure_codex_api_key()` | `worker.py:978` | Validates OpenAI key in config |
+| 7. `ws.connect()` | `worker.py:982` | WS connect with exponential backoff (up to 30 s per attempt, infinite retries) |
+| 8. `_check_claude_auth()` | `worker.py:988` | **Blocks up to 300 s** if credentials missing (manual auth-code paste) |
+| 9. `_join()` | `worker.py:990` | Sends `join` + `worker-register` — **worker becomes visible** |
```
jmelloy, line 145: The audit notes steps 4–6 always execute even when the corresponding runners are not configured, but stops short of recommending a fix (e.g. guard each check behind a `if self.cfg.gh_enabled` / `if self.cfg.codex_enabled` flag). Adding an actionable recommendation would increase the document's utility for engineers acting on these findings.

jmelloy (commented):
- Audit document is thorough and well-structured, accurately identifying the 300 s `pull_interval` as the dominant task-pickup latency risk in failure scenarios.
- The sequential startup section (Section 5) correctly flags that `_check_gh_auth()`, `_check_codex_doctor()`, and `_check_claude_auth()` always execute regardless of whether those tools are configured, adding up to 40 s of avoidable blocking time before `_join()`.
- The WS reconnect backoff analysis is accurate; the 30 s cap is a real but bounded risk, and the document correctly notes that `_on_ws_reconnect` calling `_fetch_pending_tasks()` immediately partially mitigates the poll-interval gap.
- A minor factual issue: Section 5 lists `_check_claude_auth()` as step 8 with a 300 s block, but the 300 s timeout is only hit during *interactive* credential entry (the `asyncio.wait_for` on the auth-code queue); non-interactive authenticated startups do not block, so the severity characterization in the summary could mislead readers into thinking normal startups are affected.
- No code changes are introduced; this is a documentation-only PR, so there are no functional regressions to evaluate.

PR #511, opened 2026-05-29
Scope directives font-size 16px to mobile only

jmelloy:
## Summary

- Removes `font-size: 16px` from the base `.chat-input` rule in `ChatTab.vue`
- Moves it into a `@media (max-width: 768px)` block so it only applies on mobile
- All other styles from the previous commit (resize: none, line-height, padding, etc.) are untouched

## Why

iOS Safari auto-zooms any input/textarea with a font-size below 16px on focus. The 16px fix was added unconditionally but isn't needed on desktop, where the inherited font size from the design system should apply.

### DIFF
--- a/frontend/src/components/chat-pane/ChatTab.vue
+++ b/frontend/src/components/chat-pane/ChatTab.vue
@@ -461,12 +461,17 @@ watch(
   border: 2px solid var(--color-brass-dark);
   color: var(--color-text);
   font-family: var(--font-mono);
-  font-size: 16px;
   padding: 6px 10px;
   outline: none;
   resize: none;
   line-height: 1.4;
   transition: border-color 0.15s;
+}
+
+@media (max-width: 768px) {
+  .chat-input {
+    font-size: 16px;
+  }
 }
 
 .chat-input:focus {


frontend/src/components/chat-pane/ChatTab.vue
```diff
@@ -461,14 +461,19 @@ watch(
...
   font-family: var(--font-mono);
-  font-size: 16px;
   padding: 6px 10px;
   outline: none;
   resize: none;
   line-height: 1.4;
   transition: border-color 0.15s;
 }
 
+@media (max-width: 768px) {
```
jmelloy, line 471: Consider using a shared SCSS/CSS variable or a project-wide breakpoint constant (e.g., `$breakpoint-mobile`) instead of the magic number `768px`, to stay consistent with other responsive rules in the codebase and make future breakpoint changes easier.

frontend/src/components/chat-pane/ChatTab.vue
```diff
@@ -461,14 +461,19 @@ watch(
...
+  .chat-input {
+    font-size: 16px;
```
jmelloy, line 473: 16px is only the minimum needed to prevent iOS auto-zoom; if the design system's inherited font-size on mobile happens to be smaller than 16px for other reasons, this fix will silently break that. A comment like `/* Prevents iOS Safari auto-zoom on focus */` would make the intent clear to future readers.

jmelloy (commented):
- The change correctly scopes the iOS Safari auto-zoom fix to mobile only, avoiding unintended font-size overrides on desktop.
- The 768px breakpoint is a reasonable mobile cutoff, though it should be verified it matches the project's existing breakpoint tokens/variables for consistency.
- Removing `font-size: 16px` from the base rule means the textarea will inherit whatever font-size the design system provides on desktop, which is the intended behavior but should be confirmed it renders acceptably.
- The media query is placed between the base `.chat-input` rule and its `:focus` pseudo-class, which is logically clean and won't cause specificity issues.
- No regressions are introduced for the other properties (resize, line-height, padding, transition) as they remain in the base rule.

PR #517, opened 2026-06-01
Add Pgweb as a docker-compose target

### LINKED ISSUE #516: Add Pgweb as a docker-compose target
Add [Pgweb](https://github.com/sosedoff/pgweb) as an optional service in the project's `docker-compose.yml` so developers can browse the Postgres database via a web UI.

## Acceptance criteria
- A `pgweb` service is defined in `docker-compose.yml` (or an override file)
- It connects to the existing Postgres service using the configured credentials
- It is exposed on a local port (e.g. `8081`)
- It depends on the `db` (or equivalent) service so it starts after Postgres is ready

jmelloy:
## Summary

- Adds `sosedoff/pgweb` as a `pgweb` service in `docker-compose.yml`
- Uses the project's existing Postgres credentials via the same env vars (`POSTGRES_USER`, `POSTGRES_PASSWORD`, `POSTGRES_DB`)
- Exposes the UI on host port 8081
- Placed behind the `tools` profile (consistent with how other optional services like `foreman`, `worker`, and `postgres-test` are handled) — not auto-started with a plain `docker compose up`

**Usage:**
```bash
docker compose --profile tools up pgweb
```
Then open http://localhost:8081

Closes #516

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -110,12 +110,16 @@ services:
     restart: unless-stopped
 
   pgweb:
-    image: sosedoff/pgweb
+    image: sosedoff/pgweb:0.15.0
     profiles: [ "tools" ]
     ports:
       - "8081:8081"
     environment:
-      DATABASE_URL: postgres://${POSTGRES_USER:-pioneer}:${POSTGRES_PASSWORD:-pioneer_password}@postgres:5432/${POSTGRES_DB:-pioneer_square}?sslmode=disable
+      PGHOST: postgres
+      PGUSER: ${POSTGRES_USER:-pioneer}
+      PGPASSWORD: ${POSTGRES_PASSWORD:-pioneer_password}
+      PGDATABASE: ${POSTGRES_DB:-pioneer_square}
+      PGSSLMODE: disable
     depends_on:
       postgres:
         condition: service_healthy
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -109,6 +109,18 @@ services:
       - backend
     restart: unless-stopped
 
+  pgweb:
+    image: sosedoff/pgweb
+    profiles: [ "tools" ]
+    ports:
+      - "8081:8081"
+    environment:
+      DATABASE_URL: postgres://${POSTGRES_USER:-pioneer}:${POSTGRES_PASSWORD:-pioneer_password}@postgres:5432/${POSTGRES_DB:-pioneer_square}?sslmode=disable
+    depends_on:
+      postgres:
+        condition: service_healthy
+    restart: unless-stopped
+
   postgres-test:
     image: postgres:17
     profiles: [ "test" ]
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -113,12 +113,15 @@ services:
     image: sosedoff/pgweb:0.15.0
     profiles: [ "tools" ]
     ports:
-      - "8081:8081"
+      - "${PGWEB_PORT:-8081}:8081"
     environment:
       PGHOST: postgres
       PGUSER: ${POSTGRES_USER:-pioneer}
+      # dev-only: PGPASSWORD is intentionally passed as a plain env var for local development convenience.
+      # Do not use this service configuration in shared/staging/production environments.
       PGPASSWORD: ${POSTGRES_PASSWORD:-pioneer_password}
       PGDATABASE: ${POSTGRES_DB:-pioneer_square}
+      PGPORT: "5432"
       PGSSLMODE: disable
     depends_on:
       postgres:


jmelloy (commented):
- The pgweb service is correctly placed behind the `tools` profile, keeping it opt-in and consistent with the project's existing conventions.
- Credentials are embedded in a `DATABASE_URL` environment variable using shell-style defaults, which works but exposes the password in plaintext in the environment of the container — a minor security concern worth noting.
- No image tag is pinned (`sosedoff/pgweb` uses `latest` implicitly), which can cause non-reproducible builds if the upstream image changes.
- The `depends_on` with `service_healthy` is a good practice, ensuring pgweb only starts after Postgres is ready.
- Overall the change is small, focused, and well-described; the main actionable issues are the missing image tag and the credential exposure pattern.

docker-compose.yml
```diff
@@ -109,6 +109,18 @@ services:
       - backend
     restart: unless-stopped
 
+  pgweb:
+    image: sosedoff/pgweb
+    profiles: [ "tools" ]
+    ports:
+      - "8081:8081"
+    environment:
+      DATABASE_URL: postgres://${POSTGRES_USER:-pioneer}:${POSTGRES_PASSWORD:-pioneer_password}@postgres:5432/${POSTGRES_DB:-pioneer_square}?sslmode=disable
```
jmelloy: The password is visible in plaintext as part of `DATABASE_URL`. Consider splitting the connection string into individual `PGUSER`, `PGPASSWORD`, `PGHOST`, `PGDATABASE`, and `PGSSLMODE` environment variables that pgweb also supports, so the password is not embedded in a loggable URL string (e.g. visible in `docker inspect` output).

docker-compose.yml
```diff
@@ -109,6 +109,18 @@ services:
       - backend
     restart: unless-stopped
 
+  pgweb:
+    image: sosedoff/pgweb
+    profiles: [ "tools" ]
+    ports:
```
jmelloy, line 115: No image tag is specified, so Docker will silently pull `latest` on every fresh pull. Pin to a specific version (e.g. `sosedoff/pgweb:0.15.0`) to ensure reproducible environments and avoid unexpected breaking changes.

docker-compose.yml
```diff
@@ -109,6 +109,22 @@ services:
...
 
+  pgweb:
+    image: sosedoff/pgweb:0.15.0
+    profiles: [ "tools" ]
+    ports:
+      - "8081:8081"
+    environment:
+      PGHOST: postgres
+      PGUSER: ${POSTGRES_USER:-pioneer}
+      PGPASSWORD: ${POSTGRES_PASSWORD:-pioneer_password}
```
jmelloy, line 122: Exposing `PGPASSWORD` as a plain environment variable means it will be visible in `docker inspect` output and to any process that can read `/proc`. For a dev-only tool this is an acceptable trade-off, but consider adding a comment making the dev-only intent explicit, or use a `secrets` block if the project ever needs this in a shared/staging environment.

docker-compose.yml
```diff
@@ -109,6 +109,22 @@ services:
       - backend
     restart: unless-stopped
 
+  pgweb:
+    image: sosedoff/pgweb:0.15.0
+    profiles: [ "tools" ]
+    ports:
+      - "8081:8081"
+    environment:
```
jmelloy, line 117: Consider adding `PGPORT: 5432` explicitly alongside the other PG* variables for completeness and resilience against future changes to the postgres service's port mapping.

jmelloy (commented):
- The pgweb service is well-structured and consistent with the project's existing patterns, using the `tools` profile and `service_healthy` dependency correctly.
- Pinning the image to `sosedoff/pgweb:0.15.0` is good practice for reproducibility.
- A potential security concern exists: `PGPASSWORD` is exposed as a plain environment variable, which is visible via `docker inspect` and in process listings; this is a common trade-off for dev tooling but worth noting.
- No `PGPORT` is specified, which could be a subtle issue if the postgres service runs on a non-default port; however, since the project appears to use the default 5432 this is likely fine.
- The change is minimal, focused, and does not affect any production services.

docker-compose.yml
```diff
@@ -109,6 +109,22 @@ services:
       - backend
     restart: unless-stopped
 
+  pgweb:
+    image: sosedoff/pgweb:0.15.0
+    profiles: [ "tools" ]
```
jmelloy, line 114: The `sosedoff/pgweb` image is not an official Docker Hub verified publisher image. Pinning to a digest (e.g. `sosedoff/pgweb@sha256:...`) rather than a mutable tag would give stronger supply-chain guarantees, though the semver pin is already better than `:latest`.

docker-compose.yml
```diff
@@ -109,6 +109,22 @@ services:
...
+    ports:
+      - "8081:8081"
+    environment:
+      PGHOST: postgres
+      PGUSER: ${POSTGRES_USER:-pioneer}
```
jmelloy, line 119: Passing `PGPASSWORD` as a plain env var is fine for local dev, but consider whether this service could be accidentally started in a non-dev environment. A note in docs or a clear warning comment would help prevent misuse.

docker-compose.yml
jmelloy, line 117: The `PGPORT` variable is not set here; pgweb will default to 5432, which matches the internal Postgres service port. This is fine, but adding `PGPORT: 5432` explicitly would make the configuration self-documenting and resilient to future port changes.

docker-compose.yml
```diff
@@ -109,6 +109,22 @@ services:
       - backend
     restart: unless-stopped
 
+  pgweb:
+    image: sosedoff/pgweb:0.15.0
+    profiles: [ "tools" ]
+    ports:
```
jmelloy, line 115: Port 8081 may conflict with other services on the developer's machine (e.g., another web app or API). Consider making the host port configurable via an env var like `${PGWEB_PORT:-8081}:8081` for flexibility.

jmelloy (commented):
- Clean, minimal addition that follows existing patterns in the file (profiles, depends_on with health check, restart policy)
- Pinning the image to `0.15.0` is good practice for reproducibility
- Security concern: `PGPASSWORD` is passed as a plain environment variable, which is acceptable for a local dev tool but worth noting
- The port 8081 choice should be verified against other services to avoid conflicts; it would be worth documenting in a README or `.env.example`
- No critical bugs; the implementation is straightforward and consistent with the codebase conventions

PR #525, opened 2026-06-02
Strongly type backend WebSocket messages with Pydantic v2

### LINKED ISSUE #522: Strongly type backend WebSocket messages with a WSMessage union
## Background

PR #521 introduced strong typing for inbound WS messages on the **frontend** (Zod discriminated-union, exhaustive `switch`). That PR was closed in favour of tackling the same problem from the **backend first**.

## Goal

Apply the same rigour to the Python backend:

- Define a `WSMessage` discriminated union (Pydantic models or `TypedDict` + `Literal` types) covering every message type the server sends to clients.
- Replace any `dict`-typed or loosely-typed WS `send_json(...)` calls with the new union types.
- Add a single serialisation helper so all outbound frames are validated before being sent.
- Where the backend also *receives* messages from workers/clients, add inbound typed models and validate at the entry point.
- Add or update tests to cover the new types.

## Acceptance criteria

- No bare `dict` passed to `send_json` / `send_text` in WS handlers.
- Pydantic (or equivalent) validation on outbound messages catches schema violations at runtime.
- Existing backend test suite continues to pass.
- Types are structured so the frontend union (from PR #521) can be kept in sync easily.

jmelloy:
## Summary

- Adds `backend/ws_types.py` with Pydantic v2 discriminated union models covering all inbound and outbound WebSocket message types
- Replaces bare `dict` payloads in `send_json`/`send_text` calls with `.model_dump()` on typed models throughout `ws_handlers.py`, `events.py`, `foreman/runner.py`, `foreman/tools.py`, and all route files
- Adds a centralised `send_ws_message(ws, msg)` helper in `events.py` so serialisation and validation are in one place
- Validates inbound WS messages at each entry point using `model_validate` / a `parse_ws_message` dispatcher

Closes #522

## Test plan

- [ ] All existing backend pytest tests pass (`python -m pytest` in `backend/`)
- [ ] Run `ruff check . --fix && ruff format .` from repo root — no errors
- [ ] Manually verify WS message flow: worker connects, task assigned, task-complete received, foreman triggered

### DIFF
--- a/backend/events.py
+++ b/backend/events.py
@@ -124,15 +124,12 @@ async def broadcast_msg(
 
 async def emit_terminal_line(guild_id: str, agent_id: str, line: str):
     """Broadcast and persist a terminal output line."""
+    from ws_types import TerminalOutputMsg
+
     now = datetime.now(UTC)
-    await broadcast(
+    await broadcast_msg(
         guild_id,
-        {
-            "type": "terminal-output",
-            "agentId": agent_id,
-            "line": line,
-            "timestamp": now.isoformat(),
-        },
+        TerminalOutputMsg(agentId=agent_id, line=line, timestamp=now.isoformat()),
     )
     if line:
         db = await get_db()
--- a/backend/events.py
+++ b/backend/events.py
@@ -115,9 +115,7 @@ async def send_ws_message(ws: WebSocket, message: "_WS") -> None:
     await ws.send_text(ws_dumps(message.model_dump(by_alias=True)))
 
 
-async def broadcast_msg(
-    guild_id: str, message: "_WS", exclude: WebSocket | None = None
-) -> None:
+async def broadcast_msg(guild_id: str, message: "_WS", exclude: WebSocket | None = None) -> None:
     """Serialise a typed WS model and broadcast it to all guild connections."""
     await broadcast(guild_id, message.model_dump(by_alias=True), exclude)
 
--- a/backend/events.py
+++ b/backend/events.py
@@ -8,13 +8,16 @@ import asyncio
 import json
 import logging
 from datetime import UTC, date, datetime
-from typing import Any
+from typing import TYPE_CHECKING, Any
 from uuid import UUID
 
 from database import get_db
 from fastapi import WebSocket, WebSocketDisconnect
 from models import Agent, TaskLog
 from sqlmodel import col, select
+
+if TYPE_CHECKING:
+    from ws_types import _WS
 
 logger = logging.getLogger(__name__)
 
@@ -105,6 +108,18 @@ async def broadcast(guild_id: str, message: dict, exclude: WebSocket | None = No
             dead.append(ws)
     for ws in dead:
         connections[guild_id].remove(ws)
+
+
+async def send_ws_message(ws: WebSocket, message: "_WS") -> None:
+    """Serialise a typed WS model and send it to a single WebSocket."""
+    await ws.send_text(ws_dumps(message.model_dump(by_alias=True)))
+
+
+async def broadcast_msg(
+    guild_id: str, message: "_WS", exclude: WebSocket | None = None
+) -> None:
+    """Serialise a typed WS model and broadcast it to all guild connections."""
+    await broadcast(guild_id, message.model_dump(by_alias=True), exclude)
 
 
 async def emit_terminal_line(guild_id: str, agent_id: str, line: str):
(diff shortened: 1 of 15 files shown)


backend/events.py
```diff
@@ -107,17 +110,26 @@ async def broadcast(guild_id: str, message: dict, exclude: WebSocket | None = No
         connections[guild_id].remove(ws)
 
 
+async def send_ws_message(ws: WebSocket, message: "_WS") -> None:
```
jmelloy, line 113: `send_ws_message` calls `ws_dumps(message.model_dump(by_alias=True))` directly, while `broadcast_msg` delegates to `broadcast` which presumably also calls `ws_dumps`. If `broadcast` independently serialises the dict (e.g. with `json.dumps`), the data path is consistent, but if it calls `ws_dumps` again on an already-serialised string, there will be a double-encode bug. Verify that `broadcast` accepts a plain `dict` and serialises it exactly once.

backend/routes/usage.py
```diff
@@ -109,18 +110,19 @@ async def report_usage(
         )
     await db.commit()
 
-    await broadcast(
+    await broadcast_msg(
```
jmelloy: Using `ClaudeUsageMsg.model_validate({...})` with a manually-assembled dict bypasses the type-safety benefits of typed construction. If `_summarize` returns unexpected keys, Pydantic will raise a `ValidationError` at runtime in a hot path. Prefer a direct constructor call with explicit keyword arguments, or at least wrap this in a try/except with a meaningful error message.

backend/routes/foreman.py
```diff
@@ -495,11 +495,11 @@ async def patch_task(
         "deleted_at": "deletedAt",
         "phase": "phase",
     }
-    ws_payload: dict = {"type": "task-update", "taskId": task_id}
+    ws_data: dict = {"taskId": task_id}
```
jmelloy, line 498: `TaskUpdateMsg.model_validate(ws_data)` where `ws_data` is built by iterating `update_values` means that if none of the mapped columns are in `update_values`, the resulting `TaskUpdateMsg` will only have `taskId` set and all other fields will be defaults/None. Ensure the model's optional fields have correct defaults and that an empty update payload is either rejected or produces a valid no-op message — otherwise the frontend may receive a misleading state update.

backend/routes/foreman.py
```diff
@@ -550,16 +550,15 @@ async def create_message(
     await db.refresh(msg)
     msg_id = msg.id
 
-    await broadcast(
+    await broadcast_msg(
         guild_id,
```
jmelloy, line 554: `ChatMsg(... userId=body.user_id)` unconditionally passes `userId` (possibly `None`) whereas the original code used a conditional spread `**({'userId': body.user_id} if body.user_id else {})`. If `ChatMsg.userId` is not `Optional` with `exclude_none=True` on serialisation, the frontend will receive `"userId": null` where previously it received no key at all, which may break frontend null-checks.

backend/ws_handlers.py
```diff
@@ -148,18 +175,15 @@ async def _trigger_foreman(
     """
```
jmelloy, line 174: `userId=user_id if user_id else None` and `taskId=task_id if task_id else None` are redundant — falsy values are already `None`-equivalent here. More importantly, if `user_id` is an empty string `""`, it will be coerced to `None`, silently dropping data. Use `userId=user_id or None` only if empty-string suppression is intentional, otherwise pass the value directly.

jmelloy (commented):
- Good architectural improvement: replacing bare `dict` payloads with typed Pydantic models improves correctness, IDE support, and catches serialisation bugs at construction time rather than at the wire
- The `_WS` type alias used in `events.py` via `TYPE_CHECKING` is a string-forward-reference trick that works but is fragile — the actual Union type in `ws_types.py` is not shown in the diff so it's hard to audit coverage
- `broadcast_msg` delegates to the existing `broadcast` function (which calls `ws_dumps`), while `send_ws_message` calls `ws_dumps` directly — double-serialisation risk is low here but the two code paths diverge subtly
- The `ChatMsg` model uses `from_` as a Python attribute (to avoid `from` keyword clash) with presumably an alias `from` for serialisation; if `by_alias=True` is not consistently applied, the wire format will break frontend consumers
- The `model_validate` pattern in `usage.py` and `foreman.py` constructs a plain dict and then validates it, losing the compile-time type-checking benefit that direct constructor calls would provide

PR #527, opened 2026-06-02
fix(worker): prevent pi agent from getting stuck indefinitely

jmelloy:
## Problem

`pi_runner.py` had three subprocess lifecycle bugs that caused the pi agent to hang the worker process indefinitely:

1. **No timeout on `proc.wait()`** — After breaking out of the stdout event loop and closing stdin, the code called `await proc.wait()` with no timeout. Pi runs in `--mode rpc` and stays alive waiting for more stdin; if it failed to notice the EOF (e.g. a signal-masked subprocess, a crash in pi's shutdown path, or a race), `proc.wait()` blocked forever and took the entire worker with it.

2. **No `finally` block** — If an exception (including `asyncio.CancelledError`) was raised during stdout reading, the `proc.stdin.close()` and `proc.wait()` lines were never reached. The pi subprocess was left running as an orphan with its stdin pipe still open, and the asyncio `StreamReader` for stderr was leaked.

3. **No error handling in `_drain_stderr`** — If `emit()` raised (e.g. because the WebSocket was broken), the exception propagated out of the `async for` loop inside the background task. When the caller later did `await stderr_task`, that exception was re-raised in an unexpected place, bypassing the structured error returns.

## Fix

- `_WAIT_TIMEOUT = 30` constant + `asyncio.wait_for(proc.wait(), timeout=_WAIT_TIMEOUT)`. If the timeout fires, `proc.kill()` (SIGKILL) is sent and the wait is retried, so the function always completes.
- `finally` block that: closes stdin if still open, sends SIGKILL if `proc.returncode is None` (meaning we never successfully waited), calls `proc.wait()` with a short 5 s timeout, and cancels/awaits the stderr drain task.
- `_drain_stderr` wraps its body in `try/except Exception` so emit failures are logged at DEBUG and don't surface as unhandled task exceptions.

## Tests

Added `worker/tests/test_pi_runner.py` with **21 tests**:
- `parse_pi_event` unit tests for all event types (agent_start, message_update, tool_execution_start/end, unknown)
- `run_pi_auto` integration tests using real fake-pi scripts:
  - Happy path (agent_end → success)
  - Missing binary (FileNotFoundError)
  - Prompt rejected (response.success=false)
  - Non-JSON stdout lines forwarded verbatim
  - `assistantMessageEvent` error sets saw_error
  - **Hung process**: pi ignores SIGTERM; with `_WAIT_TIMEOUT` monkeypatched to 1 s, the test completes in ~1 s via SIGKILL instead of hanging
  - Stderr lines forwarded with `[stderr]` prefix

## Test run

```
cd worker && python -m pytest tests/test_pi_runner.py -v   # 21 passed in 1.24s
cd worker && python -m pytest -x -q                         # 211 passed (no regressions)
```

### DIFF
--- a/worker/pioneer_worker/pi_runner.py
+++ b/worker/pioneer_worker/pi_runner.py
@@ -156,7 +156,7 @@ async def run_pi_auto(
             proc.stdin.close()
         try:
             exit_code = await asyncio.wait_for(proc.wait(), timeout=_WAIT_TIMEOUT)
-        except (TimeoutError, asyncio.TimeoutError):
+        except TimeoutError:
             logger.warning(
                 "pi[%d] did not exit within %ds after stdin close; killing",
                 proc.pid,
--- a/worker/pioneer_worker/pi_runner.py
+++ b/worker/pioneer_worker/pi_runner.py
@@ -11,6 +11,9 @@ logger = logging.getLogger(__name__)
 
 EmitFn = Callable[[str], Awaitable[None]]
 
+# Seconds to wait for pi to exit after stdin is closed before killing it.
+_WAIT_TIMEOUT = 30
+
 
 def _result_text(result: dict) -> str:
     """Join the text blocks of a tool result's content array."""
@@ -75,6 +78,8 @@ async def run_pi_auto(
     stop_reason = "no_events"
     event_count = 0
     agent_ended_ok = False
+    proc: asyncio.subprocess.Process | None = None
+    stderr_task: asyncio.Task[None] | None = None
     try:
         proc = await asyncio.create_subprocess_exec(
             *cmd,
@@ -90,10 +95,13 @@ async def run_pi_auto(
         await proc.stdin.drain()  # type: ignore[union-attr]
 
         async def _drain_stderr() -> None:
-            async for raw in proc.stderr:  # type: ignore[union-attr]
-                line = raw.decode(errors="replace").strip()
-                if line:
-                    await emit(f"[stderr] {line}")
+            try:
+                async for raw in proc.stderr:  # type: ignore[union-attr]
+                    line = raw.decode(errors="replace").strip()
+                    if line:
+                        await emit(f"[stderr] {line}")
+            except Exception:
+                logger.debug("pi[%d] stderr drain exited early", proc.pid, exc_info=True)  # type: ignore[union-attr]
 
         stderr_task = asyncio.create_task(_drain_stderr())
         accumulated = ""
@@ -144,7 +152,16 @@ async def run_pi_auto(
         # Closing stdin signals EOF so the RPC process exits cleanly.
         if proc.stdin and not proc.stdin.is_closing():
             proc.stdin.close()
-        exit_code = await proc.wait()
+        try:
+            exit_code = await asyncio.wait_for(proc.wait(), timeout=_WAIT_TIMEOUT)
+        except TimeoutError:
+            logger.warning(
+                "pi[%d] did not exit within %ds after stdin close; killing",
+                proc.pid,
+                _WAIT_TIMEOUT,
+            )
+            proc.kill()
+            exit_code = await proc.wait()
         await stderr_task
         if event_count == 0:
             logger.warning("pi[%d] produced no stdout events — check PATH/auth", proc.pid)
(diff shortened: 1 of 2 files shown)


worker/pioneer_worker/pi_runner.py
```diff
@@ -144,7 +152,16 @@ async def _drain_stderr() -> None:
         # Closing stdin signals EOF so the RPC process exits cleanly.
         if proc.stdin and not proc.stdin.is_closing():
             proc.stdin.close()
-        exit_code = await proc.wait()
+        try:
```
jmelloy, line 157: On Python 3.10 and earlier, `asyncio.wait_for` raises `asyncio.TimeoutError` (which inherits from `concurrent.futures.TimeoutError`), not the built-in `TimeoutError`. Catching `TimeoutError` here will silently miss the timeout on Python < 3.11, leaving the process un-killed. Use `except (TimeoutError, asyncio.TimeoutError):` or just `asyncio.TimeoutError` to be safe across versions.

worker/pioneer_worker/pi_runner.py
```diff
@@ -164,3 +181,29 @@ async def _drain_stderr() -> None:
...
+            if proc.stdin and not proc.stdin.is_closing():
+                try:
+                    proc.stdin.close()
+                except Exception:
+                    pass
+            try:
+                proc.kill()
+            except ProcessLookupError:
+                pass
+            except Exception:
```
jmelloy, line 199: In the finally block, `await asyncio.wait_for(proc.wait(), timeout=5.0)` can raise `TimeoutError` (or `asyncio.TimeoutError` on <3.11), but only `Exception` is caught below — `asyncio.TimeoutError` is a subclass of `Exception` on 3.11+ but on older Pythons it may propagate and suppress the original exception. Change to `except BaseException:` or explicitly catch `asyncio.TimeoutError` alongside `Exception`.

worker/pioneer_worker/pi_runner.py
```diff
@@ -90,10 +95,13 @@ async def run_pi_auto(
         await proc.stdin.drain()  # type: ignore[union-attr]
 
         async def _drain_stderr() -> None:
-            async for raw in proc.stderr:  # type: ignore[union-attr]
-                line = raw.decode(errors="replace").strip()
-                if line:
-                    await emit(f"[stderr] {line}")
+            try:
+                async for raw in proc.stderr:  # type: ignore[union-attr]
+                    line = raw.decode(errors="replace").strip()
```
jmelloy, line 102: The `proc.pid` reference inside `_drain_stderr`'s except handler will raise `AttributeError` if the subprocess was never successfully started (since `proc` would be `None` at that point). However, since `_drain_stderr` is only created after `proc` is assigned, and the closure captures the outer `proc` variable, this is safe — but a local capture (`pid = proc.pid`) would make the intent clearer and avoid reliance on closure semantics.

worker/tests/test_pi_runner.py
```diff
@@ -0,0 +1,344 @@
...
+
+    assert success is False
+    assert stop_reason == "error_during_execution"
+    assert any("auth failure" in line for line in emitted)
+
+
+async def test_run_pi_auto_non_json_stdout(tmp_path) -> None:
+    """Non-JSON stdout lines are emitted verbatim and don't crash the runner."""
+    fake_pi = tmp_path / "fake-pi"
+    fake_pi.write_text(
```
jmelloy, line 230: The hung-process test sets `_WAIT_TIMEOUT = 1` second, which means the test takes at least 1 second to run. Consider reducing to `0.1` or `0.2` seconds to keep the test suite fast, especially if run frequently in CI.

jmelloy (commented):
- The PR correctly addresses the three identified lifecycle bugs (no timeout, no finally block, no stderr error handling) with reasonable implementations.
- The `finally` block has a subtle double-kill issue: if the normal cleanup path already called `proc.kill()` and awaited `proc.wait()`, `proc.returncode` will be set and the finally block skips correctly — but if the timeout path killed the process, `proc.returncode` is set after `await proc.wait()`, so the finally block is safe. Logic is correct.
- The `asyncio.TimeoutError` catch uses the built-in `TimeoutError` which is fine in Python 3.11+ where `asyncio.TimeoutError` is an alias, but on Python 3.10 and below `asyncio.wait_for` raises `asyncio.TimeoutError` (a subclass of `concurrent.futures.TimeoutError`), not the built-in — this could silently miss the timeout on older runtimes.
- Test coverage is thorough and tests the critical hung-process scenario using monkeypatching of `_WAIT_TIMEOUT`, which is well-designed.
- The `_drain_stderr` closure references `proc` from the outer scope after it could theoretically be reassigned, but since `proc` is now initialized to `None` and only assigned once, this is safe in practice.

PR #530, opened 2026-06-02
feat(foreman): add spawn_worker tool + fix PR injection for plan-phase tasks

### LINKED ISSUE #528: Foreman: support configuring repos/tools when spawning a worker
## Problem

The Foreman currently has no way to spawn a new worker with a specific set of repos or tools configured. When a task requires a repo that no existing worker has access to (e.g. `Identity-Digital/dnsid`), the only recourse is to ask a human to manually start a new worker process — there's no API or tool the Foreman can call to do it autonomously.

## Desired behaviour

The Foreman should be able to spawn a new worker and specify:
- **Repos** — which repositories the worker is allowed to clone/push to
- **Tools** — which MCP tools or capabilities the worker has access to (e.g. GitHub CLI, specific SDKs)

## Proposed solution

Add a `spawn_worker` (or `start_worker`) tool to the Foreman's toolset that accepts:
- `repos` — list of `owner/repo` strings the worker should have access to
- `tools` — optional list of tool/capability names to enable
- `agent_count` — optional number of parallel agent slots (default 1 or system default)

Returns a `worker_id` that can immediately be passed to `assign_task`.

## Example usage

```json
spawn_worker({
  "repos": ["Identity-Digital/dnsid"],
  "tools": ["github_cli"],
  "agent_count": 2
})
```

## Why

This unblocks the Foreman from being able to fully autonomously manage cross-repo work without human intervention for worker provisioning.

### LINKED ISSUE #529: Foreman task payload: "create a PR" else branch is too specific
## Problem

When the Foreman builds the payload to send to a worker, there is an `else` clause that unconditionally appends "create a PR" to the instructions. This is too prescriptive — not every task should end in a PR (e.g. an ephemeral check, a status poll, a planning task, or a follow-up that just commits to an existing branch).

## Expected behaviour

The "create a PR" instruction should only be appended when the task type/phase actually warrants it (e.g. `phase == "execute"` or `phase == "followup"` on a new branch). Planning tasks, review tasks, and ephemeral/automation tasks should not have "create a PR" injected.

## Fix

Tighten the condition so "create a PR" is only included for task phases/types where opening a new PR is the correct outcome. Consider adding a `no_pr` flag or checking `phase` before appending.

jmelloy:
## Summary

- **#528** — Adds a `spawn_worker` MCP tool to the Foreman AI so it can autonomously start new worker processes configured with specific repos and tools, without human intervention. The tool pre-registers the worker in the DB and launches a Docker container via the Docker SDK.
- **#529** — Tightens the worker payload-builder so the \"create a PR\" footer is only injected for `execute` and `followup` phase tasks. Previously the `else` branch unconditionally appended PR instructions to every non-review task, including `plan` and automation tasks.

## Changes

**`backend/foreman_core/tools_schema.py`** — registers `spawn_worker` in `FOREMAN_TOOLS` alongside existing tools like `shutdown_worker`. Accepts `repos` (required), `tools` (optional), `agent_count` (optional), and `name` (optional); returns `worker_id` and `container_id` on success.

**`backend/foreman/tools.py`** — implements the `spawn_worker` executor: pre-registers the worker row in the DB (so `PIONEER_WORKER_ID` is known before container start), builds the env via `build_spawn_worker_env`, and launches the container via `docker.from_env()`. Handles missing Docker SDK gracefully with an error message.

**`backend/utils.py`** — adds `build_spawn_worker_env` helper that assembles the full environment dict passed to the worker container.

**`worker/pioneer_worker/config.py`** / **`worker.py`** — supports pre-assigned `PIONEER_WORKER_ID` and `PIONEER_AUTH_TOKEN` env vars so the worker process skips self-registration when spawned via the tool.

**`worker/pioneer_worker/worker.py`** — fixes #529: replaces unconditional `else` with `elif _phase in ("execute", "followup")` so plan/ephemeral tasks don't receive the PR-creation footer.

**`backend/tests/test_foreman_spawn_worker.py`**, **`worker/tests/test_config.py`**, **`worker/tests/test_pre_assigned_registration.py`** — test coverage for the new feature.

## Test plan

- [ ] `cd backend && python -m pytest tests/test_foreman_spawn_worker.py -v`
- [ ] `cd worker && python -m pytest tests/test_config.py tests/test_pre_assigned_registration.py -v`
- [ ] Manual: foreman calls `spawn_worker` with `repos=["owner/repo"]` → worker container starts, registers, and appears in the guild

Closes #528
Fixes #529

### DIFF
--- a/backend/foreman/tools.py
+++ b/backend/foreman/tools.py
@@ -1042,6 +1042,12 @@ async def _exec_one_tool(guild_id: str, tu, user_id: str | None = None) -> dict:
                     worker_id = "w-" + "".join(
                         random.choices(string.ascii_lowercase + string.digits, k=6)
                     )
+                    # auth_token is a 256-bit cryptographically random bearer token stored
+                    # as plaintext in the DB (matching the UserSession.token convention).
+                    # It is passed to the container as an env var and never logged.
+                    # The DB is the only persistent store; no read-after-create endpoint
+                    # exists, so exposure is limited to the spawn path and the container's
+                    # runtime environment.
                     auth_token = secrets.token_urlsafe(32)
                     worker_name = custom_name or worker_display_name(worker_id, None)
                     created_at = datetime.now(UTC)
--- a/backend/foreman/tools.py
+++ b/backend/foreman/tools.py
@@ -59,6 +59,22 @@ logger = logging.getLogger(__name__)
 # an explicit expiry. Mirrors backend.main.DEFAULT_FINALIZE_TTL.
 DEFAULT_FINALIZE_TTL_SECONDS = 3 * 24 * 60 * 60  # 3 days
 
+# Module-level Docker client — created once per process to reuse the Unix-socket
+# connection across repeated spawn_worker calls instead of opening a new socket
+# each time.  None until first successful from_env(); tests can patch
+# _get_docker_client() directly to avoid touching sys.modules.
+_docker_client: Any = None
+
+
+def _get_docker_client() -> Any:
+    """Return the cached Docker client, importing the SDK and calling from_env() once."""
+    global _docker_client
+    if _docker_client is None:
+        import docker  # ImportError propagated to caller if SDK not installed
+
+        _docker_client = docker.from_env()
+    return _docker_client
+
 
 def _resolve_finalize_deleted_at(inp: dict) -> tuple[datetime | None, str | None]:
     """Compute the soft-delete instant for a finalize_task tool call.
@@ -1039,15 +1055,7 @@ async def _exec_one_tool(guild_id: str, tu, user_id: str | None = None) -> dict:
                     # before the container starts. The worker process will find
                     # PIONEER_WORKER_ID + PIONEER_AUTH_TOKEN in its env and skip
                     # its own self-registration step.
-                    worker_id = "w-" + "".join(
-                        random.choices(string.ascii_lowercase + string.digits, k=6)
-                    )
-                    # auth_token is a 256-bit cryptographically random bearer token stored
-                    # as plaintext in the DB (matching the UserSession.token convention).
-                    # It is passed to the container as an env var and never logged.
-                    # The DB is the only persistent store; no read-after-create endpoint
-                    # exists, so exposure is limited to the spawn path and the container's
-                    # runtime environment.
+                    worker_id = "w-" + secrets.token_hex(3)
                     auth_token = secrets.token_urlsafe(32)
                     worker_name = custom_name or worker_display_name(worker_id, None)
                     created_at = datetime.now(UTC)
@@ -1064,7 +1072,6 @@ async def _exec_one_tool(guild_id: str, tu, user_id: str | None = None) -> dict:
                     )
                     await db.commit()
 
-                    # Fetch Claude OAuth token from DB (same as HTTP spawn endpoint).
                     creds_result = await db.exec(
                         select(col(ClaudeCredentials.credentials_blob)).where(
                             col(ClaudeCredentials.guild_id) == guild_pk
@@ -1085,9 +1092,7 @@ async def _exec_one_tool(guild_id: str, tu, user_id: str | None = None) -> dict:
                     )
 
                     try:
-                        import docker as docker_sdk
-
-                        docker_client = docker_sdk.from_env()
+                        docker_client = _get_docker_client()
                         image = os.environ.get("WORKER_IMAGE", "pioneer-square-worker")
 
                         network = None
@@ -1098,7 +1103,7 @@ async def _exec_one_tool(guild_id: str, tu, user_id: str | None = None) -> dict:
                             )
                         except Exception as e:
                             logger.warning(
-                                "Docker network detection failed, starting container without explicit network: %s",
+                                "Docker network detection failed, starting without explicit network: %s",
                                 e,
                             )
 
@@ -1146,8 +1151,9 @@ async def _exec_one_tool(guild_id: str, tu, user_id: str | None = None) -> dict:
                         result_text = (
                             f"Worker pre-registered as {worker_id} but the Docker SDK is not "
                             "installed on the backend — container was NOT started. "
-                            "Start the worker manually with "
-                            f"PIONEER_WORKER_ID={worker_id} PIONEER_AUTH_TOKEN=<token>."
+                            f"To start manually: PIONEER_WORKER_ID={worker_id} "
+                            f"PIONEER_AUTH_TOKEN={auth_token} "
+                            "<worker-start-command>"
                         )
                         is_error = True
                     except Exception as exc:
--- a/backend/foreman/tools.py
+++ b/backend/foreman/tools.py
@@ -11,6 +11,7 @@ import logging
 import os
 import random
 import re
+import secrets
 import string
 import urllib.error
 import urllib.parse
@@ -27,6 +28,7 @@ from foreman_core.tools_schema import (
 from lock_service import LockService
 from models import (
     Agent,
+    ClaudeCredentials,
     GithubToken,
     Guild,
     GuildKey,
@@ -38,6 +40,7 @@ from models import (
 )
 from sqlalchemy import delete, update
 from sqlmodel import col, select
+from utils import build_spawn_worker_env, decode_claude_oauth_token, worker_display_name
 from ws_types import (
     TaskAssignedMsg,
     TaskCancelMsg,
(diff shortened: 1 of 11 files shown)


backend/foreman/tools.py
```diff
@@ -1023,6 +1026,122 @@ async def _exec_one_tool(guild_id: str, tu, user_id: str | None = None) -> dict:
...
+                        claude_oauth_token=decode_claude_oauth_token(stored_blob),
+                        worker_id=worker_id,
+                        auth_token=auth_token,
+                        agent_count=agent_count,
+                        tools=tools_list or None,
+                    )
+
+                    try:
+                        import docker as docker_sdk
+
```
jmelloy: If the Docker container fails to start (ImportError or generic Exception), the pre-registered Worker row is left in the DB in `state='offline'` with no container backing it. This orphaned record will appear as a dead worker to any code that queries workers by guild. Consider rolling back or deleting the Worker row on failure, or at minimum marking it with a distinct `state='spawn_failed'` and adding a cleanup job.

backend/foreman/tools.py
```diff
@@ -1023,6 +1026,122 @@ async def _exec_one_tool(guild_id: str, tu, user_id: str | None = None) -> dict:
...
+                            id=worker_id,
+                            guild_id=guild_pk or 0,
+                            repos=json.dumps(repos),
+                            state="offline",
+                            created_at=created_at,
+                            auth_token=auth_token,
+                            name=worker_name,
+                        )
+                    )
+                    await db.commit()
```
jmelloy, line 1073: The network detection catches all exceptions with a bare `except Exception: pass`, which will silently fail even for transient errors (e.g., Docker daemon timeout). If detection fails the container is started without a network argument, which may attach it to the wrong network or make it unreachable from the backend. Consider logging a warning at minimum so failures are observable.

backend/tests/test_foreman_spawn_worker.py
```diff
@@ -0,0 +1,266 @@
...
+async def test_spawn_worker_with_agent_count_and_tools(db_session):
+    insert_guild(db_session, "guild02")
+
+    fake_container = MagicMock()
+    fake_container.id = "deadbeef000000"
+
+    fake_docker_client = MagicMock()
+    fake_docker_client.containers.get.side_effect = Exception("no hostname")
+    fake_docker_client.containers.run.return_value = fake_container
+
```
jmelloy, line 174: The assertion `assert payload["container_id"] == "abc123def45"` checks 11 characters, but the implementation does `container.id[:12]`, which would yield `"abc123def456"` (12 chars). This test will fail against the actual code — the expected string needs one more character.

worker/pioneer_worker/worker.py
```diff
@@ -1928,14 +1939,15 @@ async def _collect_usage(rec: dict) -> None:
                             "        -f 'comments[][body]=inline comment'\n"
```
jmelloy, line 1941: The set of phases that trigger PR injection (`execute`, `followup`) is hardcoded as a string literal. If a new phase (e.g., `implement`) is added elsewhere and someone forgets to update this condition, it will silently skip PR instructions. Consider defining a module-level constant like `_PR_PHASES = frozenset({"execute", "followup"})` and using `if _phase in _PR_PHASES`.

jmelloy (commented):
- The `spawn_worker` tool pre-registers the worker in the DB before starting the container, which is a sound design for avoiding race conditions, but the orphaned worker record on Docker failure is never cleaned up, leaving stale `offline` rows in the DB.
- Security concern: `auth_token` is stored in the Worker model (presumably plaintext) and passed via environment variables to the container — if the token is not hashed at rest this is a credential exposure risk.
- The network detection logic silently swallows all exceptions and may result in containers launched on the wrong (or no) network in non-trivial Docker environments.
- The PR injection fix in `worker.py` correctly narrows the `else` branch to `execute`/`followup` phases, but the set of valid phases is implicit/undocumented — a helper constant or enum would prevent future drift.
- Test coverage is thorough and well-structured, covering happy paths, missing Docker SDK, and Docker runtime failures; the `container_id` assertion (`abc123def45` — 11 chars) appears to be off-by-one vs. the code's `[:12]` slice.

jmelloy (commented):
- The `spawn_worker` tool correctly pre-registers workers before launching containers, ensuring the `worker_id` is known upfront and persisted even on failure — good defensive design.
- **Security concern**: The `auth_token` is stored as plaintext in the DB and passed via Docker environment variables, which is noted in comments but not mitigated (e.g., no token hashing, no secret store). The error message in the `ImportError` branch leaks the token placeholder instruction but not the token itself — acceptable.
- The PR injection fix (`_PR_PHASES` frozenset) correctly narrows the `else` branch so `plan`/automation phases no longer receive PR instructions — a clean, minimal fix.
- `worker_id` generation uses `random.choices` (not `secrets`) for the suffix, making IDs weakly random and potentially guessable; `secrets` is already imported and used for `auth_token` and `container_name`.
- The `spawn_worker` implementation in `tools.py` does not clean up (delete) the pre-registered `Worker` DB row when the Docker SDK is missing (ImportError path sets state to `spawn_failed` but the HTTP route does the same, so behavior is consistent — however the foreman has no way to retry without orphaned rows accumulating).

backend/foreman/tools.py
```diff
@@ -1023,6 +1026,143 @@ async def _exec_one_tool(guild_id: str, tu, user_id: str | None = None) -> dict:
                         f" Reason: {reason}" if reason else ""
                     )
 
+            elif tu.name == "spawn_worker":
+                repos = inp.get("repos") or []
+                tools_list: list[str] = inp.get("tools") or []
+                agent_count: int | None = inp.get("agent_count")
+                custom_name: str | None = inp.get("name")
```
jmelloy, line 1049: `random.choices` is not cryptographically random. Since `secrets` is already imported and used for `auth_token` and `container_name`, use `secrets.token_hex(3)` or similar for `worker_id` as well to avoid guessable identifiers.

backend/foreman/tools.py
```diff
@@ -1023,6 +1026,143 @@ async def _exec_one_tool(guild_id: str, tu, user_id: str | None = None) -> dict:
...
+
+                    env = build_spawn_worker_env(
+                        guild_id=guild_id,
+                        repos=repos,
+                        worker_name=worker_name,
+                        source_env=dict(os.environ),
+                        claude_oauth_token=decode_claude_oauth_token(stored_blob),
+                        worker_id=worker_id,
+                        auth_token=auth_token,
+                        agent_count=agent_count,
```
jmelloy, line 1090: The `ImportError` branch sets state to `spawn_failed` but the error message tells the foreman to start the worker manually with `PIONEER_AUTH_TOKEN=<token>` — the actual token is never surfaced, so the manual fallback instruction is misleading and unactionable. Consider omitting the manual-start hint or surfacing the token securely.

backend/routes/workers.py
```diff
@@ -144,12 +147,33 @@ async def spawn_worker_container(
...
+    db.add(
+        Worker(
+            id=worker_id,
+            guild_id=guild_pk,
+            repos=json.dumps(data.repos),
+            state="offline",
+            created_at=datetime.now(UTC),
+            auth_token=auth_token,
+            name=worker_name,
+        )
```
jmelloy, line 164: The worker row is committed before the Docker call. If the process crashes between `await db.commit()` and the Docker run (not just an exception but a hard crash), the row remains in `offline` state indefinitely with no container, silently orphaning it. Consider a background cleanup job or a shorter-lived `pending` state with a TTL.

worker/pioneer_worker/config.py
```diff
@@ -216,6 +216,13 @@ def _resolve(p: str) -> str:
             )
         _openai_api_key = _env_val  # None if var is absent, key string if present
 
+    _max_agents_env = os.environ.get("PIONEER_MAX_AGENTS")
```
jmelloy, line 222: `int(_max_agents_env)` will raise `ValueError` if `PIONEER_MAX_AGENTS` is set to a non-integer string (e.g., `"auto"`). Wrap in a try/except or use a helper that falls back to the default (4) on parse failure.

backend/foreman/tools.py
```diff
@@ -1023,6 +1026,143 @@ async def _exec_one_tool(guild_id: str, tu, user_id: str | None = None) -> dict:
...
+                            id=worker_id,
+                            guild_id=guild_pk or 0,
+                            repos=json.dumps(repos),
+                            state="offline",
+                            created_at=created_at,
+                            auth_token=auth_token,
+                            name=worker_name,
+                        )
+                    )
+                    await db.commit()
```
jmelloy, line 1073: The Docker client and image are constructed inside the `elif` block on every call. If the foreman spawns many workers in quick succession, each call creates a new `docker.from_env()` client (which opens a socket). Consider reusing a module-level client or at least adding a comment that this is intentionally short-lived.

PR #533, opened 2026-06-03
Worker spawn: env passthrough + remember last settings

### LINKED ISSUE #531: Worker spawn dialogue: env passthrough + remember last settings per guild
## Feature Request

Enhance the "spawn worker" dialogue with two improvements:

### 1. Environment Variable Passthrough
Add a UI section to the spawn worker dialogue that allows users to specify environment variables to pass into the worker container (e.g. `API_KEY=abc123`, `DEBUG=true`). These should be injected into the Docker container at launch time.

### 2. Remember Last Used Settings Per Guild
Persist the last-used values for the following fields, scoped per guild, so they are pre-populated the next time the dialogue is opened:
- **Tools** (e.g. claude, codex, pi)
- **Environment variables**
- **Repositories**

Settings should be stored in `localStorage` (or equivalent) keyed by guild ID so different guilds retain independent defaults.

## Acceptance Criteria
- [ ] Spawn dialogue has an "Environment Variables" input section (key=value pairs)
- [ ] Env vars are passed through to the spawned Docker container
- [ ] On dialogue open, tools/env/repos fields are pre-populated from the last saved values for the current guild
- [ ] Settings are saved on spawn (not just on close)

jmelloy:
## Summary

- **Env var passthrough**: Added an "Env Vars" section to the spawn worker dialog with a dynamic add/remove row UI (`KEY = value`). These are sent as `env_vars` in the API payload and injected into the Docker container alongside the standard Pioneer vars. Pioneer's required vars (`PIONEER_GUILD_ID`, `PIONEER_AUTH_TOKEN`, etc.) always take precedence over user-supplied values.
- **Remember last settings**: After a successful spawn, the chosen tools, repos, and env vars are saved to `localStorage` under `pioneer_square:spawn_settings:{guildId}`. When the dialog reopens for the same guild, those values are pre-populated.

## Changes

- `backend/utils.py`: Added `extra_env: dict[str, str] | None` parameter to `build_spawn_worker_env`; merged in before Pioneer's required vars so they cannot be overridden.
- `backend/routes/workers.py`: Added `env_vars: dict[str, str] | None` to `SpawnWorkerRequest`; passed through to `build_spawn_worker_env`.
- `frontend/src/components/sidebar/SpawnWorkerForm.vue`: Env var UI (add/remove rows), `loadSavedSettings` / `saveSettings` helpers using localStorage, restore on mount.
- `backend/tests/test_foreman_spawn_worker.py`: Two new unit tests — `test_build_spawn_worker_env_extra_env_passthrough` and `test_build_spawn_worker_env_extra_env_pioneer_vars_win`.

## Test plan

- [ ] `cd backend && python -m pytest tests/test_foreman_spawn_worker.py -v` — 7 pure unit tests pass
- [ ] `cd frontend && npm test && npm run type-check` — all 94 frontend tests pass, types clean
- [ ] Open spawn dialog → add env var rows → launch → reopen dialog → fields pre-populated
- [ ] Verify env var `MY_SECRET=hunter2` appears inside the spawned container

Closes #531


backend/utils.py
```diff
@@ -263,12 +264,20 @@ def build_spawn_worker_env(
     foreman), the worker process skips its own self-registration and uses
     these credentials directly.  This lets callers know the worker_id before
     the container starts.
+
+    *extra_env* — additional key/value pairs supplied by the user at spawn
+    time (e.g. via the UI).  These are merged in first so that Pioneer's own
+    required vars (PIONEER_GUILD_ID, PIONEER_AUTH_TOKEN, etc.) always win.
     """
```
jmelloy, line 271: There is no validation that `extra_env` keys are valid environment variable names (non-empty, no `=` character, etc.). A key containing `=` would produce a malformed env entry when passed to Docker. Consider validating keys before merging.

backend/routes/workers.py
```diff
@@ -53,6 +53,7 @@ class SpawnWorkerRequest(BaseModel):
     name: str | None = None
     tools: list[str] | None = None
     agent_count: int | None = None
+    env_vars: dict[str, str] | None = None
```
jmelloy, line 61: No limit on the number of entries or size of values in `env_vars`. A malicious or buggy client could send arbitrarily large payloads that get injected verbatim into the container environment. Consider adding a max-entries/max-value-length validator.

frontend/src/components/sidebar/SpawnWorkerForm.vue
```diff
@@ -109,8 +145,36 @@ const launchedWorkerId = ref('')
...
+
+function saveSettings(guildId: string) {
+  const settings: SpawnSettings = {
+    repos: selectedRepos.value,
+    tools: selectedTools.value,
+    envVars: envVars.value.filter((e) => e.key.trim() !== ''),
+  }
+  localStorage.setItem(SETTINGS_KEY(guildId), JSON.stringify(settings))
+}
+
```
jmelloy, line 169: Env var values (which may be secrets or tokens) are persisted to `localStorage` after every successful spawn via `saveSettings`. This means sensitive values survive browser sessions and are accessible to any JS running on the same origin. Consider either excluding env var values from persistence (save only keys as hints) or prompting the user before saving.

frontend/src/components/sidebar/SpawnWorkerForm.vue
```diff
@@ -166,21 +230,34 @@ function toggleTool(tool: string) {
   }
 }
 
+function addEnvVar() {
+  envVars.value.push({ key: '', value: '' })
+}
+
+function removeEnvVar(idx: number) {
```
jmelloy, line 244: `envVarsPayload` builds the object with `e.key.trim()` as the key but `e.value` (untrimmed) as the value. If a user accidentally adds leading/trailing whitespace to a value (common in password fields), it will be passed as-is to the container. Consider trimming values too, or at minimum documenting the behaviour.

jmelloy (commented):
- **Core feature is sound**: The env var passthrough correctly merges user-supplied vars before Pioneer's required vars, ensuring Pioneer vars always win. Tests cover both the passthrough and override-prevention cases.
- **Security concern**: User-supplied env var values are stored in `localStorage` without any sanitization or size limit, which could be a vector for storing sensitive secrets in an insecure location (browser storage is readable by JS on the same origin).
- **Missing input validation on backend**: The `env_vars` dict has no validation on key format (e.g., no restriction on empty strings, whitespace-only keys, or keys containing `=`), and no limit on the number of entries or value sizes.
- **localStorage persistence of secrets**: Env var values (potentially secrets/tokens) are persisted to localStorage after a successful spawn. This could inadvertently cache sensitive credentials across sessions.
- **Formatting-only changes are large noise**: A significant portion of the diff is pure reformatting (Prettier/line-length adjustments) unrelated to the feature, making review harder. These should ideally be in a separate commit.

backend/utils.py
```diff
@@ -263,12 +264,24 @@ def build_spawn_worker_env(
...
-    env: dict[str, str] = {
-        "PIONEER_BACKEND_URL": source_env.get("WORKER_BACKEND_URL", "http://backend:8000"),
-        "PIONEER_GUILD_ID": guild_id,
-        "PIONEER_REPOS": ",".join(repos),
-    }
+    if extra_env:
```
jmelloy, line 272: The key validation here (`not key or '=' in key`) is also performed in the Pydantic validator, creating duplication. More importantly, neither location rejects keys containing whitespace, null bytes (`\x00`), or newlines — all of which can cause issues with Docker's env injection or container runtime. Consider using a strict allowlist regex like `^[A-Za-z_][A-Za-z0-9_]*$`.

frontend/src/components/sidebar/SpawnWorkerForm.vue
```diff
@@ -109,8 +145,39 @@ const launchedWorkerId = ref('')
...
+    envVars: envVars.value
+      .filter((e) => e.key.trim() !== '')
+      .map((e) => ({ key: e.key.trim(), value: '' })),
+  }
+  localStorage.setItem(SETTINGS_KEY(guildId), JSON.stringify(settings))
+}
+
 onMounted(async () => {
-  selectedRepos.value = [...ghStore.selectedRepos]
+  const guild = guildStore.currentGuild
```
jmelloy, line 171: Env var values are intentionally not saved (`value: ''` is always stored), but there is no UI hint telling the user this. A user expecting their API keys or secrets to be remembered will be silently surprised. Consider adding a tooltip or hint text such as "Keys only are remembered; values must be re-entered each session."

frontend/src/components/sidebar/SpawnWorkerForm.vue
```diff
@@ -109,8 +145,39 @@ const launchedWorkerId = ref('')
...
+    if (saved.repos?.length) selectedRepos.value = saved.repos
+    if (saved.tools?.length) selectedTools.value = saved.tools
+    if (saved.envVars?.length) envVars.value = saved.envVars
+  } else {
+    selectedRepos.value = [...ghStore.selectedRepos]
+  }
+
   if (ghStore.repos.length === 0 && ghStore.token) {
     loadingRepos.value = true
     await ghStore.fetchRepos()
```
jmelloy, line 187: Saved repo names are not validated against the currently available repos for the guild. If a repo is deleted or access is revoked, it will still appear pre-selected, potentially causing a confusing spawn failure. Consider filtering `saved.repos` against `ghStore.repos` before applying them.

jmelloy (commented):
- **Security concern**: Env var keys are validated twice (in the Pydantic validator and again in `build_spawn_worker_env`), but the backend key validation only checks for empty string and `=` — it does not reject keys with null bytes, whitespace-only strings, or other shell-unsafe characters that could cause issues depending on the Docker API layer.
- **Privacy/security design decision**: `saveSettings` deliberately strips env var values before persisting to localStorage (only keys are saved), which is a reasonable choice but is not surfaced to the user in the UI — users may be surprised their values aren't remembered and re-enter them each time.
- **Correctness issue**: When loading saved settings, the filter `saved.repos?.length` only checks if the saved array is non-empty, but does not verify that saved repo names still exist in the current guild's available repos — stale/deleted repos will silently appear selected.
- **Missing validation**: There is no validation of env var key characters beyond `=` and empty string. Keys with spaces, null bytes, or newlines would pass validation but could cause Docker API errors or unexpected behavior at runtime.
- **Code quality**: The bulk of the diff is pure formatting/style changes (Prettier reformatting), which obscures the functional changes and makes the PR harder to review; these should ideally be in a separate commit.

backend/routes/workers.py
```diff
@@ -48,11 +48,30 @@ class WorkerCreate(BaseModel):
...
+
+
 class SpawnWorkerRequest(BaseModel):
     repos: list[str]
     name: str | None = None
     tools: list[str] | None = None
     agent_count: int | None = None
+    env_vars: dict[str, str] | None = None
+
+    @field_validator("env_vars")
```
jmelloy, line 63: Env var key validation does not reject keys with whitespace or control characters (e.g. `' KEY'`, `'KEY\n'`). These would pass the current check but may cause silent failures or injection issues in the Docker runtime. A stricter regex check (e.g. `re.match(r'^[A-Za-z_][A-Za-z0-9_]*$', key)`) would be safer.

frontend/src/components/sidebar/SpawnWorkerForm.vue
```diff
@@ -109,8 +145,39 @@ const launchedWorkerId = ref('')
...
+    repos: selectedRepos.value,
+    tools: selectedTools.value,
+    // Save only keys (not values) — values may contain secrets/tokens.
+    envVars: envVars.value
+      .filter((e) => e.key.trim() !== '')
+      .map((e) => ({ key: e.key.trim(), value: '' })),
+  }
+  localStorage.setItem(SETTINGS_KEY(guildId), JSON.stringify(settings))
+}
+
```
jmelloy, line 169: Env var values are intentionally discarded before saving (`value: ''`), so on reload users see their keys pre-populated but must re-enter all values. This is a significant UX surprise that should at minimum be documented in a comment or surfaced via a UI hint (e.g., 'Values are not saved for security reasons').

jmelloy (commented):
- **Env var passthrough** is well-implemented with clear precedence rules ensuring Pioneer's required vars always win; validation is applied at both the Pydantic layer and the utility function.
- **Security concern**: `saveSettings` intentionally strips env var values before persisting to localStorage, but this means the "remember last settings" feature only restores keys with empty values — users may be confused when values are blank on reload and re-enter them, partially defeating the UX goal.
- **Duplicate validation**: Key validation (`not key or '=' in key`) is performed twice — once in the Pydantic `field_validator` and again inside `build_spawn_worker_env`. The utility-level check is defensive but adds noise; at minimum it should be documented why it exists independently.
- **Missing key format validation**: Neither the Pydantic validator nor the utility checks that env var keys conform to POSIX conventions (alphanumeric + underscore, not starting with a digit). Arbitrary Unicode or shell-special characters in keys could cause Docker/runtime issues.
- **Cosmetic/formatting changes dominate the diff** (Prettier reformatting across many Vue files), which is unrelated to the feature and makes the diff harder to review; these should ideally be in a separate commit.

backend/routes/workers.py
```diff
@@ -48,11 +48,30 @@ class WorkerCreate(BaseModel):
...
 
+_MAX_ENV_VARS = 20
+_MAX_ENV_VALUE_LEN = 4096
+
+
 class SpawnWorkerRequest(BaseModel):
     repos: list[str]
     name: str | None = None
     tools: list[str] | None = None
     agent_count: int | None = None
```
jmelloy, line 60: Key validation only checks for empty string and '=' but doesn't enforce POSIX env var naming (must match `[A-Za-z_][A-Za-z0-9_]*`). Keys with spaces, Unicode, or starting with digits will pass validation here but may silently fail or behave unexpectedly inside the Docker container.

backend/utils.py
```diff
@@ -263,12 +264,24 @@ def build_spawn_worker_env(
     foreman), the worker process skips its own self-registration and uses
     these credentials directly.  This lets callers know the worker_id before
     the container starts.
+
```
jmelloy, line 267: This second key-validation block duplicates the check already performed by the Pydantic `field_validator`. While defensive programming is good, if this path can be reached without going through `SpawnWorkerRequest` (e.g. direct calls from tests or other code paths), it should raise a more specific exception type (e.g. `ValueError` is fine, but the message format differs from the Pydantic one, which could confuse callers). Consider consolidating into a shared helper.

backend/routes/workers.py
```diff
@@ -48,11 +48,30 @@ class WorkerCreate(BaseModel):
...
 
 
+_MAX_ENV_VARS = 20
+_MAX_ENV_VALUE_LEN = 4096
+
+
 class SpawnWorkerRequest(BaseModel):
     repos: list[str]
     name: str | None = None
     tools: list[str] | None = None
```
jmelloy, line 59: There is no validation that env var keys don't collide with each other case-insensitively on case-insensitive OSes (Windows), and more importantly there's no check preventing keys like `PATH`, `HOME`, or `LD_PRELOAD` that could have security or stability implications in the container environment.

PR #534, opened 2026-06-03
Don't save env var values to localStorage

jmelloy:
## Summary

Follow-up to #533, which added env var passthrough for workers and noted saving values to localStorage as a security concern in review.

- Replaces `envVars?: EnvPair[]` in `SpawnSettings` with `envVarKeys?: string[]` — the type itself now enforces that only keys are persisted, not values
- `saveSettings` maps env var entries to just their trimmed key strings before writing to localStorage
- `onMounted` restores saved keys as empty-value `EnvPair` objects, so the UI reminds the user which vars are expected while requiring values to be re-entered each session
- Existing hint text ("Keys are remembered between sessions; values must be re-entered each time (not saved for security)") continues to explain the behaviour

The previous approach zeroed out values at the call site (`value: ''`) but still typed the field as `EnvPair[]`, leaving an implicit risk of accidentally persisting values in a future refactor. The new `string[]` type makes the security boundary structurally impossible to violate.

## Test plan

- [ ] Spawn a worker with env vars — keys should be remembered after page reload, values should be blank
- [ ] Existing localStorage data with old `envVars` key is silently ignored (field absent → no env vars restored, user adds them fresh)
- [ ] Spawning without env vars still works
- [ ] Adding/removing env var rows in the form works as before

### DIFF
--- a/frontend/src/components/sidebar/SpawnWorkerForm.vue
+++ b/frontend/src/components/sidebar/SpawnWorkerForm.vue
@@ -131,7 +131,7 @@ interface SpawnSettings {
   // Only keys are persisted — values are never stored to localStorage.
   envVarKeys?: string[]
   // Legacy field from before the security fix — used only for migration.
-  envVars?: Array<{ key: string; value: string }>
+  envVars?: EnvPair[]
 }
 
 const SETTINGS_KEY = (guildId: string) => `pioneer_square:spawn_settings:${guildId}`
@@ -177,11 +177,13 @@ onMounted(async () => {
       selectedRepos.value = saved.repos.filter((r) => availableRepoNames.has(r))
     }
     if (saved.tools?.length) selectedTools.value = saved.tools
-    const keys = (saved.envVarKeys ?? saved.envVars?.map((e: { key: string }) => e.key) ?? [])
+    const isLegacy = saved.envVarKeys == null && saved.envVars != null
+    const keys = (saved.envVarKeys ?? saved.envVars?.map((e) => e.key) ?? [])
       .filter((k): k is string => typeof k === 'string')
       .map((k) => k.trim())
       .filter((k) => k !== '')
     if (keys.length) envVars.value = keys.map((k) => ({ key: k, value: '' }))
+    if (isLegacy && guild) saveSettings(guild.id)
   } else {
     selectedRepos.value = [...ghStore.selectedRepos]
   }
--- a/frontend/src/components/sidebar/SpawnWorkerForm.vue
+++ b/frontend/src/components/sidebar/SpawnWorkerForm.vue
@@ -205,7 +205,8 @@ onMounted(async () => {
     // Gate migration on keys.length — avoids a pointless save when legacy entry has no usable data.
     if (isLegacy && keys.length) {
       if (guild) {
-        // Intentionally migrate legacy envVars format on mount — safe to overwrite as values were never stored.
+        // Remove the stale legacy entry (envVars format) and replace it with the migrated envVarKeys format.
+        localStorage.removeItem(SETTINGS_KEY(guild.id))
         saveSettings(guild.id)
       } else {
         // Without guild the localStorage key is unknown, so the stale legacy entry cannot be removed here.
--- a/frontend/src/components/sidebar/SpawnWorkerForm.vue
+++ b/frontend/src/components/sidebar/SpawnWorkerForm.vue
@@ -178,18 +178,37 @@ onMounted(async () => {
       selectedRepos.value = saved.repos.filter((r) => availableRepoNames.has(r))
     }
     if (saved.tools?.length) selectedTools.value = saved.tools
-    const isLegacy = saved.envVarKeys == null && saved.envVars != null
-    const keys = (saved.envVarKeys ?? saved.envVars?.map((e) => e.key) ?? [])
-      .filter((k): k is string => typeof k === 'string')
-      .map((k) => k.trim())
-      .filter((k) => k !== '')
-    if (keys.length) envVars.value = keys.map((k) => ({ key: k, value: '' }))
+    // Legacy when envVars is present and envVarKeys is absent or empty (catches partially-migrated writes too).
+    const isLegacy = saved.envVars != null && (saved.envVarKeys == null || saved.envVarKeys.length === 0)
+
+    let keys: string[]
     if (isLegacy) {
+      // Strip any runtime `value` fields that may exist in pre-fix persisted data (JSON.parse bypasses TS types).
+      keys = (saved.envVars ?? [])
+        .map((e) => {
+          const entry = e as Record<string, unknown>
+          delete entry.value
+          return entry.key
+        })
+        .filter((k): k is string => typeof k === 'string')
+        .map((k) => k.trim())
+        .filter((k) => k !== '')
+    } else {
+      keys = (saved.envVarKeys ?? [])
+        .filter((k): k is string => typeof k === 'string')
+        .map((k) => k.trim())
+        .filter((k) => k !== '')
+    }
+
+    if (keys.length) envVars.value = keys.map((k) => ({ key: k, value: '' }))
+
+    // Gate migration on keys.length — avoids a pointless save when legacy entry has no usable data.
+    if (isLegacy && keys.length) {
       if (guild) {
         // Intentionally migrate legacy envVars format on mount — safe to overwrite as values were never stored.
         saveSettings(guild.id)
       } else {
-        // guild unavailable; delete the stale legacy entry so it doesn't persist across sessions.
+        // Without guild the localStorage key is unknown, so the stale legacy entry cannot be removed here.
         console.warn('[SpawnWorkerForm] Could not migrate legacy envVars localStorage entry: guild unavailable.')
       }
     }
--- a/frontend/src/components/sidebar/SpawnWorkerForm.vue
+++ b/frontend/src/components/sidebar/SpawnWorkerForm.vue
@@ -178,38 +178,48 @@ onMounted(async () => {
       selectedRepos.value = saved.repos.filter((r) => availableRepoNames.has(r))
     }
     if (saved.tools?.length) selectedTools.value = saved.tools
-    // Legacy when envVars is present and envVarKeys is absent or empty (catches partially-migrated writes too).
-    const isLegacy = saved.envVars != null && (saved.envVarKeys == null || saved.envVarKeys.length === 0)
+    // Derive legacyKeys and isLegacy together so migration only fires when pre-migration data
+    // is present. Using !('envVarKeys' in saved) ensures already-migrated records (where
+    // envVarKeys may be []) are never re-migrated.
+    const legacyKeys = Array.isArray(saved.envVars)
+      ? saved.envVars
+          .map((o) => {
+            // TS `value?: never` structurally prevents values from being re-introduced via the
+            // legacy path, but JSON.parse bypasses TS types — delete at runtime too.
+            delete (o as any).value
+            return (o as any).key
+          })
+          .filter(Boolean)
+          .filter((k): k is string => typeof k === 'string')
+          .map((k) => k.trim())
+          .filter((k) => k !== '')
+      : []
+    const isLegacy = legacyKeys.length > 0 && !('envVarKeys' in saved)
 
     let keys: string[]
     if (isLegacy) {
-      // Strip any runtime `value` fields that may exist in pre-fix persisted data (JSON.parse bypasses TS types).
-      keys = (saved.envVars ?? [])
-        .map((e) => {
-          const entry = e as Record<string, unknown>
-          delete entry.value
-          return entry.key
-        })
-        .filter((k): k is string => typeof k === 'string')
-        .map((k) => k.trim())
-        .filter((k) => k !== '')
+      keys = legacyKeys
     } else {
       keys = (saved.envVarKeys ?? [])
         .filter((k): k is string => typeof k === 'string')
         .map((k) => k.trim())
         .filter((k) => k !== '')
+      // Edge case: envVarKeys present but empty while legacy envVars still has real keys
+      // (partial migration). Use legacy keys rather than silently dropping them.
+      if (keys.length === 0 && legacyKeys.length > 0) {
+        keys = legacyKeys
+      }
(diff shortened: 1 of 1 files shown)


frontend/src/components/sidebar/SpawnWorkerForm.vue
```diff
@@ -177,7 +175,8 @@ onMounted(async () => {
       selectedRepos.value = saved.repos.filter((r) => availableRepoNames.has(r))
     }
     if (saved.tools?.length) selectedTools.value = saved.tools
-    if (saved.envVars?.length) envVars.value = saved.envVars
+    if (saved.envVarKeys?.length)
```
jmelloy: Keys restored from localStorage are not trimmed before being used, unlike the save path which calls `.trim()`. A manually edited or legacy localStorage entry with padded keys would silently populate the UI with whitespace in key names. Consider adding `.map((k) => k.trim())` here.

jmelloy: There is no migration from the old `saved.envVars` field to the new `saved.envVarKeys`. Users who saved settings under the previous format will silently lose their stored env var keys after this change ships. A one-time migration (reading `saved.envVars?.map(e => e.key)` as a fallback) would preserve continuity.

jmelloy: No type guard checks that elements of `saved.envVarKeys` are actually strings before mapping over them. If localStorage is corrupted or tampered with, non-string values would produce unexpected UI state. A simple `.filter((k): k is string => typeof k === 'string')` before the map would harden this.

jmelloy (commented):
- **Security improvement**: The refactor correctly moves from `EnvPair[]` (with zeroed-out values) to `string[]` for persisted data, making the security boundary a structural guarantee rather than a runtime convention — this is a meaningful improvement over the previous approach.
- **Backward compatibility gap**: The migration from `envVars` to `envVarKeys` is silent; users with existing localStorage entries using the old `envVars` key will silently lose their saved environment variable keys on next load with no migration path.
- **No input sanitization on restore**: Keys read back from localStorage are used directly without any trimming or validation, unlike the save path which trims keys — a manually crafted or corrupted localStorage entry could introduce whitespace-padded keys into the UI state.
- **Minor robustness**: The `onMounted` restore path lacks a guard for malformed localStorage data (e.g. `envVarKeys` containing non-string array elements), though this is a low-severity edge case consistent with the rest of the file's existing patterns.

jmelloy (commented):
- Good security improvement: changing `envVars?: EnvPair[]` to `envVarKeys?: string[]` makes the type system enforce the security boundary rather than relying on call-site discipline.
- Migration path for the legacy `envVars` field is handled correctly in `onMounted`, extracting only the key strings and discarding any previously-persisted values.
- The runtime type guard `.filter((k): k is string => typeof k === 'string')` is a nice defensive touch against malformed localStorage data.
- The legacy `envVars` field is retained in the `SpawnSettings` interface solely for migration reads, which is appropriate, but there is no mechanism to clean it up from localStorage after migration, meaning old value-containing entries persist indefinitely.
- Overall the change is minimal, focused, and accomplishes the stated goal with no obvious regressions.

frontend/src/components/sidebar/SpawnWorkerForm.vue
```diff
@@ -177,7 +177,11 @@ onMounted(async () => {
       selectedRepos.value = saved.repos.filter((r) => availableRepoNames.has(r))
     }
     if (saved.tools?.length) selectedTools.value = saved.tools
-    if (saved.envVars?.length) envVars.value = saved.envVars
+    const keys = (saved.envVarKeys ?? saved.envVars?.map((e: { key: string }) => e.key) ?? [])
+      .filter((k): k is string => typeof k === 'string')
+      .map((k) => k.trim())
+      .filter((k) => k !== '')
+    if (keys.length) envVars.value = keys.map((k) => ({ key: k, value: '' }))
   } else {
```
jmelloy, line 226: After migrating from the legacy `envVars` field, the old entry (which may contain blank-but-present `value` fields) is never removed from localStorage. Consider re-calling `saveSettings` immediately after a successful migration so the stored object is rewritten in the new format and the legacy `envVars` key is permanently dropped.

frontend/src/components/sidebar/SpawnWorkerForm.vue
```diff
@@ -128,7 +128,10 @@ interface EnvPair {
 interface SpawnSettings {
   repos?: string[]
   tools?: string[]
-  envVars?: EnvPair[]
+  // Only keys are persisted — values are never stored to localStorage.
```
jmelloy, line 131: The legacy `envVars` field is typed as `Array<{ key: string; value: string }>` rather than `EnvPair[]`. These are structurally identical — consider reusing the existing `EnvPair` interface (or an explicit `Pick`) to avoid the duplication and keep a single source of truth for the shape.

frontend/src/components/sidebar/SpawnWorkerForm.vue
```diff
@@ -177,7 +177,13 @@ onMounted(async () => {
       selectedRepos.value = saved.repos.filter((r) => availableRepoNames.has(r))
     }
```
jmelloy, line 179: The `isLegacy` flag is evaluated before the `guild` null-guard that wraps `saveSettings`, so if `guild` is falsy the migration never fires and the old `envVars` entry remains in localStorage indefinitely. Consider logging a warning or handling this case so stale legacy data doesn't persist across sessions when the guild is temporarily unavailable.

frontend/src/components/sidebar/SpawnWorkerForm.vue
```diff
@@ -128,6 +128,9 @@ interface EnvPair {
 interface SpawnSettings {
   repos?: string[]
   tools?: string[]
+  // Only keys are persisted — values are never stored to localStorage.
```
jmelloy, line 131: The `envVars?: EnvPair[]` field remains writable in the interface with only a comment to discourage writing values. Consider making the intent stronger — e.g. `envVars?: { key: string; value?: never }[]` or a `Readonly` wrapper — so the type system prevents accidentally persisting values through this legacy field in a future refactor.

frontend/src/components/sidebar/SpawnWorkerForm.vue
```diff
@@ -177,7 +177,13 @@ onMounted(async () => {
...
+    const isLegacy = saved.envVarKeys == null && saved.envVars != null
+    const keys = (saved.envVarKeys ?? saved.envVars?.map((e) => e.key) ?? [])
+      .filter((k): k is string => typeof k === 'string')
+      .map((k) => k.trim())
+      .filter((k) => k !== '')
+    if (keys.length) envVars.value = keys.map((k) => ({ key: k, value: '' }))
+    if (isLegacy && guild) saveSettings(guild.id)
   } else {
     selectedRepos.value = [...ghStore.selectedRepos]
   }
```
jmelloy, line 228: The auto-migration `saveSettings(guild.id)` overwrites localStorage immediately on mount without user interaction. If the user opened the page in a read-only context or there is a concurrent tab, this silent write could cause surprising behaviour. A comment explaining this is intentional would help future maintainers.

jmelloy (commented):
- The core security improvement is sound: replacing `EnvPair[]` with `string[]` makes the type system enforce that values are never persisted, which is a meaningful structural improvement over the previous zeroing-at-call-site approach.
- The legacy migration path (`isLegacy` + auto-save) is a nice touch that silently upgrades old localStorage entries, though it introduces a subtle edge case if `guild` is null at mount time.
- The `isLegacy` flag is declared before the `guild` nullcheck and the settings parse, but `saveSettings` is only called when `guild` is truthy — this is correct but worth noting the migration silently skips if guild is unavailable.
- Keeping the old `envVars` field in the `SpawnSettings` interface as a legacy migration field is reasonable, but its continued presence means future developers may not realise it should never be written to, only read during migration.
- Overall the change is well-scoped and the runtime behaviour (restoring keys as empty-value pairs) matches the existing hint text shown to users.

frontend/src/components/sidebar/SpawnWorkerForm.vue
```diff
@@ -128,7 +128,11 @@ interface EnvPair {
 interface SpawnSettings {
   repos?: string[]
   tools?: string[]
```
jmelloy, line 130: The `value?: never` on the legacy type means `e.key` is fine to access, but if any stored legacy object happens to have a `value` field (from before the fix), TypeScript won't surface that at runtime. Consider whether you want a runtime check or `delete obj.value` during migration to be truly defensive.

frontend/src/components/sidebar/SpawnWorkerForm.vue
```diff
@@ -177,7 +178,21 @@ onMounted(async () => {
       selectedRepos.value = saved.repos.filter((r) => availableRepoNames.has(r))
     }
```
jmelloy, line 179: `isLegacy` is set before `keys` is computed and before the `if (keys.length)` guard. If `saved.envVars` exists but all keys are empty/blank, `isLegacy` is still `true` and a migration save is triggered unnecessarily. Consider gating the migration on `keys.length > 0` or checking after filtering.

frontend/src/components/sidebar/SpawnWorkerForm.vue
```diff
@@ -177,7 +178,21 @@ onMounted(async () => {
...
     if (saved.tools?.length) selectedTools.value = saved.tools
-    if (saved.envVars?.length) envVars.value = saved.envVars
+    const isLegacy = saved.envVarKeys == null && saved.envVars != null
+    const keys = (saved.envVarKeys ?? saved.envVars?.map((e) => e.key) ?? [])
+      .filter((k): k is string => typeof k === 'string')
+      .map((k) => k.trim())
+      .filter((k) => k !== '')
+    if (keys.length) envVars.value = keys.map((k) => ({ key: k, value: '' }))
```
jmelloy, line 214: When `guild` is unavailable the code logs a warning but leaves the stale legacy entry in localStorage. This means every subsequent mount will hit the same warning indefinitely. Consider removing the stale key with `localStorage.removeItem(SETTINGS_KEY(guildId))` instead of (or in addition to) logging.

jmelloy (commented):
- The core security fix is sound: changing `envVarKeys?: string[]` makes the type system structurally enforce that values are never persisted, which is a meaningful improvement over the previous approach of zeroing values at the call site.
- Legacy migration logic is correct and defensive — it normalises old `envVars` entries to just keys, re-saves in the new format, and gracefully handles the case where `guild` is unavailable.
- The `value?: never` trick on the legacy `envVars` field is clever but may cause a TypeScript compile error when `saved.envVars?.map((e) => e.key)` is called, since `e.key` is valid but the type is narrower than a plain `EnvPair` — this should be verified.
- The runtime `.filter((k): k is string => typeof k === 'string')` guard is good defensive programming against malformed localStorage data, ensuring no crash if the stored format is unexpected.
- No values are ever written to storage in the new path, and the UI hint text already explains the re-entry requirement, so the user-facing behaviour is consistent with the implementation.

frontend/src/components/sidebar/SpawnWorkerForm.vue
```diff
@@ -177,7 +178,21 @@ onMounted(async () => {
...
+    if (isLegacy) {
+      if (guild) {
+        // Intentionally migrate legacy envVars format on mount — safe to overwrite as values were never stored.
+        saveSettings(guild.id)
+      } else {
+        // guild unavailable; delete the stale legacy entry so it doesn't persist across sessions.
+        console.warn('[SpawnWorkerForm] Could not migrate legacy envVars localStorage entry: guild unavailable.')
```
jmelloy, line 223: The comment says 'delete the stale legacy entry so it doesn't persist across sessions', but there is no `localStorage.removeItem(SETTINGS_KEY(guild.id))` call here — the legacy data is actually left untouched. Either remove the misleading comment or add the removal. However, without `guild`, you don't have the key to remove, so the comment should simply acknowledge the limitation rather than claiming deletion occurs.

frontend/src/components/sidebar/SpawnWorkerForm.vue
```diff
@@ -177,7 +178,21 @@ onMounted(async () => {
       selectedRepos.value = saved.repos.filter((r) => availableRepoNames.has(r))
     }
     if (saved.tools?.length) selectedTools.value = saved.tools
```
jmelloy, line 180: If `saved.envVarKeys` is an empty array `[]` and `saved.envVars` is also present (e.g. from a partially migrated write), `isLegacy` will be `false` and the legacy `envVars` will be silently ignored. Consider checking `isLegacy` before the `??` fallback, or deriving keys and legacy status together to avoid this edge case.

frontend/src/components/sidebar/SpawnWorkerForm.vue
jmelloy, line 130: The `envVars?: { key: string; value?: never }[]` legacy type is good for preventing future misuse, but it will cause a TypeScript error if any existing serialised data with a non-undefined `value` is read back and assigned to this type. Since this field is only read from `JSON.parse`, TypeScript won't catch violations at runtime — the guard on line 182 (`typeof k === 'string'`) handles keys but there is no equivalent runtime assertion that `value` is absent, which is fine given values are never used, but worth a brief comment clarifying the type is for write-side safety only.

jmelloy (commented):
- The `value?: never` trick on the legacy `envVars` field is a clever structural enforcement that prevents values from being re-introduced, though the migration path is well-handled
- The migration logic correctly falls back to the legacy `envVars` field via `??` chaining, but the `isLegacy` detection could silently lose data if both fields are somehow present (unlikely but worth noting)
- The `.filter((k): k is string => typeof k === 'string')` runtime guard is a good defensive measure against malformed localStorage data
- The guild-unavailable branch deletes the stale entry via a warning but doesn't actually call `localStorage.removeItem`, leaving the legacy data intact contrary to what the comment implies
- Overall this is a clean, well-motivated refactor that makes the security boundary structurally explicit rather than relying on call-site discipline

jmelloy:
Closing this PR as it has been superseded by #535, which implemented moving env var values from localStorage to the server-side database. The changes in that PR cover the same functionality addressed here.

PR #535, opened 2026-06-03
feat(spawn): save env var values in DB, not localStorage

jmelloy:
## Problem

PR #533 added env var passthrough to the spawn-worker form but intentionally blanked env var values before saving to localStorage (security concern: tokens could sit in browser storage). This meant users had to re-enter all env var values every session.

## Solution

Move spawn-worker form settings (repos, tools, **full** env var key-value pairs) to the database, scoped per guild × user.

### Backend
- **`backend/models.py`**: New `UserSpawnSettings` table (`user_spawn_settings`) with columns `guild_id` (FK), `user_id` (FK), `settings_json` (JSON text), `updated_at`. Composite PK on `(guild_id, user_id)`.
- **`backend/alembic/versions/20260603_000000_create_user_spawn_settings.py`**: Alembic migration.
- **`backend/routes/workers.py`**: Two new authenticated endpoints:
  - `GET /guilds/{guild_id}/spawn-settings` — returns saved settings dict (or `{}` if none)
  - `PUT /guilds/{guild_id}/spawn-settings` — upserts settings with full-replace (PUT) semantics; validates env var key format (same `^[A-Za-z_][A-Za-z0-9_]*$` regex as spawn)

### Frontend (`SpawnWorkerForm.vue`)
- Replaced `localStorage` read/write with `api()` calls to the new endpoints.
- Removed the `.map((e) => ({ key: e.key.trim(), value: '' }))` stripping that blanked env var values.
- Fetch repos before loading saved settings so the available-repo filter has data.
- Updated hint text: values are now persisted server-side.

## Known Limitations / Security Notes

**Env var values (including secrets/tokens) are stored as plaintext JSON in the database.**

This is an improvement over localStorage (browser-local, easily inspectable by any JS on the page, cleared on logout) but it is **not a final solution** for sensitive credentials:

- Anyone with DB read access can see the values.
- There is a `TODO` comment in `backend/routes/workers.py` at the `save_spawn_settings` endpoint marking this for follow-up.

**Encryption at rest is a required follow-up before this feature is used with production secrets.** Until then, treat the DB column as only marginally safer than localStorage — better isolation, worse if the DB is compromised.

## Test plan
- [ ] Spawn a worker with env vars → refresh the page → form should restore keys **and** values
- [ ] Env vars with invalid keys (e.g. `123bad`, `has space`) should be rejected by PUT with 422
- [ ] `ruff check . && ruff format .` — clean
- [ ] Backend tests: `cd backend && python -m pytest`

### DIFF
--- a/backend/models.py
+++ b/backend/models.py
@@ -1,6 +1,6 @@
 from datetime import UTC, datetime
 
-from sqlalchemy import Column, DateTime, Index, or_, text
+from sqlalchemy import Column, DateTime, Index, Text, or_, text
 from sqlmodel import Field, SQLModel, col
 
 
@@ -378,7 +378,8 @@ class UserSpawnSettings(SQLModel, table=True):
     # SECURITY: this column may contain sensitive values (API tokens, credentials,
     # secret env vars). Operators must restrict DB-level read access to this table.
     # TODO: add application-layer encryption for envVars before storing.
-    settings_json: str = Field(default="{}")
+    # TODO: tracked in issue #537
+    settings_json: str = Field(default="{}", sa_column=Column(Text, server_default="{}"))
     updated_at: datetime = Field(
         default_factory=lambda: datetime.now(UTC),
         sa_column=Column(DateTime(timezone=True), nullable=False),
--- a/backend/models.py
+++ b/backend/models.py
@@ -363,6 +363,22 @@ class ClaudeUsage(SQLModel, table=True):
     created_at: datetime = Field(sa_column=Column(DateTime(timezone=True), nullable=False))
 
 
+class UserSpawnSettings(SQLModel, table=True):
+    """Per-user, per-guild spawn-worker preferences (repos, tools, env vars).
+
+    Stored server-side so env var values survive browser sessions without
+    being exposed in localStorage.
+    """
+
+    __tablename__ = "user_spawn_settings"  # type: ignore[assignment]
+
+    guild_id: int = Field(foreign_key="guilds.id", primary_key=True)
+    user_id: str = Field(foreign_key="users.id", primary_key=True)
+    # JSON blob: {repos: [...], tools: [...], envVars: [{key, value}, ...]}
+    settings_json: str = Field(default="{}")
+    updated_at: datetime = Field(sa_column=Column(DateTime(timezone=True), nullable=False))
+
+
 class PushToken(SQLModel, table=True):
     """APNs (or, in the future, FCM) device token registered by the iOS app.
 
--- a/backend/models.py
+++ b/backend/models.py
@@ -379,7 +379,10 @@ class UserSpawnSettings(SQLModel, table=True):
     # secret env vars). Operators must restrict DB-level read access to this table.
     # TODO: add application-layer encryption for envVars before storing.
     settings_json: str = Field(default="{}")
-    updated_at: datetime = Field(sa_column=Column(DateTime(timezone=True), nullable=False))
+    updated_at: datetime = Field(
+        default_factory=lambda: datetime.now(UTC),
+        sa_column=Column(DateTime(timezone=True), nullable=False),
+    )
 
 
 class PushToken(SQLModel, table=True):
--- a/backend/models.py
+++ b/backend/models.py
@@ -375,6 +375,9 @@ class UserSpawnSettings(SQLModel, table=True):
     guild_id: int = Field(foreign_key="guilds.id", primary_key=True)
     user_id: str = Field(foreign_key="users.id", primary_key=True)
     # JSON blob: {repos: [...], tools: [...], envVars: [{key, value}, ...]}
+    # SECURITY: this column may contain sensitive values (API tokens, credentials,
+    # secret env vars). Operators must restrict DB-level read access to this table.
+    # TODO: add application-layer encryption for envVars before storing.
     settings_json: str = Field(default="{}")
     updated_at: datetime = Field(sa_column=Column(DateTime(timezone=True), nullable=False))
 
--- a/backend/routes/workers.py
+++ b/backend/routes/workers.py
@@ -378,15 +378,15 @@ class EnvVarPair(BaseModel):
 
 
 class SaveSpawnSettingsRequest(BaseModel):
-    repos: list[str] | None = None
-    tools: list[str] | None = None
-    envVars: list[EnvVarPair] | None = None
+    # All fields default to empty so a PUT always replaces the full settings object
+    # (no partial-update / PATCH semantics).
+    repos: list[str] = []
+    tools: list[str] = []
+    envVars: list[EnvVarPair] = []
 
     @field_validator("envVars")
     @classmethod
-    def validate_env_var_keys(cls, v: list[EnvVarPair] | None) -> list[EnvVarPair] | None:
-        if v is None:
-            return v
+    def validate_env_var_keys(cls, v: list[EnvVarPair]) -> list[EnvVarPair]:
         if len(v) > _MAX_SETTINGS_ENV_VARS:
             raise ValueError(f"Too many env vars (max {_MAX_SETTINGS_ENV_VARS})")
         for pair in v:
@@ -449,19 +449,25 @@ async def save_spawn_settings(
         )
     )
     row = result.one_or_none()
-    incoming = data.model_dump(exclude_none=True)
+    # model_dump() without exclude_none gives full-replace PUT semantics: every field
+    # is always written, so callers cannot do partial updates through this endpoint.
+    incoming = data.model_dump()
     now = datetime.now(UTC)
+    # TODO: encrypt env var values at rest before production use (tracked in GH issue #537).
+    # Currently stored as plaintext JSON — an improvement over localStorage but not
+    # suitable for highly sensitive credentials until encryption is added.
+    settings_blob = json.dumps(incoming)
     if row is None:
         db.add(
             UserSpawnSettings(
                 guild_id=guild_pk,
                 user_id=github_user_id,
-                settings_json=json.dumps(incoming),
+                settings_json=settings_blob,
                 updated_at=now,
             )
         )
     else:
-        row.settings_json = json.dumps(incoming)
+        row.settings_json = settings_blob
         row.updated_at = now
     await db.commit()
     return {"status": "saved"}
--- a/backend/routes/workers.py
+++ b/backend/routes/workers.py
@@ -365,9 +365,9 @@ async def list_tasks(
     return [row_to_dict(t) for t in result.all()]
 
 
-_MAX_SETTINGS_ENV_VARS = 100
-_MAX_SETTINGS_ENV_KEY_LEN = 1000
-_MAX_SETTINGS_ENV_VALUE_LEN = 1000
+_MAX_SETTINGS_ENV_VARS = 50
+_MAX_SETTINGS_ENV_KEY_LEN = 256
+_MAX_SETTINGS_ENV_VALUE_LEN = 1024
 
 
 class EnvVarPair(BaseModel):
--- a/backend/routes/workers.py
+++ b/backend/routes/workers.py
@@ -19,7 +19,7 @@ from auth_deps import get_guild_pk, require_member
 from database import get_db_dep
 from events import broadcast_msg, emit_terminal_line, pending_claude_auth
 from fastapi import APIRouter, Depends, HTTPException
-from models import ClaudeCredentials, Task, Worker, live_tasks_filter
+from models import ClaudeCredentials, Task, UserSpawnSettings, Worker, live_tasks_filter
 from pydantic import BaseModel, field_validator
 from sqlalchemy import update
 from sqlmodel import col, select
@@ -363,6 +363,90 @@ async def list_tasks(
         .order_by(col(Task.created_at).desc())
     )
     return [row_to_dict(t) for t in result.all()]
+
+
+class EnvVarPair(BaseModel):
+    key: str
+    value: str
+
+
+class SaveSpawnSettingsRequest(BaseModel):
+    repos: list[str] | None = None
+    tools: list[str] | None = None
+    envVars: list[EnvVarPair] | None = None
+
+    @field_validator("envVars")
+    @classmethod
+    def validate_env_var_keys(cls, v: list[EnvVarPair] | None) -> list[EnvVarPair] | None:
+        if v is None:
+            return v
+        for pair in v:
+            if pair.key and not re.match(r"^[A-Za-z_][A-Za-z0-9_]*$", pair.key):
+                raise ValueError(
+                    f"Invalid env var key: {pair.key!r}. Must match ^[A-Za-z_][A-Za-z0-9_]*$"
+                )
+        return v
+
+
+@router.get("/guilds/{guild_id}/spawn-settings")
+async def get_spawn_settings(
+    guild_id: str,
+    github_user_id: str = Depends(require_member()),
+    db: AsyncSession = Depends(get_db_dep),
+):
+    """Return the current user's saved spawn-worker settings for this guild."""
+    guild_pk = await get_guild_pk(db, guild_id)
+    if guild_pk is None:
+        raise HTTPException(status_code=404, detail="Guild not found")
+    result = await db.exec(
+        select(UserSpawnSettings).where(
+            col(UserSpawnSettings.guild_id) == guild_pk,
+            col(UserSpawnSettings.user_id) == github_user_id,
+        )
+    )
+    row = result.one_or_none()
+    if row is None:
+        return {}
+    try:
+        return json.loads(row.settings_json)
+    except json.JSONDecodeError:
+        return {}
+
+
+@router.put("/guilds/{guild_id}/spawn-settings")
+async def save_spawn_settings(
+    guild_id: str,
+    data: SaveSpawnSettingsRequest,
+    github_user_id: str = Depends(require_member()),
+    db: AsyncSession = Depends(get_db_dep),
+):
+    """Persist the current user's spawn-worker settings for this guild."""
+    guild_pk = await get_guild_pk(db, guild_id)
+    if guild_pk is None:
+        raise HTTPException(status_code=404, detail="Guild not found")
+    result = await db.exec(
+        select(UserSpawnSettings).where(
+            col(UserSpawnSettings.guild_id) == guild_pk,
+            col(UserSpawnSettings.user_id) == github_user_id,
+        )
+    )
+    row = result.one_or_none()
+    settings_json = json.dumps(data.model_dump(exclude_none=True))
+    now = datetime.now(UTC)
+    if row is None:
+        db.add(
+            UserSpawnSettings(
+                guild_id=guild_pk,
+                user_id=github_user_id,
+                settings_json=settings_json,
+                updated_at=now,
+            )
+        )
+    else:
+        row.settings_json = settings_json
+        row.updated_at = now
+    await db.commit()
+    return {"status": "saved"}
 
 
 @router.post("/guilds/{guild_id}/workers/{worker_id}/message")
--- a/backend/routes/workers.py
+++ b/backend/routes/workers.py
@@ -8,6 +8,7 @@ the worker process connects via WebSocket (``join`` message in ws_handlers.py).
 from __future__ import annotations
 
 import json
+import logging
 import os
 import random
 import re
@@ -33,6 +34,7 @@ from utils import (
 from ws_handlers import _resolve_user_identifier
 from ws_types import TaskAssignedMsg, WorkerMessageMsg
 
+logger = logging.getLogger(__name__)
 router = APIRouter()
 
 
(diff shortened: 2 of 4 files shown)


backend/routes/workers.py
```diff
@@ -365,6 +365,90 @@ async def list_tasks(
...
+    try:
+        return json.loads(row.settings_json)
+    except json.JSONDecodeError:
+        return {}
+
+
+@router.put("/guilds/{guild_id}/spawn-settings")
+async def save_spawn_settings(
+    guild_id: str,
+    data: SaveSpawnSettingsRequest,
```
jmelloy, line 437: Using `exclude_none=True` means a PUT with only `repos` set will produce `{"repos": [...]}` and completely overwrite (drop) any previously saved `tools` and `envVars`. Consider loading the existing row first and merging, or document that clients must always send the full settings object. As written, partial updates silently delete data.

backend/routes/workers.py
```diff
@@ -365,6 +365,90 @@ async def list_tasks(
...
+class SaveSpawnSettingsRequest(BaseModel):
+    repos: list[str] | None = None
+    tools: list[str] | None = None
+    envVars: list[EnvVarPair] | None = None
+
+    @field_validator("envVars")
+    @classmethod
+    def validate_env_var_keys(cls, v: list[EnvVarPair] | None) -> list[EnvVarPair] | None:
+        if v is None:
+            return v
```
jmelloy: There is no limit on the number of env var pairs or the size of individual values. A client could send a very large payload that gets stored verbatim. Add a max-count check (e.g. `len(v) > 100`) and consider a max value length to prevent storage abuse.

backend/models.py
```diff
@@ -363,6 +363,22 @@ class ClaudeUsage(SQLModel, table=True):
...
 
+class UserSpawnSettings(SQLModel, table=True):
+    """Per-user, per-guild spawn-worker preferences (repos, tools, env vars).
+
+    Stored server-side so env var values survive browser sessions without
+    being exposed in localStorage.
+    """
+
+    __tablename__ = "user_spawn_settings"  # type: ignore[assignment]
+
```
jmelloy, line 374: `settings_json` stores secret values (API tokens, credentials) in plaintext. Consider encrypting the field at the application layer before persisting, or at minimum document the data-classification risk so operators know this column must be protected at the DB level.

jmelloy (commented):
- **Security concern**: Env var values (potentially containing secrets/tokens) are now stored in the database server-side, which is a deliberate trade-off acknowledged in the PR. However, there is no encryption at rest for `settings_json`, meaning any DB access exposes all secrets in plaintext.
- **Race condition in `onMounted`**: Repos are fetched before saved settings are loaded, but the filter `saved.repos.filter(r => availableRepoNames.has(r))` now depends on repos being available — the ordering change is correct, but if `fetchRepos()` fails silently the saved repos will all be filtered out with no user feedback.
- **PUT is not truly an upsert — it overwrites entire settings**: Sending `PUT` with only `repos` will silently drop previously saved `tools` and `envVars` because `exclude_none=True` on a partial payload produces an incomplete dict. There is no merge/patch semantics.
- **No size or count validation on the saved payload**: A malicious or buggy client could send thousands of env var pairs or extremely large values, bloating the `settings_json` column without any server-side limit.
- **`saveSettings` is called after a successful launch but swallows all errors silently**: If the save fails (e.g. network error, validation error), the user gets no feedback and their settings are silently lost for the next session.

frontend/src/components/sidebar/SpawnWorkerForm.vue
```diff
@@ -181,12 +188,6 @@ onMounted(async () => {
...
   }
-
-  if (ghStore.repos.length === 0 && ghStore.token) {
-    loadingRepos.value = true
-    await ghStore.fetchRepos()
-    loadingRepos.value = false
-  }
 })
 
 function toggleRepo(fullName: string) {
```
jmelloy, line 208: If `fetchRepos()` throws or returns an empty list due to a transient error, `availableRepoNames` will be empty and all saved repos will be silently filtered out, falling through to `selectedRepos.value = [...ghStore.selectedRepos]` without any user-visible warning. Consider showing an error or skipping the filter when the repo list is unexpectedly empty.

backend/routes/workers.py
```diff
@@ -365,6 +365,90 @@ async def list_tasks(
...
+    )
+    row = result.one_or_none()
+    if row is None:
+        return {}
+    try:
+        return json.loads(row.settings_json)
+    except json.JSONDecodeError:
+        return {}
+
+
```
jmelloy, line 433: No payload size limit is enforced. A malicious authenticated user can store arbitrarily large JSON (huge env var values or thousands of pairs). Add a max item count (e.g. 50 pairs) and max value length validation in `SaveSpawnSettingsRequest` or `validate_env_var_keys`.

backend/routes/workers.py
```diff
@@ -365,6 +365,90 @@ async def list_tasks(
...
+            if pair.key and not re.match(r"^[A-Za-z_][A-Za-z0-9_]*$", pair.key):
+                raise ValueError(
+                    f"Invalid env var key: {pair.key!r}. Must match ^[A-Za-z_][A-Za-z0-9_]*$"
+                )
+        return v
+
+
+@router.get("/guilds/{guild_id}/spawn-settings")
+async def get_spawn_settings(
+    guild_id: str,
```
jmelloy, line 408: `require_member()` only checks that the user is authenticated; it does not verify they are a member of *this specific* `guild_id`. If `get_guild_pk` succeeds for any guild, any authenticated user can read or overwrite another guild's spawn settings. Ensure the membership check is scoped to the resolved `guild_pk`.

frontend/src/components/sidebar/SpawnWorkerForm.vue
```diff
@@ -146,32 +144,41 @@ const launchedWorkerId = ref('')
...
+    await api(`/guilds/${guildId}/spawn-settings`, {
+      method: 'PUT',
+      json: {
+        repos: selectedRepos.value,
+        tools: selectedTools.value,
+        envVars: envVars.value.filter((e) => e.key.trim() !== ''),
+      },
+    })
+  } catch {
+    // Settings persistence failure is non-fatal — the spawn already succeeded.
```
jmelloy, line 170: Repos are now fetched *before* `loadSavedSettings` is awaited, which is correct for filtering, but if `fetchRepos()` throws (network error), `loadingRepos` stays `true` indefinitely. Wrap in a `try/finally` to ensure `loadingRepos.value = false` is always executed.

backend/models.py
```diff
@@ -363,6 +363,22 @@ class ClaudeUsage(SQLModel, table=True):
...
+    guild_id: int = Field(foreign_key="guilds.id", primary_key=True)
+    user_id: str = Field(foreign_key="users.id", primary_key=True)
+    # JSON blob: {repos: [...], tools: [...], envVars: [{key, value}, ...]}
+    settings_json: str = Field(default="{}")
```
jmelloy: `updated_at` has no `default` or `default_factory`, so instantiating `UserSpawnSettings` without passing `updated_at` will raise a validation error at runtime. Either add `default_factory=lambda: datetime.now(UTC)` or document that callers must always supply it explicitly (the route does, but it's a fragile contract).

jmelloy (commented):
- **Security concern**: Storing env var values (potentially API tokens/secrets) server-side in plaintext JSON shifts the attack surface from browser localStorage to the database — this is an improvement in some ways but the PR description should explicitly acknowledge that the DB now holds secrets and appropriate encryption-at-rest should be ensured.
- **Race condition in onMounted**: Repos are fetched before saved settings are loaded, but `selectedRepos` filtering depends on `availableRepoNames` being populated; if `fetchRepos` fails silently the filter will drop all saved repos without warning.
- **Missing authorization check**: The `GET` and `PUT` endpoints only verify guild membership (`require_member()`) but do not verify that the authenticated user is actually a member of the guild associated with `guild_id` — `get_guild_pk` just resolves the ID without enforcing membership per guild.
- **No size/content limit on `settings_json`**: An authenticated user could PUT arbitrarily large payloads (thousands of env vars with large values), causing unbounded DB row growth; a max-length or max-count validation is absent.
- **`updated_at` not set on model default**: The `UserSpawnSettings` model has no `default_factory` for `updated_at`, so constructing an instance without explicitly passing it will fail at the DB level rather than being caught by Pydantic.

backend/routes/workers.py
```diff
@@ -365,6 +365,110 @@ async def list_tasks(
...
+@router.put("/guilds/{guild_id}/spawn-settings")
+async def save_spawn_settings(
+    guild_id: str,
+    data: SaveSpawnSettingsRequest,
+    github_user_id: str = Depends(require_member()),
+    db: AsyncSession = Depends(get_db_dep),
+):
+    """Persist the current user's spawn-worker settings for this guild."""
+    guild_pk = await get_guild_pk(db, guild_id)
+    if guild_pk is None:
```
jmelloy, line 443: Merge via `existing.update(incoming)` means fields can only be added or overwritten, never deleted. A user who clears all repos (sends `repos: []`) will have `repos` set to `[]` in DB, but a user who wants to remove the `repos` key entirely cannot do so. Consider replacing the entire document instead of merging, or add explicit null-means-delete semantics.

backend/routes/workers.py
```diff
@@ -365,6 +365,110 @@ async def list_tasks(
...
+    @field_validator("envVars")
+    @classmethod
+    def validate_env_var_keys(cls, v: list[EnvVarPair] | None) -> list[EnvVarPair] | None:
+        if v is None:
+            return v
+        if len(v) > _MAX_SETTINGS_ENV_VARS:
+            raise ValueError(f"Too many env vars (max {_MAX_SETTINGS_ENV_VARS})")
+        for pair in v:
+            if len(pair.key) > _MAX_SETTINGS_ENV_KEY_LEN:
+                raise ValueError(
```
jmelloy, line 394: The error message leaks the env var key name back to the client: `f"Env var value for {pair.key!r} exceeds max length"`. For sensitive var names (e.g., `STRIPE_SECRET_KEY`) this is a minor but unnecessary information disclosure; omit the key name from the error.

backend/routes/workers.py
```diff
@@ -365,6 +365,110 @@ async def list_tasks(
...
+):
+    """Return the current user's saved spawn-worker settings for this guild."""
+    guild_pk = await get_guild_pk(db, guild_id)
+    if guild_pk is None:
+        raise HTTPException(status_code=404, detail="Guild not found")
+    result = await db.exec(
+        select(UserSpawnSettings).where(
+            col(UserSpawnSettings.guild_id) == guild_pk,
+            col(UserSpawnSettings.user_id) == github_user_id,
+        )
```
jmelloy, line 420: `settings_json` is stored as plaintext JSON containing API tokens and secrets. The inline TODO acknowledges this, but given the PR's stated motivation is avoiding secrets in browser storage, shipping without at-rest encryption trades one risk for another. This should be resolved (e.g., via column-level encryption or a secrets manager) before the feature is considered production-ready.

frontend/src/components/sidebar/SpawnWorkerForm.vue
```diff
@@ -146,47 +144,57 @@ const launchedWorkerId = ref('')
...
```
jmelloy, line 170: When `availableRepoNames.size === 0` after a successful fetch, the code silently falls back to using all saved repos without validation. This can cause the spawn request to include repos the user no longer has access to. A safer fallback is to show a warning to the user and use an empty selection rather than bypassing the access check.

jmelloy (commented):
- **Security improvement**: Moving env var values from localStorage to server-side DB is the right direction, but the PR's own TODO acknowledges that values are stored in plaintext — this should be flagged as a prerequisite blocker for production use with sensitive credentials.
- **Merge semantics bug**: The `PUT` endpoint merges incoming data over existing data (`existing.update(incoming)`), meaning individual fields (e.g., `repos`) can never be cleared/reset to empty once set — a client sending `{repos: []}` will have `repos` excluded via `exclude_none=True` only if it's `None`, but an empty list is serialized and would overwrite; however a user trying to save zero repos would be excluded since `exclude_none` doesn't exclude empty lists, making the merge behavior inconsistent.
- **No authorization check**: The endpoint only verifies guild membership (`require_member()`) but does not verify that `github_user_id` actually belongs to the calling user — if `require_member()` returns the authenticated user's ID this is fine, but this assumption should be explicit in the code or docs.
- **Value length limits are very permissive**: `_MAX_SETTINGS_ENV_VALUE_LEN = 1000` chars per value × 100 vars = up to 100KB of arbitrary user-controlled data stored per guild×user combination; no total payload size cap exists beyond what the framework provides by default.
- **Frontend repo-filtering fallback is unsafe**: When the repo list is empty after fetch, saved repos are restored unfiltered with only a `console.warn`, meaning stale or invalid repo names will be silently passed to the spawn payload.

backend/routes/workers.py
```diff
@@ -365,6 +365,110 @@ async def list_tasks(
...
+    )
+    row = result.one_or_none()
+    incoming = data.model_dump(exclude_none=True)
+    now = datetime.now(UTC)
+    if row is None:
+        db.add(
+            UserSpawnSettings(
```
jmelloy, line 462: Using `existing.update(incoming)` makes it impossible for the client to delete a field (e.g. clear all env vars by sending `envVars: []`). Since `model_dump(exclude_none=True)` is used, `None` values are dropped. A full-replace strategy (overwrite `row.settings_json` entirely with `incoming` rather than merging) would be simpler and correct; alternatively document and handle explicit deletion.

backend/models.py
```diff
@@ -363,6 +363,25 @@ class ClaudeUsage(SQLModel, table=True):
...
+    __tablename__ = "user_spawn_settings"  # type: ignore[assignment]
+
+    guild_id: int = Field(foreign_key="guilds.id", primary_key=True)
+    user_id: str = Field(foreign_key="users.id", primary_key=True)
+    # JSON blob: {repos: [...], tools: [...], envVars: [{key, value}, ...]}
+    # SECURITY: this column may contain sensitive values (API tokens, credentials,
+    # secret env vars). Operators must restrict DB-level read access to this table.
+    # TODO: add application-layer encryption for envVars before storing.
+    settings_json: str = Field(default="{}")
+    updated_at: datetime = Field(sa_column=Column(DateTime(timezone=True), nullable=False))
```
jmelloy: `updated_at` has no `default` or `default_factory`, so constructing a `UserSpawnSettings` without passing `updated_at` will fail. Add `default_factory=lambda: datetime.now(UTC)` (or a `sa_column` with `server_default`) to make it safe.

backend/routes/workers.py
```diff
@@ -365,6 +365,110 @@ async def list_tasks(
...
+        for pair in v:
+            if len(pair.key) > _MAX_SETTINGS_ENV_KEY_LEN:
+                raise ValueError(
+                    f"Env var key exceeds max length ({_MAX_SETTINGS_ENV_KEY_LEN} chars)"
+                )
+            if len(pair.value) > _MAX_SETTINGS_ENV_VALUE_LEN:
+                raise ValueError(
+                    f"Env var value for {pair.key!r} exceeds max length ({_MAX_SETTINGS_ENV_VALUE_LEN} chars)"
+                )
+            if pair.key and not re.match(r"^[A-Za-z_][A-Za-z0-9_]*$", pair.key):
```
jmelloy, line 399: Env var values up to 1000 chars are accepted and stored verbatim in plaintext JSON. The model's SECURITY comment acknowledges this, but the TODO for application-layer encryption should be tracked as a follow-up issue before this ships to production, since credentials (API tokens, private keys) in a DB column are a high-value target.

frontend/src/components/sidebar/SpawnWorkerForm.vue
```diff
@@ -146,47 +144,57 @@ const launchedWorkerId = ref('')
...
+  }
+
+  const saved = guild ? await loadSavedSettings(guild.id) : null
+
+  if (saved && (saved.repos?.length || saved.tools?.length || saved.envVars?.length)) {
     if (saved.repos?.length) {
       const availableRepoNames = new Set(ghStore.repos.map((r) => r.full_name))
-      selectedRepos.value = saved.repos.filter((r) => availableRepoNames.has(r))
+      if (availableRepoNames.size === 0) {
+        console.warn(
```
jmelloy: When `availableRepoNames.size === 0` after a fetch attempt, saved repos are silently preserved. This can silently pass inaccessible repos to the spawn payload. At minimum show a user-visible warning, and consider not pre-populating `selectedRepos` in this case rather than guessing.

jmelloy (commented):
- Moving env var secrets from localStorage to the database is a sound security improvement, but the TODO for application-layer encryption is critical — secrets are now stored in plaintext in the DB, which shifts rather than eliminates the risk.
- The `save_spawn_settings` PUT endpoint merges incoming data with existing data using `dict.update()`, meaning there is no way to delete a key (e.g. remove all env vars) once it is set — a full-replace or explicit null-clearing strategy is needed.
- The `updated_at` field has no `default` or `default_factory` in the SQLModel definition, so creating a `UserSpawnSettings` object without supplying it explicitly will raise a runtime error; the alembic migration also lacks a `server_default` for this column.
- The repo-filter fallback in the frontend (`availableRepoNames.size === 0`) silently restores potentially stale saved repos without any visible warning to the user, which could lead to workers being spawned against repos the user no longer has access to.
- `_MAX_SETTINGS_ENV_VALUE_LEN` of 1000 characters is quite short for realistic API tokens or certificate values; consider raising or at least documenting the rationale.

jmelloy (commented):
- **Security improvement**: Moving env var values from localStorage to the database is the right direction, but the values are stored as plaintext JSON in the DB — the TODO for application-layer encryption should be tracked as a high-priority follow-up before any sensitive credentials are stored.
- **Merge semantics bug**: The PUT endpoint merges incoming settings with existing ones via `existing.update(incoming)`, meaning users can never delete a previously-saved repo, tool, or env var — only add/overwrite. A full-replace semantics (overwrite the whole blob) would be safer and more predictable.
- **Race condition on mount**: In `onMounted`, repos are fetched and then saved settings are restored, but if `loadSavedSettings` resolves before `fetchRepos` completes (possible if the API call races), the filter `availableRepoNames.size === 0` workaround silently skips validation. The fetch should complete before the settings load, not run in parallel.
- **`saveSettings` is fire-and-forget on failure**: Failures are silently swallowed; users will lose their env var values with no indication. At minimum a console.warn or UI toast would help with debugging.
- **Env var key length limits are very permissive**: 1000-character env var keys and values, combined with up to 100 pairs, means up to ~200 KB of data per user per guild can be stored — consider tightening these limits.

backend/routes/workers.py
```diff
@@ -365,6 +365,110 @@ async def list_tasks(
...
+):
+    """Persist the current user's spawn-worker settings for this guild."""
+    guild_pk = await get_guild_pk(db, guild_id)
+    if guild_pk is None:
+        raise HTTPException(status_code=404, detail="Guild not found")
+    result = await db.exec(
+        select(UserSpawnSettings).where(
+            col(UserSpawnSettings.guild_id) == guild_pk,
+            col(UserSpawnSettings.user_id) == github_user_id,
+        )
```
jmelloy, line 449: The merge via `existing.update(incoming)` makes it impossible to delete previously-saved repos, tools, or env vars — whatever was saved before is sticky. For a settings endpoint the expected semantics should be a full replace (just overwrite `row.settings_json = json.dumps(incoming)`), not a merge. Currently a user who removes a repo from their selection and saves will see it re-appear next session.

backend/routes/workers.py
```diff
@@ -365,6 +365,110 @@ async def list_tasks(
...
+            if len(pair.value) > _MAX_SETTINGS_ENV_VALUE_LEN:
+                raise ValueError(
+                    f"Env var value for {pair.key!r} exceeds max length ({_MAX_SETTINGS_ENV_VALUE_LEN} chars)"
+                )
+            if pair.key and not re.match(r"^[A-Za-z_][A-Za-z0-9_]*$", pair.key):
+                raise ValueError(
+                    f"Invalid env var key: {pair.key!r}. Must match ^[A-Za-z_][A-Za-z0-9_]*$"
+                )
+        return v
+
```
jmelloy, line 404: If `row.settings_json` is somehow corrupt in the GET endpoint, the response silently returns `{}` with a 200, masking a data integrity issue. Consider logging the JSONDecodeError at warning level so operators can detect corruption.

frontend/src/components/sidebar/SpawnWorkerForm.vue
```diff
@@ -146,47 +144,57 @@ const launchedWorkerId = ref('')
...
 
 onMounted(async () => {
   const guild = guildStore.currentGuild
-  const saved = guild ? loadSavedSettings(guild.id) : null
 
-  if (saved) {
+  if (ghStore.repos.length === 0 && ghStore.token) {
+    loadingRepos.value = true
+    await ghStore.fetchRepos()
+    loadingRepos.value = false
```
jmelloy: The `console.warn` fallback that preserves saved repos without validating them against available repos is a hidden footgun: if the repo fetch genuinely failed (network error, token expired), stale or invalid repos will be silently passed to the spawn call. The fetch result (success vs failure) should be checked explicitly rather than using list size as a proxy.

backend/models.py
jmelloy: The `settings_json` field uses `default="{}"` on the Python/SQLModel side but is defined as `sa.Text()` without an explicit `sa_column` — the SQLAlchemy column won't have a `server_default`, only a Python-side default. This is inconsistent with the migration which sets `server_default="{}"`. If a row is ever inserted bypassing SQLModel the column could end up NULL despite `nullable` being unset.

PR #536, opened 2026-06-03
Don't save env var values to localStorage

jmelloy:
Follow-up to #533.

## Summary

- `saveSettings` in `SpawnWorkerForm.vue` already stripped env var values before saving (from #533). This PR adds **defence-in-depth** by also stripping values in `loadSavedSettings`, so any stale/external localStorage data with values is cleared on load — the user always sees their key names but must re-enter values each session.
- Adds `SpawnWorkerForm.spec.ts` with 5 component tests covering:
  - `saveSettings` stores keys with blank values (never the actual value)
  - `saveSettings` drops blank-key entries
  - `loadSavedSettings` strips values even if stored data contains them
  - `loadSavedSettings` returns null for each non-object JSON shape (parametrized with `it.each`)
  - `loadSavedSettings` strips values across multiple env var entries

## Changes

- `frontend/src/components/sidebar/SpawnWorkerForm.vue` — add defensive value strip in `loadSavedSettings`; fix spurious `console.warn` for settings saved before env var support (undefined is now silently treated as `[]`)
- `frontend/src/components/sidebar/__tests__/SpawnWorkerForm.spec.ts` — 5 component tests; parametrized non-object JSON cases use `it.each`; assert `localStorage.getItem` is non-null before parsing in saveSettings tests
- `frontend/src/components/sidebar/spawnWorkerFormConstants.ts` — add JSDoc explaining the file exists so tests can import the key without string duplication

### DIFF
--- a/frontend/src/components/sidebar/SpawnWorkerForm.vue
+++ b/frontend/src/components/sidebar/SpawnWorkerForm.vue
@@ -115,6 +115,7 @@ import { useGuildStore } from '../../stores/guild'
 import { useGitHubStore } from '../../stores/github'
 import { api } from '../../utils/api'
 import { groupAndSortRepos } from '../../utils/repoGroups'
+import { SETTINGS_KEY } from './spawnWorkerFormConstants'
 
 const emit = defineEmits<{ (e: 'launched'): void }>()
 
@@ -133,8 +134,6 @@ interface SpawnSettings {
   tools?: string[]
   envVars?: EnvPair[]
 }
-
-const SETTINGS_KEY = (guildId: string) => `pioneer_square:spawn_settings:${guildId}`
 
 const selectedRepos = ref<string[]>([])
 const name = ref('')
@@ -158,6 +157,7 @@ function loadSavedSettings(guildId: string): SpawnSettings | null {
     if (Array.isArray(parsed.envVars)) {
       parsed.envVars = parsed.envVars.map((e) => ({ key: e.key, value: '' }))
     } else {
+      console.warn('Unexpected envVars shape in localStorage, resetting:', parsed.envVars)
       parsed.envVars = []
     }
     return parsed
--- a/frontend/src/components/sidebar/SpawnWorkerForm.vue
+++ b/frontend/src/components/sidebar/SpawnWorkerForm.vue
@@ -76,6 +76,7 @@
           <input
             v-model="pair.key"
             class="spawn-input spawn-env-input spawn-env-key"
+            data-testid="spawn-env-key"
             placeholder="KEY"
             type="text"
           />
@@ -83,6 +84,7 @@
           <input
             v-model="pair.value"
             class="spawn-input spawn-env-input spawn-env-val"
+            data-testid="spawn-env-val"
             placeholder="value"
             type="text"
           />
@@ -90,11 +92,12 @@
             ×
           </button>
         </div>
-        <button class="pixel-btn spawn-env-add" @click="addEnvVar" type="button">+ Add</button>
+        <button class="pixel-btn spawn-env-add" data-testid="spawn-env-add" @click="addEnvVar" type="button">+ Add</button>
       </div>
       <div class="spawn-actions">
         <button
           class="pixel-btn spawn-launch-btn"
+          data-testid="spawn-launch-btn"
           :disabled="spawning || selectedRepos.length === 0"
           @click="launch"
         >
@@ -152,8 +155,10 @@ function loadSavedSettings(guildId: string): SpawnSettings | null {
     if (!raw) return null
     const parsed = JSON.parse(raw) as SpawnSettings
     // Strip values on load as defence-in-depth; values are never persisted.
-    if (parsed.envVars) {
+    if (Array.isArray(parsed.envVars)) {
       parsed.envVars = parsed.envVars.map((e) => ({ key: e.key, value: '' }))
+    } else {
+      parsed.envVars = []
     }
     return parsed
   } catch {
--- a/frontend/src/components/sidebar/SpawnWorkerForm.vue
+++ b/frontend/src/components/sidebar/SpawnWorkerForm.vue
@@ -152,15 +152,17 @@ function loadSavedSettings(guildId: string): SpawnSettings | null {
   try {
     const raw = localStorage.getItem(SETTINGS_KEY(guildId))
     if (!raw) return null
-    const parsed = JSON.parse(raw) as SpawnSettings
+    const parsed = JSON.parse(raw)
+    if (parsed === null || typeof parsed !== 'object' || Array.isArray(parsed)) return null
+    const settings = parsed as SpawnSettings
     // Strip values on load as defence-in-depth; values are never persisted.
-    if (Array.isArray(parsed.envVars)) {
-      parsed.envVars = parsed.envVars.map((e) => ({ key: e.key, value: '' }))
+    if (Array.isArray(settings.envVars)) {
+      settings.envVars = settings.envVars.map((e) => ({ key: e.key, value: '' }))
     } else {
-      console.warn('Unexpected envVars shape in localStorage, resetting:', parsed.envVars)
-      parsed.envVars = []
+      console.warn('Unexpected envVars shape in localStorage, resetting:', settings.envVars)
+      settings.envVars = []
     }
-    return parsed
+    return settings
   } catch {
     return null
   }
--- a/frontend/src/components/sidebar/SpawnWorkerForm.vue
+++ b/frontend/src/components/sidebar/SpawnWorkerForm.vue
@@ -149,7 +149,13 @@ const groupedRepos = computed(() => groupAndSortRepos(ghStore.repos))
 function loadSavedSettings(guildId: string): SpawnSettings | null {
   try {
     const raw = localStorage.getItem(SETTINGS_KEY(guildId))
-    return raw ? (JSON.parse(raw) as SpawnSettings) : null
+    if (!raw) return null
+    const parsed = JSON.parse(raw) as SpawnSettings
+    // Strip values on load as defence-in-depth; values are never persisted.
+    if (parsed.envVars) {
+      parsed.envVars = parsed.envVars.map((e) => ({ key: e.key, value: '' }))
+    }
+    return parsed
   } catch {
     return null
   }
--- a/frontend/src/components/sidebar/SpawnWorkerForm.vue
+++ b/frontend/src/components/sidebar/SpawnWorkerForm.vue
@@ -158,10 +158,11 @@ function loadSavedSettings(guildId: string): SpawnSettings | null {
     // Strip values on load as defence-in-depth; values are never persisted.
     if (Array.isArray(settings.envVars)) {
       settings.envVars = settings.envVars.map((e) => ({ key: e.key, value: '' }))
-    } else {
-      console.warn('Unexpected envVars shape in localStorage, resetting:', settings.envVars)
+    } else if (settings.envVars !== undefined) {
+      console.warn('Unexpected envVars shape, resetting:', settings.envVars)
       settings.envVars = []
     }
+    // if undefined: silently default to []
     return settings
   } catch {
     return null
--- a/frontend/src/components/sidebar/__tests__/SpawnWorkerForm.spec.ts
+++ b/frontend/src/components/sidebar/__tests__/SpawnWorkerForm.spec.ts
@@ -52,10 +52,13 @@ describe('SpawnWorkerForm env var localStorage safety', () => {
       await wrapper.find('[data-testid="spawn-env-key"]').setValue('MY_SECRET')
       await wrapper.find('[data-testid="spawn-env-val"]').setValue('super-secret-value')
 
+      // Clicking the launch button triggers saveSettings as a side effect.
       await wrapper.find('[data-testid="spawn-launch-btn"]').trigger('click')
       await flushPromises()
 
-      const saved = JSON.parse(localStorage.getItem(SETTINGS_KEY) || '{}')
+      const raw = localStorage.getItem(SETTINGS_KEY)
+      expect(raw).not.toBeNull()
+      const saved = JSON.parse(raw!)
       expect(saved.envVars).toEqual([{ key: 'MY_SECRET', value: '' }])
     })
 
@@ -73,10 +76,13 @@ describe('SpawnWorkerForm env var localStorage safety', () => {
       await valInputs[0].setValue('some-value')
       // leave keyInputs[1] blank
 
+      // Clicking the launch button triggers saveSettings as a side effect.
       await wrapper.find('[data-testid="spawn-launch-btn"]').trigger('click')
       await flushPromises()
 
-      const saved = JSON.parse(localStorage.getItem(SETTINGS_KEY) || '{}')
+      const raw = localStorage.getItem(SETTINGS_KEY)
+      expect(raw).not.toBeNull()
+      const saved = JSON.parse(raw!)
       expect(saved.envVars).toEqual([{ key: 'VALID_KEY', value: '' }])
     })
   })
@@ -101,17 +107,17 @@ describe('SpawnWorkerForm env var localStorage safety', () => {
       expect(valInput.value).toBe('')
     })
 
-    it('returns null and renders defaults when localStorage contains non-object JSON', async () => {
-      for (const bad of ['null', '42', '"a string"', 'true', '[]']) {
+    it.each(['null', '42', '"a string"', 'true', '[]'])(
+      'returns null and renders defaults when localStorage contains %s',
+      async (bad) => {
         localStorage.setItem(SETTINGS_KEY, bad)
         const wrapper = createWrapper()
         await flushPromises()
         // Should fall back to default (selected repos from store, no env vars)
         expect(wrapper.findAll('[data-testid="spawn-env-key"]')).toHaveLength(0)
-        localStorage.clear()
         wrapper.unmount()
-      }
-    })
+      },
+    )
 
     it('returns null and renders defaults when localStorage contains malformed JSON', async () => {
       localStorage.setItem(SETTINGS_KEY, '{not valid json}')
--- a/frontend/src/components/sidebar/__tests__/SpawnWorkerForm.spec.ts
+++ b/frontend/src/components/sidebar/__tests__/SpawnWorkerForm.spec.ts
@@ -45,13 +45,13 @@ describe('SpawnWorkerForm env var localStorage safety', () => {
       const wrapper = createWrapper()
       await flushPromises()
 
-      await wrapper.find('.spawn-env-add').trigger('click')
+      await wrapper.find('[data-testid="spawn-env-add"]').trigger('click')
       await wrapper.vm.$nextTick()
 
-      await wrapper.find('.spawn-env-key').setValue('MY_SECRET')
-      await wrapper.find('.spawn-env-val').setValue('super-secret-value')
+      await wrapper.find('[data-testid="spawn-env-key"]').setValue('MY_SECRET')
+      await wrapper.find('[data-testid="spawn-env-val"]').setValue('super-secret-value')
 
-      await wrapper.find('.spawn-launch-btn').trigger('click')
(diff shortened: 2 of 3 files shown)


frontend/src/components/sidebar/SpawnWorkerForm.vue
```diff
@@ -149,7 +149,13 @@ const groupedRepos = computed(() => groupAndSortRepos(ghStore.repos))
 function loadSavedSettings(guildId: string): SpawnSettings | null {
   try {
     const raw = localStorage.getItem(SETTINGS_KEY(guildId))
-    return raw ? (JSON.parse(raw) as SpawnSettings) : null
+    if (!raw) return null
+    const parsed = JSON.parse(raw) as SpawnSettings
+    // Strip values on load as defence-in-depth; values are never persisted.
+    if (parsed.envVars) {
+      parsed.envVars = parsed.envVars.map((e) => ({ key: e.key, value: '' }))
```
jmelloy: The cast `JSON.parse(raw) as SpawnSettings` provides no runtime safety. A corrupted or externally-written localStorage value that passes JSON parsing but doesn't conform to `SpawnSettings` (e.g. `envVars` being a non-array) will bypass the `if (parsed.envVars)` guard and could throw or silently produce wrong state. Consider adding a runtime shape check or wrapping the `parsed.envVars.map` in an additional `Array.isArray` guard.

frontend/src/components/sidebar/__tests__/SpawnWorkerForm.spec.ts
```diff
@@ -0,0 +1,125 @@
...
+    localStorage.clear()
+  })
+
+  describe('saveSettings', () => {
+    it('stores env var keys with empty values, not the actual secret values', async () => {
+      const wrapper = createWrapper()
+      await flushPromises()
+
+      await wrapper.find('.spawn-env-add').trigger('click')
+      await wrapper.vm.$nextTick()
```
jmelloy, line 50: Tests depend heavily on CSS class selectors like `.spawn-env-add`, `.spawn-env-key`, `.spawn-env-val`, and `.spawn-launch-btn`. These are presentation-layer details not visible in this diff. Consider using `data-testid` attributes instead to decouple test stability from styling changes.

frontend/src/components/sidebar/__tests__/SpawnWorkerForm.spec.ts
```diff
@@ -0,0 +1,125 @@
...
+      await flushPromises()
+
+      const keyInput = wrapper.find('.spawn-env-key').element as HTMLInputElement
+      const valInput = wrapper.find('.spawn-env-val').element as HTMLInputElement
+      expect(keyInput.value).toBe('MY_VAR')
+      expect(valInput.value).toBe('')
+    })
+
+    it('restores multiple keys all with empty values', async () => {
+      localStorage.setItem(
```
jmelloy, line 130: The stored JSON in this test omits the `tools` field that the first load test includes. If `loadSavedSettings` or the component initialisation path expects `tools` to be present this test may not accurately reflect real usage; add `tools: []` for consistency and to avoid masking regressions.

jmelloy (commented):
- The defence-in-depth approach in `loadSavedSettings` is sound and correctly addresses the stale localStorage data scenario described in the PR.
- The type cast `JSON.parse(raw) as SpawnSettings` has no runtime validation; a malformed or attacker-controlled localStorage entry could cause unexpected behaviour downstream even after the env-var strip.
- Test coverage is good and directly exercises the security invariant, but the tests rely on CSS class selectors (`.spawn-env-key`, `.spawn-env-val`, `.spawn-env-add`, `.spawn-launch-btn`) that are not shown in the diff — if those classes change the tests will silently break without any indication of why.
- The `tools` property is omitted from the stored JSON in the last test (`restores multiple keys all with empty values`), which may mask a subtle regression if `loadSavedSettings` has any logic that depends on `tools` being present.
- No negative/edge-case tests cover `envVars: null`, `envVars: []`, or a completely missing `envVars` field, leaving the optional-chaining guard (`if (parsed.envVars)`) untested.

jmelloy (commented):
- Good defence-in-depth approach: stripping env var values on load ensures stale localStorage data never leaks secrets into the UI, complementing the existing strip-on-save logic.
- The `loadSavedSettings` defensively resets `envVars` to `[]` when the field is missing/non-array, which is a safe fallback but silently discards any partially-valid structure — acceptable given the security goal.
- Test coverage is solid: all four cases map directly to the two code paths in both `saveSettings` and `loadSavedSettings`, and use `data-testid` attributes properly.
- Minor concern: the `loadSavedSettings` tests rely on the component mounting with the pre-seeded `currentGuild.id` matching `GUILD_ID` to hit the correct localStorage key — this coupling is implicit and could silently break if store initialisation changes.
- No security regressions found; the change is strictly additive and the diff is minimal and focused.

frontend/src/components/sidebar/__tests__/SpawnWorkerForm.spec.ts
```diff
@@ -0,0 +1,126 @@
...
+      await flushPromises()
+
+      await wrapper.find('[data-testid="spawn-env-add"]').trigger('click')
+      await wrapper.vm.$nextTick()
+
+      await wrapper.find('[data-testid="spawn-env-key"]').setValue('MY_SECRET')
+      await wrapper.find('[data-testid="spawn-env-val"]').setValue('super-secret-value')
+
+      await wrapper.find('[data-testid="spawn-launch-btn"]').trigger('click')
+      await flushPromises()
```
jmelloy, line 57: After triggering two `spawn-env-add` clicks there is a single `$nextTick` await before querying inputs. With two reactive additions it would be safer to call `await flushPromises()` (or two `$nextTick` ticks) to ensure both renders are fully flushed before `findAll`.

frontend/src/components/sidebar/__tests__/SpawnWorkerForm.spec.ts
```diff
@@ -0,0 +1,126 @@
...
+
+vi.mock('../../../utils/api', () => ({
+  api: vi.fn().mockResolvedValue({ worker_id: 'test-worker' }),
+}))
+
+const GUILD_ID = 'guild-1'
+const SETTINGS_KEY = `pioneer_square:spawn_settings:${GUILD_ID}`
+const TEST_REPOS = [{ full_name: 'owner/repo', language: 'TypeScript' }]
+
+function createWrapper() {
```
jmelloy, line 15: The `SETTINGS_KEY` constant duplicates the production key template (`pioneer_square:spawn_settings:${GUILD_ID}`). If the production key format ever changes the tests will silently pass against a stale key while the real key goes untested. Consider importing and invoking the same `SETTINGS_KEY` helper from the component, or at minimum add a comment flagging this coupling.

frontend/src/components/sidebar/SpawnWorkerForm.vue
```diff
@@ -149,7 +152,15 @@ const groupedRepos = computed(() => groupAndSortRepos(ghStore.repos))
...
   try {
     const raw = localStorage.getItem(SETTINGS_KEY(guildId))
-    return raw ? (JSON.parse(raw) as SpawnSettings) : null
+    if (!raw) return null
+    const parsed = JSON.parse(raw) as SpawnSettings
+    // Strip values on load as defence-in-depth; values are never persisted.
+    if (Array.isArray(parsed.envVars)) {
+      parsed.envVars = parsed.envVars.map((e) => ({ key: e.key, value: '' }))
+    } else {
+      parsed.envVars = []
```
jmelloy: When `parsed.envVars` is not an array it is reset to `[]`, silently discarding whatever was stored. This is the right security default, but it also means a corrupt/unexpected type (e.g. an object) causes a silent no-op instead of a detectable error. A `console.warn` here would help developers notice unexpected localStorage shapes during development without affecting production behaviour.

frontend/src/components/sidebar/SpawnWorkerForm.vue
```diff
@@ -149,7 +151,18 @@ const groupedRepos = computed(() => groupAndSortRepos(ghStore.repos))
...
     const raw = localStorage.getItem(SETTINGS_KEY(guildId))
-    return raw ? (JSON.parse(raw) as SpawnSettings) : null
+    if (!raw) return null
+    const parsed = JSON.parse(raw)
+    if (parsed === null || typeof parsed !== 'object' || Array.isArray(parsed)) return null
+    const settings = parsed as SpawnSettings
+    // Strip values on load as defence-in-depth; values are never persisted.
+    if (Array.isArray(settings.envVars)) {
+      settings.envVars = settings.envVars.map((e) => ({ key: e.key, value: '' }))
+    } else {
```
jmelloy: The `else` branch fires whenever `settings.envVars` is `undefined` (i.e. any settings object saved before env var support existed). This will log a spurious warning for all legacy users. Consider only warning for truly unexpected shapes — e.g. `else if (settings.envVars !== undefined)` — and silently defaulting to `[]` when the field is simply absent.

frontend/src/components/sidebar/__tests__/SpawnWorkerForm.spec.ts
```diff
@@ -0,0 +1,146 @@
+import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest'
```
jmelloy, line 1: The PR description states '4 component tests' but the spec file contains 5 `it` blocks. Please update the description (or the test count) to avoid confusion for future reviewers.

frontend/src/components/sidebar/__tests__/SpawnWorkerForm.spec.ts
```diff
@@ -0,0 +1,146 @@
...
+      const wrapper = createWrapper()
+      await flushPromises()
+
+      await wrapper.find('[data-testid="spawn-env-add"]').trigger('click')
+      await wrapper.vm.$nextTick()
+
+      await wrapper.find('[data-testid="spawn-env-key"]').setValue('MY_SECRET')
+      await wrapper.find('[data-testid="spawn-env-val"]').setValue('super-secret-value')
+
+      await wrapper.find('[data-testid="spawn-launch-btn"]').trigger('click')
```
jmelloy, line 56: The `saveSettings` test triggers the launch button to indirectly call `saveSettings`. If `launch` fails or its behaviour changes (e.g. validation prevents the save), the test silently passes with an empty localStorage and the assertion `saved.envVars` would evaluate against `{}`, hiding the failure. Consider asserting `localStorage.getItem(SETTINGS_KEY)` is not null before parsing, or expose/spy on `saveSettings` directly.

jmelloy (commented):
- Good security defence-in-depth: stripping env var values on load prevents stale secrets from being exposed even if older code wrote them to localStorage.
- The `else` branch in `loadSavedSettings` that resets `envVars` to `[]` when it is not an array is correct, but the `console.warn` will fire for the common case where `envVars` is simply `undefined` (absent from the stored object), which will produce noisy warnings for any settings saved before env var support was added.
- The test suite is well-structured and covers the important cases, but the description says '4 passing' while the spec file contains 5 `it` blocks — a minor documentation inconsistency.
- Extracting `SETTINGS_KEY` to a constants file is a clean refactor that enables sharing between the component and tests without circular imports.
- The `loadSavedSettings` guard correctly rejects `null`, non-objects, and arrays, which is a solid improvement over the original bare cast.

frontend/src/components/sidebar/SpawnWorkerForm.vue
```diff
@@ -149,7 +151,18 @@ const groupedRepos = computed(() => groupAndSortRepos(ghStore.repos))
...
+      console.warn('Unexpected envVars shape in localStorage, resetting:', settings.envVars)
```
jmelloy: The `else` branch triggers a `console.warn` whenever `envVars` is `undefined`, which is the normal case for any settings object saved before env var support was added. This will produce noisy false-positive warnings for all existing users. Consider only warning (or resetting) when `envVars` is present but not an array: `else if (settings.envVars !== undefined) { console.warn(...); settings.envVars = [] }`.

frontend/src/components/sidebar/__tests__/SpawnWorkerForm.spec.ts
jmelloy, line 1: The PR description states "4 component tests" but the spec file contains 5 `it` blocks. This is a minor docs inconsistency but worth correcting to keep the description accurate.

frontend/src/components/sidebar/__tests__/SpawnWorkerForm.spec.ts
```diff
@@ -0,0 +1,146 @@
...
+
+  describe('saveSettings', () => {
+    it('stores env var keys with empty values, not the actual secret values', async () => {
+      const wrapper = createWrapper()
+      await flushPromises()
+
+      await wrapper.find('[data-testid="spawn-env-add"]').trigger('click')
+      await wrapper.vm.$nextTick()
+
+      await wrapper.find('[data-testid="spawn-env-key"]').setValue('MY_SECRET')
```
jmelloy, line 52: The `saveSettings` test triggers the launch button to indirectly invoke `saveSettings`. If `saveSettings` is ever decoupled from the launch flow (e.g. auto-save on change), this test will silently stop covering the intended behaviour. Consider testing `saveSettings` more directly, or at least add a comment explaining the indirect invocation.

jmelloy (commented):
- The defence-in-depth approach in `loadSavedSettings` is sound and correctly strips env var values from any stale localStorage data before they are used.
- The `else` branch in `loadSavedSettings` that resets `envVars` to `[]` fires when `envVars` is `undefined` (i.e. never set), which is a normal case for existing saved settings without env vars — the `console.warn` will produce noisy false-positive warnings for legitimate data.
- Extracting `SETTINGS_KEY` into a separate constants file is a clean refactor that enables test isolation without mocking the whole module.
- The test suite is thorough and covers the critical security scenarios; however, the test descriptions say "4 passing" but the spec file contains 5 `it` blocks, which is a minor documentation inconsistency.
- The `loadSavedSettings` defensive strip mutates the parsed object in place before returning it, which is fine here but slightly surprising style — returning a new object would be more explicit.

frontend/src/components/sidebar/SpawnWorkerForm.vue
```diff
@@ -149,7 +151,18 @@ const groupedRepos = computed(() => groupAndSortRepos(ghStore.repos))
...
+      settings.envVars = []
```
jmelloy, line 163: The `else` branch fires when `settings.envVars` is `undefined` (i.e. the stored object simply has no `envVars` key), which is a perfectly normal case for settings saved before env-var support was added. This logs a spurious warning and resets to `[]` unnecessarily. Consider only warning/resetting for truthy non-array values: `if (settings.envVars !== undefined && !Array.isArray(settings.envVars))`.

frontend/src/components/sidebar/__tests__/SpawnWorkerForm.spec.ts
```diff
@@ -0,0 +1,146 @@
...
+
+      await wrapper.find('[data-testid="spawn-launch-btn"]').trigger('click')
+      await flushPromises()
+
+      const saved = JSON.parse(localStorage.getItem(SETTINGS_KEY) || '{}')
+      expect(saved.envVars).toEqual([{ key: 'MY_SECRET', value: '' }])
+    })
+
+    it('does not save entries with blank keys', async () => {
+      const wrapper = createWrapper()
```
jmelloy, line 66: The test clicks the launch button to trigger `saveSettings`, but the launch also fires an API call that emits `launched`. If the parent ever conditionally unmounts the form on launch, the assertion on `localStorage` could race. Consider spying on `saveSettings` directly or checking `localStorage` before `flushPromises` resolves the API mock to make the intent clearer.

frontend/src/components/sidebar/__tests__/SpawnWorkerForm.spec.ts
```diff
@@ -0,0 +1,146 @@
...
+      expect(valInput.value).toBe('')
+    })
+
+    it('returns null and renders defaults when localStorage contains non-object JSON', async () => {
+      for (const bad of ['null', '42', '"a string"', 'true', '[]']) {
+        localStorage.setItem(SETTINGS_KEY, bad)
+        const wrapper = createWrapper()
+        await flushPromises()
+        // Should fall back to default (selected repos from store, no env vars)
+        expect(wrapper.findAll('[data-testid="spawn-env-key"]')).toHaveLength(0)
```
jmelloy, line 117: The parametrized loop mutates `localStorage` and mounts/unmounts wrappers inside a single `it` block. If one iteration throws, subsequent iterations are skipped silently and the test still reports a single failure with no indication of which bad value caused it. Prefer individual `it` cases or `it.each` for clearer failure attribution.

jmelloy (commented):
- Good defence-in-depth approach: stripping env var values on both save and load ensures stale localStorage data never surfaces secrets in the UI.
- The `else` branch in `loadSavedSettings` resets `envVars` to `[]` when the field is absent or has an unexpected type (e.g. `undefined`), which silently discards any legitimately absent `envVars` field and logs a spurious warning — the absent case should be handled separately.
- Extracting `SETTINGS_KEY` to a constants file is a clean refactor that enables test imports without mounting the component.
- Tests are well-structured and cover the main scenarios, but the describe block advertises "4 passing" tests while the spec file contains 5 `it` blocks.
- Minor: `data-testid` attributes on the launch button and env inputs are a useful addition for testability, though the launch-button testid is not exercised by the load-path tests.

frontend/src/components/sidebar/SpawnWorkerForm.vue
jmelloy: The `else` branch resets `envVars` to `[]` and emits a `console.warn` even when `settings.envVars` is simply `undefined` (i.e., settings saved before env var support existed). This is a valid, expected state and shouldn't be treated as an unexpected shape. Consider only warning/resetting when `settings.envVars` is defined but not an array: `else if (settings.envVars !== undefined) { console.warn(...); settings.envVars = [] }`.

frontend/src/components/sidebar/__tests__/SpawnWorkerForm.spec.ts
jmelloy, line 56: The `createWrapper` helper reads `selectedRepos` from the pinia store initial state, but the component likely derives its own `selectedRepos` ref independently. If the component doesn't consume `github.selectedRepos` from the store directly, the `launch` button may remain disabled (`selectedRepos.length === 0`) and `saveSettings` may never be called, causing these tests to silently pass for the wrong reason. Verify that the button is actually not disabled when the test clicks it, or assert on the disabled state first.

frontend/src/components/sidebar/spawnWorkerFormConstants.ts
```diff
@@ -0,0 +1 @@
+export const SETTINGS_KEY = (guildId: string) => `pioneer_square:spawn_settings:${guildId}`
```
jmelloy, line 2: The file exports a single function with no other content. Consider co-locating this with other constants if more are expected to follow, or documenting why this warrants its own module — currently its only purpose is to make the key testable, which could also be achieved by exporting it from the Vue SFC's `<script>` block or a broader `constants.ts` file.

jmelloy (commented):
- The defence-in-depth approach in `loadSavedSettings` is sound and correctly strips env var values from any stale localStorage data, closing the gap left by #533.
- Extracting `SETTINGS_KEY` to a shared constants file is a good refactor that enables testing without duplicating the key string.
- The `else` branch in `loadSavedSettings` silently resets `envVars` to `[]` when `envVars` is absent/undefined (not just malformed), which may be an over-aggressive reset for settings saved before env var support was added — these would previously have loaded fine with the implicit undefined.
- Test coverage is thorough for the localStorage safety cases; however the test suite describes 4 tests in the PR description but the spec file contains 5 `it` blocks (the non-object JSON and malformed JSON cases are listed separately).
- Minor: the `console.warn` in the else branch will fire for any valid pre-existing settings object that simply has no `envVars` property, potentially causing noise in production logs for a normal/expected condition.

jmelloy:
Closing this PR as it has been superseded by #535, which implemented moving env var values from localStorage to the server-side database. The changes in that PR cover the same functionality addressed here.

PR #538, opened 2026-06-03
fix: resolve merge conflicts from PR #532 (model/provider overrides)

jmelloy:
## Summary

This PR resolves the merge conflicts from PR #532 ("Add pi auth flow and per-task model/provider overrides") so the model/provider override changes can be merged.

- Rebased the three commits from PR #532 onto current `main`
- Resolved conflict in `backend/foreman/tools.py`: merged `model`/`provider` fields into the `TaskFollowupMsg` typed object (updated `ws_types.py` to add those fields)
- Resolved conflict in `backend/ws_handlers.py`: removed `pending_pi_auth` import (reverted with pi auth), kept `send_ws_message` which is still used
- Resolved conflict in `backend/routes/workers.py`: kept `broadcast_msg` style from main, dropped `pending_pi_auth`
- Resolved conflict in `worker/pioneer_worker/worker.py`: kept Control API methods (already in main), removed `_check_pi_auth` and `_run_pi_login` (pi auth revert)

## What's included

The three commits from PR #532:
1. `Add pi auth flow and per-task model/provider overrides`
2. `Fix model/provider silently dropped in task dispatch paths`
3. `Revert pi auth flow — keep model/provider overrides only`

Net result: per-task `model` and `provider` overrides are wired through the full dispatch path, pi auth flow is not included.


worker/pioneer_worker/worker.py
```diff
@@ -1966,12 +1970,22 @@ async def _collect_usage(rec: dict) -> None:
                         openai_api_key=self.cfg.openai_api_key,
```
jmelloy, line 1970: The precedence `task.get('model') or self.cfg.pi_model` will fall through to the config default whenever the task field is an empty string, which is fine here because `or None` is applied at ingestion in `tools.py`. However, if the ingestion guard is ever removed, a task explicitly sent with `model=''` will silently use the config default instead of sending no override. Consider using `task.get('model') if task.get('model') is not None else self.cfg.pi_model` to be explicit.

backend/foreman_core/tools_schema.py
```diff
@@ -70,6 +70,14 @@
                     "enum": ["claude", "codex", "pi"],
                     "description": "Coding agent to use. Default: claude.",
                 },
+                "model": {
+                    "type": "string",
+                    "description": "Model override for the chosen tool (e.g. 'claude-opus-4-8', 'o4-mini', 'gpt-4o'). Omit to use the worker's configured default.",
+                },
+                "provider": {
```
jmelloy, line 77: The example model name `claude-opus-4-8` appears to be a typo (likely `claude-opus-4` or `claude-3-5-sonnet`). Incorrect examples in the schema description can mislead the LLM when selecting model names.

backend/foreman_core/tools_schema.py
```diff
@@ -70,6 +70,14 @@
...
+                    "type": "string",
+                    "description": "Provider override for pi tasks (e.g. 'anthropic', 'openai', 'google'). Ignored for claude and codex.",
+                },
                 "parent_task_id": {
```
jmelloy, line 81: The `provider` field description says it is 'Ignored for claude and codex', yet the field is accepted, stored in the DB, and forwarded in all WebSocket messages for those tool types without any validation or warning. Consider either rejecting/warning when `provider` is supplied with a non-pi tool, or removing the 'ignored' caveat and documenting actual behaviour.

backend/alembic/versions/20260601_000000_add_model_provider_to_tasks.py
```diff
@@ -0,0 +1,31 @@
...
+"""
+
+from __future__ import annotations
+
+from collections.abc import Sequence
+
+import sqlalchemy as sa
+from alembic import op
+
+revision: str = "20260601_000000_add_model_provider_to_tasks"
```
jmelloy, line 18: The revision ID and file name use the date 2026-06-01, which is a future date. While Alembic uses the `down_revision` chain rather than filenames for ordering, a future-dated ID is unconventional and may confuse tooling or contributors. Consider using the actual current date or an auto-generated UUID-style ID.

jmelloy (commented):
- The PR cleanly extracts model/provider override fields from PR #532, threading them through the DB model, migration, API routes, WebSocket types, foreman tools, and worker — a well-scoped change with no leftover pi-auth artifacts visible in the diff.
- Precedence logic for model/provider (`task.get('model') or self.cfg.pi_model`) silently drops a task-level empty-string override in favour of the config default; however since `or None` is applied at ingestion this is unlikely to be a real issue in practice.
- The `tools_schema.py` description contains a typo (`claude-opus-4-8` looks like a mangled model name) and the `provider` field is documented as ignored for claude/codex but still accepted and stored for them — no validation or warning is applied.
- `TaskAssignedWS` in `frontend/src/types.ts` gains `tool`, `model`, and `provider` fields but they are not reflected in any analogous `TaskFollowupWS` frontend type (if one exists), which could cause silent type gaps on the frontend.
- The migration revision ID uses a date-based string (`20260601_000000_...`) with a future date (2026), which is inconsistent with typical timestamp-based alembic IDs and may cause ordering surprises if new migrations are added before that date.

PR #541, opened 2026-06-03
Include tools in worker-online event (#540)

### LINKED ISSUE #540: Include available tools/runners in worker-online event
## Summary

When a worker comes online and emits the `[worker-online]` event, it should include the list of tools/runners available on that worker (e.g. `claude`, `codex`, `pi`).

## Current behavior

The `[worker-online]` message currently looks like:
```
[worker-online] worker_id=w-831859 repos=jmelloy/pioneer-square agent_count=4
```

## Desired behavior

It should also include a `tools` field listing the enabled agent runners:
```
[worker-online] worker_id=w-831859 repos=jmelloy/pioneer-square agent_count=4 tools=claude,codex,pi
```

## Motivation

The Foreman needs to know which tools are available on each worker so it can route tasks to workers that support the required agent runner. This also helps with debugging and operational visibility.

## Scope

- Update the worker registration/startup code to include `tools` in the online event payload
- Update the Foreman's `[worker-online]` handling to read and display the `tools` field
- Update the worker state model to persist available tools alongside repos and agent_count

jmelloy:
## Summary

- **Worker**: At startup, after auth checks, calls `_detect_available_tools()` which uses `shutil.which()` to detect which runner binaries (claude/codex/pi) are on PATH and stores the result in `self._available_tools`. Both `worker-register` sends (initial join and repo-refresh) now include a `tools` field.
- **Backend model**: Added `tools: str` column to the `workers` table (JSON array, defaults to `[]`), with a new Alembic migration (`20260603_000001_add_tools_to_workers`).
- **WS handler**: `handle_worker_register` now reads the `tools` field from the message, persists it to the DB, and appends `tools=claude,codex,pi` to the `[worker-online]` foreman event string (omitted when empty for backwards compatibility).
- **Foreman state**: Both the embedded foreman (`backend/foreman/runner.py`) and the standalone foreman (`foreman/pioneer_foreman/runner.py`) now include `tools` in the workers JSON block passed to the Foreman AI. The `foreman/state` API endpoint also exposes the field.
- **Test**: `test_worker_online_notifies_foreman` now sends `"tools": ["claude", "codex"]` in the `worker-register` message and asserts `tools=claude,codex` appears in the triggered foreman message.

## Test plan

- [ ] `cd backend && python -m pytest tests/test_worker_foreman_notify.py -v`
- [ ] Start a worker and confirm the `[worker-online]` log line shows `tools=claude` (or whichever runners are installed)

Closes #540

### DIFF
--- a/backend/alembic/versions/20260603_000001_add_tools_to_workers.py
+++ b/backend/alembic/versions/20260603_000001_add_tools_to_workers.py
@@ -1,0 +1,32 @@
+"""Add tools column to workers table.
+
+Revision ID: 20260603_000001_add_tools_to_workers
+Revises: 20260603_000000_create_user_spawn_settings
+Create Date: 2026-06-03
+
+Stores the JSON-serialised list of available tool runners (e.g. ["claude", "codex", "pi"])
+that the worker reported at registration time.
+"""
+
+from __future__ import annotations
+
+from collections.abc import Sequence
+
+import sqlalchemy as sa
+from alembic import op
+
+revision: str = "20260603_000001_add_tools_to_workers"
+down_revision: str | Sequence[str] | None = "20260603_000000_create_user_spawn_settings"
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+
+def upgrade() -> None:
+    op.add_column(
+        "workers",
+        sa.Column("tools", sa.Text(), nullable=False, server_default="'[]'"),
+    )
+
+
+def downgrade() -> None:
+    op.drop_column("workers", "tools")
--- a/backend/alembic/versions/20260603_000001_add_tools_to_workers.py
+++ b/backend/alembic/versions/20260603_000001_add_tools_to_workers.py
@@ -16,7 +16,7 @@ import sqlalchemy as sa
 from alembic import op
 
 revision: str = "20260603_000001_add_tools_to_workers"
-down_revision: str | Sequence[str] | None = "20260603_000000_create_user_spawn_settings"
+down_revision: str | Sequence[str] | None = "20260601_000000_add_model_provider_to_tasks"
 branch_labels: str | Sequence[str] | None = None
 depends_on: str | Sequence[str] | None = None
 
--- a/backend/ws_handlers.py
+++ b/backend/ws_handlers.py
@@ -559,8 +559,9 @@ async def handle_worker_register(ctx: WSContext, data: dict) -> None:
     if not worker_id:
         return
     repos = data.get("repos") or []
+    tools = data.get("tools") or []
     user_ident = data.get("user")
-    update_vals: dict = {"repos": json.dumps(repos)}
+    update_vals: dict = {"repos": json.dumps(repos), "tools": json.dumps(tools)}
     if user_ident:
         resolved = await _resolve_user_identifier(ctx.db, user_ident)
         if resolved:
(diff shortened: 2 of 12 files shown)


worker/pioneer_worker/worker.py
```diff
@@ -358,6 +360,20 @@ async def _ensure_codex_api_key(self) -> None:
                 "Set openai_api_key in the [codex] config block or the OPENAI_API_KEY env var."
             )
 
+    def _detect_available_tools(self) -> None:
```
jmelloy, line 363: `shutil.which()` returns `None` when given a full absolute path to a non-existent file, but also returns the path itself when the file exists and is executable — regardless of PATH. If `cfg.claude_path` is an absolute path (a common configuration), this check behaves differently than intended: it will report the tool as available whenever the binary exists, not when it is on PATH. Consider separating the "is it a plain name?" vs "absolute path" cases, or documenting the expected format of these config fields.

backend/ws_handlers.py
```diff
@@ -559,8 +559,9 @@ async def handle_worker_register(ctx: WSContext, data: dict) -> None:
     if not worker_id:
         return
```
jmelloy, line 560: There is no validation or sanitisation of the `tools` list received from the worker. An untrusted or compromised worker could send arbitrary strings (including spaces, `=` signs, or SQL-special characters) that get stored in the DB and embedded verbatim into the foreman event string. At minimum, whitelist the accepted tool names (e.g. `{"claude", "codex", "pi"}`) before persisting.

worker/pioneer_worker/worker.py
```diff
@@ -358,6 +360,20 @@ async def _ensure_codex_api_key(self) -> None:
...
+        """Populate self._available_tools based on which runner binaries are on PATH."""
```
jmelloy: `import shutil` inside the method body works but is unconventional; standard Python style places all imports at the module top level. Move it there to improve readability and avoid re-importing on every call.

backend/alembic/versions/20260603_000001_add_tools_to_workers.py
```diff
@@ -0,0 +1,32 @@
...
+
+revision: str = "20260603_000001_add_tools_to_workers"
+down_revision: str | Sequence[str] | None = "20260603_000000_create_user_spawn_settings"
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+
+def upgrade() -> None:
+    op.add_column(
+        "workers",
```
jmelloy, line 26: The `server_default` value `"'[]'"` (a SQL literal with embedded single quotes) is correct for SQLite/PostgreSQL but can be surprising and fragile. Consider using `sa.text("'[]'")`  explicitly to make the intent clearer and ensure cross-dialect compatibility.

jmelloy (commented):
- The PR correctly threads tool detection through worker startup, registration, DB persistence, and foreman state, providing a clean end-to-end feature.
- Tool detection uses binary names from config paths (e.g. `self.cfg.claude_path`) as arguments to `shutil.which()`, which is correct only when those config values are plain binary names; full paths passed to `shutil.which()` return `None` unless the path itself is executable and on PATH, potentially causing silent false-negatives.
- The `tools` field in the foreman event string uses a space-delimited format (`tools=claude,codex`) appended without a leading space separator matching the rest of the string, which works but is fragile; a structural format (JSON) would be more robust for future parsing.
- No input validation is performed on the `tools` list received from the worker in `ws_handlers.py`; a malicious or misconfigured worker could inject arbitrary strings into the DB and foreman event string.
- The Alembic migration correctly adds a `NOT NULL` column with a server default, and the model definition mirrors it, though the `import shutil` inside the method body is a minor style issue (should be at module level).

PR #543, opened 2026-06-03
Fix pi_runner asyncio buffer overflow (#542)

### LINKED ISSUE #542: pi_runner crashes with ValueError when output line exceeds asyncio stream buffer limit
## Bug

The pi agent runner crashes when a subprocess stdout line exceeds asyncio's default 64 KB `StreamReader` limit:

```
asyncio.exceptions.LimitOverrunError: Separator is not found, and chunk exceed the limit
...
ValueError: Separator is not found, and chunk exceed the limit
```

**File:** `pioneer_worker/pi_runner.py`, line 117, in `run_pi_auto`

```python
async for raw in proc.stdout:  # iterates via readline(), default limit=64KB
```

## Root Cause

`asyncio.create_subprocess_exec` / `asyncio.create_subprocess_shell` creates a `StreamReader` with a default `limit=65536` (64 KB). Any single line of stdout longer than this raises `LimitOverrunError`, which propagates as a `ValueError` and crashes the runner.

## Fix

Pass a larger `limit` when creating the subprocess, e.g.:

```python
limit = 10 * 1024 * 1024  # 10 MB
proc = await asyncio.create_subprocess_exec(
    ...,
    stdout=asyncio.subprocess.PIPE,
    stderr=asyncio.subprocess.PIPE,
    limit=limit,
)
```

Alternatively, read stdout with `proc.stdout.read` / `proc.stdout.readuntil` in a loop and handle `LimitOverrunError` gracefully so the runner can continue rather than crash.

jmelloy:
## Summary

- Passes `limit=10 * 1024 * 1024` (10 MB) to `asyncio.create_subprocess_exec` so the subprocess `StreamReader` can handle stdout lines larger than asyncio's default 64 KB limit.
- Wraps the `async for raw in proc.stdout` loop in a `try/except (asyncio.LimitOverrunError, ValueError)` block — if a line still somehow exceeds the limit, the error is logged and emitted to the caller rather than crashing the whole runner.
- Adds `test_run_pi_auto_large_line_does_not_crash`: a fake pi script that writes a 2 MB line followed by `agent_end`, verifying the runner completes cleanly.

Closes #542

### DIFF
--- a/worker/pioneer_worker/pi_runner.py
+++ b/worker/pioneer_worker/pi_runner.py
@@ -128,11 +128,12 @@ async def run_pi_auto(
                     exc,
                 )
                 await emit(f"[pi] ✗ stdout line too large, skipping: {exc}")
-                # Consume the rest of the oversized line so the stream stays usable.
+                # Drain the rest of the oversized line one byte at a time so we
+                # stop exactly at the \n and don't consume bytes from the next line.
                 try:
                     while True:
-                        chunk = await proc.stdout.read(65536)  # type: ignore[union-attr]
-                        if not chunk or b"\n" in chunk:
+                        byte = await proc.stdout.read(1)  # type: ignore[union-attr]
+                        if not byte or byte == b"\n":
                             break
                 except Exception:
                     pass
--- a/worker/pioneer_worker/pi_runner.py
+++ b/worker/pioneer_worker/pi_runner.py
@@ -96,8 +96,13 @@ async def run_pi_auto(
             stdin=asyncio.subprocess.PIPE,
             stdout=asyncio.subprocess.PIPE,
             stderr=asyncio.subprocess.PIPE,
-            limit=_STREAM_LIMIT,
         )
+        # asyncio.create_subprocess_exec silently ignores the `limit` kwarg, so
+        # patch the StreamReader directly after creation.
+        if proc.stdout:
+            proc.stdout._limit = _STREAM_LIMIT
+        if proc.stderr:
+            proc.stderr._limit = _STREAM_LIMIT
         logger.info("pi subprocess started pid=%s", proc.pid)
 
         rpc_msg = json.dumps({"type": "prompt", "message": description}) + "\n"
--- a/worker/pioneer_worker/pi_runner.py
+++ b/worker/pioneer_worker/pi_runner.py
@@ -93,6 +93,7 @@ async def run_pi_auto(
             stdin=asyncio.subprocess.PIPE,
             stdout=asyncio.subprocess.PIPE,
             stderr=asyncio.subprocess.PIPE,
+            limit=10 * 1024 * 1024,  # 10 MB — prevents LimitOverrunError on large lines
         )
         logger.info("pi subprocess started pid=%s", proc.pid)
 
@@ -114,48 +115,54 @@ async def run_pi_auto(
         stderr_task = asyncio.create_task(_drain_stderr())
         accumulated = ""
 
-        async for raw in proc.stdout:  # type: ignore[union-attr]
-            line_str = raw.decode(errors="replace").strip()
-            if not line_str:
-                continue
-            event_count += 1
-            logger.debug("pi[%d] stdout#%d: %s", proc.pid, event_count, line_str)
-            try:
-                event = json.loads(line_str)
-            except json.JSONDecodeError:
-                await emit(line_str)
-                continue
-            etype = event.get("type")
-            if (
-                etype == "response"
-                and event.get("command") == "prompt"
-                and not event.get("success", True)
-            ):
-                # Prompt was rejected before acceptance.
-                err = event.get("error", "prompt rejected")
-                await emit(f"[pi] ✗ {err}")
-                saw_error = True
-            if etype == "agent_end":
-                saw_agent_end = True
-                if accumulated.strip():
-                    last_text = accumulated
-                accumulated = ""
-            if etype == "message_update":
-                ame = event.get("assistantMessageEvent", {})
-                if ame.get("type") == "error":
+        try:
+            async for raw in proc.stdout:  # type: ignore[union-attr]
+                line_str = raw.decode(errors="replace").strip()
+                if not line_str:
+                    continue
+                event_count += 1
+                logger.debug("pi[%d] stdout#%d: %s", proc.pid, event_count, line_str)
+                try:
+                    event = json.loads(line_str)
+                except json.JSONDecodeError:
+                    await emit(line_str)
+                    continue
+                etype = event.get("type")
+                if (
+                    etype == "response"
+                    and event.get("command") == "prompt"
+                    and not event.get("success", True)
+                ):
+                    # Prompt was rejected before acceptance.
+                    err = event.get("error", "prompt rejected")
+                    await emit(f"[pi] ✗ {err}")
                     saw_error = True
-                    reason = ame.get("reason", "error")
-                    await emit(f"[pi] ✗ {reason}")
-            text, accumulated = parse_pi_event(event, accumulated)
-            if text:
-                await emit(text)
-            if etype == "message_update" and accumulated.strip():
-                last_text = accumulated
-            # We only send a single prompt; pi RPC mode stays alive waiting
-            # for more stdin after the run completes, so stop reading once the
-            # agent (or a fatal error) has finished.
-            if saw_agent_end or saw_error:
-                break
+                if etype == "agent_end":
+                    saw_agent_end = True
+                    if accumulated.strip():
+                        last_text = accumulated
+                    accumulated = ""
+                if etype == "message_update":
+                    ame = event.get("assistantMessageEvent", {})
+                    if ame.get("type") == "error":
+                        saw_error = True
+                        reason = ame.get("reason", "error")
+                        await emit(f"[pi] ✗ {reason}")
+                text, accumulated = parse_pi_event(event, accumulated)
+                if text:
+                    await emit(text)
+                if etype == "message_update" and accumulated.strip():
+                    last_text = accumulated
+                # We only send a single prompt; pi RPC mode stays alive waiting
+                # for more stdin after the run completes, so stop reading once the
+                # agent (or a fatal error) has finished.
+                if saw_agent_end or saw_error:
+                    break
+        except (asyncio.LimitOverrunError, ValueError) as exc:
+            logger.error(
+                "pi[%d] stdout line exceeded StreamReader limit: %s", proc.pid, exc
+            )
+            await emit(f"[pi] ✗ stdout line too large, skipping remainder: {exc}")
 
         # Closing stdin signals EOF so the RPC process exits cleanly.
         if proc.stdin and not proc.stdin.is_closing():
--- a/worker/pioneer_worker/pi_runner.py
+++ b/worker/pioneer_worker/pi_runner.py
@@ -14,6 +14,9 @@ EmitFn = Callable[[str], Awaitable[None]]
 # Seconds to wait for pi to exit after stdin is closed before killing it.
 _WAIT_TIMEOUT = 30
 
+# StreamReader buffer size — large enough to handle big tool outputs.
+_STREAM_LIMIT = 10 * 1024 * 1024  # 10 MB
+
 
 def _result_text(result: dict) -> str:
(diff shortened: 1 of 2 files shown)


worker/pioneer_worker/pi_runner.py
```diff
@@ -114,48 +115,54 @@ async def _drain_stderr() -> None:
...
+                if saw_agent_end or saw_error:
+                    break
+        except (asyncio.LimitOverrunError, ValueError) as exc:
+            logger.error(
+                "pi[%d] stdout line exceeded StreamReader limit: %s", proc.pid, exc
+            )
+            await emit(f"[pi] ✗ stdout line too large, skipping remainder: {exc}")
 
         # Closing stdin signals EOF so the RPC process exits cleanly.
         if proc.stdin and not proc.stdin.is_closing():
```
jmelloy, line 186: `asyncio.create_subprocess_exec` does not accept a `limit` keyword argument directly — it belongs on `asyncio.open_connection` / `StreamReaderProtocol`. In CPython this kwarg is silently ignored, meaning the actual StreamReader limit remains 64 KB and the fix does not work. Verify with `proc.stdout._limit` in a test, or use `asyncio.subprocess.Process` with a manually constructed `StreamReaderProtocol`.

worker/pioneer_worker/pi_runner.py
```diff
@@ -114,48 +115,54 @@ async def _drain_stderr() -> None:
...
+                # We only send a single prompt; pi RPC mode stays alive waiting
+                # for more stdin after the run completes, so stop reading once the
+                # agent (or a fatal error) has finished.
+                if saw_agent_end or saw_error:
+                    break
+        except (asyncio.LimitOverrunError, ValueError) as exc:
+            logger.error(
+                "pi[%d] stdout line exceeded StreamReader limit: %s", proc.pid, exc
+            )
+            await emit(f"[pi] ✗ stdout line too large, skipping remainder: {exc}")
```
jmelloy: The `except` block exits the entire `async for` loop on the first oversized line, so any subsequent lines — including `agent_end` — are never processed. `saw_agent_end` stays `False`, and the caller likely receives a misleading `no_events` or error outcome even when the subprocess completed successfully. Consider skipping only the offending chunk via `readuntil`/`read` with a size cap, or at minimum setting `saw_error = True` and breaking cleanly before emitting.

worker/tests/test_pi_runner.py
```diff
@@ -320,6 +320,36 @@ async def emit(line: str) -> None:
...
+    async def emit(line: str) -> None:
+        emitted.append(line)
+
+    # Should complete without raising — either success or graceful error.
+    success, stop_reason, _ = await run_pi_auto(
+        "do the work", str(tmp_path), emit=emit, pi_path=os.fspath(fake_pi)
+    )
+
+    # The runner must not crash; success is acceptable if the limit is raised,
+    # but a logged error is also fine as long as it returns cleanly.
```
jmelloy: The assertion accepts `"success"`, `"error_during_execution"`, or `"no_events"`, making the test pass regardless of whether the fix actually works. Because the 2 MB line is below the new 10 MB limit, the expected outcome is deterministically `"success"`; asserting exactly that would catch regressions where the limit parameter is silently ignored.

jmelloy (commented):
- The 10 MB `limit` parameter correctly addresses the root cause by raising asyncio's `StreamReader` buffer ceiling, and the fallback `except` block provides a safety net for any remaining edge cases.
- The `except (asyncio.LimitOverrunError, ValueError)` handler aborts the entire stdout loop on a single oversized line, meaning subsequent valid lines (including `agent_end`) are silently dropped and state flags like `saw_agent_end` are never set — the caller may receive a misleading outcome.
- The test's assertion `assert stop_reason in ("success", "error_during_execution", "no_events")` is too permissive; since the 2 MB line is well within the new 10 MB limit, the test should expect a deterministic `"success"` outcome, making it a meaningful regression guard.
- The `limit` parameter is passed to `asyncio.create_subprocess_exec` but that function does not accept a `limit` kwarg directly; it must be forwarded through the subprocess protocol/transport layer. Depending on the Python version and implementation this may silently be ignored or raise a `TypeError`.
- No explicit upper bound is enforced on individual line size before decoding/JSON-parsing, so a legitimately large but valid JSON blob will be decoded into memory in full; a memory-pressure consideration worth documenting.

worker/pioneer_worker/pi_runner.py
```diff
@@ -94,6 +97,12 @@ async def run_pi_auto(
             stdout=asyncio.subprocess.PIPE,
             stderr=asyncio.subprocess.PIPE,
         )
+        # asyncio.create_subprocess_exec silently ignores the `limit` kwarg, so
```
jmelloy: Mutating `proc.stdout._limit` is accessing a private, undocumented CPython implementation detail that can break silently on any Python version upgrade. `asyncio.create_subprocess_exec` accepts `limit=` as a keyword argument in Python 3.10+ — pass `limit=_STREAM_LIMIT` directly to the call above instead, and only fall back to this mutation (with a comment) if you need to support older versions.

worker/pioneer_worker/pi_runner.py
```diff
@@ -114,7 +123,27 @@ async def _drain_stderr() -> None:
...
+        while True:
+            try:
+                raw = await proc.stdout.readline()  # type: ignore[union-attr]
+            except (asyncio.LimitOverrunError, ValueError) as exc:
+                logger.error(
+                    "pi[%d] stdout line exceeded StreamReader limit, skipping line: %s",
+                    proc.pid,
+                    exc,
+                )
+                await emit(f"[pi] ✗ stdout line too large, skipping: {exc}")
```
jmelloy, line 130: This drain loop is incorrect. After `LimitOverrunError`, the oversized line is still sitting in the StreamReader's internal buffer; calling `read(65536)` will return chunks but a subsequent `readline()` will immediately raise `LimitOverrunError` again on the remaining bytes of the same line. You should call `proc.stdout.readline()` inside the drain loop (catching `LimitOverrunError` again) or use `proc.stdout.read(n)` repeatedly and track newlines manually until a `\n` is consumed from the buffer.

worker/tests/test_pi_runner.py
```diff
@@ -320,6 +320,73 @@ async def emit(line: str) -> None:
...
+    )
+
+    # The 2 MB line is below the 10 MB _STREAM_LIMIT, so the runner completes cleanly.
+    assert stop_reason == "success"
+
+
+async def test_run_pi_auto_oversized_line_continues_to_agent_end(tmp_path, monkeypatch) -> None:
+    """After an oversized line (LimitOverrunError), subsequent lines including
+    agent_end must still be processed — the runner must not hang or lose events."""
+    import pioneer_worker.pi_runner as pi_runner_mod
```
jmelloy: Monkeypatching `pi_runner_mod._STREAM_LIMIT` after the module is imported does not affect the subprocess that is about to be created because the `_limit` mutation on `proc.stdout` happens inside `run_pi_auto` using whatever value `_STREAM_LIMIT` holds at call time — but the subprocess itself reads from the OS pipe, and the StreamReader limit is set at subprocess-creation time. Verify this test actually exercises the `LimitOverrunError` path (add an assertion on `emitted` before the `stop_reason` check, and consider a debug log inside the except block to confirm it fires).

jmelloy (commented):
- The core fix (increasing the StreamReader limit) is implemented by directly mutating the private `_limit` attribute on `asyncio.StreamReader`, which is fragile and undocumented; the `limit` kwarg is actually supported by `create_subprocess_exec` in Python 3.10+ and should be used instead, with the monkey-patch as a documented fallback.
- The oversized-line recovery logic in the `except` block is incorrect: after a `LimitOverrunError`, the unconsumed bytes are still buffered in the StreamReader and must be drained using `readline(limit=...)` or `read()` until a newline is found, but the current `read(65536)` loop will re-trigger `LimitOverrunError` on the next `readline()` call if the line is extremely large, potentially causing an infinite loop or swallowed data.
- The second test (`test_run_pi_auto_oversized_line_continues_to_agent_end`) monkeypatches `_STREAM_LIMIT` but the subprocess has already been created with the original limit at that point — the monkeypatch only affects the module-level constant read at process start, so it may not reliably trigger the error path it claims to test.
- Both new tests lack the `@pytest.mark.asyncio` decorator (or equivalent runner registration), which is required by the existing test suite style shown in the diff context; they will likely be silently skipped rather than executed.
- Overall the intent of the PR is correct and the test coverage is appreciated, but the private-attribute mutation and the broken drain loop are production-quality concerns that should be addressed before merging.

worker/pioneer_worker/pi_runner.py
jmelloy: The comment is incorrect: `asyncio.create_subprocess_exec` does accept and honour the `limit` keyword argument (it is forwarded to the underlying `StreamReader`). The private `_limit` mutation is both unnecessary and fragile — it depends on a CPython implementation detail that is not part of the public API. Replace with `limit=_STREAM_LIMIT` in the `create_subprocess_exec` call and remove these lines.

worker/pioneer_worker/pi_runner.py
```diff
@@ -114,7 +123,27 @@ async def _drain_stderr() -> None:
...
+                # Consume the rest of the oversized line so the stream stays usable.
+                try:
+                    while True:
```
jmelloy, line 134: This drain loop is incorrect. After a `LimitOverrunError`, `asyncio.StreamReader` has already buffered data up to the limit; calling `read(65536)` will consume bytes from the *start of the next line*, not just the remainder of the oversized one. The correct recovery is `await proc.stdout.readuntil(b'\n')` wrapped in another `LimitOverrunError` catch, or simply calling `readline()` again (the error already consumed up to the limit boundary). As written, this silently drops the beginning of the following line, which could be `agent_end`, causing the runner to hang.

worker/tests/test_pi_runner.py
```diff
@@ -320,6 +320,73 @@ async def emit(line: str) -> None:
...
+    fake_pi = tmp_path / "fake-pi"
+    fake_pi.write_text(
+        f"""#!{sys.executable}
+import json, sys
+# This 256-byte line exceeds the patched 128-byte limit.
+print("x" * 256, flush=True)
+# agent_end must still be received even after the oversized line.
+print(json.dumps({{"type": "agent_end"}}), flush=True)
+sys.exit(0)
+""",
```
jmelloy: Monkeypatching `_STREAM_LIMIT` at the module level only affects the `proc.stdout._limit = _STREAM_LIMIT` assignment that runs *after* subprocess creation. This means the test is entirely dependent on the private `_limit` hack working correctly, making the test brittle. If the implementation is fixed to pass `limit=` to `create_subprocess_exec`, this test will stop triggering the overflow path without further changes.

jmelloy (commented):
- **Critical bug**: The comment on line 100 claims `asyncio.create_subprocess_exec` silently ignores the `limit` kwarg, but that is false — `limit` is a supported parameter passed through to `asyncio.create_subprocess_shell`/`exec`. The private `_limit` mutation is unnecessary, fragile, and relies on CPython internals that could break across Python versions.
- **Broken oversized-line recovery**: The drain loop after a `LimitOverrunError` uses `read()` (byte-oriented) to skip to the next newline, but this is unreliable — `read()` may return data from the *next* well-formed line, silently discarding valid messages including a potential `agent_end`, risking a hang or lost results.
- **Test design flaw**: `test_run_pi_auto_oversized_line_continues_to_agent_end` monkeypatches `_STREAM_LIMIT` but the subprocess is already created with the original limit baked into the `StreamReader`; the patched value only affects `proc.stdout._limit = _STREAM_LIMIT` post-creation, so the test is actually exercising the `_limit` mutation path correctly — but this makes the test tightly coupled to the private-attribute hack.
- **Missing `limit` kwarg on `create_subprocess_exec`**: Even if the private `_limit` patch works today, the proper fix is simply passing `limit=_STREAM_LIMIT` to `create_subprocess_exec`, which is documented and supported, making the post-creation mutation entirely redundant.
- **Error recovery emits raw exception text to callers**: Surfacing `exc` (the `LimitOverrunError`/`ValueError` message) directly in the user-visible `emit` string leaks internal implementation details; a sanitised message without the exception repr would be safer.

PR #545, opened 2026-06-03
Pi runner: error reporting, cost tracking, remove no-session

### LINKED ISSUE #544: Audit pi runner: proper error/crash reporting + cost tracking + remove no-session mode
## Summary

The `pi_runner.py` needs several reliability and observability improvements:

### 1. Error & Crash Reporting
- Audit how `pi_runner.py` handles subprocess crashes, non-zero exit codes, and exceptions
- Ensure errors are surfaced back to the caller (web service / task system) rather than silently swallowed
- Check that stderr from the `pi` process is captured and forwarded as an error chunk
- Verify that partial output followed by a crash is handled gracefully (not just EOF-silenced)

### 2. Remove "no session" mode
- Identify and remove any code paths that allow running without a session
- Ensure every invocation is tied to a tracked session

### 3. Cost Tracking
- Add cost tracking for `pi` invocations — parse token/cost info from `pi` output or API responses
- Store/emit cost data alongside task results so it's visible in the UI or logs
- Follow the pattern used by other runners (claude, codex) if they already track cost

### Acceptance Criteria
- All crashes and non-zero exits from `pi` are reported as errors upstream
- No code path allows a session-less invocation
- Cost per invocation is captured and stored/emitted
- Tests cover the new error reporting and cost tracking paths

jmelloy:
## Summary

- **Error & crash reporting**: `run_pi_auto` now emits an explicit `[pi] ✗ process exited with non-zero code N` chunk whenever `pi` exits non-zero without a clean `agent_end`. Stderr was already forwarded via `[stderr]` prefix. Partial output before a crash is preserved in `last_text`.
- **Remove no-session mode**: Dropped `--no-session` from the `pi` command. `run_pi_auto` now extracts `session_id` from `agent_start` and `session` events (both `session_id` and `sessionId` spellings). The return type is now a 4-tuple `(success, stop_reason, last_text, session_id)` matching `claude_runner`. `worker.py` captures the session ID into `resume_session_id` so it appears in task-completion messages and usage reports.
- **Cost tracking**: Added `on_usage: UsageFn | None` parameter to `run_pi_auto`. Per-message token counts from `message_update` events are forwarded as `{"kind": "api_call", ...}` records; final cost/token data from `agent_end` is forwarded as `{"kind": "result", ...}`. Both camelCase (`inputTokens`) and snake_case (`input_tokens`) field names are supported. `worker.py` passes `on_usage=_collect_usage` so pi usage is included in the existing usage-reporting pipeline.
- **Tests**: All 14 existing tests updated for the new 4-tuple return value; 10 new tests added covering crash/non-zero exit, session ID extraction (both field spellings), absence of `--no-session` flag, and cost/usage callback for `agent_end`, `message_update`, snake_case fields, and the no-callback path.

## Test plan

- [x] `pytest worker/tests/test_pi_runner.py` — 33/33 passed
- [x] `python -c "from pioneer_worker import worker"` — imports cleanly

Closes #544

### DIFF
--- a/worker/pioneer_worker/pi_runner.py
+++ b/worker/pioneer_worker/pi_runner.py
@@ -226,7 +226,7 @@ async def run_pi_auto(
                             "model": event.get("model"),
                             "cost_usd": cost,
                             "stop_reason": "success",
-                            **( usage_data or {}),
+                            **(usage_data or {}),
                         }
                         await on_usage(rec)
             if etype == "message_update":
--- a/worker/pioneer_worker/pi_runner.py
+++ b/worker/pioneer_worker/pi_runner.py
@@ -37,23 +37,19 @@ def _parse_pi_usage(event: dict) -> dict | None:
     usage = event.get("usage") or event.get("tokenUsage")
     if not isinstance(usage, dict):
         return None
+    v = usage.get("inputTokens")
+    input_tokens = v if v is not None else usage.get("input_tokens", 0)
+    v = usage.get("outputTokens")
+    output_tokens = v if v is not None else usage.get("output_tokens", 0)
+    v = usage.get("cacheReadInputTokens")
+    cache_read = v if v is not None else usage.get("cache_read_input_tokens", 0)
+    v = usage.get("cacheCreationInputTokens")
+    cache_creation = v if v is not None else usage.get("cache_creation_input_tokens", 0)
     return {
-        "input_tokens": int(
-            usage.get("inputTokens") or usage.get("input_tokens") or 0
-        ),
-        "output_tokens": int(
-            usage.get("outputTokens") or usage.get("output_tokens") or 0
-        ),
-        "cache_read_input_tokens": int(
-            usage.get("cacheReadInputTokens")
-            or usage.get("cache_read_input_tokens")
-            or 0
-        ),
-        "cache_creation_input_tokens": int(
-            usage.get("cacheCreationInputTokens")
-            or usage.get("cache_creation_input_tokens")
-            or 0
-        ),
+        "input_tokens": int(input_tokens),
+        "output_tokens": int(output_tokens),
+        "cache_read_input_tokens": int(cache_read),
+        "cache_creation_input_tokens": int(cache_creation),
     }
 
 
@@ -285,13 +281,15 @@ async def run_pi_auto(
             stop_reason = "success"
         # Emit an explicit error when the subprocess exits non-zero without a
         # clean agent_end so callers see a failure rather than a silent EOF.
+        # Guard with `not saw_error` to avoid emitting a second error when the
+        # message_update handler already forwarded one.
         agent_ended_ok = saw_agent_end and not saw_error
-        if exit_code != 0 and not agent_ended_ok:
+        if exit_code != 0 and not agent_ended_ok and not saw_error:
             msg = f"[pi] ✗ process exited with non-zero code {exit_code}"
             logger.error("pi[%d] %s", proc.pid if proc else "?", msg)
             await emit(msg)
-            if not saw_error:
-                stop_reason = "error_during_execution"
+        if exit_code != 0 and not agent_ended_ok:
+            stop_reason = "error_during_execution"
         return (exit_code == 0 and agent_ended_ok), stop_reason, last_text, session_id
     except FileNotFoundError:
         logger.error("`pi` CLI not found on PATH")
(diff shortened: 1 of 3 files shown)


worker/pioneer_worker/pi_runner.py
```diff
@@ -27,6 +28,35 @@ def _result_text(result: dict) -> str:
...
+def _parse_pi_usage(event: dict) -> dict | None:
+    """Extract token counts from a pi event's usage block, if present.
+
+    Pi may use camelCase or snake_case field names depending on version.
+    Returns None when no usage block is found.
+    """
+    usage = event.get("usage") or event.get("tokenUsage")
+    if not isinstance(usage, dict):
+        return None
+    return {
```
jmelloy, line 48: Using `or` to chain fallbacks is buggy when token counts can legitimately be `0`: `usage.get('inputTokens') or usage.get('input_tokens') or 0` will skip `input_tokens` if `inputTokens` exists but is `0`. Use `is not None` checks instead, e.g. `v = usage.get('inputTokens'); result = v if v is not None else usage.get('input_tokens', 0)`.

worker/pioneer_worker/pi_runner.py
```diff
@@ -204,16 +283,24 @@ async def _drain_stderr() -> None:
             stop_reason = "success"
         elif exit_code == 0:
             stop_reason = "success"
+        # Emit an explicit error when the subprocess exits non-zero without a
+        # clean agent_end so callers see a failure rather than a silent EOF.
         agent_ended_ok = saw_agent_end and not saw_error
-        return (exit_code == 0 and agent_ended_ok), stop_reason, last_text
+        if exit_code != 0 and not agent_ended_ok:
+            msg = f"[pi] ✗ process exited with non-zero code {exit_code}"
```
jmelloy, line 288: The non-zero exit error is emitted even when `saw_error` is already True (a `message_update` error was already forwarded). This means callers receive two error messages. Consider guarding with `if exit_code != 0 and not agent_ended_ok and not saw_error:` and keeping the `stop_reason` update separate.

worker/pioneer_worker/pi_runner.py
```diff
@@ -165,12 +220,36 @@ async def _drain_stderr() -> None:
                 if accumulated.strip():
                     last_text = accumulated
                 accumulated = ""
+                # Emit final usage record if pi includes cost/token data here.
+                if on_usage is not None:
+                    usage_data = _parse_pi_usage(event)
```
jmelloy, line 221: The `agent_end` usage/cost block is nested inside the `etype == 'response' and command == 'prompt'` branch (based on surrounding diff context), not the `agent_end` event. Verify this is intentional — if pi emits usage on `agent_end` events, this block will never fire, silently dropping cost data.

worker/pioneer_worker/worker.py
```diff
@@ -2032,14 +2032,17 @@ async def _collect_usage(rec: dict) -> None:
...
+                    success, stop_reason, last_msg, _pi_session_id = await pi_runner.run_pi_auto(
                         current_desc,
                         primary_wt,
                         emit=emit,
                         pi_path=self.cfg.pi_path,
                         model=_pi_model,
                         provider=_pi_provider,
+                        on_usage=_collect_usage,
                     )
+                    if _pi_session_id:
```
jmelloy: `resume_session_id` is assigned from `_pi_session_id` only when truthy, but if a previous Claude run already set `resume_session_id`, a subsequent pi run with no session event will leave the stale Claude session ID in place. Consider explicitly setting `resume_session_id = _pi_session_id` (even if None) to avoid cross-runner contamination.

jmelloy (commented):
- The PR successfully adds error reporting, cost tracking, and session ID extraction with comprehensive test coverage for all three features.
- A subtle bug exists in `_parse_pi_usage`: using `or` to chain fallbacks means a value of `0` in the first field will incorrectly fall through to the next field name, potentially reading the wrong key.
- The `agent_end` usage handling fires inside the `etype == 'response'` block but checks `etype == 'message_update'` separately — however the usage block for `agent_end` is actually triggered on `agent_end` events processed earlier in the loop via a different branch not clearly shown, which may cause confusion or missed events depending on the full code structure.
- The non-zero exit error emission at the end correctly guards with `not agent_ended_ok`, but also fires when `saw_error` is True (already emitted an error), leading to a double-error scenario that could confuse callers.
- Test coverage is thorough and well-structured; the fake-pi scripts are a clean pattern for subprocess testing.

Issue #546, opened 2026-06-03
Debug endpoint: unified task timeline from DB, foreman, API calls, GitHub events


jmelloy:
## Summary

Add a debug endpoint (e.g. `GET /api/debug/task/{task_id}`) that assembles a single cohesive narrative of everything that happened for a given task.

## Data sources to aggregate

1. **Database** — task record, state transitions, branch/PR URL, worker assignment, timestamps
2. **Foreman logs** — all foreman messages/decisions referencing this task_id
3. **API call log** — every tool call made by the foreman or worker for this task (create_task, assign_task, send_followup, redirect_task, etc.)
4. **Worker logs** — agent subprocess output, stdout/stderr, exit codes
5. **Commits** — git commits on the task's branch (via GitHub API)
6. **GitHub PR timeline** — PR events, review submissions, status checks, webhooks received (check_runs, check_suites, pull_request events), comments
7. **Cost/token usage** — any recorded cost data tied to the task session

## Output format

Return a chronologically sorted list of events, each with:
- `timestamp`
- `source` (db | foreman | worker | github | api)
- `event_type` (e.g. `task_created`, `pr_opened`, `check_run_completed`, `commit_pushed`, etc.)
- `summary` (human-readable one-liner)
- `detail` (full raw payload, collapsible in UI)

## UI (optional stretch)

If there's a task detail page, add a "Timeline" tab that renders these events as a vertical timeline with icons per source.

## Purpose

Helps operators understand exactly what happened for a task end-to-end — useful for debugging failures, reviewing agent behaviour, and auditing cost.

PR #547, opened 2026-06-04
Debug endpoint: unified task timeline (#546)

jmelloy:
## Summary

- Adds `GET /guilds/{guild_id}/tasks/{task_id}/debug` that returns a unified, chronologically sorted task timeline
- Aggregates 5 data sources: DB task record + lifecycle events, worker/agent logs (`task_logs`), stored GitHub webhook events (`github_events`), Claude usage/cost records (`claude_usage`), and live GitHub API data (PR metadata, commits, reviews, check runs, comments, timeline events)
- GitHub API calls are lazy — skipped entirely when the task has no `pr_url`; individual API failures surface as `warnings` in the response rather than causing a 500
- Token resolution: uses `GITHUB_TOKEN` env var if set, otherwise falls back to the requesting user's OAuth token from `github_tokens`
- Follows existing route patterns (guild-scoped path, `require_member()` auth, `get_db_dep`, `row_to_dict`)

## Response shape

```json
{
  "task": { /* full task record */ },
  "timeline": [
    {
      "timestamp": "ISO-8601",
      "source": "db | worker | foreman | api | github",
      "event_type": "task_created | log | pr_opened | commit_pushed | check_run_completed | ...",
      "summary": "Human-readable one-liner",
      "detail": { /* raw payload */ }
    }
  ],
  "warnings": ["..."]   // optional
}
```

## Test plan

- `test_debug_requires_auth` — 401 without Bearer token
- `test_debug_unknown_guild_returns_403_or_404` — non-existent guild
- `test_debug_unknown_task_returns_404` — valid guild, no such task
- `test_debug_basic_structure` — full response shape, sorted timeline, required fields
- `test_debug_includes_usage` — Claude usage rows appear with correct summary
- `test_debug_task_in_wrong_guild_returns_404` — cross-guild task isolation
- `test_debug_non_member_forbidden` — 403 for non-members
- `test_debug_skips_github_when_no_pr_url` — no warnings or GitHub entries when no PR

Tests require `docker compose up -d postgres-test` (localhost:5433). GitHub API calls are not exercised in tests because test tasks have no `pr_url`.

### DIFF
--- a/backend/routes/debug.py
+++ b/backend/routes/debug.py
@@ -1,0 +1,445 @@
+"""Debug endpoint: unified task timeline from all data sources.
+
+Returns a single chronologically sorted list of everything that happened for a
+given task: DB state transitions, worker logs, GitHub webhook events, Claude
+usage records, and live GitHub API data (PR metadata, commits, reviews, checks,
+comments) if the task has a ``pr_url``.
+"""
+
+from __future__ import annotations
+
+import json
+import logging
+import os
+import re
+from datetime import UTC, datetime
+from typing import Any
+
+import httpx
+from auth_deps import get_guild_pk, require_member
+from database import get_db_dep
+from fastapi import APIRouter, Depends, HTTPException
+from models import ClaudeUsage, GithubEvent, GithubToken, Task, TaskLog
+from sqlmodel import col, select
+from sqlmodel.ext.asyncio.session import AsyncSession
+from utils import row_to_dict
+
+router = APIRouter()
+logger = logging.getLogger(__name__)
+
+_GITHUB_API = "https://api.github.com"
+
+
+# ---------------------------------------------------------------------------
+# Helpers
+# ---------------------------------------------------------------------------
+
+
+def _parse_pr_url(pr_url: str) -> tuple[str, str, int] | None:
+    """Return (owner, repo, pr_number) from a GitHub PR URL, or None."""
+    m = re.match(r"https://github\.com/([^/]+)/([^/]+)/pull/(\d+)", pr_url)
+    if not m:
+        return None
+    return m.group(1), m.group(2), int(m.group(3))
+
+
+def _iso(dt: datetime | None) -> str | None:
+    """Return an ISO-8601 UTC string for *dt*, or None."""
+    if dt is None:
+        return None
+    if dt.tzinfo is None:
+        dt = dt.replace(tzinfo=UTC)
+    return dt.isoformat()
+
+
+def _parse_ts(ts: str | None) -> datetime:
+    """Parse an ISO-8601 timestamp string for sorting; unknown → far future."""
+    if not ts:
+        return datetime.max.replace(tzinfo=UTC)
+    try:
+        return datetime.fromisoformat(ts.replace("Z", "+00:00"))
+    except ValueError:
+        return datetime.max.replace(tzinfo=UTC)
+
+
+async def _gh_get(client: httpx.AsyncClient, url: str, **kwargs: Any) -> tuple[Any, str | None]:
+    """GET *url* via *client*; return (data, error_string)."""
+    try:
+        r = await client.get(url, **kwargs)
+        r.raise_for_status()
+        return r.json(), None
+    except httpx.HTTPStatusError as exc:
+        return None, f"{url}: HTTP {exc.response.status_code}"
+    except httpx.RequestError as exc:
+        return None, f"{url}: {exc}"
+
+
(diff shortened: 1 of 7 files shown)


jmelloy (commented):
- Overall the implementation is clean, well-structured, and follows existing route patterns; auth, guild scoping, and graceful GitHub API failure handling are all correct.
- **Security concern**: The GITHUB_TOKEN environment variable (likely a server-side installation/bot token) is exposed in responses via the `detail` field of GitHub API events which echo back raw PR data, and more critically the token selection logic can silently fall back to another user's OAuth token if the env var is absent — the token ownership lookup should be scoped to the authenticated user which it is, but the env var token sharing across all users is worth flagging.
- **Data leak risk**: The `detail` field on GitHub API timeline entries includes the full raw API response, which may contain private email addresses, internal reviewer notes, and other sensitive fields that callers shouldn't receive wholesale.
- The `_parse_ts` helper uses `datetime.max.replace(tzinfo=UTC)` but `datetime.max` cannot have a timezone applied on Python < 3.11 in some edge cases; more importantly, timeline entries with `None` timestamps from live GitHub data will silently sort to the end rather than being flagged.
- The PR timeline API endpoint uses the deprecated `mockingbird-preview` Accept header and fires without combining it with the base headers (the `headers=` kwarg to `_gh_get` replaces rather than merges the client's default headers, dropping the Authorization header for that one request).

backend/routes/debug.py
```diff
@@ -0,0 +1,445 @@
...
+            )
+            if err:
+                warnings.append(f"Check runs fetch failed: {err}")
+            elif check_data:
+                for cr in check_data.get("check_runs", []):
+                    ts = cr.get("completed_at") or cr.get("started_at")
+                    status = cr.get("status", "")
+                    conclusion = cr.get("conclusion") or status
+                    etype = "check_run_completed" if status == "completed" else "check_run_started"
+                    events.append(
```
jmelloy, line 242: The `headers=tl_headers` kwarg passed to `_gh_get` is forwarded directly to `client.get(..., headers=tl_headers)`, which **overrides** (not merges) the `AsyncClient`'s default headers. This means the `Authorization` token is silently dropped for the timeline request, causing 401s for private repos. Merge the headers instead: `headers={**base_headers, **tl_headers}` and pass them at call-site or construct the merged dict inside `_fetch_github_timeline`.

backend/routes/debug.py
```diff
@@ -0,0 +1,445 @@
...
+            params={"per_page": 100},
+        )
+        if err:
+            warnings.append(f"Commits fetch failed: {err}")
+        elif commits:
+            for c in commits:
+                commit_info = c.get("commit") or {}
+                ts = (commit_info.get("committer") or {}).get("date") or (
+                    commit_info.get("author") or {}
+                ).get("date")
```
jmelloy, line 181: The full raw GitHub API `detail` payload is returned to the caller. For commits and PR data this can include committer/author email addresses, sensitive CI output, and internal org data. Consider stripping or projecting to only the fields actually needed for debugging rather than echoing the entire API response.

backend/routes/debug.py
```diff
@@ -0,0 +1,445 @@
...
+):
+    """Unified debug timeline for a task.
+
+    Aggregates all data sources — DB rows, worker logs, GitHub webhook events,
+    Claude usage records, and live GitHub API data — into a single list sorted
+    ascending by timestamp.
+
+    Response shape::
+
+        {
```
jmelloy, line 350: The `github_user_id` comes from `require_member()` which is the authenticated user's ID — the token lookup is correctly scoped. However, `GITHUB_TOKEN` is a shared server credential (typically a bot/installation token with broad repo access) being used on behalf of individual user requests. If the intent is to show only what the requesting user can see, using the server token silently elevates access. Add a comment or, if this is intentional, document it explicitly.

backend/routes/debug.py
```diff
@@ -0,0 +1,445 @@
...
+        }
+    """
+    guild_pk = await get_guild_pk(db, guild_id)
+    if guild_pk is None:
+        raise HTTPException(status_code=404, detail="Guild not found")
+
+    # Include soft-deleted rows so the debug view shows completed tasks.
+    result = await db.exec(
+        select(Task).where(col(Task.id) == task_id, col(Task.guild_id) == guild_pk)
+    )
```
jmelloy, line 371: The source field for Claude usage entries is hardcoded to `"worker"` but the PR description lists `"api"` as a possible source value. Using a dedicated source like `"claude"` would make it easier to filter these entries on the client side and matches the conceptual separation from worker log lines.

PR #548, opened 2026-06-04
Replace .execute() with .exec() to remove test warnings

jmelloy:
## Summary

- SQLModel's `AsyncSession.execute()` is decorated with `@deprecated` and generates `DeprecationWarning` at runtime whenever it's called, polluting test output with noise warnings
- Replaced all 9 call sites in application code with `AsyncSession.exec()`, which is SQLModel's recommended API and is not deprecated
- Files changed: `backend/push.py`, `backend/oauth.py`, `backend/ws_handlers.py`, `backend/routes/push.py`

## What was NOT changed

- `backend/alembic/env.py` — uses `await conn.execute()` on an Alembic `AsyncConnection`, not a SQLModel session
- `backend/alembic/versions/*.py` — all use `op.execute()` on Alembic's migration operations
- `backend/tests/helpers.py` and all test files — use `from sqlalchemy.orm import Session` (plain SQLAlchemy), whose `execute()` method is **not** deprecated

## Test plan

- [ ] Existing test suite should pass without the `DeprecationWarning: session.execute() is deprecated` warnings
- [ ] Result handling is preserved: `exec()` returns the same `Row`-like objects for multi-column `select()` expressions, so index access (`row[0]`) and named attribute access (`row.token`) both continue to work

### DIFF
--- a/backend/oauth.py
+++ b/backend/oauth.py
@@ -179,7 +179,7 @@ async def create_session(code: str, state: str) -> dict:
                 "updated_at": stmt.excluded.updated_at,
             },
         )
-        await db.execute(stmt)
+        await db.exec(stmt)
 
         user_stmt = pg_insert(User).values(
             id=github_user_id,
@@ -201,7 +201,7 @@ async def create_session(code: str, state: str) -> dict:
                 "updated_at": user_stmt.excluded.updated_at,
             },
         )
-        await db.execute(user_stmt)
+        await db.exec(user_stmt)
         db.add(UserSession(token=login_token, github_user_id=github_user_id, created_at=now))
         await db.commit()
     finally:
--- a/backend/push.py
+++ b/backend/push.py
@@ -112,7 +112,7 @@ async def _delete_tokens(tokens: list[str]) -> None:
         return
     db = await get_db()
     try:
-        await db.execute(delete(PushToken).where(PushToken.token.in_(tokens)))
+        await db.exec(delete(PushToken).where(PushToken.token.in_(tokens)))
         await db.commit()
     finally:
         await db.close()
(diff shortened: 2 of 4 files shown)


jmelloy (commented):
- The migration from `execute()` to `exec()` is correct in intent, but `SQLModel`'s `exec()` returns a different result type (`ScalarResult` for scalar selects vs `CursorResult` for DML), which can silently break result-handling code that calls `.all()` or row-indexing on the returned object.
- In `backend/push.py` the pattern `result.all()` followed by `row[0]` is risky: `exec()` on a multi-column or scalar select returns a `ScalarResult`, whose `.all()` yields scalars, not tuples, so `row[0]` would raise `TypeError` at runtime.
- DML statements (`delete`, `pg_insert` upserts) passed to `exec()` are not officially supported by SQLModel's `exec()` — it is designed for `select` statements; passing DML may work today via internal fall-through but is undocumented and fragile.
- The PR description's test plan is truncated, suggesting the description was not fully reviewed before submission.
- No tests were added or modified to explicitly assert the absence of `DeprecationWarning`, so the fix cannot be automatically verified to have achieved its stated goal.

jmelloy (commented):
- The migration from `session.execute()` to `session.exec()` is the correct approach for SQLModel's async API and will eliminate the deprecation warnings as intended.
- **Critical bug in `push.py`**: `exec()` returns a SQLModel `ScalarResult` (not a `CursorResult`), so `.all()` yields scalar values directly rather than row-tuples; the existing `row[0]` indexing will raise a `TypeError` at runtime.
- The upsert statements in `oauth.py` and `routes/push.py` use raw SQLAlchemy `pg_insert()` constructs — SQLModel's `exec()` is designed for ORM-style `select()`/`delete()` statements and may not handle these dialect-specific DML constructs correctly, potentially causing silent failures or errors.
- The PR description and test plan are incomplete (cut off mid-sentence), but the exclusions for Alembic and plain-SQLAlchemy test helpers are correctly identified and left unchanged.
- No security concerns introduced; the change is purely at the ORM API layer.

PR #562, opened 2026-06-04
Per-guild asyncio.Lock for run_foreman_ai (#553)

### LINKED ISSUE #553: Concurrent `run_foreman_ai` coroutines can double-assign workers (no guild-level serialization)
## Problem

When multiple events fire simultaneously (e.g. three workers complete tasks at once), three independent `run_foreman_ai` coroutines are spawned via `asyncio.create_task`. Each independently reads the same DB snapshot of idle workers and can independently call `assign_task` on the same worker — resulting in double-assignment or duplicate task rows.

There is **no per-guild serialization** of foreman runs. The only existing concurrency guard is a per-task lock inside `send_followup`; `assign_task` and `create_task` have none.

## Root cause

`spawn()` in `util/tasks.py` is a thin `asyncio.create_task` wrapper. Multiple callers — the `_poll_loop`, the WebSocket event handler, and webhook handlers — can all trigger `run_foreman_ai` concurrently for the same guild.

## Fix

Introduce a **per-guild asyncio lock (or distributed `LockService` key)** that serializes `run_foreman_ai` invocations at the guild level. Options:

1. A simple `asyncio.Lock` keyed by `guild_id` in an in-process dict (works for the embedded foreman).
2. A `LockService`-based distributed lock with key `foreman:{guild_id}` (required if the standalone foreman is ever horizontally scaled).

If a second invocation arrives while one is already running, it should either be **dropped** (the poll will re-trigger) or **queued** (one at a time, not fan-out).

## Related
- #550 (assign_task missing lock guard — partial fix, but guild-level serialization is the real root cause)
- See `docs/foreman-threading-analysis.md` for full analysis

jmelloy:
## Summary

- Adds `_guild_locks: dict[str, asyncio.Lock]` in `backend/foreman/runner.py` — lazily created per guild via `setdefault`.
- `run_foreman_ai` (public entry point) checks `lock.locked()` before acquiring; if the guild's foreman is already running the new invocation is **dropped** with an INFO log. The poll loop will re-trigger on the next tick, so no work is permanently lost. Dropping (vs. queuing) keeps memory bounded and avoids stale DB snapshots piling up under load.
- The implementation body is extracted to `_run_foreman_ai`; all existing call sites (poll loop, WebSocket handler, webhook debounce, task follow-up endpoint) go through the same public wrapper automatically — no call-site changes needed.
- Adds `backend/tests/test_foreman_guild_lock.py` with 4 tests covering: concurrent same-guild drops second, concurrent different guilds both run, lock released after normal completion, lock released after exception.

## Test plan

- [ ] `uv run python -m pytest backend/tests/test_foreman_guild_lock.py -v` — all 4 pass
- [ ] `uv run python -m pytest backend/tests/test_foreman_runner.py backend/tests/test_spawn_helper.py -v` — all existing tests pass
- [ ] Manual smoke: trigger multiple rapid foreman events for the same guild and confirm only one run executes at a time (second is logged as dropped)

Closes #553

### DIFF
--- a/backend/foreman/runner.py
+++ b/backend/foreman/runner.py
@@ -3,7 +3,6 @@
 import asyncio
 import json
 import logging
-import weakref
 from datetime import UTC, datetime
 
 from auth_deps import get_guild_pk
@@ -46,11 +45,8 @@ _poll_tasks: dict[str, "asyncio.Task[None]"] = {}
 # for a guild, new invocations are dropped rather than queued — the poll loop
 # will re-trigger on the next tick.  Dropping (vs. queuing) keeps memory
 # bounded and avoids stale snapshots piling up under load.
-#
-# WeakValueDictionary is used so that lock entries are GC-d automatically
-# once no coroutine holds a strong reference to the lock (i.e. after the
-# foreman run finishes and run_foreman_ai returns).  No manual pruning needed.
-_guild_locks: weakref.WeakValueDictionary[str, asyncio.Lock] = weakref.WeakValueDictionary()
+# Entries are popped in the finally block after each run so the dict stays small.
+_guild_locks: dict[str, asyncio.Lock] = {}
 
 # Module-level client — reused across calls so the underlying httpx connection
 # pool isn't thrown away every invocation. Lazily initialised so import works
@@ -310,16 +306,13 @@ async def run_foreman_ai(
 ) -> None:
     """Serialise per-guild and delegate to ``_run_foreman_ai``.
 
-    Uses a per-guild ``asyncio.Lock`` stored in a ``WeakValueDictionary`` (so
-    entries are GC-d automatically when no coroutine holds the lock).  The
-    acquire is attempted with ``asyncio.wait_for(..., timeout=0)`` so the
-    check-and-acquire is atomic — no TOCTOU window between the check and the
-    actual acquire.  If the lock is already held, the new invocation is dropped
-    and the caller returns immediately.  The poll loop will re-trigger on the
-    next tick, so no work is permanently lost.
+    Uses a per-guild ``asyncio.Lock`` stored in ``_guild_locks``.  The acquire
+    is attempted with ``asyncio.wait_for(..., timeout=0)`` so the check-and-
+    acquire is atomic — no TOCTOU window.  If the lock is already held the new
+    invocation is dropped; the poll loop re-triggers on the next tick.  The
+    dict entry is popped in the finally block so the dict stays small.
     """
-    lock = asyncio.Lock()
-    lock = _guild_locks.setdefault(guild_id, lock)
+    lock = _guild_locks.setdefault(guild_id, asyncio.Lock())
     try:
         await asyncio.wait_for(lock.acquire(), timeout=0)
     except asyncio.TimeoutError:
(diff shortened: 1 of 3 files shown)


jmelloy (commented):
- The per-guild lock mechanism is correctly implemented using `asyncio.Lock` with a drop-on-busy strategy, which is appropriate for a poll-driven system where stale work can be safely discarded.
- There is a subtle but real TOCTOU race: `lock.locked()` is checked and then `async with lock:` is entered as two separate steps; another coroutine could acquire the lock between those two operations, meaning the drop check is not atomic and a second invocation could slip through.
- The `_guild_locks` dict grows unboundedly — locks are added for every guild that ever triggers foreman but are never removed, which is a minor memory leak in long-running deployments with many transient guilds.
- Tests are well-structured and cover the four critical scenarios (drop, isolation, release after success, release after exception); the use of `asyncio.sleep(0)` to yield control is correct for single-threaded asyncio testing.
- The `uv.lock` file committed to the repository is unexpected and may cause issues if it conflicts with an existing lock file strategy; it should be reviewed for intentionality.

backend/foreman/runner.py
```diff
@@ -296,6 +302,30 @@ async def run_foreman_ai(
...
     user_id: str | None = None,
+) -> None:
+    """Serialise per-guild and delegate to ``_run_foreman_ai``.
+
+    Uses a per-guild ``asyncio.Lock``.  If the lock is already held (i.e. a
+    foreman run is in progress for this guild) the new invocation is dropped
+    and the caller returns immediately.  The poll loop will re-trigger on the
+    next tick, so no work is permanently lost.
+    """
+    lock = _guild_locks.setdefault(guild_id, asyncio.Lock())
```
jmelloy, line 317: TOCTOU race: `lock.locked()` and `async with lock:` are not atomic. Between the check and the acquisition, another coroutine could acquire the lock, allowing two concurrent runs for the same guild. Use `lock.acquire()` with `asyncio.wait_for(..., timeout=0)` or attempt acquisition via a non-blocking helper to make this truly atomic.

backend/foreman/runner.py
```diff
@@ -296,6 +302,30 @@ async def run_foreman_ai(
     human_message: str,
     extra_context: str = "",
     user_id: str | None = None,
+) -> None:
+    """Serialise per-guild and delegate to ``_run_foreman_ai``.
+
```
jmelloy, line 308: `_guild_locks` grows indefinitely — one `asyncio.Lock` object is added per guild and never evicted. For deployments with many short-lived guilds this is a slow memory leak. Consider using a `weakref`-based structure or periodically pruning locks for guilds with no active tasks.

backend/tests/test_foreman_guild_lock.py
```diff
@@ -0,0 +1,153 @@
...
+queued) while invocations for different guilds proceed independently.
+"""
+
+from __future__ import annotations
+
+import asyncio
+import os
+import sys
+from unittest.mock import AsyncMock, patch
+
```
jmelloy: `sys.path.insert` is a fragile way to set up imports; it mutates global state and can interfere with other tests. Prefer a `conftest.py` with `pythonpath` configuration or a `pyproject.toml`/`pytest.ini` `pythonpath` setting instead.

PR #563, opened 2026-06-04
feat: per-guild foreman config in DB

### LINKED ISSUE #561: Move foreman configuration into guild settings (per-guild DB config)
## Summary

Instead of running a separate foreman process with its own config file/env vars, expose all foreman settings directly in the guild settings and store them per-guild in the database.

## Motivation

- A single backend can host multiple guilds, each with different foreman behaviour (model, prompt, periodic-check interval, repo list, etc.).
- Operators shouldn't need to manage a separate foreman process or environment per guild — everything should be configurable from the guild settings UI and stored in the DB.

## What to implement

1. **Audit foreman config** — identify every setting the foreman process currently reads (model, system prompt, primary repo, periodic-check interval, tool allowlist, DNSid identity, etc.).
2. **DB schema** — add a `foreman_config` JSONB column (or a dedicated `foreman_settings` table) to the `guilds` table. Include a migration.
3. **API** — expose GET/PATCH endpoints (e.g. `GET /api/guilds/{id}/foreman-config` and `PATCH /api/guilds/{id}/foreman-config`) so the frontend and CLI can read/write these settings.
4. **Runtime** — when the foreman AI runs for a guild, load its config from the DB instead of from environment variables or a separate process config.
5. **Frontend** — add a "Foreman" section to the guild settings page (or a dedicated settings tab) where operators can view and edit the config fields.
6. **Migration path** — if a guild has no DB config, fall back to the current env-var / process defaults so existing deployments aren't broken.

## Files to look at

- `backend/foreman/` — current foreman process entry point and config loading
- `backend/models/guild.py` (or equivalent) — Guild DB model
- `backend/api/guilds.py` (or equivalent) — guild API routes
- `frontend/src/` — guild settings UI

## Acceptance criteria

- [ ] All foreman settings are stored per-guild in the database.
- [ ] A guild with no DB config falls back to existing defaults.
- [ ] The guild settings UI has a section to view/edit foreman config.
- [ ] No separate foreman process is required; the backend drives the foreman using DB config.
- [ ] Includes a DB migration.

jmelloy:
## Summary

- Adds `foreman_config` JSONB column to the `guilds` table via an Alembic migration
- Exposes `GET /api/guilds/{id}/foreman-config` and `PATCH /api/guilds/{id}/foreman-config` endpoints for reading/writing per-guild foreman settings
- Foreman runner now loads its config (model, system prompt, primary repo, periodic-check interval, tool allowlist) from the DB at runtime, falling back to existing env-var defaults so no existing deployments break
- Adds a "Foreman" settings section to the guild settings UI in `TopBar.vue`

Closes #561

## Test plan

- [ ] Run `alembic upgrade head` — `guilds.foreman_config` column should be created
- [ ] `GET /api/guilds/{id}/foreman-config` returns `{}` for a guild with no config set
- [ ] `PATCH /api/guilds/{id}/foreman-config` with `{"model": "claude-opus-4-8"}` persists and is returned by GET
- [ ] Foreman runner uses DB config values when present; falls back to env vars when not set
- [ ] Guild settings UI shows the Foreman section and allows editing fields
- [ ] Existing guilds with no DB config continue to work as before (env-var fallback)

### DIFF
--- a/backend/alembic/versions/20260604_000000_add_foreman_config_to_guilds.py
+++ b/backend/alembic/versions/20260604_000000_add_foreman_config_to_guilds.py
@@ -1,0 +1,33 @@
+"""Add foreman_config column to guilds table.
+
+Revision ID: 20260604_000000_add_foreman_config_to_guilds
+Revises: 20260603_000001_add_tools_to_workers
+Create Date: 2026-06-04
+
+Stores per-guild foreman AI configuration (model, provider, system prompt suffix,
+poll intervals, max rounds) as a JSON text column. NULL = use process-level env-var
+defaults, so existing deployments are unaffected.
+"""
+
+from __future__ import annotations
+
+from collections.abc import Sequence
+
+import sqlalchemy as sa
+from alembic import op
+
+revision: str = "20260604_000000_add_foreman_config_to_guilds"
+down_revision: str | Sequence[str] | None = "20260603_000001_add_tools_to_workers"
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+
+def upgrade() -> None:
+    op.add_column(
+        "guilds",
+        sa.Column("foreman_config", sa.Text(), nullable=True),
+    )
+
+
+def downgrade() -> None:
+    op.drop_column("guilds", "foreman_config")
--- a/backend/alembic/versions/20260604_000000_add_foreman_config_to_guilds.py
+++ b/backend/alembic/versions/20260604_000000_add_foreman_config_to_guilds.py
@@ -25,7 +25,7 @@ depends_on: str | Sequence[str] | None = None
 def upgrade() -> None:
     op.add_column(
         "guilds",
-        sa.Column("foreman_config", sa.Text(), nullable=True),
+        sa.Column("foreman_config", sa.JSON(), nullable=True),
     )
 
 
--- a/backend/foreman/runner.py
+++ b/backend/foreman/runner.py
@@ -7,7 +7,7 @@ import os
 from datetime import UTC, datetime
 
 from auth_deps import get_guild_pk
-from database import get_db
+from database import AsyncSessionLocal, get_db
 from events import broadcast_msg
 from foreman.prompt import build_state_preamble, build_system_blocks, build_system_prompt
 from foreman.tools import exec_tools
@@ -29,10 +29,6 @@ from foreman_core.message_utils import (
 )
 from foreman_core.tools_schema import FOREMAN_TOOLS
 from models import Agent, ForemanTurn, Guild, GuildMember, Message, Task, Worker, live_tasks_filter
-
-_FOREMAN_CONFIG_FIELDS = frozenset(
-    {"model", "provider", "system_prompt_suffix", "max_rounds", "poll_min_interval", "poll_max_interval"}
-)
 from sqlalchemy import delete, func
 from sqlmodel import col, select
 from util.tasks import spawn
@@ -60,21 +56,13 @@ def _get_anthropic_client():
 
 
 async def _load_foreman_config(guild_id: str) -> dict:
-    """Load per-guild foreman config from the DB. Returns {} if unset or unparseable."""
-    db = await get_db()
-    try:
+    """Load per-guild foreman config from the DB. Returns {} if unset."""
+    async with AsyncSessionLocal() as db:
         result = await db.exec(
             select(col(Guild.foreman_config)).where(col(Guild.guild_id) == guild_id)
         )
         raw = result.one_or_none()
-        if raw:
-            try:
-                return json.loads(raw)
-            except (json.JSONDecodeError, TypeError):
-                pass
-        return {}
-    finally:
-        await db.close()
+        return raw if isinstance(raw, dict) else {}
 
 
 async def _get_guild_user_id(guild_id: str) -> str | None:
@@ -207,10 +195,7 @@ async def _poll_loop(guild_id: str) -> None:
     The foreman-poll-status broadcast is sent after each poll so the UI can
     display the countdown to the *next* check.
     """
-    cfg = await _load_foreman_config(guild_id)
-    poll_min = int(cfg.get("poll_min_interval", POLL_MIN_SECS))
-    poll_max = int(cfg.get("poll_max_interval", POLL_MAX_SECS))
-    interval = poll_min
+    interval = POLL_MIN_SECS
     while True:
         try:
             await asyncio.sleep(interval)
@@ -222,6 +207,10 @@ async def _poll_loop(guild_id: str) -> None:
             return
 
         try:
+            # Reload config each cycle so guild owners see changes without restarting.
+            cfg = await _load_foreman_config(guild_id)
+            poll_max = int(cfg.get("poll_max_interval", POLL_MAX_SECS))
+
             # If an external foreman is connected for this guild it owns the
             # poll loop — skip the embedded run to avoid double-triggering.
             from events import foreman_connections
--- a/backend/foreman/runner.py
+++ b/backend/foreman/runner.py
@@ -3,6 +3,7 @@
 import asyncio
 import json
 import logging
+import os
 from datetime import UTC, datetime
 
 from auth_deps import get_guild_pk
@@ -28,6 +29,10 @@ from foreman_core.message_utils import (
 )
 from foreman_core.tools_schema import FOREMAN_TOOLS
 from models import Agent, ForemanTurn, Guild, GuildMember, Message, Task, Worker, live_tasks_filter
+
+_FOREMAN_CONFIG_FIELDS = frozenset(
+    {"model", "provider", "system_prompt_suffix", "max_rounds", "poll_min_interval", "poll_max_interval"}
+)
 from sqlalchemy import delete, func
 from sqlmodel import col, select
 from util.tasks import spawn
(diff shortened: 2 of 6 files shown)


backend/foreman/runner.py
```diff
@@ -54,6 +59,24 @@ def _get_anthropic_client():
     return _anthropic_client
 
 
+async def _load_foreman_config(guild_id: str) -> dict:
+    """Load per-guild foreman config from the DB. Returns {} if unset or unparseable."""
```
jmelloy: `get_db()` is called without a request context, creating an untracked session that is closed in the `finally` block but never added to any session-management scope. Under concurrent load this will leak connections if an exception is raised before `db.close()` (e.g. inside `json.loads`). Use the existing `get_db_dep` dependency pattern or a dedicated async context manager instead.

backend/foreman/runner.py
```diff
@@ -179,12 +202,15 @@ async def _update_turn_tokens(turn_id: int, input_tokens: int, output_tokens: in
 async def _poll_loop(guild_id: str) -> None:
     """Background loop: check non-terminal tasks and call the foreman if any are found.
 
-    Interval doubles each cycle from POLL_MIN_SECS up to POLL_MAX_SECS.
-    Cancelled (via reset_foreman_poll) whenever a significant event arrives.
+    Interval doubles each cycle from POLL_MIN_SECS up to POLL_MAX_SECS (or per-guild
```
jmelloy, line 193: Config is loaded once when `_poll_loop` starts and cached in `poll_min`/`poll_max` for the lifetime of the loop. A guild owner changing these values in the UI will see no effect until the foreman is restarted. Either reload config each cycle or document this limitation clearly.

backend/routes/guilds.py
```diff
@@ -368,6 +377,61 @@ async def get_webhook_secret(
...
+    db: AsyncSession = Depends(get_db_dep),
+):
+    """Return the guild's foreman configuration (owner only)."""
+    guild_pk = await get_guild_pk(db, guild_id)
+    if guild_pk is None:
+        raise HTTPException(status_code=404, detail="Guild not found")
+    res = await db.exec(select(Guild).where(col(Guild.id) == guild_pk))
+    guild = res.one_or_none()
+    if not guild:
+        raise HTTPException(status_code=404, detail="Guild not found")
```
jmelloy, line 391: The `get_guild_pk` + `select(Guild)` pattern issues two queries where one would suffice — just `select(Guild).where(Guild.guild_id == guild_id)` directly. More importantly, there is no validation that `guild_id` in the URL belongs to the authenticated user's guild membership before `require_member("owner")` resolves it; confirm `get_guild_pk` + `require_member` together prevent IDOR here.

backend/routes/guilds.py
```diff
@@ -368,6 +377,61 @@ async def get_webhook_secret(
...
+    config: dict = {}
+    if guild.foreman_config:
+        try:
+            config = json.loads(guild.foreman_config)
+        except json.JSONDecodeError:
+            config = {}
```
jmelloy: No length or content validation on `system_prompt_suffix` (could be megabytes) and no whitelist/enum validation on `provider` (arbitrary string reaches `make_anthropic_client`). Add `Field(max_length=...)` to `ForemanConfigUpdate` and validate `provider` against an explicit set of supported values.

backend/alembic/versions/20260604_000000_add_foreman_config_to_guilds.py
```diff
@@ -0,0 +1,33 @@
...
+from alembic import op
+
+revision: str = "20260604_000000_add_foreman_config_to_guilds"
+down_revision: str | Sequence[str] | None = "20260603_000001_add_tools_to_workers"
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+
+def upgrade() -> None:
+    op.add_column(
```
jmelloy, line 26: Column is added as `sa.Text()` but the PR description advertises JSONB. For PostgreSQL, using `sa.dialects.postgresql.JSONB` would allow server-side JSON operators and avoid the manual `json.loads`/`json.dumps` dance in every caller. If cross-DB compatibility is required at least document the deliberate choice of Text over JSONB.

jmelloy (commented):
- Migration adds `foreman_config` as `Text` (not JSONB) despite the PR description mentioning JSONB; the model field likewise uses `Text`, so JSON parsing on every read/write is manual and error-prone — a native `JSON`/`JSONB` column would be cleaner and safer.
- `_load_foreman_config` in `runner.py` opens a new DB session via `get_db()` on every poll-loop iteration and every `run_foreman_ai` call; this bypasses the request-scoped session pattern used everywhere else and could exhaust the connection pool under load.
- The poll-loop loads foreman config once at startup and never refreshes it, meaning poll-interval changes in the DB won't take effect until the loop is cancelled and restarted — this is inconsistent with the per-call reload done in `run_foreman_ai`.
- No input validation is performed on security-sensitive fields: `provider` accepts arbitrary strings, `system_prompt_suffix` has no length cap, and numeric fields (`max_rounds`, `poll_min_interval`, `poll_max_interval`) are only range-limited in the frontend HTML — the backend `ForemanConfigUpdate` model has no `Field(ge=...)` constraints.
- `loadForemanConfig` is only triggered when the user opens the Foreman panel (`watch(showForemanConfig, ...)`); if the user opens settings without toggling the Foreman section, the reactive state is never populated, so the initial guild-switch reset (which blanks the fields but never fetches) means the save button could silently null-out all fields.

backend/foreman/runner.py
```diff
@@ -54,6 +59,24 @@ def _get_anthropic_client():
     return _anthropic_client
 
 
+async def _load_foreman_config(guild_id: str) -> dict:
```
jmelloy, line 58: `_load_foreman_config` opens a new DB session via `get_db()` each call and is invoked twice per `run_foreman_ai` execution (once in `_poll_loop`, once inside `run_foreman_ai`). This bypasses the request-scoped session pattern used elsewhere, risks connection-pool exhaustion, and means poll config is read once at loop start and never refreshed. Consider passing the config in or reading it once with the existing session.

backend/foreman/runner.py
```diff
@@ -28,6 +29,10 @@
 )
 from foreman_core.tools_schema import FOREMAN_TOOLS
 from models import Agent, ForemanTurn, Guild, GuildMember, Message, Task, Worker, live_tasks_filter
+
+_FOREMAN_CONFIG_FIELDS = frozenset(
```
jmelloy: `_FOREMAN_CONFIG_FIELDS` is defined but never referenced anywhere in the codebase — remove it to avoid confusion.

backend/routes/guilds.py
```diff
@@ -368,6 +377,61 @@ async def get_webhook_secret(
     return {"guild_id": guild_id, "webhook_secret": secret}
 
 
+@router.get("/api/guilds/{guild_id}/foreman-config")
+async def get_foreman_config(
+    guild_id: str,
+    github_user_id: str = Depends(require_member("owner")),
+    db: AsyncSession = Depends(get_db_dep),
```
jmelloy, line 385: `get_foreman_config` fetches the guild twice: once via `get_guild_pk` and again with a full `select(Guild)`. You can simplify to a single query that fetches the full `Guild` row and derives the PK check from it, eliminating the redundant round-trip (the same applies to `update_foreman_config`).

backend/routes/guilds.py
```diff
@@ -368,6 +377,61 @@ async def get_webhook_secret(
...
+        raise HTTPException(status_code=404, detail="Guild not found")
+    res = await db.exec(select(Guild).where(col(Guild.id) == guild_pk))
+    guild = res.one_or_none()
+    if not guild:
+        raise HTTPException(status_code=404, detail="Guild not found")
+    config: dict = {}
+    if guild.foreman_config:
+        try:
+            config = json.loads(guild.foreman_config)
+        except json.JSONDecodeError:
```
jmelloy: There is no input validation on `ForemanConfigUpdate` values — `poll_min_interval=0`, `max_rounds=0`, or an unrecognised `provider` string will be stored and silently cause runtime errors in the runner. Add `Field(gt=0)` / `Literal` / `validator` constraints on the Pydantic model.

backend/alembic/versions/20260604_000000_add_foreman_config_to_guilds.py
```diff
@@ -0,0 +1,33 @@
...
+        "guilds",
```
jmelloy, line 27: The column is created as `sa.Text()` (plain string) rather than `sa.JSON` or `postgresql.JSONB`. This means all JSON serialisation/deserialisation is manual and scattered across multiple call sites; using a native JSON column type would let SQLAlchemy handle encoding, enable server-side JSON operators, and eliminate the repeated `json.loads`/`json.dumps` guards.

jmelloy (commented):
- Migration uses `sa.Text()` instead of `sa.JSON`/`JSONB`, and the SQLModel field is also `str`, meaning JSON encoding/decoding is manually handled everywhere rather than delegated to SQLAlchemy — this creates multiple parse/serialize sites and drift risk.
- `_load_foreman_config` in `runner.py` calls `await get_db()` which opens a new session each invocation (and is called twice per `run_foreman_ai` call — once for poll config, once for AI config), but never adds the session to any context; this is a resource-management concern and could exhaust the connection pool under load.
- The `_FOREMAN_CONFIG_FIELDS` frozenset is defined between imports and the rest of the module-level code, and it is never actually used anywhere — dead code.
- No validation is applied to user-supplied config values (e.g. `poll_min_interval` could be set to 0 or negative, `max_rounds` to 0, `provider` to an arbitrary string); the backend should validate/clamp these before persisting.
- `loadForemanConfig` is only called when the section is toggled open, so if a user opens the panel, edits, saves, closes, and reopens, they see the updated values — but the initial guild-switch watcher clears all fields without re-fetching, meaning if the panel was already open when switching guilds it will show stale blank state until toggled.

backend/routes/guilds.py
```diff
@@ -57,6 +58,15 @@ class GuildUpdate(BaseModel):
     version: str | None = None
 
 
+class ForemanConfigUpdate(BaseModel):
+    model: str | None = None
+    provider: Literal["anthropic", "openai", "google"] | None = None
```
jmelloy: this should include bedrock

PR #564, opened 2026-06-04
Wrap blocking GitHub/Docker calls in asyncio.to_thread (#559)

### LINKED ISSUE #559: GitHub/Docker tool calls block the asyncio event loop (synchronous calls in async context)
## Problem

Several tool handlers in `foreman/tools.py` make synchronous blocking calls to the GitHub REST API and Docker SDK directly on the asyncio event loop thread. Because asyncio is single-threaded, a single slow GitHub API call (e.g. fetching a large diff or creating an issue) stalls **all other coroutines** in the process — including WebSocket heartbeats, incoming webhook handling, and other foreman tool calls.

- GitHub calls (`get_github_issue`, `list_github_prs`, `review_pr_internal`, etc.) use `PyGithub` or `httpx` in blocking mode
- Docker SDK calls (`spawn_worker`, `shutdown_worker`) are also synchronous
- These are nominally wrapped in `asyncio.to_thread` in some places but not consistently throughout the tool handler layer
- A 10-second GitHub API timeout would freeze the entire event loop for 10 seconds

## Fix

Audit every tool handler in `foreman/tools.py` (and `ws_handlers.py`) that calls an external service. Ensure all blocking I/O is wrapped in `asyncio.to_thread(...)` or replaced with an async client (`httpx.AsyncClient`, `gidgethub`, etc.). Add a global timeout (e.g. 30s) on all external calls so a hung upstream cannot stall the loop indefinitely.

## Related
- `backend/foreman/tools.py`
- #555 (sequential tool calls — fixing that makes this worse without this fix)

jmelloy:
## Summary

- Add `EXTERNAL_CALL_TIMEOUT = 30` constant and `_to_thread()` helper that combines `asyncio.to_thread` with `asyncio.wait_for(timeout=EXTERNAL_CALL_TIMEOUT)` in `backend/foreman/tools.py`
- Fix two previously unguarded blocking Docker calls in `spawn_worker`: `_get_docker_client()` (first call opens a socket) and `docker_client.containers.get()` (network round-trip to Docker daemon)
- Replace all existing bare `asyncio.to_thread()` calls with `_to_thread()` to apply the 30-second timeout guard uniformly — covers GitHub REST API, GitHub GraphQL, A2A agent card/task calls, and Docker `containers.run()`
- `ws_handlers.py` required no changes — all I/O there is already async (DB, WebSocket)

## Details

A single slow or hung GitHub/Docker call on the event loop thread previously froze all coroutines — WebSocket heartbeats, webhook handlers, everything. The fix has two layers:

1. **Thread offload** (`asyncio.to_thread`): every blocking call runs in the thread pool, releasing the event loop immediately.
2. **Hard timeout** (`asyncio.wait_for`): a 30-second cap ensures a hung upstream can stall at most one tool call rather than blocking indefinitely. `TimeoutError` is caught by the existing `except Exception` handlers in `_exec_one_tool` and returned as a tool error.

## Test plan

- [ ] Existing tests pass (`ruff check` ✅)
- [ ] `spawn_worker` Docker path: `containers.get()` and `_get_docker_client()` now go through `_to_thread`
- [ ] All GitHub tool paths (`list_github_issues`, `get_github_issue`, `get_pr_status`, `review_pr`, `review_pr_internal`, etc.) use `_to_thread`
- [ ] A2A `call_agent` card fetch and task post use `_to_thread`
- [ ] Timeout: simulate a hung GitHub API call and verify `TimeoutError` surfaces as a tool error within 30 s instead of blocking the loop

Closes #559

### DIFF
--- a/backend/foreman/tools.py
+++ b/backend/foreman/tools.py
@@ -65,6 +65,11 @@ DEFAULT_FINALIZE_TTL_SECONDS = 3 * 24 * 60 * 60  # 3 days
 # event loop indefinitely.
 EXTERNAL_CALL_TIMEOUT = 30
 
+# Separate, larger timeout for docker_client.containers.run (detach=True).
+# The Docker API call itself should complete once the daemon confirms the container
+# started, but can legitimately take longer than a GitHub REST call on a loaded host.
+CONTAINER_RUN_TIMEOUT = 600
+
 # Module-level Docker client — created once per process to reuse the Unix-socket
 # connection across repeated spawn_worker calls instead of opening a new socket
 # each time.  None until first successful from_env(); tests can patch
@@ -72,18 +77,28 @@ EXTERNAL_CALL_TIMEOUT = 30
 _docker_client: Any = None
 
 
-def _get_docker_client() -> Any:
-    """Return the cached Docker client, importing the SDK and calling from_env() once."""
-    global _docker_client
-    if _docker_client is None:
-        import docker  # ImportError propagated to caller if SDK not installed
+async def _get_docker_client() -> Any:
+    """Return the cached Docker client, importing the SDK and calling from_env() once.
 
-        _docker_client = docker.from_env()
+    Skips the thread hop on subsequent calls once the client is cached.
+    """
+    global _docker_client
+    if _docker_client is not None:
+        return _docker_client
+    import docker  # ImportError propagated to caller if SDK not installed
+
+    _docker_client = await _to_thread(docker.from_env)
     return _docker_client
 
 
 async def _to_thread(fn, /, *args, **kwargs):
-    """Off-load a blocking I/O call to the thread pool with a global timeout."""
+    """Run a blocking callable in a thread pool with a global timeout.
+
+    WARNING: asyncio cancellation/timeout does NOT interrupt the underlying thread.
+    If fn has side effects (e.g. spawning containers, open sockets), callers must
+    handle cleanup in an except block — the thread will continue running to
+    completion even after a TimeoutError is raised to the caller.
+    """
     return await asyncio.wait_for(
         asyncio.to_thread(fn, *args, **kwargs),
         timeout=EXTERNAL_CALL_TIMEOUT,
@@ -1168,7 +1183,7 @@ async def _exec_one_tool(guild_id: str, tu, user_id: str | None = None) -> dict:
                     )
 
                     try:
-                        docker_client = await _to_thread(_get_docker_client)
+                        docker_client = await _get_docker_client()
                         image = os.environ.get("WORKER_IMAGE", "pioneer-square-worker")
 
                         network = None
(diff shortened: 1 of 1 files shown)


backend/foreman/tools.py
```diff
@@ -76,6 +82,14 @@ def _get_docker_client() -> Any:
     return _docker_client
 
 
+async def _to_thread(fn, /, *args, **kwargs):
```
jmelloy, line 94: `asyncio.wait_for` cancels the awaitable but the underlying `asyncio.to_thread` worker thread is not interruptible — on timeout, the blocking call keeps running in the background. For `containers.run` especially, this means the container continues executing while the caller receives a `TimeoutError`, resulting in orphaned containers. Consider tracking spawned containers and cleaning them up on timeout, or using a separate process/signal mechanism.

jmelloy (commented):
- The `_to_thread` helper correctly addresses the core problem: blocking Docker/GitHub calls previously ran on the event loop thread, freezing all coroutines. The two newly wrapped calls (`_get_docker_client` and `docker_client.containers.get`) are legitimate fixes.
- **Critical concern**: `docker_client.containers.run` with `detach=False` (blocking mode) can legitimately take far longer than 30 seconds for a real workload. The timeout will cancel the `asyncio.wait_for` future, but the underlying thread continues running — the container keeps executing while the caller gets a `TimeoutError`, leading to orphaned containers and lost results.
- The 30-second timeout is applied uniformly to all call sites including `containers.run`, but Docker container execution is fundamentally unbounded in duration. A single constant is inappropriate for both a quick API call and a potentially long-running container.
- `asyncio.wait_for` cancels the *coroutine* wrapping the thread, but `asyncio.to_thread` threads are not interruptible — the blocking thread runs to completion regardless. This means timeouts prevent the caller from waiting but do not free resources, and error handling downstream may be confused by a `TimeoutError` while the container is still alive.
- No tests are added or updated to cover the new `_to_thread` helper or the timeout behavior, making it hard to verify correctness under edge cases.

backend/foreman/tools.py
```diff
@@ -1185,9 +1201,7 @@ async def _exec_one_tool(guild_id: str, tu, user_id: str | None = None) -> dict:
                         if network:
                             run_kwargs["network"] = network
 
-                        container = await asyncio.to_thread(
-                            docker_client.containers.run, **run_kwargs
-                        )
+                        container = await _to_thread(docker_client.containers.run, **run_kwargs)
```
jmelloy: `docker_client.containers.run` can legitimately take minutes for real worker jobs. Applying the same 30-second `EXTERNAL_CALL_TIMEOUT` here as for GitHub REST calls will cause spurious timeouts for normal container execution. This call site should either not use `_to_thread`, or use a much larger timeout constant specific to container operations.

backend/foreman/tools.py
jmelloy, line 94: The helper should be documented to note that it does NOT interrupt the thread on timeout — callers that rely on side-effect cleanup (e.g., containers, open sockets) on cancellation will not get it. At minimum, add a docstring warning about this semantics gap.

backend/foreman/tools.py
```diff
@@ -1154,12 +1168,14 @@ async def _exec_one_tool(guild_id: str, tu, user_id: str | None = None) -> dict:
                     )
 
                     try:
-                        docker_client = _get_docker_client()
+                        docker_client = await _to_thread(_get_docker_client)
```
jmelloy: `_get_docker_client` uses a module-level singleton (`_docker_client`). Wrapping it in `_to_thread` is correct for the first call (socket open), but subsequent calls return immediately from cache and the thread overhead is unnecessary. Consider checking the cached value before offloading, or restructuring the singleton initialization to be async-safe.

PR #566, opened 2026-06-04
Fix foreman lock key to guild_id:user_id (#565)

### LINKED ISSUE #565: Fix foreman lock key to use guild_id:user_id instead of guild_id
## Problem

The per-guild asyncio.Lock introduced in PR #562 uses only `guild_id` as the lock key. This is too broad — if multiple users in the same guild can trigger `run_foreman_ai`, they'll unnecessarily serialize on the same lock.

## Fix

Change the lock key from `guild_id` to `f"{guild_id}:{user_id}"` (or equivalent tuple key) so locks are scoped per user per guild.

## Files to look at
- `backend/foreman/` — wherever the per-guild lock dict and `run_foreman_ai` entry point live (introduced in PR #562)

jmelloy:
## Summary

- Changed `_guild_locks` key from `guild_id` (a plain string) to `(guild_id, user_id)` (a tuple) in `backend/foreman/runner.py`
- Updated the type annotation from `dict[str, asyncio.Lock]` to `dict[tuple[str, str | None], asyncio.Lock]`
- Updated the log message and docstring to reflect per-(guild, user) scope
- No call-site changes needed — `user_id` was already threaded through to `run_foreman_ai`

## Why

PR #562 introduced a per-guild lock that unnecessarily serialized all users in the same guild. Two different users triggering `run_foreman_ai` concurrently would block each other even though their foreman runs are completely independent.

Closes #565

### DIFF
--- a/backend/foreman/runner.py
+++ b/backend/foreman/runner.py
@@ -315,6 +315,10 @@ async def run_foreman_ai(
     and cooperative, so no other coroutine can run between the check and the
     acquire (there is no ``await`` between them).
     """
+    # When user_id is None (system-triggered/poll runs), all such invocations
+    # within the same guild share key (guild_id, None) and serialize against
+    # each other.  This is intentional — system runs are per-guild work and
+    # should not overlap with themselves.
     lock_key = (guild_id, user_id)
     lock = _guild_locks.setdefault(lock_key, asyncio.Lock())
     if lock.locked():
--- a/backend/foreman/runner.py
+++ b/backend/foreman/runner.py
@@ -42,11 +42,11 @@ POLL_MAX_SECS = 3600  # maximum poll interval: 60 minutes
 _poll_tasks: dict[str, "asyncio.Task[None]"] = {}
 
 # Per-guild locks to serialise foreman runs.  If a run is already in progress
-# for a guild, new invocations are dropped rather than queued — the poll loop
-# will re-trigger on the next tick.  Dropping (vs. queuing) keeps memory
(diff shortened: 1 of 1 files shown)


backend/foreman/runner.py
```diff
@@ -304,29 +304,32 @@ async def run_foreman_ai(
     extra_context: str = "",
     user_id: str | None = None,
 ) -> None:
-    """Serialise per-guild and delegate to ``_run_foreman_ai``.
+    """Serialise per-(guild, user) and delegate to ``_run_foreman_ai``.
```
jmelloy, line 307: When `user_id` is `None` (e.g. system-triggered runs), all such invocations across the guild share the key `(guild_id, None)` and will serialize against each other. This may be intentional, but it's worth a comment clarifying whether `None` is a realistic case and whether that serialization is acceptable or should be handled differently (e.g. using a sentinel UUID).

jmelloy (commented):
- The fix correctly narrows the lock granularity from per-guild to per-(guild, user), eliminating unnecessary serialization between independent users in the same guild.
- The type annotation `dict[tuple[str, str | None], asyncio.Lock]` correctly reflects that `user_id` can be `None`, though callers should be aware that all `None`-user invocations in the same guild will still serialize against each other.
- The atomicity comment about `lock.locked()` + `lock.acquire()` remains valid since there is no `await` between the check and acquire.
- Memory management via `_guild_locks.pop(lock_key, None)` in the `finally` block is preserved correctly with the new tuple key.
- No call-site changes were required, confirming `user_id` was already available at the invocation point.

PR #572, opened 2026-06-05
fix(foreman): correct task lifecycle and add error handling for PR events

### LINKED ISSUE #570: Foreman: finalize task on PR merge (not agent completion), improve error handling
## Problem

The foreman is currently marking tasks complete when the coding agent finishes its work, rather than waiting for:
1. The branch to be verified / pushed
2. A PR to be opened
3. The PR to be merged

Additionally, error handling appears to be insufficient — errors from workers/agents are not being surfaced or handled properly.

## Expected Behaviour

- Tasks should remain in `awaiting-review` state after the agent finishes and a PR is opened
- The foreman should listen for the `pull_request/closed` GitHub webhook event with `merged=true` and only then call `finalize_task`
- Errors from agent runs (e.g. CI failures, push failures, API errors) should be caught and surfaced — either by transitioning the task to a failed state or by sending a follow-up with the error details

## Acceptance Criteria

- [ ] `finalize_task` is only called after a PR is merged (via `pull_request/closed` + `merged=true` webhook), not when the agent subprocess exits
- [ ] Task lifecycle: `working` → `awaiting-review` (PR open) → `complete` (PR merged)
- [ ] Errors during agent execution cause the task to enter a `failed` state or trigger a follow-up with the error details
- [ ] Unmerged PR closures are handled (transition to failed/abandoned)

jmelloy:
## Summary

Fixes two related issues with how the foreman manages task lifecycle and errors (closes #570).

**Issue 1 — Tasks marked complete too early:**
- `finalize_task` was being called by the AI as soon as the coding agent finished, before any PR was opened or merged
- Added deterministic auto-finalization in the GitHub webhook handler: when a `pull_request/closed` event arrives with `merged=true`, the task is immediately transitioned to `done` without requiring an AI decision
- Updated the foreman summary message for merged/closed PRs to reflect that the transition is automatic, so the AI no longer tries to call `finalize_task` itself

**Issue 2 — Errors not surfaced:**
- Unmerged PR closures now automatically transition the task to `failed` (via `_auto_fail_task_on_pr_close` in the webhook handler)
- `finalize_task` now accepts an `outcome` parameter (`"done"` | `"failed"`) so the foreman AI can explicitly mark errored/abandoned tasks as failed
- Invalid `outcome` values fall back to `"done"` to avoid breaking existing callers
- The `finalize_task` tool description now warns the AI not to call it for tasks with an open PR (those are handled by webhooks)

Both lifecycle helpers (`_auto_finalize_task_on_pr_merge`, `_auto_fail_task_on_pr_close`) release task locks, discard queued follow-up events, and broadcast `TaskUpdateMsg` / `TaskFinalizeMsg` so the frontend updates immediately.

## Test plan
- [ ] `test_finalize_task_marks_failed_with_outcome` — `finalize_task(outcome="failed")` sets `state=failed`
- [ ] `test_finalize_task_invalid_outcome_defaults_to_done` — bogus outcome value falls back to `done`
- [ ] `test_pr_merge_webhook_auto_finalizes_task` — merged PR webhook sets task `state=done`
- [ ] `test_pr_close_unmerged_webhook_auto_fails_task` — closed-without-merge webhook sets task `state=failed`
- [ ] Existing webhook and foreman tests continue to pass

### DIFF
--- a/backend/foreman/tools.py
+++ b/backend/foreman/tools.py
@@ -1114,6 +1114,8 @@ async def _exec_one_tool(guild_id: str, tu, user_id: str | None = None) -> dict:
 
             elif tu.name == "finalize_task":
                 task_id = inp["task_id"]
+                raw_outcome = inp.get("outcome", "done")
+                outcome = raw_outcome if raw_outcome in ("done", "failed") else "done"
                 deleted_at, err = _resolve_finalize_deleted_at(inp)
                 if err:
                     result_text = err
@@ -1131,12 +1133,12 @@ async def _exec_one_tool(guild_id: str, tu, user_id: str | None = None) -> dict:
                             update(Task)
                             .where(col(Task.id) == task_id)
                             .values(
-                                state="done",
+                                state=outcome,
                                 finished_at=finished_at,
                                 deleted_at=deleted_at,
                             )
                         )
-                        # Discard any queued follow-up events — the task is done.
+                        # Discard any queued follow-up events — the task is closed.
                         await db.exec(delete(TaskEvent).where(col(TaskEvent.task_id) == task_id))
                         await LockService(db).release(f"task:{task_id}")
                         await db.commit()
@@ -1148,7 +1150,7 @@ async def _exec_one_tool(guild_id: str, tu, user_id: str | None = None) -> dict:
                             guild_id,
                             TaskUpdateMsg(
                                 taskId=task_id,
-                                state="done",
+                                state=outcome,
                                 finishedAt=finished_at.isoformat(),
                                 deletedAt=deleted_at.isoformat()
                                 if deleted_at is not None
@@ -1156,7 +1158,7 @@ async def _exec_one_tool(guild_id: str, tu, user_id: str | None = None) -> dict:
                             ),
                         )
                         result_text = (
-                            f"Task {task_id} finalized; soft-delete at "
+                            f"Task {task_id} finalized as {outcome}; soft-delete at "
                             f"{deleted_at.isoformat() if deleted_at is not None else 'unknown'}."
                         )
 
--- a/backend/foreman/tools.py
+++ b/backend/foreman/tools.py
@@ -1116,13 +1116,19 @@ async def _exec_one_tool(guild_id: str, tu, user_id: str | None = None) -> dict:
                 task_id = inp["task_id"]
                 raw_outcome = inp.get("outcome", "done")
                 outcome = raw_outcome if raw_outcome in ("done", "failed") else "done"
+                if outcome != raw_outcome:
+                    logger.warning(
+                        "finalize_task: unknown outcome %r, defaulting to 'done'", raw_outcome
+                    )
                 deleted_at, err = _resolve_finalize_deleted_at(inp)
                 if err:
                     result_text = err
                     is_error = True
                 else:
                     result = await db.exec(
-                        select(Task).where(col(Task.id) == task_id, col(Task.guild_id) == guild_pk)
+                        select(Task)
+                        .where(col(Task.id) == task_id, col(Task.guild_id) == guild_pk)
+                        .with_for_update()
                     )
                     task = result.one_or_none()
                     if not task:
--- a/backend/routes/webhooks.py
+++ b/backend/routes/webhooks.py
@@ -241,28 +241,40 @@ async def _auto_finalize_task_on_pr_merge(
 ) -> bool:
     """Directly transition a task to 'done' when its PR is merged.
 
-    Skips tasks already in a terminal state. Returns True if finalization occurred.
+    Uses a single conditional UPDATE to prevent TOCTOU races with concurrent
+    state transitions. Returns True if finalization occurred.
     Releases any in-progress lock, discards queued follow-up events, and broadcasts
     TaskFinalizeMsg + TaskUpdateMsg so the frontend reflects the change immediately.
     """
-    result = await db.exec(
-        select(col(Task.worker_id), col(Task.state)).where(
+    # Fetch worker_id for the broadcast; existence check only — state filtering
+    # is delegated entirely to the conditional UPDATE to eliminate SELECT→UPDATE TOCTOU.
+    row_result = await db.exec(
+        select(col(Task.id), col(Task.worker_id)).where(
             col(Task.id) == task_id, col(Task.guild_id) == guild_pk
         )
     )
-    row = result.one_or_none()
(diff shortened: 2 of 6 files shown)


backend/routes/webhooks.py
```diff
@@ -227,6 +228,98 @@ async def schedule(
...
+    guild_pk: int,
+    guild_id: str,
+    task_id: str,
+) -> bool:
+    """Directly transition a task to 'failed' when its PR is closed without merging.
+
+    Skips tasks already in a terminal state. Returns True if the transition occurred.
+    Releases any in-progress lock and broadcasts TaskUpdateMsg so the frontend updates.
+    """
+    state = await db.exec(
```
jmelloy: `_auto_fail_task_on_pr_close` does not delete `TaskEvent` rows for the task, unlike `_auto_finalize_task_on_pr_merge`. Orphaned events could cause the foreman to process follow-ups for an already-failed task. Add `await db.exec(delete(TaskEvent).where(col(TaskEvent.task_id) == task_id))` before the commit, matching the merge path.

backend/routes/webhooks.py
```diff
@@ -227,6 +228,98 @@ async def schedule(
...
+        guild_id,
+        TaskUpdateMsg(
+            taskId=task_id,
+            state="done",
+            finishedAt=now.isoformat(),
+            deletedAt=deleted_at.isoformat(),
+        ),
+    )
+    return True
+
```
jmelloy, line 293: The state read and write in `_auto_fail_task_on_pr_close` (and similarly in `_auto_finalize_task_on_pr_merge`) are not atomic — another request could transition the task between the SELECT and the UPDATE. Consider using a conditional UPDATE with `WHERE state NOT IN (terminal states)` and checking `rowcount` instead of a separate SELECT, which would eliminate the TOCTOU race.

jmelloy (commented):
- The core lifecycle fix is sound: moving task finalization out of the AI decision path and into deterministic webhook handlers eliminates the race condition where tasks were marked complete before PR events arrived.
- A significant race condition exists in `_auto_fail_task_on_pr_close`: the state check and update are not atomic, meaning a concurrent operation could change state between the SELECT and UPDATE, potentially overwriting a valid terminal state.
- `_auto_fail_task_on_pr_close` does not discard queued `TaskEvent` rows unlike `_auto_finalize_task_on_pr_merge`, leaving orphaned events that could trigger spurious foreman activity after a task is failed.
- The `TaskFinalizeMsg` broadcast is missing from `_auto_fail_task_on_pr_close`, creating an inconsistency where the frontend may not properly clean up worker state on failure compared to the merge path.
- Test coverage is solid with terminal-state guard tests and both lifecycle paths covered; the new `outcome` parameter fallback behavior is well-tested.

backend/routes/webhooks.py
```diff
@@ -227,6 +228,98 @@ async def schedule(
...
+    now = datetime.now(UTC)
+    deleted_at = now + timedelta(seconds=_DEFAULT_FAIL_TTL_SECS)
+
+    await db.exec(
+        update(Task)
+        .where(col(Task.id) == task_id)
+        .values(state="failed", finished_at=now, deleted_at=deleted_at)
+    )
+    await LockService(db).release(f"task:{task_id}")
+    await db.commit()
```
jmelloy, line 339: `_auto_fail_task_on_pr_close` never broadcasts `TaskFinalizeMsg`, but `_auto_finalize_task_on_pr_merge` does. If `TaskFinalizeMsg` is used by the frontend to release/return a worker to the idle pool, skipping it on the failure path will leave the worker in a stale state. The two paths should broadcast the same set of messages for consistency.

backend/routes/webhooks.py
jmelloy: `_auto_fail_task_on_pr_close` does not call `db.exec(delete(TaskEvent)...)` before committing, unlike `_auto_finalize_task_on_pr_merge`. Any queued follow-up `TaskEvent` rows will survive and may be dispatched later against a task already in `failed` state. Add the same `delete(TaskEvent)` step here.

backend/routes/webhooks.py
```diff
@@ -227,6 +228,98 @@ async def schedule(
...
+    result = await db.exec(
+        select(col(Task.worker_id), col(Task.state)).where(
+            col(Task.id) == task_id, col(Task.guild_id) == guild_pk
+        )
+    )
+    row = result.one_or_none()
+    if not row or row[1] in _WEBHOOK_TERMINAL_STATES:
+        return False
+
+    worker_id = row[0]
```
jmelloy: Both `_auto_finalize_task_on_pr_merge` and `finalize_task` in `tools.py` update the same row without a pessimistic lock (e.g. `SELECT ... FOR UPDATE`). If a webhook arrives while the AI is mid-execution of `finalize_task`, both writes can race. The terminal-state guard only prevents the second writer if the first has already committed — add `FOR UPDATE` or use `UPDATE ... WHERE state NOT IN (terminal states)` with row-count checking.

backend/foreman/tools.py
```diff
@@ -1114,6 +1114,8 @@ async def _exec_one_tool(guild_id: str, tu, user_id: str | None = None) -> dict:
 
             elif tu.name == "finalize_task":
                 task_id = inp["task_id"]
+                raw_outcome = inp.get("outcome", "done")
```
jmelloy, line 1117: The silent fallback `outcome = raw_outcome if raw_outcome in ("done", "failed") else "done"` hides unexpected values without any log warning. This will make future prompt regressions hard to diagnose; at minimum add `logger.warning("finalize_task: unknown outcome %r, defaulting to 'done'", raw_outcome)`.

jmelloy (commented):
- The PR correctly moves task lifecycle transitions (merge→done, close→failed) from AI-driven to deterministic webhook-driven logic, which is a sound architectural improvement.
- A race condition exists between the AI calling `finalize_task` and the webhook handler calling `_auto_finalize_task_on_pr_merge`/`_auto_fail_task_on_pr_close` — both paths update state without a DB-level row lock, so concurrent execution could result in a lost update or double-broadcast.
- `_auto_fail_task_on_pr_close` does not discard queued `TaskEvent` rows (unlike `_auto_finalize_task_on_pr_merge`), meaning stale follow-up events could be replayed after a task is marked failed.
- The fallback for invalid `outcome` values silently swallows the bad input and uses `'done'`; this masks potential prompt/schema bugs and should at minimum be logged as a warning.
- Test coverage for the new webhook paths is solid, including the terminal-state guard, though there is no test for the race condition between AI tool call and webhook arriving simultaneously.

PR #576, opened 2026-06-05
feat(foreman): capture API request IDs and response metadata in foreman turns

### LINKED ISSUE #575: Save API request IDs and response metadata in foreman turns
## Summary

When the foreman calls external APIs (GitHub, task management, worker assignment, etc.), the response metadata — particularly request IDs, rate-limit headers, and other diagnostics — should be persisted alongside the foreman turn record.

## Motivation

- **Debuggability**: API request IDs allow correlating foreman actions with provider-side logs (e.g. GitHub API request IDs, Anthropic API IDs).
- **Auditability**: Knowing exactly which API calls were made during a turn, and what they returned, makes replaying or diagnosing failures much easier.
- **Rate-limit awareness**: Storing rate-limit headers lets the system detect and react to exhaustion before it causes failures.

## Proposed approach

1. Identify where foreman turns are created/stored in the codebase.
2. For each tool call made during a turn, capture:
   - The tool name and input params (likely already stored)
   - The response — including any `x-request-id`, `x-github-request-id`, or equivalent headers
   - HTTP status code (for REST calls)
   - Timestamp of the call
3. Store this metadata in the foreman turn record (e.g. a `api_calls` JSON column, or a separate `foreman_turn_api_calls` table).
4. Surface request IDs in error messages / logs so they're easy to find.

## Where to look

- Foreman turn creation/storage logic
- Tool call wrappers / MCP tool implementations
- GitHub API client (look for where `x-github-request-id` headers are available)

jmelloy:
## Summary

- Adds an `api_calls_json` TEXT column to `foreman_turns` (via Alembic migration `20260605_000001`) to store per-HTTP-call metadata alongside each tool-result turn.
- Instruments all four GitHub HTTP helpers (`_gh_api`, `_gh_api_post`, `_gh_api_diff`, `_gh_graphql`) to record `x-request-id`, `x-github-request-id`, HTTP status code, path, and wall-clock timestamp on every call — including error responses.
- Uses a `ContextVar` (`_api_calls_ctx`) so concurrent tool executions each collect their own metadata without cross-contamination; `asyncio.to_thread` copies the context automatically so thread-pool calls participate too.
- Strips `api_calls` from tool-result blocks before they are sent to the Anthropic API (the API only accepts `type/tool_use_id/content/is_error`), then persists the metadata on the DB turn row.
- Surfaces GitHub request IDs in error result text so failures are immediately visible in logs without needing to query the database.
- Propagates the full pipeline to the standalone foreman process (`foreman/pioneer_foreman/`) via `ForemanHTTPClient.save_turn` and `runner._save_turn`.
- Exposes `api_calls_json` in the foreman history REST endpoints so callers can retrieve the metadata.

Closes #575

### DIFF
--- a/backend/foreman/runner.py
+++ b/backend/foreman/runner.py
@@ -156,6 +156,7 @@ async def _save_turn(
     *,
     is_tool_response: bool = False,
     parent_id: int | None = None,
+    api_calls: list | None = None,
 ) -> int:
     """Persist one turn to the DB. Returns the new row's id."""
     db = await get_db()
@@ -169,6 +170,7 @@ async def _save_turn(
             is_tool_response=1 if is_tool_response else 0,
             parent_id=parent_id,
             created_at=datetime.now(UTC),
+            api_calls_json=json.dumps(api_calls) if api_calls else None,
         )
         db.add(turn)
         await db.commit()
@@ -574,12 +576,20 @@ async def _run_foreman_ai(
             tool_results = await exec_tools(guild_id, tool_uses, user_id=user_id)
             # Truncate verbose results; filter to only IDs in the current batch so
             # stale results that survived history trimming are never persisted.
+            # Extract api_calls metadata before building trimmed — the Anthropic API
+            # only accepts {type, tool_use_id, content, is_error} in tool_result blocks.
             current_tool_use_ids = {tu.id for tu in tool_uses}
-            trimmed = [
-                {**r, "content": truncate_tool_result(r["content"])} if r.get("content") else r
-                for r in tool_results
-                if r.get("tool_use_id") in current_tool_use_ids
-            ]
+            all_api_calls: list[dict] = []
+            trimmed: list[dict] = []
+            for r in tool_results:
+                if r.get("tool_use_id") not in current_tool_use_ids:
+                    continue
+                for call in r.get("api_calls") or []:
+                    all_api_calls.append({"tool_use_id": r["tool_use_id"], **call})
+                entry = {k: v for k, v in r.items() if k != "api_calls"}
+                if entry.get("content"):
+                    entry = {**entry, "content": truncate_tool_result(entry["content"])}
+                trimmed.append(entry)
 
             # Broadcast tool-result events
             _now = datetime.now(UTC)
@@ -651,6 +661,7 @@ async def _run_foreman_ai(
                 trimmed,
                 is_tool_response=True,
                 parent_id=asst_turn_id,
+                api_calls=all_api_calls if all_api_calls else None,
             )
             logger.info(
                 "guild=%s round %d: %d tool call(s) dispatched: %s",
@@ -662,6 +673,21 @@ async def _run_foreman_ai(
                         "tool_use_id": r["tool_use_id"],
                         "name": r.get("name"),
                         "is_error": r.get("is_error", False),
+                        **(
+                            {
+                                "x_github_request_id": next(
+                                    (
+                                        c["x_github_request_id"]
+                                        for c in all_api_calls
+                                        if c.get("tool_use_id") == r["tool_use_id"]
+                                        and c.get("x_github_request_id")
+                                    ),
+                                    None,
+                                )
+                            }
+                            if r.get("is_error", False)
+                            else {}
+                        ),
                     }
                     for r in trimmed
                 ],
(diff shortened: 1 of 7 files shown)


backend/foreman/tools.py
```diff
@@ -746,6 +795,11 @@ async def _exec_one_tool(guild_id: str, tu, user_id: str | None = None) -> dict:
     inp = tu.input
     result_text = ""
     is_error = False
+    # Set up a fresh per-call collector for GitHub API response metadata.
```
jmelloy: `_api_calls_ctx.set(_api_call_log)` sets the ContextVar in the current task's context without saving the token returned by `.set()`, so there is no way to restore the previous value. If `_exec_one_tool` is ever called recursively or the context is reused across invocations, the collector list will be wrong. Save the token (`tok = _api_calls_ctx.set(...)`) and call `_api_calls_ctx.reset(tok)` in a `finally` block to guarantee cleanup.

backend/routes/foreman.py
```diff
@@ -266,6 +267,7 @@ async def get_foreman_history_for_user(
             "is_tool_response": bool(t.is_tool_response),
             "parent_id": t.parent_id,
```
jmelloy, line 268: This endpoint returns `api_calls_json` as a raw JSON string, while `get_foreman_history` in runner.py deserialises it to a list. Clients consuming this REST endpoint will receive a string they must parse themselves. Consider deserialising here with `json.loads(t.api_calls_json) if t.api_calls_json else None` for consistency.

backend/routes/foreman.py
```diff
@@ -330,6 +333,8 @@ async def create_foreman_turn(
 
     Response: ``{ "id": int, "created_at": str }``
     """
+    import json as _json
```
jmelloy, line 336: `import json as _json` inside the function body is unnecessary and unconventional — `json` is already imported at module level in most Python files and has no meaningful import cost. Move it to the top-level imports.

backend/foreman/runner.py
```diff
@@ -662,6 +673,21 @@ async def _run_foreman_ai(
                         "tool_use_id": r["tool_use_id"],
                         "name": r.get("name"),
                         "is_error": r.get("is_error", False),
+                        **(
+                            {
+                                "x_github_request_id": next(
+                                    (
+                                        c["x_github_request_id"]
+                                        for c in all_api_calls
+                                        if c.get("tool_use_id") == r["tool_use_id"]
```
jmelloy: The log entry only picks the *first* `x_github_request_id` across all api_calls for an errored tool result. If a tool made multiple HTTP calls before failing, only one request ID is surfaced. Consider collecting all IDs for the given `tool_use_id` (e.g., as a list) so no diagnostic information is silently dropped.

jmelloy (commented):
- The ContextVar approach for collecting API call metadata is sound, but `_api_calls_ctx.set()` in `_exec_one_tool` mutates the *current* task's context permanently rather than using a scoped token — concurrent coroutines sharing a context (e.g., if tasks are created without `copy_context`) could see stale collectors.
- Duplicate code: the tool-result extraction loop (strip `api_calls`, truncate content, aggregate `all_api_calls`) is copy-pasted verbatim across `backend/foreman/runner.py` and `foreman/pioneer_foreman/runner.py` with no shared helper, increasing maintenance risk.
- The REST endpoint `get_foreman_history_for_user` returns the raw `api_calls_json` string to clients rather than deserialising it, while the internal `get_foreman_history` function deserialises it — inconsistency could confuse API consumers.
- `import json as _json` is placed inside a function body in `backend/routes/foreman.py` rather than at the module top-level, which is an unusual and unnecessary pattern since `json` is a stdlib module with negligible import cost.
- Error surfacing of GitHub request IDs in result text is useful, but the log entry in runner.py that picks the *first* `x_github_request_id` for an error turn only includes one ID even when multiple calls failed, potentially hiding relevant IDs.

foreman/pioneer_foreman/runner.py
```diff
@@ -265,11 +267,19 @@ async def run_foreman_ai(
```
jmelloy, line 284: The tool-result extraction loop is an exact copy of the one in `backend/foreman/runner.py`. This duplication will diverge over time. Extract a shared utility function (e.g., `split_tool_results(tool_results, current_ids)`) returning `(trimmed, all_api_calls)` and import it in both runners.

jmelloy:
This needs to focus on anthropic api calls, not GitHub ones

PR #577, opened 2026-06-05
Increase foreman max sleep from 1h to 4h

jmelloy:
## Summary

- `backend/foreman/runner.py`: raise `POLL_MAX_SECS` from 3600 → 14400 seconds
- `foreman/pioneer_foreman/config.py`: raise the `poll_max_interval` dataclass default and the TOML parsing fallback from 3600 → 14400 seconds

Allows the foreman to back off to a 4-hour poll interval during long idle periods instead of capping at 1 hour.

## Test plan

- [ ] Confirm `POLL_MAX_SECS` and `poll_max_interval` read 14400 in both modules
- [ ] Run existing foreman tests (`pytest backend/tests/test_foreman.py`) — no behavioural changes expected other than the higher cap

### DIFF
--- a/backend/foreman/runner.py
+++ b/backend/foreman/runner.py
@@ -37,7 +37,7 @@ from ws_types import ChatMsg, ForemanPollStatusMsg
 logger = logging.getLogger(__name__)
 
 POLL_MIN_SECS = 60  # initial poll interval: 1 minute
-POLL_MAX_SECS = 3600  # maximum poll interval: 60 minutes
+POLL_MAX_SECS = 14400  # maximum poll interval: 4 hours
 
 # Per-guild background poll task registry
 _poll_tasks: dict[str, "asyncio.Task[None]"] = {}
--- a/foreman/pioneer_foreman/config.py
+++ b/foreman/pioneer_foreman/config.py
@@ -43,7 +43,7 @@ class Config:
     history_limit: int = 40
     # Poll settings
     poll_min_interval: int = 60
-    poll_max_interval: int = 3600
+    poll_max_interval: int = 14400
     # Logging
     log_level: str = "INFO"
 
@@ -200,7 +200,7 @@ def load(explicit_path: str | None = None, overrides: dict | None = None) -> Con
             overrides.get("poll_min_interval", poll_block.get("min_interval", 60))
         ),
         poll_max_interval=int(
-            overrides.get("poll_max_interval", poll_block.get("max_interval", 3600))
+            overrides.get("poll_max_interval", poll_block.get("max_interval", 14400))
         ),
         log_level=log_level.upper(),
         config_path=cfg_path,


foreman/pioneer_foreman/config.py
```diff
@@ -200,7 +200,7 @@ def load(explicit_path: str | None = None, overrides: dict | None = None) -> Con
             overrides.get("poll_min_interval", poll_block.get("min_interval", 60))
         ),
         poll_max_interval=int(
-            overrides.get("poll_max_interval", poll_block.get("max_interval", 3600))
+            overrides.get("poll_max_interval", poll_block.get("max_interval", 14400))
```
jmelloy, line 203: The hardcoded fallback here (14400) and the dataclass default on line 46 are now in sync, which is good. However, consider extracting a single module-level constant (e.g., `DEFAULT_POLL_MAX_INTERVAL = 14400`) shared by both sites to prevent them drifting apart again in the future.

foreman/pioneer_foreman/config.py
```diff
@@ -43,7 +43,7 @@ class Config:
     history_limit: int = 40
     # Poll settings
     poll_min_interval: int = 60
-    poll_max_interval: int = 3600
+    poll_max_interval: int = 14400
```
jmelloy, line 46: There is no guard ensuring `poll_max_interval > poll_min_interval` after construction. A `__post_init__` validator raising `ValueError` when `poll_max_interval <= poll_min_interval` would make misconfiguration fail fast rather than silently producing a broken backoff loop.

backend/foreman/runner.py
```diff
@@ -37,7 +37,7 @@
 logger = logging.getLogger(__name__)
 
 POLL_MIN_SECS = 60  # initial poll interval: 1 minute
-POLL_MAX_SECS = 3600  # maximum poll interval: 60 minutes
+POLL_MAX_SECS = 14400  # maximum poll interval: 4 hours
```
jmelloy, line 40: With a 4-hour maximum back-off, any work item that arrives shortly after the interval resets to the cap will be delayed up to 4 hours. If there is no event-driven (e.g., webhook or queue-based) wake-up mechanism, please confirm that a 4-hour detection latency is acceptable for the relevant use cases.

jmelloy (commented):
- Change is minimal and internally consistent: both `POLL_MAX_SECS` in `runner.py` and the `poll_max_interval` default/fallback in `config.py` are updated to the same value (14400), avoiding a split-brain scenario.
- No validation exists to enforce that `poll_min_interval < poll_max_interval`; while not broken by this PR, it remains a latent misconfiguration risk.
- The 4-hour cap significantly increases the worst-case detection latency for new work arriving during an idle period; callers should ensure an event-driven wake-up path exists or that the business requirement tolerates up to 4-hour delays.
- No tests are added or updated to assert the new cap value, so a future accidental revert would not be caught automatically.
- The PR description mentions a TOML parsing fallback, which is accurate — the hardcoded fallback in the `load()` function is correctly updated alongside the dataclass default.

backend/foreman/runner.py
jmelloy, line 40: Consider adding a brief inline note that this value should stay in sync with `poll_max_interval` in `foreman/pioneer_foreman/config.py`. The two constants serve the same purpose in different modules, and a future edit to one without the other would silently create divergent behavior.

foreman/pioneer_foreman/config.py
jmelloy, line 46: No test asserts the new default value. The test plan mentions running existing tests but those won't catch a regression back to 3600. A small parametrized unit test checking `Config().poll_max_interval == 14400` and `load()` returning 14400 when `max_interval` is absent would make this verifiable in CI.

foreman/pioneer_foreman/config.py
jmelloy, line 203: The fallback chain `overrides.get(..., poll_block.get(..., 14400))` is correct, but if a user's TOML explicitly sets `max_interval = 3600` they will silently keep the old cap. This is expected behavior, but worth documenting in a changelog or migration note so operators know to audit their TOML files if they want the new cap.

jmelloy (commented):
- Change is minimal and focused: two constants updated across two files, keeping them in sync with each other.
- No tests are added or updated to assert the new 14400 value, leaving the stated test plan item unverified in the PR itself.
- The TOML fallback default in `load()` is correctly updated to match the dataclass default, avoiding a silent mismatch between programmatic and file-based configuration.
- A 4-hour idle cap is a significant operational change; dependent systems (alerting, liveness checks, SLAs) that assume a maximum poll latency of 1 hour may be silently broken and should be reviewed.
- No backward-compatibility or migration concern for existing TOML configs that rely on the old default — the new default only kicks in when `max_interval` is absent from the file, which is the correct behavior.

PR #580, opened 2026-06-05
fix(worker): WebSocket ping/pong timeout, fast wake reconnect, reconnect tagging (#579)

### LINKED ISSUE #579: Audit worker WebSocket ping/pong and reconnect logic — flapping on flaky WiFi
## Problem

When the foreman host's laptop goes to sleep (lid closed, edge-of-WiFi) the worker WebSocket connection flaps repeatedly. This suggests the ping/pong heartbeat and/or the worker-online registration on reconnect may be doing something dumb — e.g. re-registering as a new worker on every reconnect, resetting state incorrectly, or not handling the sleep/wake cycle gracefully.

## What to audit

1. **WebSocket ping/pong** — find where the worker sends/receives pings. Verify the timeout is reasonable (not too aggressive for a sleeping host). Check whether missed pings are treated as a fatal disconnect or handled gracefully.
2. **Reconnect / re-registration logic** — on reconnect, does the worker re-send a full `worker-online` registration? Does the foreman treat it as a brand-new worker (clearing task state) or as a resumed connection? Ensure it's the latter.
3. **Worker state on foreman side** — verify the foreman doesn't reset or duplicate worker records on reconnect. Check that in-progress task assignments aren't lost when the worker briefly drops and reconnects.
4. **Sleep/wake handling** — consider whether the client should detect a stale connection after wake and cleanly reconnect rather than relying on ping timeout.

## Expected outcome

- Worker reconnects transparently after a sleep/wake cycle without losing task state or re-registering as a new worker.
- Ping timeouts are tuned so a sleeping laptop doesn't cause repeated `worker-offline`/`worker-online` churn.
- Any bugs found should be fixed; otherwise document the intended behaviour.

jmelloy:
## Summary

- **Ping timeout too short**: `WSClient` was using `ping_interval=20, ping_timeout=20` — any network blip or laptop sleep >20 s would close the connection and trigger a full offline/online cycle. Changed defaults to `ping_interval=60, ping_timeout=30` (matching the backend's 90 s offline threshold). Both values are now configurable WSClient constructor parameters.

- **Fast wake-up recovery**: `_heartbeat` now checks if the underlying socket is already `CLOSED` before sending. If so, it proactively closes it so the `messages()` loop reconnects on the very next `recv()` instead of waiting up to another 60 s for the next ping cycle.

- **Reconnect tagging in foreman trigger**: `handle_join` reads the worker's DB state *before* the upsert. When the worker was already `"online"` (fast reconnect — backend never marked it offline), `WSContext.worker_was_online = True`. `handle_worker_register` appends `reconnect=true` to the `[worker-online]` foreman trigger so the AI knows not to re-assign in-flight tasks.

- **Comment accuracy**: Updated `HEARTBEAT_INTERVAL_SECONDS` comment to document the three-level liveness design: heartbeats (25 s) drive the backend sweeper (90 s threshold); transport-level ping (60 s / 30 s) only closes dead TCP connections.

## Verification

**Task state survives reconnect** — confirmed. Task records are keyed by `worker_id`, not `agent_id`. The `_on_ws_reconnect` hook re-sends `join` + `worker-register` + the current `agent-state` for any non-idle slots, restoring the backend's view of in-progress work without losing the task assignment.

**No duplicate worker records** — confirmed. `handle_join` uses `pg_insert(...).on_conflict_do_update(...)` (upsert) for agents and a targeted `UPDATE` for workers — reconnect never creates new rows.

**No task state lost on reconnect** — confirmed. The task record (`Task.worker_id`, `Task.state`) is independent of the agent record. The stale-task watchdog only fires after 90 s without an active agent, giving the reconnecting worker plenty of time to re-announce.

## Tests added

- `test_ws_client_default_ping_settings` — asserts 60 s / 30 s defaults
- `test_ws_client_configurable_ping_settings` — asserts custom values accepted
- `test_heartbeat_closes_stale_socket_on_wake` — asserts `ws.close()` called when socket is CLOSED
- `test_worker_online_fast_reconnect_tags_reconnect` — `reconnect=true` present for fast reconnect
- `test_worker_online_fresh_join_no_reconnect_tag` — no `reconnect=true` for offline→online

Closes #579

### DIFF
--- a/backend/ws_handlers.py
+++ b/backend/ws_handlers.py
@@ -222,6 +222,12 @@ class WSContext:
     joined_agents: set[str] = field(default_factory=set)
     ws_user_id: str | None = None
     guild_pk: int | None = None
+    # Set to True by handle_join when the joining worker was already "online"
+    # in the DB before this connection arrived (i.e. a fast reconnect where the
+    # backend never marked it offline).  handle_worker_register uses this flag
+    # to tag the foreman trigger so the AI can distinguish a reconnect from a
+    # genuine first-time registration.
+    worker_was_online: bool = False
 
 
 # ---------------------------------------------------------------------------
@@ -243,6 +249,18 @@ async def handle_join(ctx: WSContext, data: dict) -> None:
     joined_at = datetime.now(UTC)
     is_external_foreman = agent_type == "foreman" and data.get("external") is True
     if not is_external_foreman:
+        if agent_type == "worker" and worker_id:
+            # Check the worker's current DB state *before* the upsert so
+            # handle_worker_register can tag reconnects.  "online" means the
+            # worker reconnected so fast that the backend never marked it
+            # offline — the foreman should not treat this as a brand-new join.
+            prev_state_res = await ctx.db.exec(
+                select(col(Worker.state)).where(
+                    col(Worker.id) == worker_id, col(Worker.guild_id) == ctx.guild_pk
+                )
+            )
+            ctx.worker_was_online = prev_state_res.one_or_none() == "online"
+
         stmt = pg_insert(Agent).values(
             id=agent_id,
             guild_id=ctx.guild_pk,
@@ -586,10 +604,16 @@ async def handle_worker_register(ctx: WSContext, data: dict) -> None:
     )
     agent_count = count_res.one()
     tools_suffix = f" tools={tools_str}" if tools_str else ""
+    # Tag fast reconnects so the foreman AI knows the worker never actually went
+    # offline from the backend's perspective (e.g. brief WiFi blip or laptop
+    # sleep shorter than the ping timeout).  Without this tag, the foreman may
+    # treat the reconnect as a fresh join and re-assign tasks that are still
+    # in flight.
+    reconnect_suffix = " reconnect=true" if ctx.worker_was_online else ""
     await _trigger_foreman(
         ctx.guild_id,
         "worker-online",
-        f"[worker-online] worker_id={worker_id} repos={repos_str} agent_count={agent_count}{tools_suffix}",
+        f"[worker-online] worker_id={worker_id} repos={repos_str} agent_count={agent_count}{tools_suffix}{reconnect_suffix}",
         task_name=f"foreman.worker-online:{worker_id}",
     )
 
--- a/backend/ws_handlers.py
+++ b/backend/ws_handlers.py
@@ -270,10 +270,13 @@ async def handle_join(ctx: WSContext, data: dict) -> None:
             # handle_worker_register can tag reconnects.  "online" means the
             # worker reconnected so fast that the backend never marked it
             # offline — the foreman should not treat this as a brand-new join.
+            # SELECT FOR UPDATE serializes concurrent reconnects from the same
+            # worker so both can't read "online", set worker_was_online=True,
+            # and emit reconnect=true simultaneously (TOCTOU fix).
             prev_state_res = await ctx.db.exec(
-                select(col(Worker.state)).where(
-                    col(Worker.id) == worker_id, col(Worker.guild_id) == ctx.guild_pk
-                )
+                select(col(Worker.state))
+                .where(col(Worker.id) == worker_id, col(Worker.guild_id) == ctx.guild_pk)
+                .with_for_update()
             )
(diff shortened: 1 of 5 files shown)


backend/ws_handlers.py
```diff
@@ -243,6 +265,18 @@ async def handle_join(ctx: WSContext, data: dict) -> None:
     joined_at = datetime.now(UTC)
     is_external_foreman = agent_type == "foreman" and data.get("external") is True
     if not is_external_foreman:
+        if agent_type == "worker" and worker_id:
+            # Check the worker's current DB state *before* the upsert so
+            # handle_worker_register can tag reconnects.  "online" means the
+            # worker reconnected so fast that the backend never marked it
+            # offline — the foreman should not treat this as a brand-new join.
```
jmelloy, line 272: There is a TOCTOU race here: if two connections from the same worker arrive nearly simultaneously, both reads of `Worker.state` may see `'online'` before either upsert completes, causing both to set `worker_was_online = True` and both to emit `reconnect=true`. Consider using a `SELECT ... FOR UPDATE` or a single upsert that returns the old state via `RETURNING` to make this atomic.

worker/pioneer_worker/worker.py
```diff
@@ -1114,8 +1118,22 @@ async def _heartbeat(self) -> None:
         on application messages. Without this loop a worker that finishes
         all its tasks would go silent and the sweeper would mark it offline
         even though the connection is healthy.
+
```
jmelloy, line 1121: Importing `_is_open` inside the loop body on every heartbeat tick is wasteful and unconventional. Move the import to the top of the file (or at least to module level in `worker.py`). Also, `_is_open` is a private symbol — consider either making it public or inlining the `state == State.CLOSED` check directly.

worker/pioneer_worker/worker.py
```diff
@@ -1114,8 +1118,22 @@ async def _heartbeat(self) -> None:
...
+        Fast wake-up recovery: if the underlying socket is already closed when
+        the heartbeat fires (e.g. the host woke from sleep after the transport-
+        level ping timed out), we proactively close it so the messages() loop
+        reconnects on its very next iteration instead of waiting up to another
+        full ping_interval.
         """
+        from .ws_client import _is_open
```
jmelloy: After calling `await self.ws.close()` for a stale socket, the loop immediately falls through to `_send(...)` in the same iteration. If `close()` doesn't raise, the send will attempt to write on a closed/None socket and likely raise an exception caught by the outer handler, but the log message will be confusing. A `continue` after `await self.ws.close()` would make the intent explicit and avoid the spurious send attempt.

backend/ws_handlers.py
```diff
@@ -622,7 +662,6 @@ async def handle_worker_disconnect(ctx: WSContext, data: dict) -> None:
             f"[worker-offline] worker_id={worker_id} reason=shutdown",
```
jmelloy, line 670: Removing `reset_foreman_poll` from `handle_worker_disconnect` changes observable behavior with no explanation of what that call was originally guarding. If the foreman poll was being reset to trigger an immediate re-evaluation after a worker went offline, removing it could delay task redistribution. At minimum this deserves a comment explaining why the reset is no longer needed here.

jmelloy (commented):
- The ping/pong timeout fix and configurable `WSClient` parameters are well-motivated and correctly implemented; the 60s/30s defaults align with the 90s backend sweeper threshold.
- The `worker_was_online` flag approach for reconnect tagging is clean, but relies on reading DB state before the upsert inside `handle_join`, which introduces a TOCTOU window under concurrent reconnects from the same worker.
- Removing `reset_foreman_poll` from `handle_worker_disconnect` is a behavioral change with no rollback path documented — the test only asserts the new behavior but doesn't explain what the original call was protecting against.
- The heartbeat fast-wake detection imports `_is_open` from `ws_client` at call time (inside the loop body) rather than at module level, which is inefficient and unusual; the private helper is also not exported in `__init__.py`.
- Test `test_worker_online_fresh_join_no_reconnect_tag` inserts the worker as `offline` and labels it "brand-new" in the docstring — these are actually different cases (never-seen vs. previously-offline), making the test intent misleading.

PR #585, opened 2026-06-05
Add api_request_log + refactor foreman_turns (#583)

### LINKED ISSUE #583: Add api_request_log table and refactor foreman_turns schema
## Summary

Add a new `api_request_log` table to capture full Anthropic API requests and responses made by the backend/foreman, and simplify the `foreman_turns` table.

### New table: `api_request_log`
Capture every outbound API call and its response. Columns should include (at minimum):
- `id` (PK)
- `request_id` (Anthropic request ID from response headers)
- `created_at`, `completed_at`
- `model`
- `system` (system prompt — moved from foreman_turns)
- `messages` (JSON — full outbound messages array)
- `response` (JSON — full response body)
- `input_tokens`, `output_tokens`, `cache_read_tokens`, `cache_write_tokens` (usage — moved from foreman_turns)
- `stop_reason`
- `task_id` (FK to tasks, nullable)
- Any other relevant metadata (temperature, max_tokens, etc.) in an `extra` JSONB column

### Refactor: `foreman_turns`
- Add `request_id` FK → `api_request_log.id`
- Add `task_id` FK → `tasks.id` (nullable)
- Remove `system` column (it's now on `api_request_log`)
- Remove usage columns (`input_tokens`, `output_tokens`, etc.) — now on `api_request_log`
- `system` prompt should be a column on the outbound row in `api_request_log`, not a separate record

### Migration
- Write an Alembic migration
- Backfill `request_id` on existing `foreman_turns` rows where possible
- Existing usage columns in `foreman_turns` can be left as nullable during transition or dropped

## Goals
- Full auditability of every Anthropic API call
- Simplified `foreman_turns` — one row per turn, system prompt inline on the request log row
- Usage/cost data centralised in `api_request_log`

jmelloy:
## Summary

- **New `api_request_log` table** (`ApiRequestLog` SQLAlchemy model): records every `anthropic.messages.create()` call the embedded foreman makes — full request (`model`, `system` prompt blocks, `messages` array, `extra` params) and full response (`request_id` header, response body as JSONB, token usage, `stop_reason`, `completed_at`). Two indexes: `task_id` and `created_at`.
- **Migration `20260605_000002_add_api_request_log`**: creates the table with proper FK to `tasks.id` and a unique constraint on `request_id` (Anthropic header value).
- **Migration `20260605_000003_refactor_foreman_turns`**: adds `foreman_turns.request_id` (FK → `api_request_log.id`) and `foreman_turns.task_id` (FK → `tasks.id`). Usage columns (`input_tokens`, `output_tokens`) are left nullable on existing rows — no data loss.
- **Foreman runner wiring** (`backend/foreman/runner.py`): `_create_api_request_log()` inserts a log row before each `messages.create()` call; `_complete_api_request_log()` updates it after with the response. `_save_turn()` now accepts `api_log_id` and `task_id` to populate the new FK columns. `_update_turn_tokens()` calls removed — usage lives in `api_request_log` going forward.
- Both the normal rounds and the safety-cap wrap-up call are fully instrumented.

## Test plan

- [ ] `docker compose up -d postgres-test && cd backend && python -m pytest` — all 119 tests pass
- [ ] Apply migrations against a live DB: `alembic upgrade head` — table created, columns added
- [ ] Downgrade roundtrip: `alembic downgrade -1` twice — columns/table dropped cleanly
- [ ] Start backend with `ANTHROPIC_API_KEY` set and send a foreman message; confirm `api_request_log` row appears with `request_id`, usage, and `completed_at` populated, and the `foreman_turns` assistant row has `request_id` FK set

Closes #583

### DIFF
--- a/backend/alembic/versions/20260605_000002_add_api_request_log.py
+++ b/backend/alembic/versions/20260605_000002_add_api_request_log.py
@@ -1,0 +1,57 @@
+"""Add api_request_log table.
+
+Revision ID: 20260605_000002_add_api_request_log
+Revises: 20260605_000001_add_api_calls_to_foreman_turns
+Create Date: 2026-06-05
+
+Records every Anthropic messages.create() call made by the foreman: the full
+request (model, system prompt, messages, extra params) and the full response
+(request_id header, response body, token usage, stop_reason). Created before
+the API call; updated after. foreman_turns.request_id FK points here.
+"""
+
+from __future__ import annotations
+
+from collections.abc import Sequence
+
+import sqlalchemy as sa
+from alembic import op
+
+revision: str = "20260605_000002_add_api_request_log"
+down_revision: str | Sequence[str] | None = "20260605_000001_add_api_calls_to_foreman_turns"
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+
+def upgrade() -> None:
+    op.create_table(
+        "api_request_log",
+        sa.Column("id", sa.Integer(), primary_key=True, autoincrement=True),
+        # Anthropic x-request-id response header; nullable until the call completes.
+        sa.Column("request_id", sa.Text(), nullable=True, unique=True),
+        sa.Column("created_at", sa.DateTime(timezone=True), nullable=False),
+        sa.Column("completed_at", sa.DateTime(timezone=True), nullable=True),
+        sa.Column("model", sa.Text(), nullable=False),
+        # Full JSON of the system blocks sent (serialized list of content block dicts).
+        sa.Column("system", sa.Text(), nullable=True),
+        # Full outbound messages array as JSONB.
+        sa.Column("messages", sa.JSON(), nullable=True),
+        # Full response body from Anthropic as JSONB.
+        sa.Column("response", sa.JSON(), nullable=True),
+        sa.Column("input_tokens", sa.Integer(), nullable=True),
+        sa.Column("output_tokens", sa.Integer(), nullable=True),
+        sa.Column("cache_read_tokens", sa.Integer(), nullable=True),
+        sa.Column("cache_write_tokens", sa.Integer(), nullable=True),
+        sa.Column("stop_reason", sa.Text(), nullable=True),
+        sa.Column("task_id", sa.Text(), sa.ForeignKey("tasks.id"), nullable=True),
+        # Extra API parameters: max_tokens, tools, tool_choice, etc.
+        sa.Column("extra", sa.JSON(), nullable=True),
+    )
+    op.create_index("ix_api_request_log_task_id", "api_request_log", ["task_id"])
+    op.create_index("ix_api_request_log_created_at", "api_request_log", ["created_at"])
+
+
+def downgrade() -> None:
+    op.drop_index("ix_api_request_log_created_at", table_name="api_request_log")
+    op.drop_index("ix_api_request_log_task_id", table_name="api_request_log")
+    op.drop_table("api_request_log")
--- a/backend/alembic/versions/20260605_000002_add_api_request_log.py
+++ b/backend/alembic/versions/20260605_000002_add_api_request_log.py
@@ -32,8 +32,8 @@ def upgrade() -> None:
         sa.Column("created_at", sa.DateTime(timezone=True), nullable=False),
         sa.Column("completed_at", sa.DateTime(timezone=True), nullable=True),
         sa.Column("model", sa.Text(), nullable=False),
-        # Full JSON of the system blocks sent (serialized list of content block dicts).
-        sa.Column("system", sa.Text(), nullable=True),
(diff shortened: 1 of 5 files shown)


backend/foreman/runner.py
```diff
@@ -532,6 +600,16 @@ async def _run_foreman_ai(
...
             )
+            await _complete_api_request_log(
+                api_log_id,
+                request_id=_anthropic_request_id,
+                response_dict=json.loads(_serialize_content(resp.content)),
+                input_tokens=_input_tokens,
+                output_tokens=_output_tokens,
+                cache_read_tokens=_cache_read,
+                cache_write_tokens=_cache_write,
+                stop_reason=resp.stop_reason,
```
jmelloy, line 615: `_serialize_content(resp.content)` serializes only the content blocks array, not the full Anthropic response object. The `response` column is intended to store the full response body, so this should use something like `resp.model_dump()` instead to capture model, usage, stop_reason, id, etc.

backend/foreman/runner.py
```diff
@@ -505,6 +567,12 @@ async def _run_foreman_ai(
                 round_num,
                 len(messages),
             )
+            api_log_id = await _create_api_request_log(
+                model=foreman_model,
+                system_blocks=system_blocks,
```
jmelloy, line 575: `_create_api_request_log` is called without `task_id` here (and at line 786), but `task_id` is available in the enclosing `_run_foreman_ai` scope. This means every log row will have `task_id=NULL`, defeating a primary purpose of the column.

backend/foreman/runner.py
```diff
@@ -180,16 +194,64 @@ async def _save_turn(
...
+    try:
+        log = ApiRequestLog(
+            created_at=datetime.now(UTC),
+            model=model,
+            system=json.dumps(system_blocks),
+            messages=messages,
+            extra=extra or None,
+            task_id=task_id,
+        )
+        db.add(log)
```
jmelloy, line 218: `return log.id or 0` silently returns 0 if the ID is falsy after commit+refresh, which would set `foreman_turns.request_id=0` — a non-existent FK value. Raise an explicit exception instead: `if log.id is None: raise RuntimeError(...)`

backend/foreman/runner.py
```diff
@@ -505,6 +567,12 @@ async def _run_foreman_ai(
...
             )
+            api_log_id = await _create_api_request_log(
+                model=foreman_model,
+                system_blocks=system_blocks,
+                messages=messages,
+                extra={"max_tokens": 1024, "tools": [t["name"] for t in FOREMAN_TOOLS]},
+            )
             _raw = await client.messages.with_raw_response.create(
                 model=foreman_model,
                 max_tokens=1024,
```
jmelloy, line 582: The `extra` dict stores `[t["name"] for t in FOREMAN_TOOLS]` (just names) rather than the actual tool definitions passed to the API. For replay or audit purposes, the full tool schemas should be stored.

backend/alembic/versions/20260605_000002_add_api_request_log.py
```diff
@@ -0,0 +1,57 @@
...
+down_revision: str | Sequence[str] | None = "20260605_000001_add_api_calls_to_foreman_turns"
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+
+def upgrade() -> None:
+    op.create_table(
+        "api_request_log",
+        sa.Column("id", sa.Integer(), primary_key=True, autoincrement=True),
+        # Anthropic x-request-id response header; nullable until the call completes.
```
jmelloy, line 30: The `system` column is `sa.Text()` in the migration but `sa.JSON()` is used for `messages` and `response`. For consistency and queryability, `system` should also be `sa.JSON()` since it stores a serialized JSON list. Currently the model serializes it with `json.dumps` before storing, which is an unnecessary double-encoding pattern.

jmelloy (commented):
- The PR introduces a well-structured `api_request_log` table and migration chain to capture full Anthropic API request/response details, replacing the previous per-turn token tracking approach.
- A significant concern is the `response_dict` serialization in `_complete_api_request_log`: `json.loads(_serialize_content(resp.content))` serializes only the content blocks, not the full response body (model, usage, stop_reason, etc.), making the `response` column misleading and incomplete.
- The `_create_api_request_log` call in `_run_foreman_ai` never passes `task_id`, so the log rows will always have `task_id=NULL` despite the column existing for that purpose — the task_id is available in the caller's scope.
- The fallback `return log.id or 0` in `_create_api_request_log` is dangerous: if `log.id` is somehow falsy after commit/refresh, callers silently use `0` as the log ID, potentially corrupting FK references.
- The `extra` field logs only tool names (`[t["name"] for t in FOREMAN_TOOLS]`) rather than the full tool schemas, which limits the usefulness of the log for replay or debugging.

PR #588, opened 2026-06-05
Validate tool availability in assign_task (#586)

### LINKED ISSUE #586: assign_task should validate requested tool is available on the target worker
## Problem

When `assign_task` is called with a `tool` parameter (e.g. `"claude"`, `"pi"`, `"codex"`), there is currently no check that the target worker actually supports that tool. Workers expose an `agent_count` and `tools` list in their registration data. If a task is assigned with `tool="codex"` to a worker that only has `["claude", "pi"]`, it will fail at runtime rather than being caught at assignment time.

## Desired Behaviour

- Before creating a worker task, check the worker's `tools` list (from its registration/state).
- If the requested `tool` is not in the worker's supported tools, raise a validation error (HTTP 400 or equivalent) with a message like: `"Worker w-xxx does not support tool 'codex'. Available tools: claude, pi"`.
- If no `tool` is specified, default to the first tool in the worker's `tools` list (or keep existing defaulting logic).

## Acceptance Criteria

- [ ] `assign_task` rejects tasks where the requested `tool` is not available on the worker, with a clear error message listing available tools.
- [ ] If `tool` is omitted, the system picks a sensible default from the worker's available tools.
- [ ] Existing tests pass; add a test covering the rejection case.

jmelloy:
## Summary

- In `assign_task`, fetch `Worker.tools` alongside `repos`/`org` and validate the requested tool against it.
- If the requested `tool` is not in the worker's registered tools list (and the list is non-empty), the call is rejected with a clear error: `"Worker w-xxx does not support tool 'codex'. Available tools: claude, pi"`.
- If `tool` is omitted, default to `worker.tools[0]` instead of hardcoding `"claude"`. Falls back to `"claude"` only for legacy workers with an empty tools list (backward-compatible).
- Extended `insert_worker` in `helpers.py` to accept a `tools` parameter so tests can set up workers with specific tool lists.
- Added two new tests in `test_foreman.py`: unsupported tool → error, omitted tool → defaults to first available.

## Test plan
- [ ] `cd backend && python -m pytest tests/test_foreman.py -k "assign_task" -v` (requires `docker compose up -d postgres-test`)
- [ ] Existing assign_task tests unaffected: workers with empty `tools=[]` continue to accept any tool (backward-compatible).

Closes #586

### DIFF
--- a/backend/foreman/tools.py
+++ b/backend/foreman/tools.py
@@ -869,7 +869,8 @@ async def _exec_one_tool(guild_id: str, tu, user_id: str | None = None) -> dict:
                         "review comments on the original PR.",
                         wid,
                     )
-                tool = inp.get("tool", "claude")
+                requested_tool: str | None = inp.get("tool")
+                tool = requested_tool or "claude"  # overwritten during worker tool validation
                 model = inp.get("model") or None
                 provider = inp.get("provider") or None
                 existing_task_id = inp.get("task_id")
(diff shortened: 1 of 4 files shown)


backend/foreman/tools.py
```diff
@@ -879,32 +880,45 @@ async def _exec_one_tool(guild_id: str, tu, user_id: str | None = None) -> dict:
...
+                        available = ", ".join(worker_tools)
                         result_text = (
-                            f"Worker {wid} cannot access repo(s) {unreachable} — "
-                            f"task NOT queued. Worker has {len(worker_repos)} registered "
-                            f"repo(s) and org={worker_org!r}. Choose a worker that has "
-                            f"access to these repos, or omit repos to use the worker's "
-                            f"full configured list."
+                            f"Worker {wid} does not support tool {requested_tool!r}. "
+                            f"Available tools: {available}"
                         )
```
jmelloy, line 900: When `requested_tool` is provided but `worker_tools` is empty (legacy worker), neither the `elif` branch nor the `else` branch runs, so `tool` keeps the value from line 872 (`requested_tool or "claude"`). This means an explicit `tool` value is silently accepted for legacy workers. This is likely the intended behavior, but it should be made explicit (e.g., add a comment or restructure so the `else` also covers the empty-tools case), otherwise future readers may accidentally break the implicit fallback.

backend/foreman/tools.py
```diff
@@ -869,7 +869,8 @@ async def _exec_one_tool(guild_id: str, tu, user_id: str | None = None) -> dict:
                         "review comments on the original PR.",
                         wid,
                     )
-                tool = inp.get("tool", "claude")
+                requested_tool: str | None = inp.get("tool")
```
jmelloy, line 872: The comment `# overwritten during worker tool validation` is inaccurate for two cases: (1) legacy workers with empty tools and an explicit `requested_tool` — the value is not overwritten; (2) `requested_tool=None` with empty tools — the fallback `"claude"` is also not overwritten. Consider removing the comment or replacing it with something like `# may be replaced below`.

backend/tests/test_foreman.py
```diff
@@ -432,6 +432,46 @@ async def test_assign_task_with_existing_task_id(self, db_session):
...
+
+    async def test_assign_task_defaults_to_first_tool(self, db_session):
+        insert_guild(db_session, "g-assign-tooldefault")
+        insert_worker(db_session, "g-assign-tooldefault", "w-tooldefault", tools='["pi", "claude"]')
+        with patch("foreman.tools.broadcast", new_callable=AsyncMock):
+            results = await exec_tools(
+                "g-assign-tooldefault",
+                [
+                    _fake_tool_use(
+                        "assign_task",
```
jmelloy, line 464: The test only asserts `"queued" in results[0]["content"].lower()` without also asserting the absence of an error flag. If the response contains both `"queued"` and an error indicator, the test would pass falsely. Consider also asserting that `"does not support" not in content` or checking `is_error` equivalent in the response.

jmelloy (commented):
- Core logic is correct: tool validation is added cleanly and the fallback chain (requested → first available → `"claude"`) is well-structured and backward-compatible.
- The repo-access check is correctly preserved under `if not is_error`, so the two validations compose properly without masking errors.
- A subtle bug exists: when `requested_tool` is provided and `worker_tools` is empty (legacy worker), `tool` is never explicitly set to `requested_tool` — it retains the value from the initial `requested_tool or "claude"` assignment, which is correct by coincidence but not by intent. The `else: tool = requested_tool` branch only runs when `worker_tools` is non-empty.
- The comment `# overwritten during worker tool validation` on the initial assignment is misleading when `requested_tool` is `None` and `worker_tools` is empty — in that case the initial value `"claude"` is never overwritten and is used as-is.
- Test coverage is solid: both the error path and the default-tool path are exercised, and the default-tool test validates the actual persisted value in the DB rather than just the response string.

PR #591, opened 2026-06-05
Fix task_id in api_request_logs and foreman turns

### LINKED ISSUE #590: api_request_logs and foreman turns not capturing task_id
## Problem

Both `api_request_logs` and foreman turn records are failing to capture `task_id`, leaving those rows unassociated with the task that generated them.

## Expected behavior

- Every row in `api_request_logs` that originates from a task-driven API call should have `task_id` populated.
- Every foreman turn (conversation turn record) should have `task_id` populated when the turn was triggered by or relates to a specific task.

## Steps to investigate

1. Find where `api_request_logs` rows are inserted — check whether `task_id` is being passed in or is being dropped.
2. Find where foreman turns are persisted — check whether `task_id` is available in context and being written.
3. Fix both write paths to propagate `task_id` correctly.
4. Add/update any relevant tests.

jmelloy:
Fixes #590 — propagates task_id correctly into api_request_logs and foreman turn records.

### DIFF
--- a/backend/routes/foreman.py
+++ b/backend/routes/foreman.py
@@ -59,6 +59,22 @@ router = APIRouter()
 logger = logging.getLogger(__name__)
 
 
+async def _require_task_in_guild(db: AsyncSession, task_id: str, guild_pk: int) -> None:
+    """Raise HTTP 404 if task_id does not exist or belongs to a different guild.
+
+    Task.id is the string PK (e.g. "t-abc123"), not an integer. guild_pk is the
+    integer FK from guilds.id. This check enforces guild ownership, which a bare
+    FK constraint cannot express.
+    """
+    result = await db.exec(
+        select(col(Task.id)).where(col(Task.id) == task_id, col(Task.guild_id) == guild_pk)
+    )
+    if result.one_or_none() is None:
+        raise HTTPException(
+            status_code=404, detail="Task not found or does not belong to this guild"
+        )
+
+
 # ---------------------------------------------------------------------------
 # Existing debug endpoints (unchanged)
 # ---------------------------------------------------------------------------
@@ -362,15 +378,7 @@ async def create_foreman_turn(
     if guild_pk is None:
         raise HTTPException(status_code=404, detail="Guild not found")
     if body.task_id is not None:
-        task_check = await db.exec(
-            select(col(Task.id)).where(
-                col(Task.id) == body.task_id, col(Task.guild_id) == guild_pk
-            )
-        )
-        if task_check.one_or_none() is None:
-            raise HTTPException(
-                status_code=404, detail="Task not found or does not belong to this guild"
-            )
+        await _require_task_in_guild(db, body.task_id, guild_pk)
     created_at = datetime.now(UTC)
     turn = ForemanTurn(
         guild_id=guild_pk,
--- a/backend/routes/foreman.py
+++ b/backend/routes/foreman.py
@@ -339,6 +339,7 @@ class ForemanTurnCreate(BaseModel):
     is_tool_response: bool = False
     parent_id: int | None = None
     api_calls: list | None = None  # per-HTTP-call metadata from tool execution
+    task_id: str | None = None
 
 
 @router.post("/guilds/{guild_id}/foreman/history")
@@ -370,6 +371,7 @@ async def create_foreman_turn(
         parent_id=body.parent_id,
         created_at=created_at,
         api_calls_json=_json.dumps(body.api_calls) if body.api_calls else None,
+        task_id=body.task_id,
     )
     db.add(turn)
     await db.commit()
--- a/backend/routes/foreman.py
+++ b/backend/routes/foreman.py
@@ -361,6 +361,16 @@ async def create_foreman_turn(
     guild_pk = await get_guild_pk(db, guild_id)
     if guild_pk is None:
         raise HTTPException(status_code=404, detail="Guild not found")
+    if body.task_id is not None:
+        task_check = await db.exec(
+            select(col(Task.id)).where(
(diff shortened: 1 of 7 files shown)


foreman/pioneer_foreman/runner.py
```diff
@@ -139,6 +141,7 @@ async def run_foreman_ai(
     primary_repo = guild_data.get("primary_repo")
     worker_rows = state.get("workers") or []
     task_rows = state.get("tasks") or []
+    _task_id: str | None = task_rows[0]["id"] if len(task_rows) == 1 else None
```
jmelloy: The heuristic `task_rows[0]['id'] if len(task_rows) == 1 else None` silently discards the task_id when more than one task is present. If the caller already knows the relevant task (e.g. it was passed as a parameter), this will produce incorrect NULL values. Consider either requiring an explicit `task_id` parameter or adding a log warning when the ambiguous multi-task case is hit.

backend/routes/foreman.py
```diff
@@ -370,6 +371,7 @@ async def create_foreman_turn(
         parent_id=body.parent_id,
```
jmelloy, line 389: There is no foreign-key or application-level validation that `body.task_id` actually exists in the tasks table for the given guild. Storing a dangling or cross-guild task_id will succeed silently. Consider adding a guard query or relying on a DB-level FK constraint.

foreman/pioneer_foreman/http_client.py
```diff
@@ -152,6 +153,8 @@ async def save_turn(
             body["parent_id"] = parent_id
         if api_calls:
             body["api_calls"] = api_calls
```
jmelloy, line 155: The `if task_id is not None` guard is correct, but it diverges from how `parent_id` is handled just above (`if parent_id`), which would also exclude `parent_id=0`. Keep the style consistent and prefer the explicit `is not None` form for both to avoid subtle bugs if IDs ever start at 0.

jmelloy (commented):
- The fix correctly propagates `task_id` through all call sites in both the backend monolith runner and the pioneer foreman runner, with consistent optional/nullable handling.
- The `_task_id` derivation logic (`task_rows[0]["id"] if len(task_rows) == 1 else None`) silently drops the task_id when multiple tasks are present; this edge-case behaviour is undocumented and could silently produce NULL task_id rows in multi-task scenarios that are legitimately tied to a single task.
- The HTTP client conditionally omits `task_id` from the request body only when it is `None`, which is correct, but the route model accepts any arbitrary string without validating it against the tasks table — an invalid/dangling task_id would be stored without error.
- Test coverage is good: both positive (task_id set) and negative (task_id null) cases are exercised at the unit and integration level for both runner variants.
- No database migration is shown in the diff; if `ForemanTurn.task_id` is a new column, the absence of an Alembic migration script is a gap that could break deployments.

foreman/pioneer_foreman/runner.py
```diff
@@ -139,6 +141,14 @@ async def run_foreman_ai(
     primary_repo = guild_data.get("primary_repo")
```
jmelloy, line 144: The single-task heuristic is fragile. When multiple tasks are present, `_task_id` is silently set to `None` and turns are stored without a `task_id`, which is exactly the bug being fixed. The caller should pass `task_id` explicitly rather than inferring it from `task_rows`, or the ambiguous case should raise an error instead of silently degrading.

backend/routes/foreman.py
```diff
@@ -360,6 +361,16 @@ async def create_foreman_turn(
     guild_pk = await get_guild_pk(db, guild_id)
     if guild_pk is None:
```
jmelloy, line 378: This extra `SELECT` on every turn creation adds a DB round-trip and couples the foreman history endpoint to the Task table. If the `task_id` is already validated when the task is fetched by the caller, this check is redundant. Consider moving validation to a shared utility or accepting the FK constraint from the DB as sufficient protection.

backend/routes/foreman.py
```diff
@@ -360,6 +361,16 @@ async def create_foreman_turn(
...
         raise HTTPException(status_code=404, detail="Guild not found")
```
jmelloy, line 379: `col(Task.id) == body.task_id` compares without also filtering on guild ownership being correct — the `col(Task.guild_id) == guild_pk` clause is there, which is good, but ensure `guild_pk` here is the integer PK and `body.task_id` is the string external ID; a type mismatch could cause the query to silently return no rows and incorrectly 404 valid tasks.

backend/tests/test_foreman_context_api.py
```diff
@@ -179,3 +179,58 @@ def test_get_foreman_context_isolates_across_guilds(client):
     )
```
jmelloy, line 179: There is no test for the cross-guild or invalid `task_id` case (expecting a 404 response). Adding a test where `task_id` belongs to a different guild would confirm the security boundary introduced by the validation query is working correctly.

jmelloy (commented):
- The PR correctly propagates `task_id` through the call stack from `runner.py` → `_save_turn` → HTTP client → API route, covering all turn types (system, user, assistant, tool response, wrap).
- The task validation added in `routes/foreman.py` is a good guard, but it adds an extra DB round-trip on every turn creation; since task ownership should already be enforced upstream, this may be unnecessary overhead.
- The `_task_id` inference logic in `foreman/pioneer_foreman/runner.py` (single-task heuristic) is fragile: when `len(task_rows) != 1` the turns are silently stored with `NULL`, which defeats the bug fix for the multi-task case.
- Tests cover the happy path and the NULL default well, but there is no test for the 404 path when an invalid/cross-guild `task_id` is supplied, nor for the ambiguous multi-task warning branch.
- No migration file is visible in the diff; if `ForemanTurn.task_id` is a new column, the absence of a DB migration would break deployments.

PR #595, opened 2026-06-06
Add boto3/Bedrock deps to requirements.txt + standalone foreman

### LINKED ISSUE #594: Add boto3/Amazon Bedrock libraries to requirements.txt and standalone foreman
Add anthropic[bedrock] into the default requirements.txt for both the backend and the foreman

## Acceptance criteria
- Both requirement files contain boto3/Bedrock deps
- No existing deps are broken

jmelloy:
## Summary

- Created root `requirements.txt` with `boto3>=1.34.0` (new file — none existed at repo root)
- Added `boto3>=1.34.0` to `backend/requirements.txt` (main service runtime deps)
- Added `boto3>=1.34.0` to `foreman/requirements.txt` (standalone foreman legacy runtime deps)
- Added `boto3>=1.34.0` to `backend/pyproject.toml` `[bedrock]` optional dep
- Added `boto3>=1.34.0` to `foreman/pyproject.toml` `[bedrock]` optional dep

`botocore` is a transitive dep of boto3 and is not listed separately. No existing pinned deps were changed.

## Test plan

- [ ] `pip install -r requirements.txt` installs boto3
- [ ] `pip install -r backend/requirements.txt` installs boto3
- [ ] `pip install -r foreman/requirements.txt` installs boto3
- [ ] `pip install -e backend/[bedrock]` installs boto3
- [ ] `pip install -e foreman/[bedrock]` installs boto3
- [ ] `boto3.client("bedrock-runtime", ...)` usable at runtime in both main service and standalone foreman

Closes #594


requirements.txt
```diff
@@ -0,0 +1 @@
+boto3>=1.34.0
```
jmelloy: A root-level `requirements.txt` with a single dependency and no comment is confusing. It's unclear what environment this is meant for (CI? local dev? deployment?). At minimum, add a comment explaining its purpose and whether it's meant to complement or replace the sub-package files.

backend/requirements.txt
```diff
@@ -9,6 +9,7 @@ sqlmodel
 alembic
 python-multipart
 anthropic
+boto3>=1.34.0
```
jmelloy: `boto3` is a large optional dependency (~10MB + botocore) that should only be needed when `FOREMAN_PROVIDER=bedrock` or similar. Installing it unconditionally for all backend environments inflates image size and install time for users who never use Bedrock. Consider moving this behind the `[bedrock]` extra and documenting that users must install it explicitly.

backend/pyproject.toml
```diff
@@ -11,7 +11,7 @@ dependencies = []
 
 [project.optional-dependencies]
 anthropic = ["anthropic>=0.49"]
```
jmelloy: `anthropic[bedrock]` already depends on `boto3` and `botocore` transitively, so this explicit entry is redundant. It's not harmful, but it creates a second version constraint to maintain. If kept, consider aligning the lower bound with what `anthropic[bedrock]` actually requires to avoid silent conflicts.

jmelloy (commented):
- The `boto3>=1.34.0` lower bound is reasonable but note that `anthropic[bedrock]` already pulls in `boto3` transitively via its own dependencies, making the explicit additions to `pyproject.toml` optional deps somewhat redundant (though explicit is generally better).
- The root `requirements.txt` containing only `boto3>=1.34.0` is concerning — a single-dep root requirements file with no context or comment could confuse contributors about its purpose and relationship to the sub-package requirement files.
- `backend/requirements.txt` lists `boto3>=1.34.0` as a hard runtime dependency for all backend installs, but Bedrock is an optional provider — this unconditionally installs a large AWS SDK (~10MB+) even for deployments that never use Bedrock.
- The `foreman/requirements.txt` already had a Bedrock section comment block, and the placement of `boto3>=1.34.0` (after the comment, before the blank line) is correct and well-structured.
- No pinning or upper bound is set anywhere; while flexible, this could lead to breakage if boto3 introduces a breaking change, especially given botocore's tight coupling.

backend/pyproject.toml
```diff
@@ -11,7 +11,7 @@ dependencies = []
 
 [project.optional-dependencies]
```
jmelloy, line 14: Just put these in regular dependencies.

PR #596, opened 2026-06-06
Fix default tool selection in assign_task / send_followup

### LINKED ISSUE #593: Bug: assign_task defaults to hardcoded "claude" when worker has no tools registered
## Summary

When `assign_task` is called without an explicit `tool` parameter, the logic _should_ default to the first tool in the worker's `tools` list. However, there are multiple hardcoded `\"claude\"` fallbacks in the path that activate when `worker.tools` is empty — meaning legacy workers or any worker with an empty tools list will always get `tool=\"claude\"`, regardless of intent.

### 1. `backend/foreman/tools.py` — `assign_task` handler (lines 872–906)

```python
# Line 872-873: initial default always "claude"
requested_tool: str | None = inp.get("tool")
tool = requested_tool or "claude"  # may be replaced below during tool validation

# Line 892-894: override only happens if worker has tools registered
worker_tools: list[str] = json.loads(worker_row.tools or "[]")
if requested_tool is None:
    tool = worker_tools[0] if worker_tools else "claude"   # <-- BUG: empty-tools fallback
```

If `worker.tools` is `"[]"` or `null`, the code falls back to `"claude"` even if the Foreman intended a different tool.

### 2. `backend/foreman/tools.py` — `create_task` handler (line 837)

```python
db.add(Task(
    ...
    tool="claude",    # hardcoded — no attempt to inherit from any context
    ...
))
```

Tasks created via `create_task` always get `tool="claude"` regardless of context.

### 3. `backend/models.py` — `Task` model (line 147)

```python
tool: str = Field(default="claude", sa_column_kwargs={"server_default": "'claude'"})
```

The database column default silently coerces any row with no explicit tool to `"claude"`.

### 4. `backend/foreman/tools.py` — `send_followup` (line 1176)

```python
tool=task_tool or "claude",
```

Follow-up tasks also default to `"claude"` if the parent task's tool field is falsy.

## The Test

`test_assign_task_defaults_to_first_tool` (`backend/tests/test_foreman.py:456`) validates the _happy path_ (worker with tools `["pi", "claude"]` → expects `tool="pi"`). That path works correctly. The bug surfaces only when `worker.tools` is empty/null (legacy workers), where the code correctly falls through to `"claude"` per line 894, but there's no way for the Foreman to override this via tool selection alone.

Additionally, `create_task` (line 837) unconditionally writes `tool="claude"` without inspecting the worker's tools list at all.

## Impact

- Any worker registered without a `tools` list gets `tool="claude"` hardcoded on every assigned task, even if the operator intends a different runner.
- The `create_task` path ignores tool selection entirely.

## Fix

1. **`assign_task` (line 894):** When `worker_tools` is empty and `requested_tool is None`, consider returning an error or using a configurable system-level default rather than silently picking `"claude"`.
2. **`create_task` (line 837):** Remove the hardcoded `tool="claude"` and either require the caller to pass a tool or derive it from guild/worker context.
3. **`Task` model default (line 147):** The `server_default` of `'claude'` is a silent footgun — document it clearly or make it `null` and enforce non-null at the application layer.
4. **`send_followup` (line 1176):** Same pattern — `tool=task_tool or "claude"` should not silently default.

## Relevant Files

| File | Line(s) | Issue |
|---|---|---|
| `backend/foreman/tools.py` | 873 | Initial `or "claude"` default in assign_task |
| `backend/foreman/tools.py` | 894 | Empty-tools fallback in assign_task |
| `backend/foreman/tools.py` | 837 | Hardcoded `tool="claude"` in create_task |
| `backend/foreman/tools.py` | 1176 | `or "claude"` fallback in send_followup |
| `backend/models.py` | 147 | `server_default="'claude'"` on Task.tool |
| `backend/tests/test_foreman.py` | 456 | Test covers happy path only (non-empty tools list) |

jmelloy:
## Summary

- **`assign_task`** (line 873): Already had the fix at lines 893–894 (`worker_tools[0] if worker_tools else "claude"`); no change needed.
- **`create_task`** (line 837): Uses `tool="claude"` as an unassigned-task placeholder; `assign_task` overwrites it with the worker's first tool when the task is later assigned — acceptable, no change.
- **`send_followup`** (line 1176): Added a Worker tools lookup for `target_worker_id` and changed the fallback from hardcoded `"claude"` to `worker_tools[0] if worker_tools else "claude"`, matching `assign_task` behaviour.

## Test plan

- [ ] `test_assign_task_defaults_to_first_tool` (existing) — confirms assign_task picks `worker_tools[0]`
- [ ] `test_send_followup_defaults_to_first_worker_tool` (new) — inserts a task with `tool=""` (falsy), worker with `tools=["pi","claude"]`, verifies `TaskFollowupMsg.tool == "pi"`

Closes #593

### DIFF
--- a/backend/foreman/tools.py
+++ b/backend/foreman/tools.py
@@ -1167,14 +1167,15 @@ async def _exec_one_tool(guild_id: str, tu, user_id: str | None = None) -> dict:
                                 ).model_dump(by_alias=True, exclude_none=True),
                             )
                             followup_worker_result = await db.exec(
-                                select(col(Worker.tools)).where(
-                                    col(Worker.id) == target_worker_id
-                                )
+                                select(col(Worker.tools)).where(col(Worker.id) == target_worker_id)
                             )
                             followup_worker_tools_json = followup_worker_result.one_or_none()
-                            followup_worker_tools: list[str] = json.loads(
-                                followup_worker_tools_json or "[]"
-                            )
+                            if isinstance(followup_worker_tools_json, str):
+                                followup_worker_tools: list[str] = json.loads(
+                                    followup_worker_tools_json or "[]"
+                                )
+                            else:
+                                followup_worker_tools = followup_worker_tools_json or []
                             await broadcast(
                                 guild_id,
                                 TaskFollowupMsg(
@@ -1182,7 +1183,12 @@ async def _exec_one_tool(guild_id: str, tu, user_id: str | None = None) -> dict:
                                     taskId=task_id,
                                     name=task_name or "",
                                     description=task_desc or "",
-                                    tool=task_tool or (followup_worker_tools[0] if followup_worker_tools else "claude"),
+                                    tool=task_tool
+                                    or (
+                                        followup_worker_tools[0]
+                                        if followup_worker_tools
+                                        else "claude"
+                                    ),
                                     model=task_model,
                                     provider=task_provider,
                                     branch=branch,
--- a/backend/foreman/tools.py
+++ b/backend/foreman/tools.py
@@ -1166,6 +1166,15 @@ async def _exec_one_tool(guild_id: str, tu, user_id: str | None = None) -> dict:
                                     finishedAt=None,
                                 ).model_dump(by_alias=True, exclude_none=True),
                             )
+                            followup_worker_result = await db.exec(
+                                select(col(Worker.tools)).where(
+                                    col(Worker.id) == target_worker_id
+                                )
+                            )
+                            followup_worker_tools_json = followup_worker_result.one_or_none()
+                            followup_worker_tools: list[str] = json.loads(
+                                followup_worker_tools_json or "[]"
+                            )
                             await broadcast(
                                 guild_id,
                                 TaskFollowupMsg(
@@ -1173,7 +1182,7 @@ async def _exec_one_tool(guild_id: str, tu, user_id: str | None = None) -> dict:
                                     taskId=task_id,
                                     name=task_name or "",
                                     description=task_desc or "",
-                                    tool=task_tool or "claude",
+                                    tool=task_tool or (followup_worker_tools[0] if followup_worker_tools else "claude"),
                                     model=task_model,
                                     provider=task_provider,
                                     branch=branch,
--- a/backend/tests/test_foreman.py
+++ b/backend/tests/test_foreman.py
@@ -505,9 +505,9 @@ class TestExecToolsDispatching:
                 ],
             )
         assert "t-flwup-tool" in results[0]["content"]
-        followup_msgs = [m for m in broadcast_calls if getattr(m, "type", None) == "task-followup"]
+        followup_msgs = [m for m in broadcast_calls if m.get("type") == "task-followup"]
         assert len(followup_msgs) == 1
-        assert followup_msgs[0].tool == "pi"
+        assert followup_msgs[0]["tool"] == "pi"
 
     async def test_send_followup_task_not_found(self, db_session):
         insert_guild(db_session, "g-followup-missing")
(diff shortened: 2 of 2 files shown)


jmelloy (commented):
- The core fix correctly mirrors `assign_task` behaviour by looking up the worker's first tool before falling back to `"claude"`, addressing the hardcoded default in `send_followup`.
- The DB query selects only `Worker.tools` (a JSON column) and uses `one_or_none()`, but the returned scalar may already be a list (depending on the ORM/driver JSON handling) rather than a raw JSON string, making the unconditional `json.loads()` call potentially fragile or error-prone.
- The new test patches `broadcast` but checks `m.get("type") == "task-followup"` on what should be a Pydantic model instance; if `broadcast` receives a model object (not a dict), `.get()` will fail silently and `followup_msgs` will always be empty, making the assertion vacuous.
- No error handling exists for the case where the DB query for `followup_worker_tools` itself raises (e.g. DB connectivity issue), which could surface an unhandled exception mid-broadcast flow.
- The fix only addresses the `send_followup` path; if other message types in `_exec_one_tool` have similar hardcoded `"claude"` fallbacks they remain unaddressed, but that is out of scope for this PR.

backend/foreman/tools.py
```diff
@@ -1166,14 +1166,23 @@ async def _exec_one_tool(guild_id: str, tu, user_id: str | None = None) -> dict:
                                     finishedAt=None,
                                 ).model_dump(by_alias=True, exclude_none=True),
                             )
+                            followup_worker_result = await db.exec(
+                                select(col(Worker.tools)).where(
+                                    col(Worker.id) == target_worker_id
+                                )
+                            )
+                            followup_worker_tools_json = followup_worker_result.one_or_none()
+                            followup_worker_tools: list[str] = json.loads(
```
jmelloy: `followup_worker_tools_json or "[]"` assumes the column value is always a raw JSON string. Depending on the SQLAlchemy dialect/driver, a JSON column may already be deserialized to a Python list, causing `json.loads()` to raise `TypeError`. Add a guard: `followup_worker_tools_json if isinstance(followup_worker_tools_json, str) else json.dumps(followup_worker_tools_json or [])`.

backend/tests/test_foreman.py
```diff
@@ -474,6 +474,41 @@ async def test_assign_task_defaults_to_first_tool(self, db_session):
...
+                "g-followup-tooldefault",
+                [
+                    _fake_tool_use(
+                        "send_followup",
+                        {"task_id": "t-flwup-tool", "instructions": "Continue work"},
+                    )
+                ],
+            )
+        assert "t-flwup-tool" in results[0]["content"]
+        followup_msgs = [m for m in broadcast_calls if m.get("type") == "task-followup"]
```
jmelloy: `broadcast_calls` will contain whatever `broadcast` is called with — likely Pydantic model instances, not dicts. Calling `.get("type")` on a model object returns `None` for every item, so `followup_msgs` will always be empty and `assert len(followup_msgs) == 1` will always fail (or never truly validate). Use `getattr(m, 'type', None) == 'task-followup'` or call `.model_dump()` first.

backend/tests/test_foreman.py
```diff
@@ -474,6 +474,41 @@ async def test_assign_task_defaults_to_first_tool(self, db_session):
...
+        assert len(followup_msgs) == 1
+        assert followup_msgs[0]["tool"] == "pi"
```
jmelloy: Consistent with the above: `followup_msgs[0]["tool"]` will raise `TypeError` if elements are Pydantic models. Use attribute access (`followup_msgs[0].tool`) or ensure dicts are captured.

PR #598, opened 2026-06-06
Fix poll loop backoff reset on foreman trigger (#556)

### LINKED ISSUE #556: Poll loop backoff resets to 60s on every foreman trigger — can spam Claude API
## Problem

The `_poll_loop` uses an exponential backoff (60s → 120s → … → 3600s) to avoid hammering Claude when there's nothing to do. However, every time an external event (WebSocket message, webhook) fires `run_foreman_ai`, the backoff counter is **reset to 60s**.

In a busy guild with frequent worker completions, this means the poll interval never grows beyond 60s — even at 3am when the foreman has nothing meaningful to do — generating unnecessary Claude API calls.

- `_poll_loop` resets `interval` to `MIN_POLL_INTERVAL` (60s) on each triggered run
- In a busy system, a new event can fire every few seconds, keeping the interval permanently at floor
- Each poll turn still hits the Claude API even if the system prompt + state yields no tool calls

## Fix Options

1. **Decouple poll reset from event-triggered runs**: only reset the backoff when the foreman actually *does something* (made a tool call), not merely when it was triggered.
2. **Idempotent trigger debounce**: if a run is already in-flight for a guild, swallow the new trigger rather than spawning a second coroutine and resetting the timer.
3. **Last-active tracking**: track `last_action_at` per guild; only reset the backoff if that timestamp is recent.

## Related
- #553 (concurrent foreman runs)

jmelloy:
## Summary

- **Decouple backoff reset from trigger events**: `reset_foreman_poll` only cancels/restarts the embedded poll loop at `POLL_MIN_SECS` when the foreman made at least one tool call within the last `POLL_MIN_SECS` seconds. When the foreman is idle (no tool calls), the existing loop is left undisturbed so the interval continues to grow normally (60 s → 120 s → … → 14400 s). Previously every `task-complete` / `agent-state` / `needs-input` event reset the timer to 60 s even when the foreman had nothing useful to do.
- **In-flight debounce**: `reset_foreman_poll` checks `_guild_locks` and returns immediately if any run is already executing for the guild. The spawned run was already triggered by `_trigger_foreman`; a second timer reset would be wasteful.
- **Last-action tracking**: new `_record_guild_action` / `_guild_active_recently` helpers track a per-guild monotonic timestamp that `_run_foreman_ai` updates whenever it processes at least one tool call.
- **Standalone foreman**: added `_in_flight` flag to `Foreman` to prevent concurrent trigger runs; `run_foreman_ai` returns `bool` (made tool calls); poll interval resets to `poll_min_interval` only when the foreman was active.

## Test plan

- [ ] `cd backend && python -m pytest tests/test_foreman_poll_backoff.py -v` — 9 new tests (all pass)
- [ ] `cd backend && python -m pytest tests/test_foreman_guild_lock.py tests/test_foreman_runner.py -v` — existing tests still pass

Closes #556

### DIFF
--- a/backend/foreman/runner.py
+++ b/backend/foreman/runner.py
@@ -4,6 +4,7 @@ import asyncio
 import json
 import logging
 import os
+import time
 from datetime import UTC, datetime
 
 from auth_deps import get_guild_pk
@@ -58,6 +59,22 @@ _poll_tasks: dict[str, "asyncio.Task[None]"] = {}
 # memory bounded and avoids stale snapshots piling up under load.
 # Entries are popped in the finally block after each run so the dict stays small.
 _guild_locks: dict[tuple[str, str | None], asyncio.Lock] = {}
+
+# Monotonic timestamp (time.monotonic()) of the last foreman run that made at
+# least one tool call, keyed by guild_id.  Used by reset_foreman_poll to decide
+# whether to reset the backoff: only reset when the foreman was recently active.
+_guild_last_action_at: dict[str, float] = {}
+
+
+def _record_guild_action(guild_id: str) -> None:
+    """Record that the foreman made tool calls for *guild_id* right now."""
+    _guild_last_action_at[guild_id] = time.monotonic()
+
+
+def _guild_active_recently(guild_id: str) -> bool:
+    """Return True if the foreman made tool calls within the last POLL_MIN_SECS."""
+    return (time.monotonic() - _guild_last_action_at.get(guild_id, 0.0)) <= POLL_MIN_SECS
+
 
 # Module-level client — reused across calls so the underlying httpx connection
 # pool isn't thrown away every invocation. Lazily initialised so import works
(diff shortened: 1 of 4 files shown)


backend/foreman/runner.py
```diff
@@ -340,16 +357,41 @@ async def _poll_loop(guild_id: str) -> None:
 
 
 def reset_foreman_poll(guild_id: str) -> None:
-    """Cancel any running poll loop for this guild and start a fresh one at POLL_MIN_SECS.
+    """Ensure a poll loop is running for this guild, resetting the backoff only when appropriate.
+
+    The backoff is only reset to POLL_MIN_SECS when the foreman was recently
```
jmelloy, line 362: Race condition: `_guild_locks` is checked here before the lock is acquired inside `_run_foreman_ai`. Between this check and when the spawned coroutine actually acquires its lock, another call to `reset_foreman_poll` will see no in-flight run and may restart the poll loop. Consider setting a per-guild 'pending' flag at task-spawn time and clearing it once the lock is acquired, or use a single asyncio.Lock per guild that is acquired before spawning.

backend/foreman/runner.py
```diff
@@ -59,6 +60,22 @@
...
+# least one tool call, keyed by guild_id.  Used by reset_foreman_poll to decide
+# whether to reset the backoff: only reset when the foreman was recently active.
+_guild_last_action_at: dict[str, float] = {}
+
+
+def _record_guild_action(guild_id: str) -> None:
+    """Record that the foreman made tool calls for *guild_id* right now."""
+    _guild_last_action_at[guild_id] = time.monotonic()
+
+
```
jmelloy, line 73: `_guild_last_action_at` is a module-level dict that is never pruned. Guilds that become inactive accumulate entries indefinitely. Add cleanup (e.g., pop the entry when `_guild_active_recently` returns False and the poll loop exits) to avoid unbounded memory growth in long-running deployments.

foreman/pioneer_foreman/foreman.py
```diff
@@ -80,6 +91,43 @@ async def _ws_send(message: dict) -> None:
...
+            async def _run_foreman(
+                guild_id: str,
+                human_message: str,
+                *,
+                extra_context: str = "",
+                user_id: str | None = None,
+                task_id: str | None = None,
+                task_name: str = "foreman.run",
+            ) -> None:
+                """Wrapper that enforces the in-flight flag and updates last_action_at."""
```
jmelloy, line 103: `self._in_flight` is a plain boolean set and cleared without any lock. While asyncio is single-threaded and this coroutine does `await` before setting `_in_flight = False`, a re-entrant trigger arriving during `await run_foreman_ai(...)` could still observe the correct value. However, if this class is ever used with threading or multiple event loops, or if `_in_flight` is checked across multiple coroutines, this becomes a race. Using `asyncio.Lock` would make the intent and safety guarantee explicit.

foreman/pioneer_foreman/runner.py
```diff
@@ -437,3 +443,6 @@ async def run_foreman_ai(
                 "createdAt": now,
```
jmelloy, line 443: The `except` block returns `False` unconditionally, which is correct. However, if a partial run made tool calls before the exception, returning `False` will suppress the backoff reset even though the foreman was active. Consider tracking `made_tool_calls` in the exception path and returning it so the backoff behaves consistently.

jmelloy (commented):
- The core fix correctly decouples backoff resets from foreman trigger events, preventing spurious Claude API calls when the foreman is idle — this is a sound design improvement.
- The in-flight debounce in `reset_foreman_poll` has a potential race: it checks `_guild_locks` but the lock is not acquired until inside `_run_foreman_ai`, so there is a window between `reset_foreman_poll` returning and the lock being acquired where a duplicate reset could slip through.
- The standalone `foreman.py` `_run_foreman` wrapper uses a per-instance `_in_flight` boolean instead of a lock, which is sufficient for a single-guild standalone process but would be unsafe if the class were ever instantiated for multiple guilds or if asyncio scheduling caused re-entry before the flag clears.
- `_guild_last_action_at` and `_guild_locks` are module-level dicts that are never pruned, creating a slow memory leak proportional to the number of distinct guild IDs seen over the process lifetime.
- The test for `_run_foreman_ai` is valuable but fragile: it patches many internal symbols by string name and relies on implementation details like `get_db` being used as a context manager-free async call, making it likely to break on minor internal refactors.

PR #601, opened 2026-06-06
Fix duplicate task trigger on issue click (#600)

### LINKED ISSUE #600: Duplicate Task Trigger on Issue Click
## Frontend Audit: Duplicate Task Trigger on Issue Click

Traced the full call path for the \"Work on issue\" flow. **Two separate actions fire in a single `onSend()` call**, which is the root cause of duplicate task spawning.

---

### The Two Paths

**Path 1 — WebSocket chat to foreman**

`frontend/src/components/chat-pane/ChatTab.vue`, lines 118–123:
```ts
guildStore.sendMessage({
  type: 'chat',
  from: 'user',
  to: 'foreman',
  content: text,   // "Work on issue #N in repo: \"title\""
})
```
The foreman AI receives this chat message, interprets it, and calls its `create_task`/`assign_task` tools — creating and assigning a task in the DB.

**Path 2 — Direct HTTP POST task assignment**

`frontend/src/components/chat-pane/ChatTab.vue`, lines 126–136:
```ts
const match = text.match(ISSUE_PATTERN)
if (match) {
  const [, issueNum, repoName] = match
  const worker = agentsStore.firstIdleWorker()
  if (worker) {
    await agentsStore.assignTask(worker.id, {
      description: text,
      issueNumber: parseInt(issueNum, 10),
      issueRepo: repoName.trim(),
    })
```
This calls `POST /guilds/{guildId}/workers/{workerId}/tasks` directly, creating **another** task in the DB — bypassing the foreman entirely.

Both paths execute inside the same `onSend()` invocation, one right after the other.

---

### Exact Conditions for the Duplicate

1. User clicks an issue row in the Issues tab (`IssuesTab.vue:27`: `@click="$emit('select-issue', issue)"`)
2. `ChatPane.selectIssue()` (`ChatPane.vue:60–63`) populates the input with `Work on issue #N in repo: "title"`
3. User presses Enter → `onSend()` fires
4. **Line 118**: chat message sent to foreman via WebSocket → foreman AI spawns Task A
5. **Line 126**: ISSUE_PATTERN matches → HTTP POST → Task B created directly

---

### Deduplication / Debounce

- **None on the frontend** between these two sub-paths.
- The backend has a per-`(guild, user)` `asyncio.Lock` in `foreman/runner.py` that drops *concurrent foreman invocations*, but the HTTP POST in Path 2 bypasses the foreman entirely, so the lock does not help here.

---

### Recommended Fix

**Remove the redundant direct `assignTask` call from `onSend()`** (`ChatTab.vue:126–162`). The foreman already handles the chat message and assigns the task via its AI tools — the frontend should not also assign directly. The system message "Task assigned to {worker}" currently produced by Path 2 can be omitted or replaced by a foreman reply.

If the intent is to keep Path 2 (fast-path assignment without foreman AI), then **remove the WebSocket send for issue-pattern messages** (skip lines 118–123 when `ISSUE_PATTERN` matches), so the foreman is not also triggered. But this silences the foreman from the conversation, which may be undesirable.

The cleanest fix is: **let the foreman own task creation; the frontend should only send the chat message** and wait for the foreman's reply.

Relevant files:
- `frontend/src/components/chat-pane/ChatTab.vue` — lines 114–163 (`onSend`)
- `frontend/src/components/ChatPane.vue` — lines 60–66 (`selectIssue`)
- `frontend/src/components/chat-pane/IssuesTab.vue` — line 27 (click handler)
- `frontend/src/stores/agents.ts` — `assignTask` (HTTP POST)
- `backend/foreman/tools.py` — `assign_task` tool (foreman-side task creation)

jmelloy:
## Summary

- Removed the `ISSUE_PATTERN` match block inside `onSend()` that was directly calling `agentsStore.assignTask(...)` via HTTP POST
- This block was firing in parallel with the WebSocket `sendMessage` call, causing two tasks to be created in the DB for every issue click
- The foreman is the authoritative task dispatcher; the frontend now only sends the chat message and lets the foreman handle task creation
- Cleaned up the now-unused `agentsStore` import, `useAgentsStore` store instance, and `ISSUE_PATTERN` constant
- Made `onSend` synchronous (removed `async`) since there's no longer any `await` in it

## Test plan

- [ ] Click an issue and press Enter — confirm exactly one task is created (via the foreman), not two
- [ ] Send a plain chat message — confirm it still works as expected
- [ ] Send non-issue text — confirm no regressions
- [ ] Frontend lint/type-check passes

Closes #600

### DIFF
--- a/frontend/src/components/chat-pane/ChatTab.vue
+++ b/frontend/src/components/chat-pane/ChatTab.vue
@@ -111,12 +111,32 @@ function onSend() {
   const text = inputText.value.trim()
   if (!text) return
 
-  guildStore.sendMessage({
+  const sent = guildStore.sendMessage({
     type: 'chat',
     from: 'user',
     to: 'foreman',
     content: text,
   })
+
+  if (sent) {
+    // Show the message immediately; the server echo will replace this entry when it arrives
+    guildStore.messages.push({
+      type: 'chat',
+      from: 'user',
+      to: 'foreman',
+      content: text,
+      createdAt: new Date().toISOString(),
+      _local: true,
+    })
+  } else {
+    guildStore.messages.push({
+      type: 'chat',
+      from: 'system',
+      content: 'Not connected — message not sent.',
+      createdAt: new Date().toISOString(),
+    })
+  }
+
   inputText.value = ''
 }
 
--- a/frontend/src/components/chat-pane/ChatTab.vue
+++ b/frontend/src/components/chat-pane/ChatTab.vue
@@ -71,15 +71,11 @@
 import { ref, computed, watch, nextTick } from 'vue'
 import { renderMarkdown } from '../../utils/markdown'
 import { useGuildStore } from '../../stores/guild'
-import { useAgentsStore } from '../../stores/agents'
 import { formatClock } from '../../utils/format'
 import { useChatGrouping, isToolUseGroup } from '../../composables/useChatGrouping'
-import type { ChatMessage, WSInbound } from '../../types'
+import type { ChatMessage } from '../../types'
 
 const guildStore = useGuildStore()
-const agentsStore = useAgentsStore()
-
-const ISSUE_PATTERN = /Work on issue #(\d+) in ([^:]+): "(.+)"/
 
 const inputText = ref('')
 const messagesEl = ref<HTMLElement | null>(null)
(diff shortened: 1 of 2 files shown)


frontend/src/components/chat-pane/ChatTab.vue
```diff
@@ -122,44 +118,6 @@ async function onSend() {
     content: text,
```
jmelloy, line 118: All user-facing feedback for the issue-assignment flow (task assigned confirmation, no idle worker warning, assignment error) has been removed with no replacement. If the foreman does not push corresponding status messages back over the WebSocket, users will have no indication of whether their issue click succeeded or failed. Ensure the foreman emits appropriate chat events, or add a loading/pending indicator here.

jmelloy, line 118: There is now no guard against the case where no idle worker is available — the message is simply sent to the foreman and nothing is shown to the user. Previously the frontend gave immediate feedback ('No idle worker available'). If this feedback is important for UX, consider keeping it or confirming the foreman always responds with a status message over the socket.

jmelloy (commented):
- Correctly removes the duplicate HTTP POST `assignTask` call that was firing in parallel with the WebSocket `sendMessage`, resolving the double-task creation bug.
- The removal of user-facing feedback messages (idle worker unavailable, task assigned confirmation, assignment error) is a regression in UX — the frontend now silently delegates to the foreman with no acknowledgement to the user if assignment fails or no worker is available.
- The cleanup of unused imports (`useAgentsStore`, `agentsStore`, `ISSUE_PATTERN`) is correct and keeps the codebase tidy.
- Making `onSend` synchronous is appropriate since no async operations remain, but if the foreman path ever needs error surfacing back to the UI this will need to be revisited.
- The fix assumes the foreman reliably handles task dispatch and surfaces errors through the WebSocket channel; this assumption should be verified or documented.

PR #606, opened 2026-06-07
feat: worker lifecycle tracking & graceful restart

### LINKED ISSUE #604: Backend should track spawned workers and gracefully restart them on new version/restart
## Summary

The backend should maintain a registry of worker processes it has spawned. When it detects a restart or a new version deployment, it should:

1. Signal all tracked workers to gracefully shut down (finish current task, then exit)
2. Start a fresh set of workers against the new version

## Motivation

Currently workers may be running against a stale backend version after a redeploy. This can cause API mismatches, stale code execution, or orphaned worker processes that never pick up the new tooling.

## Desired behaviour

- On startup, the backend records each worker process it spawns (PID, worker ID, start time, version)
- On backend restart/redeploy, it detects that existing workers were spawned by a previous version
- It sends each tracked worker a graceful-shutdown signal (e.g. `shutdown_worker` API call or SIGTERM with drain)
- Once workers have drained, a fresh set is started
- If a worker fails to shut down within a timeout, it is force-killed

## Open questions

- Where is worker spawning currently handled? (process manager, Docker, systemd, direct `subprocess`?)
- Is there a version identifier available at runtime (git SHA, package version, env var)?
- How many workers are typically spawned per backend instance?
- Should the registry be in-memory only (lost on crash) or persisted (DB / file)?

jmelloy:
## Summary

- **Schema**: adds 4 nullable columns to `workers` via Alembic migration `20260607_000000_add_worker_lifecycle_columns`: `container_id`, `spawned_version`, `started_at`, `drain_requested_at`
- **Spawn persistence**: both `foreman/tools.py:spawn_worker()` and `routes/workers.py:spawn_worker_container()` now record `container_id` (full Docker container id), `spawned_version` (backend version), and `started_at` immediately after `containers.run()` succeeds
- **Version detection** (`backend/worker_lifecycle.py:get_current_version()`): prefers the `PIONEER_VERSION` env var (set at image build time), falls back to `git rev-parse --short HEAD` for local dev, returns `None` if neither is available
- **Stale detection + drain** (`drain_stale_workers_on_startup()`): on startup, queries workers whose `spawned_version` differs from current and state is not `offline`; sends a graceful `worker-shutdown` WS signal to each; waits `PIONEER_WORKER_DRAIN_TIMEOUT` seconds (default 60, env-configurable); force-kills surviving Docker containers via `container.kill()` using the stored `container_id`; workers with no `container_id` (non-Docker/compose) are warned but not killed
- **Lifespan wiring** (`main.py`): `drain_stale_workers_on_startup()` is spawned as a background task *before* `reset_connection_state()` so pre-restart DB state is still available for stale detection; the drain task is cancelled on shutdown
- **Tests**: 12 unit tests in `tests/test_worker_lifecycle.py` covering version detection, drain skip conditions, shutdown signalling, Docker force-kill, null `container_id` handling, and timeout env var

## Backward compatibility

All new columns are nullable — existing `Worker` rows without `container_id` or `spawned_version` are skipped by the lifecycle drain (graceful signal is sent but no force-kill attempt is made).

## Test plan

- [x] `pytest tests/test_worker_lifecycle.py` — 12 new tests pass
- [x] No existing tests broken (all non-DB unit tests still pass)
- [ ] Deploy with `PIONEER_VERSION` stamped into the image and verify stale workers are drained after a rolling backend update
- [ ] Verify `PIONEER_WORKER_DRAIN_TIMEOUT=120` env var is respected

Closes #604

### DIFF
--- a/backend/tests/test_worker_lifecycle.py
+++ b/backend/tests/test_worker_lifecycle.py
@@ -185,7 +185,8 @@ async def test_drain_force_kills_surviving_container(monkeypatch):
         async def __aexit__(self, *a):
             return False
 
-    # force_kill_stale_workers opens one DB session to fetch the workers.
+    # Patch drain timeout to 0 so the polling loop is skipped; only the final
+    # container-fetch session is needed.
     sessions = [FakeDB([fake_worker])]
     session_iter = iter(sessions)
 
@@ -198,12 +199,51 @@ async def test_drain_force_kills_surviving_container(monkeypatch):
     with (
         patch("worker_lifecycle.AsyncSessionLocal", lambda: next(session_iter)),
         patch("worker_lifecycle.asyncio.sleep", AsyncMock()),
+        patch("worker_lifecycle.WORKER_DRAIN_TIMEOUT", 0.0),
         patch.dict("sys.modules", {"docker": fake_docker_module}),
     ):
         await force_kill_stale_workers(["w-stale2"])
 
     fake_docker_client.containers.get.assert_called_once_with("abc123deadbeef")
     fake_container.kill.assert_called_once()
+
+
+@pytest.mark.asyncio
+async def test_force_kill_exits_early_when_workers_self_terminate():
+    """Polling loop exits before the full drain timeout once all workers go offline."""
+
+    class FakeDB:
+        def __init__(self, rows):
+            self._rows = rows
+            self.exec = AsyncMock(
+                return_value=MagicMock(all=MagicMock(return_value=self._rows or []))
+            )
+            self.commit = AsyncMock()
+
+        async def __aenter__(self):
+            return self
+
+        async def __aexit__(self, *a):
+            return False
+
+    # First poll: worker still alive. Second poll: worker gone → early exit.
+    sessions = [FakeDB(["w-early"]), FakeDB([])]
+    session_iter = iter(sessions)
+    sleep_calls: list[float] = []
+
+    async def fake_sleep(s):
+        sleep_calls.append(s)
+
+    with (
+        patch("worker_lifecycle.AsyncSessionLocal", lambda: next(session_iter)),
+        patch("worker_lifecycle.asyncio.sleep", fake_sleep),
+        patch("worker_lifecycle.WORKER_DRAIN_TIMEOUT", 60.0),
+    ):
+        await force_kill_stale_workers(["w-early"])
+
+    # Returned after the second DB poll — only one sleep occurred (after first poll).
+    assert len(sleep_calls) == 1
+    # Docker kill was never reached because function returned early.
 
 
 @pytest.mark.asyncio
@@ -227,6 +267,7 @@ async def test_drain_skips_kill_when_no_container_id(monkeypatch):
         async def __aexit__(self, *a):
             return False
 
+    # Patch drain timeout to 0 so the polling loop is skipped.
     sessions = [FakeDB([fake_worker])]
     session_iter = iter(sessions)
 
@@ -235,6 +276,7 @@ async def test_drain_skips_kill_when_no_container_id(monkeypatch):
     with (
         patch("worker_lifecycle.AsyncSessionLocal", lambda: next(session_iter)),
         patch("worker_lifecycle.asyncio.sleep", AsyncMock()),
+        patch("worker_lifecycle.WORKER_DRAIN_TIMEOUT", 0.0),
         patch.dict("sys.modules", {"docker": fake_docker_module}),
     ):
         await force_kill_stale_workers(["w-nocontainer"])
--- a/backend/tests/test_worker_lifecycle.py
+++ b/backend/tests/test_worker_lifecycle.py
@@ -15,9 +15,11 @@ sys.path.insert(0, os.path.join(os.path.dirname(__file__), ".."))
 
 from worker_lifecycle import (  # noqa: E402
     WORKER_DRAIN_TIMEOUT,
+    _spawn_replacement_workers,
     drain_stale_workers_on_startup,
     force_kill_stale_workers,
     get_current_version,
+    record_worker_spawn,
 )
 
 # ---------------------------------------------------------------------------
@@ -201,6 +203,7 @@ async def test_drain_force_kills_surviving_container(monkeypatch):
         patch("worker_lifecycle.asyncio.sleep", AsyncMock()),
         patch("worker_lifecycle.WORKER_DRAIN_TIMEOUT", 0.0),
         patch.dict("sys.modules", {"docker": fake_docker_module}),
+        patch("worker_lifecycle._spawn_replacement_workers", AsyncMock()),
     ):
         await force_kill_stale_workers(["w-stale2"])
 
@@ -238,6 +241,7 @@ async def test_force_kill_exits_early_when_workers_self_terminate():
         patch("worker_lifecycle.AsyncSessionLocal", lambda: next(session_iter)),
         patch("worker_lifecycle.asyncio.sleep", fake_sleep),
         patch("worker_lifecycle.WORKER_DRAIN_TIMEOUT", 60.0),
+        patch("worker_lifecycle._spawn_replacement_workers", AsyncMock()),
     ):
         await force_kill_stale_workers(["w-early"])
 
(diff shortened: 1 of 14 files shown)


backend/worker_lifecycle.py
```diff
@@ -0,0 +1,167 @@
...
+        ).all()
+
+    if not rows:
+        logger.info(
+            "worker_lifecycle: no stale workers found (current version=%s)", current_version
+        )
+        return
+
+    logger.info(
+        "worker_lifecycle: %d stale worker(s) from a previous deploy — "
```
jmelloy, line 184: Workers with `NULL spawned_version` (i.e., all rows created before this migration) are silently excluded from the stale-worker query. On first deploy after the migration, any previously-running workers will never be drained. Consider adding a condition to also catch `spawned_version IS NULL` rows, or at minimum document this known gap explicitly.

backend/worker_lifecycle.py
```diff
@@ -0,0 +1,167 @@
...
+                    col(Worker.id).in_(stale_ids),
+                    col(Worker.state) != "offline",
+                )
+            )
+        ).all()
+
+    for worker in still_alive:
+        if worker.container_id:
+            # Step 4: force-kill the Docker container.
+            try:
```
jmelloy, line 258: `docker.from_env()` is called inside the loop, creating a new Docker client (and potentially a new HTTP connection) for every surviving worker. Move client instantiation above the `for` loop and reuse it across iterations.

backend/worker_lifecycle.py
```diff
@@ -0,0 +1,167 @@
...
+                await broadcast_msg(
+                    guild_slug,
+                    WorkerShutdownMsg(
+                        workerId=worker.id,
+                        reason="backend restarted: version mismatch",
+                    ),
+                )
+            except Exception:
+                logger.warning(
+                    "worker_lifecycle: could not send shutdown signal to worker %s",
```
jmelloy, line 207: There is no early-exit once all stale workers have gone offline before the drain timeout completes. The full `WORKER_DRAIN_TIMEOUT` sleep always elapses even if all workers self-terminate in 2 seconds. Consider polling at a shorter interval or using an event to short-circuit the wait.

backend/tests/test_worker_lifecycle.py
```diff
@@ -0,0 +1,261 @@
...
+
+    # Docker was never asked to kill anything.
+    fake_docker_module.from_env.assert_not_called()
+
+
+def test_drain_timeout_default():
+    assert WORKER_DRAIN_TIMEOUT == 60.0
+
+
+def test_drain_timeout_env_override(monkeypatch):
```
jmelloy, line 337: The `importlib.reload` for the timeout env-override test does not use a `try/finally`, so if the assertion fails the module is left with `WORKER_DRAIN_TIMEOUT=120` for subsequent tests. Wrap the reload and assertion in a `try/finally` or use a pytest fixture with a finalizer.

backend/worker_lifecycle.py
```diff
@@ -0,0 +1,167 @@
...
+                .join(Guild, col(Guild.id) == col(Worker.guild_id))
+                .where(
+                    col(Worker.spawned_version).isnot(None),
+                    col(Worker.spawned_version) != current_version,
+                    col(Worker.state) != "offline",
+                )
+            )
+        ).all()
+
+    if not rows:
```
jmelloy, line 177: The SQL query selects `col(Guild.guild_id).label('guild_slug')` alongside `Worker`, but `Guild.guild_id` is typically a UUID/integer PK, not the human-readable slug. If `broadcast_msg` expects a guild slug string, this may pass the wrong value and silently fail to reach any connected worker sockets.

jmelloy (commented):
- The `drain_stale_workers_on_startup()` function is spawned as a background task but its 60-second `asyncio.sleep` blocks the drain window regardless of whether workers have already gone offline, and there is no early-exit path once all stale workers self-terminate.
- Version comparison silently skips workers with `NULL spawned_version` (pre-migration rows), meaning legacy workers that were running before the migration will never be drained — this should be documented or handled explicitly.
- The Docker `client = docker.from_env()` is instantiated inside a loop per surviving worker, which is wasteful and could fail differently for each iteration; it should be created once before the loop.
- The `guild_slug` label in the SQLModel `select` query uses `col(Guild.guild_id).label('guild_slug')` but the result rows are unpacked as `(worker, guild_slug)` — SQLModel's `.all()` on a multi-column select returns tuples of the actual column values, so this relies on tuple ordering that may be fragile and is not type-safe.
- Tests use `importlib.reload` on `worker_lifecycle` to test env-var overrides of `WORKER_DRAIN_TIMEOUT`, but this mutates the module globally and can leak state into other tests if a failure prevents the cleanup `reload`; a `finally` block or fixture teardown should be used.

jmelloy:
### Code review

Found 1 issue from historical PR context:

1. Lifecycle DB update duplicated across two spawn paths with no shared helper (same pattern flagged in PR #576 review: "Duplicate code [...] with no shared helper, increasing maintenance risk")

The identical `container_id` / `spawned_version` / `started_at` update block is copy-pasted into both `backend/foreman/tools.py` and `backend/routes/workers.py`. If a third spawn path is added, or if the lifecycle fields change, both sites must be updated together. PR #576's review explicitly called out this pattern of duplication across parallel code paths as a maintenance risk.

https://github.com/jmelloy/pioneer-square/blob/1a39b5b9925f0dd24007d64999ff16b744758bf1/backend/foreman/tools.py#L748-L761

https://github.com/jmelloy/pioneer-square/blob/1a39b5b9925f0dd24007d64999ff16b744758bf1/backend/routes/workers.py#L272-L284

Fix: extract a shared `async def _record_worker_spawn(db, worker_id, container_id)` helper in `worker_lifecycle.py` and call it from both sites.

jmelloy (commented):
### Code review

Found 3 issues:

1. **`drain_stale_workers_on_startup()` is scheduled as a background task but `reset_connection_state()` runs immediately after, wiping the DB state the drain depends on.** `spawn()` calls `asyncio.create_task()`, which defers execution until the current coroutine yields. The very next line is `await reset_connection_state()`, which runs to completion and sets every worker's `state` to `'offline'` before the drain task ever executes its initial `SELECT WHERE state != 'offline'`. The drain finds zero stale workers and returns early — the entire feature is a no-op on every restart. The comment on line 340 says "BEFORE reset_connection_state()" but `spawn()` does not provide that ordering guarantee; the drain needs to be `await`ed synchronously for the detection step, or the query must not rely on `state` (e.g. filter on `spawned_version != current_version AND started_at < <cutoff>` regardless of `state`).

https://github.com/jmelloy/pioneer-square/blob/1a39b5b9925f0dd24007d64999ff16b744758bf1/backend/main.py#L340-L348

2. **The post-sleep "still alive" force-kill check is dead code for the same reason.** After the 60-second `asyncio.sleep`, the code re-queries `WHERE state != 'offline'` to find containers to force-kill. But `reset_connection_state()` set all workers to `'offline'` at startup — well before the 60-second sleep expires — so this query always returns zero rows and `container.kill()` is never called.

https://github.com/jmelloy/pioneer-square/blob/1a39b5b9925f0dd24007d64999ff16b744758bf1/backend/worker_lifecycle.py#L131-L152

3. **`docker.from_env()`, `containers.get()`, and `container.kill()` are synchronous blocking calls inside an `async` function with no `asyncio.to_thread()` wrapper.** These calls block the event loop for the duration of each Docker daemon HTTP round-trip. The rest of the codebase (`foreman/tools.py`) uses `asyncio.to_thread()` for every Docker SDK call for this exact reason.

https://github.com/jmelloy/pioneer-square/blob/1a39b5b9925f0dd24007d64999ff16b744758bf1/backend/worker_lifecycle.py#L141-L155

backend/worker_lifecycle.py
```diff
@@ -0,0 +1,167 @@
...
+
+Lifecycle steps on backend startup
+───────────────────────────────────
+1. get_current_version()  — PIONEER_VERSION env var or git short SHA
+2. detect_stale_workers() — workers whose spawned_version differs from current
+3. Send graceful worker-shutdown WS signal to each stale worker (best-effort;
+   sockets are typically dead on a cold restart, but works for hot reloads)
+4. Record drain_requested_at timestamp
+5. Wait PIONEER_WORKER_DRAIN_TIMEOUT seconds (default 60)
+6. Force-kill any surviving containers via Docker SDK using stored container_id
```
jmelloy: We don't need to force kill, just make this fire and forget. Completed work should finish.

PR #609, opened 2026-06-07
Fix foreman startup with alternate provider (Bedrock)

### LINKED ISSUE #608: Embedded foreman fails to start with alternate provider (e.g. Bedrock) due to session error
## Bug Report

When starting the embedded foreman with an alternate provider (e.g. AWS Bedrock), it fails to start due to a session error.

## Steps to Reproduce
1. Configure the embedded foreman to use an alternate provider (e.g. `bedrock`)
2. Attempt to start the foreman process
3. Observe a session-related error preventing startup

## Expected Behavior
The foreman should start successfully with any supported provider, including Bedrock.

## Actual Behavior
A session error is raised and the foreman fails to start.

## Likely Area
- Provider/session initialization code for the embedded foreman
- May involve how sessions are instantiated differently for non-default providers (Anthropic direct vs. Bedrock)

## Notes
Reported during live usage. Needs investigation of how alternate providers are initialized at foreman startup.

jmelloy:
## Summary

Fixes two bugs that prevented the embedded foreman from working when configured with an alternate provider (e.g. `FOREMAN_PROVIDER=bedrock`):

- **`backend/foreman/tools.py`** – `review_pr_internal` hardcoded `_anthropic.AsyncAnthropic()` instead of the shared `make_anthropic_client()` factory. On Bedrock deployments only AWS credentials are available (no `ANTHROPIC_API_KEY`), so the direct constructor raised `AuthenticationError` and the foreman failed.

- **`backend/foreman_core/llm.py`** – `make_anthropic_client()` resolved the provider from a module-level `FOREMAN_PROVIDER` constant captured at import time, not from `os.environ` dynamically. This was inconsistent with `get_foreman_model()` and meant tests (or environments) that set the env var after module import got the wrong client type.

## Changes

| File | Change |
|------|--------|
| `backend/foreman/tools.py` | Replace `_anthropic.AsyncAnthropic()` with `make_anthropic_client()` in `review_pr_internal` |
| `backend/foreman_core/llm.py` | Read `FOREMAN_PROVIDER` via `os.environ.get(...)` on every call instead of the cached module-level constant |
| `backend/tests/test_foreman_llm.py` | 16 new tests: provider dispatch, Bedrock region forwarding, API key passthrough, dynamic env-var read |

## Test plan

- `python -m pytest backend/tests/test_foreman_llm.py -v` → 16 passed
- `python -m pytest backend/tests/test_foreman.py::TestBuildSystemPrompt backend/tests/test_foreman.py::TestSerializeContent backend/tests/test_foreman_runner.py backend/tests/test_tool_use_parser.py backend/tests/test_tools_dedup.py -v` → 53 passed (non-DB tests)

Closes #608

### DIFF
--- a/backend/foreman/tools.py
+++ b/backend/foreman/tools.py
@@ -1725,9 +1725,11 @@ async def _exec_one_tool(guild_id: str, tu, user_id: str | None = None) -> dict:
                             head_ref = (pr_data.get("head") or {}).get("ref", "")
 
                             try:
-                                import anthropic as _anthropic
+                                # Use the shared factory so alternate providers (e.g. Bedrock)
+                                # are handled correctly — avoids hardcoding AsyncAnthropic().
+                                from foreman_core.llm import make_anthropic_client as _make_client
 
-                                _ai = _anthropic.AsyncAnthropic()
+                                _ai = _make_client()
                                 review_prompt = (
                                     "You are a thorough code reviewer. Review the following "
                                     "GitHub pull request and provide structured feedback.\n\n"
--- a/backend/foreman/tools.py
+++ b/backend/foreman/tools.py
@@ -22,7 +22,7 @@ from typing import Any
 
 from database import get_db
 from events import broadcast, broadcast_msg, emit_terminal_line
-from foreman_core.llm import get_foreman_model
+from foreman_core.llm import get_foreman_model, make_anthropic_client
 from foreman_core.message_utils import _json_default
 from foreman_core.tools_schema import (
     FOREMAN_TOOLS,  # noqa: F401 — re-exported for test compatibility
@@ -1725,11 +1725,7 @@ async def _exec_one_tool(guild_id: str, tu, user_id: str | None = None) -> dict:
                             head_ref = (pr_data.get("head") or {}).get("ref", "")
 
                             try:
-                                # Use the shared factory so alternate providers (e.g. Bedrock)
-                                # are handled correctly — avoids hardcoding AsyncAnthropic().
-                                from foreman_core.llm import make_anthropic_client as _make_client
-
-                                _ai = _make_client()
+                                _ai = make_anthropic_client()
                                 review_prompt = (
                                     "You are a thorough code reviewer. Review the following "
                                     "GitHub pull request and provide structured feedback.\n\n"
--- a/backend/tests/test_foreman_llm.py
+++ b/backend/tests/test_foreman_llm.py
@@ -1,0 +1,251 @@
+"""Tests for foreman_core/llm.py — provider selection and client factory.
+
+Verifies that make_anthropic_client() and get_foreman_model() correctly handle
+both the default Anthropic provider and alternate providers (e.g. AWS Bedrock).
+"""
+
+from __future__ import annotations
+
+import os
+import sys
+from types import SimpleNamespace
+from unittest.mock import MagicMock, patch
+
+import pytest
+
+sys.path.insert(0, os.path.join(os.path.dirname(__file__), ".."))
+
+
+# ---------------------------------------------------------------------------
+# Helpers
+# ---------------------------------------------------------------------------
+
+
+def _make_mock_anthropic(anthropic_cls=None, bedrock_cls=None):
+    """Build a minimal mock of the anthropic module."""
+    mock_mod = MagicMock()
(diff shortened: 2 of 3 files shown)


backend/tests/test_foreman_llm.py
```diff
@@ -0,0 +1,251 @@
...
+            import foreman_core.llm as llm_mod
+
+            importlib.reload(llm_mod)
+            # HAS_ANTHROPIC will be False after reload with anthropic=None in sys.modules
+            if not llm_mod.HAS_ANTHROPIC:
+                with pytest.raises(ImportError):
+                    llm_mod.make_anthropic_client()
+
+
+# ---------------------------------------------------------------------------
```
jmelloy, line 186: This guard makes the test a no-op in any environment where the `anthropic` package is installed (the common case in CI). The test should instead explicitly force `HAS_ANTHROPIC = False` and `_anthropic_mod = None` on the reloaded module before calling `make_anthropic_client()`, rather than relying on the `sys.modules` patch alone to change `HAS_ANTHROPIC` at reload time. As written, the `pytest.raises` block is never reached when `anthropic` is installed.

backend/tests/test_foreman_llm.py
```diff
@@ -0,0 +1,251 @@
...
+            importlib.reload(llm_mod)
+            llm_mod.make_anthropic_client()
+            mock_mod.AsyncAnthropic.assert_called_once()
+            mock_mod.AsyncAnthropicBedrock.assert_not_called()
+
+    def test_bedrock_env_creates_bedrock_client(self, monkeypatch):
+        """FOREMAN_PROVIDER=bedrock must produce an AsyncAnthropicBedrock client."""
+        monkeypatch.setenv("FOREMAN_PROVIDER", "bedrock")
+        mock_mod = _make_mock_anthropic()
+        with patch.dict("sys.modules", {"anthropic": mock_mod}):
```
jmelloy: Calling `importlib.reload(llm_mod)` inside a `patch.dict('sys.modules', {'anthropic': mock_mod})` context will bind `_anthropic_mod` and `HAS_ANTHROPIC` correctly for that reload, but the reload also re-executes all module-level side effects. If `foreman_core.llm` imports other modules at the top level that themselves import `anthropic`, those won't be intercepted by this patch and may raise or produce unexpected state. Consider patching at the attribute level (`patch.object(llm_mod, '_anthropic_mod', mock_mod)`) as done in `TestMakeAnthropicClientDynamicEnv`, which is the cleaner approach already demonstrated in the same file.

backend/foreman/tools.py
```diff
@@ -1725,9 +1725,11 @@ async def _exec_one_tool(guild_id: str, tu, user_id: str | None = None) -> dict:
                             head_ref = (pr_data.get("head") or {}).get("ref", "")
 
                             try:
-                                import anthropic as _anthropic
+                                # Use the shared factory so alternate providers (e.g. Bedrock)
```
jmelloy: The import of `make_anthropic_client` is placed inside a deeply nested `try` block, meaning import failures (e.g. `foreman_core` not on the path) will be silently swallowed by whatever `except` clause wraps this code. Consider hoisting the import to the top of the file alongside other `foreman_core` imports so import errors surface clearly at startup rather than at runtime during a PR review call.

jmelloy (commented):
- The core fixes are correct and well-motivated: replacing the hardcoded `AsyncAnthropic()` with `make_anthropic_client()` and switching from a module-level constant to a dynamic `os.environ.get()` call both directly address the reported bugs.
- The test suite is comprehensive and covers the main scenarios (default provider, Bedrock, explicit overrides, region forwarding, api_key handling), which is commendable for a bug-fix PR.
- A subtle but important issue exists in the test class `TestMakeAnthropicClient`: each test calls `importlib.reload(llm_mod)` but the reloaded module object is a *new* reference, while the local variable still points to it correctly — however, after the `with` block exits, the `sys.modules` entry is restored, leaving the module in a potentially inconsistent state that could affect subsequent tests.
- The `test_missing_anthropic_package_raises` test has a conditional `if not llm_mod.HAS_ANTHROPIC:` guard that silently skips the assertion when the package IS installed, making it a no-op in most CI environments where `anthropic` is present.
- The dynamic `os.environ.get()` fix is correct, but the module-level `FOREMAN_PROVIDER` constant (if it still exists elsewhere in the file) may create confusion; it would be worth removing or deprecating it to avoid future drift.

PR #622, opened 2026-06-09
Scope worker join/notification events to member guilds

### LINKED ISSUE #621: Worker join/notification events should only be sent to guilds the worker belongs to
## Problem

When a worker comes online or goes offline, the join/notification events are currently broadcast too broadly — they should only be delivered to the guilds that the worker actually belongs to.

## Expected Behaviour

- `worker-online` and `worker-offline` (and any related worker lifecycle notifications) are scoped to the worker's guild(s) only.
- Guilds that the worker does not belong to should not receive these events.

## Acceptance Criteria

- Find all places where worker join/leave/notification messages are dispatched (e.g. WebSocket broadcasts, SSE pushes, DB notification rows, etc.).
- Add a guild membership check so events are only sent to matching guilds.
- Add or update tests to cover the scoping behaviour.

jmelloy:
Fixes #621

Adds guild membership checks in ws_handlers.py so worker-online/offline events are only broadcast to guilds the worker belongs to. Includes new tests covering the scoping behaviour.

### DIFF
--- a/backend/routes/websocket.py
+++ b/backend/routes/websocket.py
@@ -227,10 +227,13 @@ async def websocket_endpoint(websocket: WebSocket, guild_id: str):
                 # Batch all stale-worker notifications into a single foreman
                 # trigger so a simultaneous mass-disconnect doesn't stampede
                 # the foreman with N concurrent embedded-foreman tasks.
-                if stale_worker_ids:
+                # Skip workers that already sent reason=shutdown via a graceful
+                # worker-disconnect message to avoid double-firing.
+                abrupt_worker_ids = stale_worker_ids - ctx.gracefully_disconnected_workers
+                if abrupt_worker_ids:
                     offline_lines = "\n".join(
                         f"[worker-offline] worker_id={wid} reason=disconnect"
-                        for wid in sorted(stale_worker_ids)
+                        for wid in sorted(abrupt_worker_ids)
                     )
                     await ws_handlers._trigger_foreman(
                         guild_id,
--- a/backend/tests/test_worker_foreman_notify.py
+++ b/backend/tests/test_worker_foreman_notify.py
@@ -263,3 +263,141 @@ def test_task_complete_max_turns_notifies_foreman(client):
     assert "I was working on the migration" in msg, (
         f"Expected lastText snippet in message, got: {msg}"
     )
+
+
+# ---------------------------------------------------------------------------
+# Guild-scoping tests (issue #621)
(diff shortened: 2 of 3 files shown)


jmelloy (commented):
## Review: Scope worker join/notification events to member guilds

**Verdict: LGTM** — the fix is correct. Ruff passes clean (`ruff check .` → All checks passed).

### What the PR does

Adds guild membership validation in three WebSocket handlers to prevent a misconfigured or misbehaving worker from injecting `worker-online`/`worker-offline` events into a guild it does not belong to:

- **`handle_join`** — early return if worker is not a member of the connected guild (primary fix; prevents agent rows and all downstream side-effects)
- **`handle_worker_register`** — early return before repos/tools update and foreman trigger
- **`handle_worker_disconnect`** — gates only the `_trigger_foreman` call; the rest of the function is already safe due to `col(Worker.guild_id) == ctx.guild_pk` filters in existing `WHERE` clauses

### Correctness

The membership check query is consistent across all three handlers:
```python
select(col(Worker.id)).where(
    col(Worker.id) == worker_id,
    col(Worker.guild_id) == ctx.guild_pk,
)
```
Using `ctx.guild_pk` (integer PK) rather than the string `ctx.guild_id` is correct and efficient.

### Minor observations (non-blocking)

1. **Asymmetric pattern in `handle_worker_disconnect`**: `handle_join` and `handle_worker_register` early-return on mismatch; `handle_worker_disconnect` runs most of its body and only guards `_trigger_foreman`. This is safe — `ctx.joined_agents` is empty (join was rejected) and the `update(Worker)` has a `guild_id` filter that makes it a no-op. But the inconsistency could surprise a reader; a short comment would help.

2. **Missing positive regression test for `worker-offline`**: There's `test_worker_online_fires_for_correct_guild` (worker connects to own guild → `worker-online` fires) but no symmetric `test_worker_offline_fires_for_correct_guild`. Low risk since the happy path is covered by existing tests, but would improve symmetry.

3. **`reset_foreman_poll` fires unconditionally**: Line 679 runs regardless of guild membership. Pre-existing behavior and low risk, but worth noting.

### Test coverage

Three new integration tests cover the critical paths:
- Foreign worker cannot trigger `worker-online` in a guild it doesn't belong to ✓
- Foreign worker cannot trigger `worker-offline` in a guild it doesn't belong to ✓  
- Regression: legitimate worker connecting to its own guild still gets `worker-online` ✓

PR #633, opened 2026-06-10
Drop finished_at, consolidate into deleted_at

### LINKED ISSUE #632: Plan: reconcile tasks.finalized_at vs tasks.deleted_at — drop one column
The `tasks` table currently has two columns that feel very similar in purpose:

- **`finalized_at`** — set when `finalize_task` is called; signals that the task reached a terminal state (done or failed)
- **`deleted_at`** — used as a soft-delete / expiry timestamp; controls when the task is hidden or purged

These two concepts overlap enough that carrying both is confusing. We need a plan for collapsing them into a single column.

## Questions to resolve

1. **Are the semantics actually distinct?** Is there a real difference between "finalized" and "soft-deleted", or do they always get set together?
2. **Which column should survive?** Options:
   - Keep `deleted_at` and drop `finalized_at` (soft-delete is the canonical pattern)
   - Keep `finalized_at` and drop `deleted_at` (better reflects domain language)
   - Merge into a single `expires_at` or `archived_at` with clear semantics
3. **What code paths write/read each column?** DB migrations, ORM model, API routes, WS types, frontend store, foreman runtime (`finalize_task`), tests.
4. **Migration strategy** — how to consolidate without data loss on live instances.

## Desired outcome

A short written plan (comment on this issue) covering: chosen column name, rationale, and the ordered list of changes needed across the stack (migration, model, API, frontend, tests).

jmelloy:
## Summary

Consolidates `tasks.finished_at` into `deleted_at`, dropping the redundant column.

## Changes
- `backend/alembic/versions/20260610_000000_drop_tasks_finished_at.py` — migration dropping `finished_at`
- `backend/ws_handlers.py` — removed `finishedAt` handling block
- `backend/routes/debug.py` — uses `deleted_at` instead of `finished_at`
- `backend/routes/foreman.py` — removed `finished_at` from `TaskPatch`, query, WS key map
- `frontend/src/types.ts` — removed `finishedAt?: string` from `TaskUpdateWS`
- `frontend/src/stores/tasks.ts` — `cancelTask` sets `deleted_at` instead of `finished_at`; WS handler updated
- `frontend/src/stores/__tests__/tasks.spec.ts` — tests updated
- `backend/tests/test_foreman.py` — `TestSummarizeTask` updated to use `deleted_at`

Closes #632

### DIFF
--- a/backend/alembic/versions/20260610_000000_drop_tasks_finished_at.py
+++ b/backend/alembic/versions/20260610_000000_drop_tasks_finished_at.py
@@ -1,0 +1,33 @@
+"""Drop tasks.finished_at column.
+
+Revision ID: 20260610_000000_drop_tasks_finished_at
+Revises: 20260608_000000_rename_guild_slug
+Create Date: 2026-06-10
+
+Consolidates the finished_at and deleted_at columns into a single deleted_at column.
+deleted_at already serves as the soft-delete expiry timestamp (set to now + 3 days on
+completion); finished_at was a redundant "when did the task complete?" marker.
+
+The completion time can be approximated as deleted_at - DEFAULT_TTL (3 days).
+"""
+
+from __future__ import annotations
+
+from collections.abc import Sequence
+
+from alembic import op
+
+revision: str = "20260610_000000_drop_tasks_finished_at"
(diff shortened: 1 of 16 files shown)


jmelloy (commented):

PR #639, opened 2026-06-11
Fix foreman debug to show next-API-call context only

### LINKED ISSUE #638: Foreman debug view shows all 25k messages instead of next API call payload
## Problem

The foreman debug panel currently displays the full message history (~25,000 messages), which is unusable. It should only show what would actually be sent on the **next API call** — i.e. the trimmed/windowed context that the foreman would submit to the LLM.

## Expected behaviour

The debug view reflects the real context window: system prompt + the most recent messages that fit within the token/message limit, exactly as they'd be serialised for the next API request.

## Suggested fix

- Pull the same message-trimming / context-windowing logic used before each API call and render that subset in the debug view
- Ideally show a summary line: e.g. `Showing 42 of 25,312 messages (last N tokens)`

jmelloy:
## Problem

The foreman debug panel was showing the entire accumulated message history (~25,000 messages), making it useless for understanding what the foreman would actually send to the LLM.

## Solution

Modified `get_foreman_history()` in `backend/foreman/runner.py` to apply the same trimming pipeline used before every real API call:

1. **`_HUMAN_TURN_WINDOW` sliding-window cutoff** — walks backwards to find the 5th-from-last non-tool-response user turn, then includes only turns from that cutoff onwards (same logic as `_load_history()`)
2. **`prune_history()`** — caps at `MAX_HISTORY_MESSAGES` (20) and ensures the list starts with a user turn
3. **`strip_orphaned_tool_results()`** — removes orphaned tool_use/tool_result blocks

The `total` count of all stored non-system turns is also returned so the debug panel can display a summary line.

### Changes

- **`backend/foreman/runner.py`** — `get_foreman_history()` now applies windowing pipeline and returns `total`
- **`backend/routes/foreman.py`** — `GET /guilds/{guild_id}/foreman/context` passes `total` through in the response
- **`frontend/src/components/DebugSidebar.vue`** — header shows "Showing 42 of 25,312 messages" when windowed, otherwise "N msgs in context"

## Testing

- `ruff check . --fix && ruff format .` — all clean
- `npm run type-check` (from `frontend/`) — no errors

Closes #638

### DIFF
--- a/backend/foreman/runner.py
+++ b/backend/foreman/runner.py
@@ -991,16 +991,19 @@ async def clear_foreman_history(guild_id: str, user_id: str) -> int:
 async def get_foreman_history(guild_id: str, user_id: str) -> dict:
     """Return stored turns structured for the debug view.
 
+    Applies the same windowing pipeline used before each real API call so the
+    debug pane shows exactly the messages that would be submitted next:
+      1. ``_HUMAN_TURN_WINDOW`` sliding-window cutoff (same as ``_load_history``)
+      2. ``prune_history`` cap (``MAX_HISTORY_MESSAGES``)
+      3. ``strip_orphaned_tool_results`` cleanup
+
     Returns::
 
         {
             "system": <str | None>,   # most-recent audit system prompt text
-            "messages": [...],        # non-system turns only, in DB order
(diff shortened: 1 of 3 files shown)


jmelloy (commented):

PR #640, opened 2026-06-11
Add optional S3 log sync to worker

### LINKED ISSUE #637: Optional S3 sync for worker task logs
## Summary

Add an optional S3 sync feature to the worker so that the raw session logs logs are periodically and/or post-task copied to a configured S3 bucket.

## Requirements

- **Opt-in**: controlled by environment variables (e.g. `LOG_S3_BUCKET`, `LOG_S3_PREFIX`, `LOG_S3_SYNC_INTERVAL_SECONDS`)
- **Periodic sync**: sync logs on a configurable interval (e.g. every 60 seconds)
- **Path convention**: logs should be keyed by something like `{prefix}/{guild_id}/{worker_id}/{session}.jsonl`
- **No hard dependency**: if S3 env vars are not set, the feature is silently disabled; no import errors if boto3 is not installed
- **Credentials**: use standard AWS credential chain (env vars, ~/.aws, instance metadata)

## Out of scope

- Log streaming (line-by-line push); batch sync is sufficient
- Any UI / retrieval tooling

jmelloy:
## Summary

- Adds `worker/pioneer_worker/s3_log_sync.py` — a self-contained `S3LogSync` class that periodically uploads a task log file to S3 in a background daemon thread, with a guaranteed final upload after task completion.
- Each task's terminal output is captured to `{work_dir}/agent.log` by wrapping the existing `emit` function.
- The S3 key follows the convention `{prefix}/{guild_id}/{worker_id}/{task_id}/agent.log`.
- Feature is entirely opt-in via three env vars (`LOG_S3_BUCKET`, `LOG_S3_PREFIX`, `LOG_S3_SYNC_INTERVAL_SECONDS`). Zero runtime cost when `LOG_S3_BUCKET` is unset.
- `boto3` gracefully degrades: if it's missing and `LOG_S3_BUCKET` is set, a warning is logged and the feature disables itself rather than crashing.
- `boto3` added as `[s3]` optional extra in `cli/pyproject.toml` (not a hard dependency).
- 14 unit tests covering: disabled paths, boto3 missing warning, env var defaults, S3 key convention, upload-on-finish, periodic sync, daemon thread, and upload failure swallowing.

## Test plan

- [ ] `cd worker && python -m pytest tests/test_s3_log_sync.py -v` — all 14 pass
- [ ] `cd worker && python -m pytest` — no regressions (pre-existing `test_no_tools_flag_gives_none` failure is unrelated to this PR)
- [ ] Manual: set `LOG_S3_BUCKET=my-bucket` and run a task; verify `agent.log` appears at the expected S3 key

Closes #637

### DIFF
--- a/cli/pyproject.toml
+++ b/cli/pyproject.toml
@@ -37,6 +37,9 @@ test = [
     "anyio[trio]>=4.0",
     "ruff",
 ]
+s3 = [
+    "boto3>=1.26",
+]
 
 [project.scripts]
 pioneer = "pioneer_cli.main:main"
--- a/worker/pioneer_worker/claude_runner.py
+++ b/worker/pioneer_worker/claude_runner.py
@@ -288,6 +288,7 @@ async def run_claude_auto(
     on_usage: UsageFn | None = None,
     claude_path: str = "claude",
     resume_session_id: str | None = None,
+    raw_log_path: str | None = None,
 ) -> tuple[bool, str, str, str | None]:
     """Run claude on *description* in *cwd*. Returns (success, stop_reason, last_assistant_text, session_id).
 
@@ -318,6 +319,12 @@ async def run_claude_auto(
     stop_reason = "no_events"
     event_count = 0
     session_id = None
+    _raw_fh = None
+    if raw_log_path:
+        try:
+            _raw_fh = open(raw_log_path, "ab")  # noqa: WPS515
+        except OSError as exc:
+            logger.warning("Could not open raw log %s: %s", raw_log_path, exc)
     try:
         proc = await asyncio.create_subprocess_exec(
             *cmd,
@@ -337,10 +344,22 @@ async def run_claude_auto(
                 line = raw.decode(errors="replace").strip()
                 if line:
                     await emit(f"[stderr] {line}")
+                if _raw_fh is not None:
+                    try:
+                        _raw_fh.write(b"[stderr] " + raw if raw.endswith(b"\n") else b"[stderr] " + raw + b"\n")
+                        _raw_fh.flush()
+                    except OSError:
+                        pass
 
         stderr_task = asyncio.create_task(_drain_stderr())
 
         async for raw in proc.stdout:  # type: ignore[union-attr]
+            if _raw_fh is not None:
+                try:
+                    _raw_fh.write(raw)
+                    _raw_fh.flush()
+                except OSError:
+                    pass
             line_str = raw.decode(errors="replace").strip()
             if not line_str:
                 continue
@@ -423,3 +442,9 @@ async def run_claude_auto(
         logger.exception("claude subprocess crashed: %s", exc)
         await emit(f"[claude] ✗ {exc}")
         return False, "error_during_execution", last_text, session_id
+    finally:
+        if _raw_fh is not None:
+            try:
+                _raw_fh.close()
+            except OSError:
+                pass
--- a/worker/pioneer_worker/claude_runner.py
+++ b/worker/pioneer_worker/claude_runner.py
@@ -346,7 +346,11 @@ async def run_claude_auto(
                     await emit(f"[stderr] {line}")
                 if _raw_fh is not None:
                     try:
-                        _raw_fh.write(b"[stderr] " + raw if raw.endswith(b"\n") else b"[stderr] " + raw + b"\n")
+                        _raw_fh.write(
+                            b"[stderr] " + raw
+                            if raw.endswith(b"\n")
+                            else b"[stderr] " + raw + b"\n"
+                        )
                         _raw_fh.flush()
                     except OSError:
                         pass
--- a/worker/pioneer_worker/codex_runner.py
+++ b/worker/pioneer_worker/codex_runner.py
@@ -110,7 +110,11 @@ async def run_codex_auto(
                     await emit(f"[stderr] {line}")
                 if _raw_fh is not None:
                     try:
-                        _raw_fh.write(b"[stderr] " + raw if raw.endswith(b"\n") else b"[stderr] " + raw + b"\n")
+                        _raw_fh.write(
+                            b"[stderr] " + raw
+                            if raw.endswith(b"\n")
+                            else b"[stderr] " + raw + b"\n"
+                        )
                         _raw_fh.flush()
                     except OSError:
                         pass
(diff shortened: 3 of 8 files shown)


jmelloy (commented):

jmelloy (commented):
## Review: Add optional S3 log sync to worker

All 14 new tests pass and there are no regressions (the sole failure is the pre-existing `test_no_tools_flag_gives_none` called out in the PR description). The implementation is clean and well-structured. A few minor observations below — none are blocking.

---

### Nits / minor observations

**1. `_s3_syncer.start()` is called outside the `try/finally` block**

`start()` fires at worker.py ~line 1946, but the `try` block that owns the `finally` (which calls `finish()`) doesn't begin until line 1988. The intervening code is only simple assignments/closures that can't realistically raise, so the risk is negligible in practice. However, if anyone adds code in that gap that throws, the background daemon thread would be leaked until process exit. Moving `start()` inside the `try` (or pulling the `try` up to enclose `start()`) would close this gap.

**2. `join()` timeout + post-join concurrent upload**

In `finish()`, `self._thread.join(timeout=15.0)` is followed unconditionally by `self._upload()`. If the join times out (e.g., a very slow network upload), the background thread may still be inside `upload_file` when `finish()` also calls `upload_file`. Since S3 uploads are idempotent and `_upload` swallows all exceptions, this is benign in practice, but worth noting if we ever add richer error reporting.

**3. Redundant `delenv`/`setenv` in `test_from_env_returns_instance_when_configured`**

```python
monkeypatch.setenv("LOG_S3_BUCKET", "my-bucket")
monkeypatch.setenv("LOG_S3_PREFIX", "my-prefix")
monkeypatch.setenv("LOG_S3_SYNC_INTERVAL_SECONDS", "30")
monkeypatch.delenv("LOG_S3_BUCKET", raising=False)  # reset  ← dead code
monkeypatch.setenv("LOG_S3_BUCKET", "my-bucket")
```

The `delenv` + re-`setenv` at the bottom is a copy-paste artifact — `LOG_S3_BUCKET` was already set two lines earlier and the `delenv` has no observable effect. Can be removed.

**4. Warning message points to `pip install boto3` but the package exposes `[s3]` extra**

The graceful-degradation warning in `from_env()` says:
```
"Install it with: pip install boto3"
```
Since `cli/pyproject.toml` now ships `pioneer-square[s3]`, mentioning both forms (`pip install boto3` or `pip install 'pioneer-square[s3]'`) would help users discover the extras-based path.

---

### What's solid

- Graceful boto3 import: zero cost when bucket env var is unset, warning (not crash) when boto3 is missing.
- `_sync_loop` uses `Event.wait(timeout=…)` — correct interruptible-sleep pattern.
- Daemon thread: won't prevent clean process exit.
- `finish()` always runs in the `finally` block for all non-early-return paths; the early-return path (no worktrees) correctly closes `_log_fh` manually and is reached before `_s3_syncer` is ever initialised, so no cleanup is needed there.
- Upload failures are swallowed with a warning — correct for a best-effort observability feature.
- 14 tests cover all the key paths: disabled, missing boto3, env var defaults, S3 key convention, upload-on-finish, periodic sync, daemon flag, and failure swallowing.

Great feature addition — the opt-in design is exactly right for an observability concern like this.

jmelloy:
## S3 Log Sync: What's Being Synced and Why It's the Right Choice

**TL;DR:** The current implementation syncs the `emit()` data via `agent.log`, which is the correct choice for operational monitoring. The `.claude/sessions` JSONL files are not written by the worker and would not be appropriate here.

---

### 1. What the current implementation syncs

The implementation syncs `agent.log` — a plaintext file written by the task-scoped `emit()` function in `worker.py`. Every call to `emit()` writes a line to both the WebSocket (for real-time frontend display) and this local file:

```python
# worker.py:1845–1851
async def emit(line: str, detail: dict | None = None, level: str = LEVEL_INFO) -> None:
    await _base_emit(line, detail, level)   # → WebSocket
    try:
        _log_fh.write(line + "\n")          # → agent.log
        _log_fh.flush()
    except Exception:
        pass
```

`S3LogSync.start()` receives this `agent.log` path and uploads it periodically to `{prefix}/{guild_id}/{worker_id}/{task_id}/agent.log`.

The log contains: task description, branch name, repo prep status, Claude/tool activity (file reads/edits/searches, thinking lines), errors, redirects, branch push confirmation, and PR URL — essentially the same stream the frontend shows in real time.

---

### 2. Trade-offs

| | `emit()` → `agent.log` | `.claude/sessions` JSONL |
|---|---|---|
| **Content** | Operational events: task lifecycle, tool activity, errors, outcomes | Full conversation transcript: every message, tool call, result, raw model output |
| **Size** | Small — concise summary of what happened | Large — full verbatim transcript, often 10–100× bigger |
| **Sensitivity** | Low — same data already shown in the frontend UI | High — includes raw model reasoning, potentially sensitive repo content seen by the model |
| **Location** | Written by the worker at a known path (`{work_dir}/agent.log`) | Written by the Claude CLI into the worktree (`.claude/projects/.../sessions/*.jsonl`) — path varies per task |
| **Availability** | Always present when a task runs | Only present if Claude CLI has session persistence enabled; requires discovery logic |
| **Fit for ops monitoring** | ✅ Purpose-built — structured around task events | ❌ Raw transcript — requires parsing to extract operational signal |

**Important note:** The worker does not reference `.claude/sessions` JSONL files anywhere in its codebase. Session IDs are tracked in memory and reported to the backend; the JSONL files are a Claude CLI artifact written into the worktree directory, not something the worker actively manages.

---

### 3. Recommendation: keep the current approach

Syncing `agent.log` (the `emit()` stream) is exactly right for the stated goal of *operational log sync for worker task logs*:

- It's already written — no additional integration needed.
- It's the same signal operators see in the UI, making it easy to correlate S3 logs with what users report.
- It's compact and low-sensitivity, appropriate for a shared S3 bucket.
- The `.claude/sessions` JSONL alternative would require path discovery, is much larger, and exposes raw model reasoning that may be sensitive.

If full conversation transcripts ever become a requirement (e.g., for audit trails or fine-tuning data), that would be a separate opt-in feature with its own S3 path and access controls — not something to conflate with operational log sync.

**No implementation changes are needed.** The current `s3_log_sync.py` correctly syncs `agent.log` and is well-suited for its purpose.

jmelloy (commented):

PR #642, opened 2026-06-11
feat(foreman): queue socket messages while a turn is being processed

### LINKED ISSUE #641: Queue incoming socket messages while foreman is processing a query
## Problem

Messages that arrive over the socket while the foreman is actively running a query (e.g. calling tools, awaiting LLM response) are currently dropped or handled immediately in a way that can interleave with the in-progress turn. They should be buffered and processed after the current turn completes.

## Desired behaviour

1. When the foreman begins processing a turn (from the first token of the LLM call until the final tool call returns and the response is sent), incoming socket messages are placed in a FIFO queue rather than dispatched immediately.
2. Once the current turn finishes, the foreman drains the queue in order, processing each buffered message as a new turn.
3. The queue should be bounded (e.g. max 100 messages) to avoid unbounded memory growth; excess messages should be logged and dropped with a warning.
4. Periodic-check / automated poll messages that arrive while a turn is running can be silently discarded (they'll fire again on the next interval) — or deprioritised to the back of the queue.

## Implementation notes

- Look at where the foreman's WebSocket/socket receive loop lives and add a queue + busy flag there.
- A simple `asyncio.Queue` (or `collections.deque` with a lock for threaded code) should suffice.
- Add a unit test that verifies messages received during a "busy" period are processed after the turn completes, in order.

jmelloy:
## Summary

- **Busy flag + queue**: Renames `_in_flight` → `_processing` and adds an `asyncio.Queue(maxsize=100)` to `Foreman`. When `_processing` is `True`, incoming `foreman-trigger` messages are buffered in the queue instead of being dispatched (or dropped).
- **Drain loop**: After each turn completes, `_run_foreman` drains the queue in FIFO order while keeping `_processing=True` throughout, so any triggers arriving during the drain are queued rather than spawning a concurrent run.
- **Bounded queue**: The queue is capped at 100 messages. An overflow logs `WARNING` and drops the message.
- **Periodic-check discard**: Triggers whose `humanMessage` contains `[periodic-check]` are silently discarded when busy — they re-fire on the next poll interval anyway.
- **Tests**: 3 new tests (`test_ws_queued_messages_processed_in_order`, `test_ws_periodic_check_discarded_when_busy`, `test_ws_queue_full_drops_message`); 2 existing tests updated to reflect the new behaviour (`_dropped_` → `_queued_`, `_in_flight` → `_processing`). All 14 tests pass.

## Test plan

```
cd foreman
python -m pytest tests/test_ws_integration.py -v   # 14 passed
ruff check . && ruff format .                       # no issues
```

Closes #641

### DIFF
--- a/foreman/pioneer_foreman/foreman.py
+++ b/foreman/pioneer_foreman/foreman.py
@@ -18,6 +18,8 @@ if TYPE_CHECKING:
 
 logger = logging.getLogger(__name__)
 
+_QUEUE_MAX = 100
+
 
 class Foreman:
     """Manages the standalone foreman lifecycle for one guild."""
@@ -25,10 +27,10 @@ class Foreman:
     def __init__(self, config: Config):
         self._config = config
         self._http: ForemanHTTPClient | None = None
-        # In-flight flag: True while a foreman run coroutine is executing.
-        # Checked before spawning a new trigger run or poll run so we never
-        # have two concurrent runs for the same guild.
-        self._in_flight: bool = False
+        # True while a foreman run (including its queue drain) is executing.
+        self._processing: bool = False
+        # Triggers buffered while _processing is True; drained FIFO after each turn.
+        self._message_queue: asyncio.Queue[dict] = asyncio.Queue(maxsize=_QUEUE_MAX)
         # Monotonic timestamp of the last run that made at least one tool call.
         # Used to decide whether to reset the poll backoff.
         self._last_action_at: float = 0.0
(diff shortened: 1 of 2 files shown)


jmelloy (commented):
## Code Review: feat(foreman): queue socket messages while a turn is being processed

**All 14 tests pass, ruff is clean. Implementation is correct — would approve.**

---

### What the PR does

Replaces the old "drop on busy" strategy with a bounded FIFO queue. When `_processing` is True, incoming `foreman-trigger` messages are either discarded (periodic-check) or buffered (everything else, up to 100 slots). After each turn `_run_foreman` drains the queue while holding `_processing=True`, so triggers arriving during the drain are themselves queued rather than spawning a concurrent run.

---

### Correctness

- **Busy flag scope**: `_processing` is set before the first turn and cleared in the outer `finally`, covering the full initial-turn + drain window. Correct.
- **TOCTOU in drain loop**: `empty()` / `get_nowait()` are safe in asyncio (no yield point between them in single-threaded event loop). The defensive `asyncio.QueueEmpty` catch is a good belt-and-suspenders measure.
- **Periodic-check filter**: Applied at enqueue time in `_handle_trigger`, so the drain loop never sees periodic-check messages. Correct.
- **Poll-loop guard**: The `if self._processing: return` guard at the top of `_run_foreman` is still needed — the poll loop calls `asyncio.create_task(_run_foreman(...))` directly, bypassing `_handle_trigger` and the queue.
- **Queue bounds**: `asyncio.QueueFull` is caught, logged at WARNING, and silently dropped. Correct.
- **Test coverage**: FIFO order, periodic-check discard, queue-full drop, flag reset, and the updated in-flight test are all covered. Solid suite.

---

### Minor nits (non-blocking)

**1. Orphan `import logging` inside `test_ws_queue_full_drops_message` (line 457)**

`logging` is already imported at the top of the file and is not used inside this function body. Leftover from an earlier draft — safe to delete.

**2. Stale log message in `_run_foreman`**

The early-return guard logs `"dropped — run already in-flight"`. This path is now only reachable from the poll-loop path (trigger messages go through `_handle_trigger` first, which queues them). The message is accurate for that path but could be clearer, e.g. `"poll trigger dropped — run already processing"`.

**3. Queue not cleared on reconnection**

`_message_queue` is allocated once in `__init__` and persists across `run()` retry cycles. If the connection drops mid-drain, any un-drained messages carry over to the next connection. This is likely benign, but worth a comment or an explicit flush at the top of `_run_connection` if stale cross-connection messages are a concern.

PR #653, opened 2026-06-16
Optional S3 sync for worker task logs (#637)

### LINKED ISSUE #637: Optional S3 sync for worker task logs

jmelloy:
## Summary

- Adds opt-in S3 upload of raw `stream-json` session logs for each worker task
- Enabled by setting `LOG_S3_BUCKET`; silently disabled if unset or if `boto3` is not installed
- Three new env vars: `LOG_S3_BUCKET`, `LOG_S3_PREFIX`, `LOG_S3_SYNC_INTERVAL_SECONDS` (default 60s)
- S3 key path: `{prefix}/{guild_id}/{worker_id}/{task_id}.jsonl`
- AWS credentials resolved via the standard chain — no hard-coded secrets

## Changes

**`worker/pioneer_worker/s3_log_sync.py`** (new)  
Self-contained helper module. `create_syncer()` reads env vars and returns an `S3LogSyncer` (or `None` if disabled). `S3LogSyncer` runs a daemon thread that periodically calls `sync_all()` and exposes `sync_file()` for immediate post-task uploads. `ImportError` on `boto3` is caught at module load; all sync methods are no-ops when unavailable.

**`worker/pioneer_worker/claude_runner.py`**  
`run_claude_auto` gains an optional `log_file_path` parameter. When set, each raw stdout line from the `claude` subprocess is appended to the file (flushed per line; append mode so redirect-loop reruns accumulate in one file).

**`worker/pioneer_worker/worker.py`**  
- `Worker.__init__` calls `create_syncer()` and stores `_s3_syncer`
- `run()` starts the syncer after registration and stops it in the `finally` block
- `_execute_task()` computes `session_log_path = {work_dir}/{guild_id}/{worker_id}/{task_id}.jsonl`, passes it to `run_claude_auto`, registers the file with the syncer on task start, does an immediate `sync_file()` when the run finishes, and unregisters in `finally`

**`worker/README.md`**  
New "Optional S3 log sync" section documents the env vars and credential chain.

**`worker/tests/test_s3_log_sync.py`** (new)  
12 unit tests covering: upload key construction, missing-prefix handling, missing-file no-op, upload error tolerance, `sync_all` across multiple files, unregister behaviour, boto3-absent no-op, thread start/stop lifecycle, and all `create_syncer` branches.

## Test plan

- [ ] `cd worker && python -m pytest tests/test_s3_log_sync.py -v` — all 12 pass
- [ ] `ruff check . && ruff format .` — clean
- [ ] Manual smoke: set `LOG_S3_BUCKET=<bucket>` and run a task; verify `.jsonl` appears at the expected S3 key

Closes #637

### DIFF
--- a/worker/README.md
+++ b/worker/README.md
@@ -116,6 +116,31 @@ The worker stays running, reconnecting if the backend restarts.
 6. Reports state changes (`working`, `done`, `failed`, branch, PR url) back
    to the backend via `task-update` WebSocket messages.
 
+## Optional S3 log sync
+
+The worker can periodically upload its raw session log files (the
+`stream-json` output captured from each `claude` subprocess) to an S3
+bucket.  This is opt-in and requires `boto3`:
+
+```bash
+pip install boto3   # or include it in your worker environment
+```
+
+Set the following environment variables before starting the worker:
+
+| Variable | Required | Default | Description |
+|---|---|---|---|
+| `LOG_S3_BUCKET` | yes | — | S3 bucket to upload logs to.  Leaving this unset disables the feature entirely. |
+| `LOG_S3_PREFIX` | no | `""` | Key prefix (e.g. `pioneer/logs`).  The final key is `{prefix}/{guild_id}/{worker_id}/{task_id}.jsonl`. |
+| `LOG_S3_SYNC_INTERVAL_SECONDS` | no | `60` | How often the background sync thread uploads in-progress log files.  A final upload also runs immediately when each task completes. |
+
+AWS credentials are resolved through the standard chain — environment
+variables (`AWS_ACCESS_KEY_ID` / `AWS_SECRET_ACCESS_KEY`), `~/.aws/credentials`,
+EC2/ECS instance metadata, etc.  No credentials are hard-coded.
+
+If `boto3` is not installed, a warning is logged at startup and the feature
+is silently disabled — the worker continues normally.
+
 ## Control API (drive a live worker without a frontend)
 
 A real worker only acts on tasks the backend/foreman assigns over WebSocket,
--- a/worker/pioneer_worker/claude_runner.py
+++ b/worker/pioneer_worker/claude_runner.py
@@ -288,6 +288,7 @@ async def run_claude_auto(
     on_usage: UsageFn | None = None,
     claude_path: str = "claude",
     resume_session_id: str | None = None,
+    log_file_path: str | None = None,
 ) -> tuple[bool, str, str, str | None]:
     """Run claude on *description* in *cwd*. Returns (success, stop_reason, last_assistant_text, session_id).
 
@@ -318,6 +319,7 @@ async def run_claude_auto(
     stop_reason = "no_events"
     event_count = 0
     session_id = None
+    _log_fh = open(log_file_path, "a", encoding="utf-8") if log_file_path else None
     try:
         proc = await asyncio.create_subprocess_exec(
             *cmd,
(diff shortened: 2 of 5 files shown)


jmelloy (commented):
### Code review

Found 3 issues:

1. `register()` called before `try/finally`, so `unregister()` is skipped on the no-worktree early-return path — the file stays in `_files` permanently and the background thread retries it forever.

The `register()` call at line 1856 is outside the `try` block that begins at line 1982. The early return at line 1936 (`if not primary_wt: return`) fires between them, bypassing the `finally` clause that holds `unregister()` at line 2220. Fix: move `register()` inside the `try` block (after line 1982), or restructure the early return so it falls through to `finally`.

https://github.com/jmelloy/pioneer-square/blob/8a025a3a7984bdfe51ca729d7637386a87a6c631/worker/pioneer_worker/worker.py#L1854-L1860

https://github.com/jmelloy/pioneer-square/blob/8a025a3a7984bdfe51ca729d7637386a87a6c631/worker/pioneer_worker/worker.py#L1934-L1943

2. `session_log_path` is registered and synced for all task types, but `log_file_path` is only passed to `run_claude_auto` — codex and pi tasks always produce an empty `.jsonl` in S3.

`run_codex_auto` (line 2045) and `run_pi_auto` (line 2063) have no `log_file_path` parameter, so nothing writes to the file. The S3 syncer still uploads it as a zero-byte object on every interval and at task completion.

https://github.com/jmelloy/pioneer-square/blob/8a025a3a7984bdfe51ca729d7637386a87a6c631/worker/pioneer_worker/worker.py#L2043-L2090

3. `open(log_file_path)` in `claude_runner.py` sits outside the `try` block — a `PermissionError` or `IsADirectoryError` there aborts the entire task instead of degrading gracefully.

The write-error path (lines ~355-358) already handles `OSError` by logging a warning and setting `_log_fh = None`, but the `open()` call at line 322 has no equivalent protection. Moving line 322 inside the `try`, or wrapping it in its own `try/except OSError`, would make it consistent.

https://github.com/jmelloy/pioneer-square/blob/8a025a3a7984bdfe51ca729d7637386a87a6c631/worker/pioneer_worker/claude_runner.py#L320-L326

jmelloy (commented):
All three issues addressed in commit 1ae0fe0:

1. **register() before try/finally** — removed the `register()` call from before the worktree-setup block and moved it inside the `try` block (after line 1982). The no-worktree early return now bypasses `register()` entirely, so `_files` is never polluted and the background thread has nothing to retry.

2. **codex/pi register zero-byte files** — gated both `register()` and `unregister()` on `tool == "claude"`. Codex and pi tasks never write to `session_log_path`, so they no longer cause spurious zero-byte uploads on every sync interval.

3. **open() outside try in claude_runner.py** — wrapped `open(log_file_path, "a", ...)` in its own `try/except OSError`. A `PermissionError` or `IsADirectoryError` now logs a warning and sets `_log_fh = None`, consistent with how write errors are already handled, instead of aborting the task.

PR #663, opened 2026-06-19
Fix Anthropic auth token not passed to API call (#660)

### LINKED ISSUE #660: Anthropic auth token not being passed correctly to API call in foreman setup
## Problem

When the Anthropic auth token is set in the foreman setup UI/config, it may not be getting passed correctly to the actual API call. This could cause authentication failures or silent fallback to a different auth method.

## Requirements

- Investigate how the Anthropic auth token is stored during foreman setup
- Trace how it flows through to the actual API call
- Fix any bug where the token is not being correctly forwarded
- Add a test or validation to confirm the token is used

jmelloy:
## Summary

- **Root cause**: `make_anthropic_client()` in `backend/foreman_core/llm.py` was building kwargs with both `api_key` (from the server's process-level `ANTHROPIC_API_KEY`) and `auth_token` (from the guild's configured `ANTHROPIC_AUTH_TOKEN`). The Anthropic SDK raises `ValueError` when both are supplied — they are mutually exclusive — so the configured auth token never reached the API call.

- **Embedded foreman fix**: `auth_token` now takes unconditional precedence over `api_key` in `make_anthropic_client()`. When `ANTHROPIC_AUTH_TOKEN` is present in the effective env (process env merged with guild `extra_env`), `api_key` is skipped entirely.

- **Standalone foreman fix**: The `pioneer-foreman.toml` config had no field for an Anthropic `auth_token`. Added `anthropic_auth_token` to `Config`, reads it from `[claude] auth_token` in TOML or `ANTHROPIC_AUTH_TOKEN` env var, and forwards it to `make_anthropic_client()` via `extra_env` in `_get_anthropic_client()`.

## Files changed

- `backend/foreman_core/llm.py` — core fix: `auth_token` wins over `api_key`
- `foreman/pioneer_foreman/config.py` — new `anthropic_auth_token` field + loading
- `foreman/pioneer_foreman/runner.py` — pass `auth_token` via `extra_env`
- `foreman/pioneer-foreman.toml.example` — document the new option
- `backend/tests/test_foreman_llm.py` — regression tests for the conflict case
- `foreman/tests/test_config.py` — tests for TOML/env loading of `auth_token`

## Test plan

- [x] `cd backend && python -m pytest tests/test_foreman_llm.py -v` — 24 passed (2 new)
- [x] `cd foreman && python -m pytest tests/test_config.py -v` — 34 passed (4 new)

Closes #660

### DIFF
--- a/backend/foreman_core/llm.py
+++ b/backend/foreman_core/llm.py
@@ -251,20 +251,24 @@ def make_anthropic_client(
 
     # Direct Anthropic API. The SDK reads ANTHROPIC_* from os.environ on its own,
     # but guild-configured env_vars never reach this process, so resolve them
-    # from the merged `env` and pass explicitly. Explicit api_key arg wins.
+    # from the merged `env` and pass explicitly.
+    # api_key and auth_token are mutually exclusive; auth_token takes precedence
+    # so a guild-configured ANTHROPIC_AUTH_TOKEN overrides a process-level
+    # ANTHROPIC_API_KEY rather than causing an SDK ValueError.
     kwargs: dict = {}
-    resolved_api_key = api_key or env.get("ANTHROPIC_API_KEY")
-    if resolved_api_key:
-        kwargs["api_key"] = resolved_api_key
     if env.get("ANTHROPIC_AUTH_TOKEN"):
         kwargs["auth_token"] = env["ANTHROPIC_AUTH_TOKEN"]
+    else:
+        resolved_api_key = api_key or env.get("ANTHROPIC_API_KEY")
+        if resolved_api_key:
+            kwargs["api_key"] = resolved_api_key
     if env.get("ANTHROPIC_BASE_URL"):
         kwargs["base_url"] = env["ANTHROPIC_BASE_URL"]
     logger.info(
         "Foreman using Anthropic API (auth=%s, base_url=%s)",
         "auth-token"
         if kwargs.get("auth_token")
-        else ("api-key" if resolved_api_key else "default"),
+        else ("api-key" if kwargs.get("api_key") else "default"),
         kwargs.get("base_url", "default"),
     )
     return _anthropic_mod.AsyncAnthropic(**kwargs)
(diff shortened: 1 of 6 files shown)


jmelloy (commented):
## Review: Fix Anthropic auth token not passed to API call (#660)

The fix is correct, well-scoped, and fully tested. All 24 backend LLM tests and 34 foreman config tests pass locally.

### What the PR does

The Anthropic SDK raises `ValueError` when both `api_key` and `auth_token` are passed to `AsyncAnthropic`. The old code unconditionally built `api_key` from the process env *before* checking `auth_token`, so any guild that set `ANTHROPIC_AUTH_TOKEN` via `extra_env` would error at client-creation time instead of authenticating. The fix gates `api_key` behind an `else`, giving `auth_token` unconditional priority — the correct SDK behaviour.

### Code quality

- **`llm.py`** — the `if/else` restructure is minimal and clear. The logging fix (switching from the now out-of-scope `resolved_api_key` to `kwargs.get("api_key")`) is a tidy bonus cleanup.
- **`config.py`** — precedence order (overrides → TOML → env var) is consistent with how every other field is loaded.
- **`runner.py`** — cache key correctly extended to include `anthropic_auth_token`. The `extra_env or None` idiom (empty dict → `None`) is intentional and correct.
- **Tests** — the two new regression tests in `test_foreman_llm.py` cover both failure modes: process-level `ANTHROPIC_API_KEY` with a guild-supplied `ANTHROPIC_AUTH_TOKEN`, and explicit `api_key` arg with a guild-supplied `ANTHROPIC_AUTH_TOKEN`. The four new config tests cover TOML loading, env-var loading, TOML-beats-env precedence, and the default-to-`None` case. Good coverage of the full configuration surface.

### One minor nit

`runner.py:35` — the cache-key comment still says `(provider, region, api_key)` but the actual key is now `(provider, region, profile, api_key, anthropic_auth_token)`. Doesn't block merging.

**LGTM** — the fix is correct, tests are solid, and the change is well-contained.

PR #670, opened 2026-06-21
feat(members): invite users by GitHub username or ID

### LINKED ISSUE #668: Add ability to invite organizations
## Summary

Add the ability to invite organizations to Pioneer Square, similar to how individual users can be invited.

## Acceptance Criteria

- Users can send an invite to an organization (by name or ID)
- Invited organizations can accept/decline the invitation
- Organization members gain appropriate access upon acceptance
- UI surfaces the invite flow for organizations
- Backend handles organization invite creation, listing, and acceptance

## Notes

This is a new feature request. Implementation should follow existing invite patterns for individual users where applicable.

jmelloy:
## Summary

- Adds a `guild_invites` table (new Alembic migration) to store pending invites keyed to a GitHub username or numeric ID
- New owner-only REST endpoints: `POST/GET/DELETE /api/guilds/{id}/invites`
- `oauth.create_session` auto-accepts pending invites the moment the invited user first logs in, adding them as a guild member
- `GuildMembers.vue` invite form now falls back to creating a pending invite when the direct add returns 404 (user hasn't logged in yet); shows a "Pending invites" section with cancel controls
- New backend tests for create/list/cancel invites and the auto-accept-on-login flow
- Updated frontend spec to cover the 404→pending-invite fallback

## How it works

1. Owner types a GitHub username (or numeric ID) in the invite form
2. Backend tries to add them directly (`POST /members`)
3. If 404 (not in DB yet), falls back to `POST /invites` — stores a `guild_invites` row
4. When the user logs in via GitHub OAuth, `create_session` matches their `github_login`/`github_id` against pending invites and materializes the `guild_members` row automatically

## Test plan

- [ ] `cd backend && docker compose up -d postgres-test && python -m pytest tests/test_user_members_api.py -v`
- [ ] `cd frontend && npm test`
- [ ] Manual: invite a username that hasn't logged in → verify pending invite appears → log in as that user → verify they appear in the members list

Closes #668

### DIFF
--- a/backend/alembic/versions/20260621_000000_add_guild_invites.py
+++ b/backend/alembic/versions/20260621_000000_add_guild_invites.py
@@ -38,8 +38,26 @@ def upgrade() -> None:
         sa.PrimaryKeyConstraint("id"),
     )
     op.create_index("ix_guild_invites_guild_id_status", "guild_invites", ["guild_id", "status"])
+    # Partial unique indexes prevent duplicate pending invites at the DB level,
+    # closing the TOCTOU race that application-level one_or_none() checks leave open.
+    op.create_index(
+        "uq_guild_invites_login_pending",
+        "guild_invites",
+        ["guild_id", "github_login"],
+        unique=True,
+        postgresql_where=sa.text("status = 'pending' AND github_login IS NOT NULL"),
+    )
+    op.create_index(
+        "uq_guild_invites_github_id_pending",
+        "guild_invites",
+        ["guild_id", "github_id"],
+        unique=True,
+        postgresql_where=sa.text("status = 'pending' AND github_id IS NOT NULL"),
+    )
 
 
 def downgrade() -> None:
+    op.drop_index("uq_guild_invites_github_id_pending", table_name="guild_invites")
+    op.drop_index("uq_guild_invites_login_pending", table_name="guild_invites")
     op.drop_index("ix_guild_invites_guild_id_status", table_name="guild_invites")
     op.drop_table("guild_invites")
--- a/backend/alembic/versions/20260621_000000_add_guild_invites.py
+++ b/backend/alembic/versions/20260621_000000_add_guild_invites.py
@@ -1,0 +1,45 @@
+"""Add guild_invites table for pending GitHub-username invites.
+
+Revision ID: 20260621_000000_add_guild_invites
+Revises: 20260619_000000_rename_claude_usage_to_llm_usage
+Create Date: 2026-06-21
+
+Adds ``guild_invites`` to track pending invites addressed to a GitHub
+username or numeric ID. When the invited user logs in, ``oauth.create_session``
+auto-accepts matching invites and adds them as guild members.
+"""
+
+from __future__ import annotations
+
+from collections.abc import Sequence
+
+import sqlalchemy as sa
+from alembic import op
+
+revision: str = "20260621_000000_add_guild_invites"
+down_revision: str | Sequence[str] | None = "20260619_000000_rename_claude_usage_to_llm_usage"
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+
+def upgrade() -> None:
+    op.create_table(
+        "guild_invites",
+        sa.Column("id", sa.Integer(), nullable=False),
+        sa.Column("guild_id", sa.Integer(), nullable=False),
+        sa.Column("github_login", sa.Text(), nullable=True),
+        sa.Column("github_id", sa.Text(), nullable=True),
+        sa.Column("role", sa.Text(), nullable=False, server_default="'member'"),
+        sa.Column("invited_by", sa.Text(), nullable=False),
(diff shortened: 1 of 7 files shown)


frontend/src/components/GuildMembers.vue
```diff
@@ -151,12 +195,43 @@ async function invite() {
     inviteHint.value = `Added ${user} as ${inviteRole.value}.`
     await load()
   } catch (e) {
-    inviteError.value = e instanceof ApiError ? e.message : 'Failed to invite member'
+    if (e instanceof ApiError && e.status === 404) {
```
jmelloy: This catches *any* 404, not just 'user not found'. A 404 from the guild itself, a misconfigured API gateway, or a future endpoint rename will silently create a pending invite instead of showing the real error. Check `e.message` or a specific error code before falling back.

backend/oauth.py
```diff
@@ -203,6 +205,31 @@ async def create_session(code: str, state: str) -> dict:
...
+        for invite in pending_res.all():
+            member_stmt = pg_insert(GuildMember).values(
+                guild_id=invite.guild_id,
+                user_id=github_user_id,
+                role=invite.role,
+                created_at=now,
+            )
+            member_stmt = member_stmt.on_conflict_do_update(
+                index_elements=["guild_id", "user_id"],
+                set_={"role": member_stmt.excluded.role},
```
jmelloy, line 232: The invite query, member upsert, and invite status update span multiple awaits without an explicit transaction/savepoint. Concurrent logins for the same user (e.g. two browser tabs) can race, resulting in duplicate GuildMember rows or invites left in 'pending' state. Wrap this block in an explicit `async with db.begin()` or use `SELECT FOR UPDATE`.

backend/alembic/versions/20260621_000000_add_guild_invites.py
```diff
@@ -0,0 +1,45 @@
...
+        "guild_invites",
+        sa.Column("id", sa.Integer(), nullable=False),
+        sa.Column("guild_id", sa.Integer(), nullable=False),
+        sa.Column("github_login", sa.Text(), nullable=True),
+        sa.Column("github_id", sa.Text(), nullable=True),
+        sa.Column("role", sa.Text(), nullable=False, server_default="'member'"),
+        sa.Column("invited_by", sa.Text(), nullable=False),
+        sa.Column("created_at", sa.DateTime(timezone=True), nullable=False),
+        sa.Column("status", sa.Text(), nullable=False, server_default="'pending'"),
+        sa.ForeignKeyConstraint(["guild_id"], ["guilds.id"]),
```
jmelloy: There is no unique constraint on `(guild_id, github_login, status)` or `(guild_id, github_id, status)`. The application-level duplicate check in `create_guild_invite` is a TOCTOU race: two concurrent requests for the same username can both pass the `one_or_none()` check and insert two pending rows. Add a partial unique index, e.g. `UNIQUE (guild_id, github_login) WHERE status = 'pending'`.

backend/tests/test_user_members_api.py
```diff
@@ -215,3 +217,250 @@ def test_api_me_excludes_legacy_guilds(client):
...
+    db_url: str, guild_id: str, github_login: str, role: str = "member"
+) -> None:
+    """Insert a pending guild invite row directly."""
+    now = datetime.now(UTC)
+    with _sync_session(db_url) as session:
+        from sqlalchemy import select as sa_select
+
+        guild_pk = session.scalar(
+            sa_select(Guild.id).where(Guild.slug == guild_id, Guild.deleted_at.is_(None))
+        )
```
jmelloy, line 376: This test duplicates the invite-acceptance logic inline rather than calling `oauth.create_session`. It will pass even if `create_session` is broken. Consider mocking the GitHub OAuth response and calling the actual `/auth/callback` endpoint, or at minimum add a separate unit test for `create_session` with a mocked DB.

jmelloy (commented):
- **Security concern**: The 404 fallback in `GuildMembers.vue` is fragile — any 404 (e.g. guild not found, network proxy 404) will silently create a pending invite instead of surfacing the real error to the user.
- **Data integrity**: The `guild_invites` migration uses `sa.Text()` for `invited_by` with a FK to `users.id`, but `users.id` is likely a text/UUID column; the FK type must match exactly. Also `github_id` is stored as `Text` but compared against a numeric GitHub user ID string — this is inconsistent with `users.github_id` which may be an integer column.
- **Race condition / missing transaction isolation**: In `oauth.create_session`, the pending invite query and the subsequent member upsert + status update are not wrapped in a single atomic operation; concurrent logins could double-insert or leave invites in an inconsistent state.
- **Test quality**: `test_pending_invite_accepted_on_login` manually re-implements the OAuth acceptance logic rather than exercising the actual `create_session` code path, so it does not truly test the auto-accept feature and could miss real bugs.
- **Missing constraint**: There is no unique constraint on `(guild_id, github_login)` or `(guild_id, github_id)` for pending invites in the migration, only an index; the application-level upsert check is therefore subject to a TOCTOU race that can create duplicate pending invites.

Issue #649, opened 2026-06-14
Design: Split Foreman context across tasks (per-task context + parent watcher)


jmelloy:
## Background

The Foreman currently runs in a single Claude context that accumulates state for all active tasks. As the number of concurrent tasks grows, this creates several problems:

- Context window fills up with unrelated task history
- Noise from one task bleeds into reasoning about another
- A single context crash or confused state affects all tasks simultaneously

## Goal

Design a system where each active task has its own isolated context, and a lightweight parent Foreman watches the overall state and orchestrates between them.

## What we want

A design doc covering:
1. How per-task child contexts are spawned and what state they carry
2. How the parent Foreman monitors child contexts (polling, events, escalation)
3. How the parent aggregates status for the human
4. Context lifecycle (creation, handoff, teardown)
5. Failure modes and recovery

Relates to the current single-context Foreman architecture in `foreman/`.

jmelloy:
# Design: Per-Task Context Splitting for the Foreman

*Branch: claude/design-foreman-per-task-context-splitting-t-fgun*  
*Date: 2026-06-14*

---

## 1. Motivation & Problem Statement

### What the Foreman does today

The Foreman is a Claude agent that manages all active tasks in a single conversation thread per guild. Every trigger — `task-complete`, `github-event`, `needs-input`, human chat, and the periodic poll — is serialized through one `_processing` lock and appended to one shared conversation history. Each turn injects the full `<state>` block (all workers, all tasks) into the user message.

### What breaks as task count grows

**Context bloat.** The state preamble grows linearly with task count. With 10 active tasks, each Claude call carries a wall of JSON that's largely irrelevant to the specific event being handled. Because `MAX_HISTORY_MESSAGES = 20` and `_HUMAN_TURN_WINDOW = 5`, real decision context gets squeezed out by task-list noise.

**Cross-task interference.** Tool results, follow-up reasoning, and GitHub event details for task A interleave with task B's context in the same conversation thread. The foreman's reasoning about "which worker to assign" or "does this PR need a follow-up" can be polluted by unrelated prior turns. A `send_followup` intended for task A might reference task B's branch name if the history is dense enough.

**Serial throughput.** The `_processing` lock ensures exactly one foreman run at a time. If task A's `task-complete` review triggers a 10-round Claude conversation (CI failure analysis, redirect, follow-up), task B's `task-complete` event waits in a 100-slot queue. With N concurrent tasks, p99 latency scales with N × average-run-duration.

[...]

PR #674, opened 2026-06-23
fix: ensure task_id is set on messages wherever context is available

jmelloy:
## Summary

Audited all `Message()` creation sites in the codebase and added `task_id` wherever the task context is already known:

- **`models.py`**: Added nullable `task_id` FK column to `Message` (→ `tasks.id`), with an index for efficient per-task message queries
- **`foreman/runner.py`**: Pass `_task_id` to tool_use, tool_result, and final-response `Message` rows — the variable was already being computed from the active task list but never forwarded to the DB rows
- **`routes/webhooks.py`**: Set `task_id` on the chat message persisted by the GitHub webhook handler and the `ci-notify` endpoint — both already resolve `task_id` from the PR coordinates
- **`routes/foreman.py`**: Add `task_id` field to `MessageCreate`; validate guild ownership before persisting; pass it through to the `Message` row (used by the standalone foreman)
- **Migration**: `20260623_000000_add_task_id_to_messages` — `ALTER TABLE messages ADD COLUMN task_id TEXT REFERENCES tasks(id)` + index

Messages that legitimately have no task context remain `NULL`: general user chat, worker-online/offline notifications, periodic-check triggers, and claude-auth prompts.

## Test plan

- [ ] `cd backend && python -m pytest -x -q` (requires postgres-test container)
- [ ] `cd worker && python -m pytest -x -q` — 235 tests pass; 1 pre-existing failure in `test_tools_detection.py` unrelated to this change
- [ ] `ruff check . --fix && ruff format .` — all checks pass

Relates to per-task context splitting (#649).

### DIFF
--- a/backend/alembic/versions/20260623_000000_add_task_id_to_messages.py
+++ b/backend/alembic/versions/20260623_000000_add_task_id_to_messages.py
@@ -1,0 +1,37 @@
+"""Add task_id column to messages table.
+
+Revision ID: 20260623_000000_add_task_id_to_messages
+Revises: 20260621_000000_add_guild_invites
+Create Date: 2026-06-23
+
+Adds a nullable ``task_id`` FK column to ``messages`` so that chat messages
+created in response to task events (tool calls, webhook events, CI results,
+foreman responses) can be linked to the task that caused them. General user
+chat and system-level messages (worker-online, periodic-check, etc.) remain
+NULL.
+"""
+
+from __future__ import annotations
+
+from collections.abc import Sequence
+
+import sqlalchemy as sa
+from alembic import op
+
+revision: str = "20260623_000000_add_task_id_to_messages"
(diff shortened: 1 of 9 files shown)


jmelloy (commented):
### Code review

Found 1 issue:

1. `_task_id` is `None` whenever a guild has 0 or 2+ active tasks, so all `Message` rows written during those foreman runs get `task_id=NULL`. The PR's goal is to tag messages with task context "wherever context is available", but the single-task heuristic silently fails the multi-task case, which is the common steady state for active guilds. The callers of `_trigger_foreman` in `ws_handlers.py` already pass a specific `task_id` for each worker event, but that value never reaches `run_foreman_ai`; the function relies entirely on this heuristic instead.

https://github.com/jmelloy/pioneer-square/blob/b781aa8b7a23efd8bd9c1fa418e30cb89c4ee2fb/backend/foreman/runner.py#L556-L561

PR #678, opened 2026-07-01
Tasks pane: tree view parented to issues/PRs (#677)

### LINKED ISSUE #677: Tasks/Issues pane: tree view with tasks parented to issues/PRs
## Summary

The tasks/issues sidebar pane should display tasks in a hierarchical tree rather than a flat list. Every task (plan, execute, review, followup) should be nested under its parent GitHub issue or PR.

### Tree structure
- Top-level nodes: GitHub issues and standalone PRs
- Child nodes: tasks that reference the issue (`issue_number`) or PR (`pr_url`)
  - Plan → Execute → Review → Followup tasks nest in order under the same issue
  - If a task has a `parent_task_id`, it nests under that parent task (e.g. a review task under its execute task)
- PRs that are not linked to an issue appear as top-level nodes with their own child tasks

### PR review parenting
- When a PR review event comes in, look up the PR's linked issue (via `closes #N` in the PR body or the GitHub API `issues` link) and parent the review task to that issue node
- If no linked issue is found, parent the review task to the PR node

### Visual design
- Collapsible tree nodes (expand/collapse per issue/PR)
- Show issue/PR title, number, and state (open/closed/merged) on each top-level node
- Show task phase badge (plan / execute / review / followup) on child task rows
- Show task state (running / awaiting-review / done / failed) inline
- Worker ID badge on running tasks

### Data
- The `parent_task_id` column already exists on tasks — use it for task→task nesting
- Add a `github_issue_number` / `github_issue_repo` lookup so the UI can group tasks by issue
- Expose a `/api/tasks/tree` endpoint (or extend the existing tasks endpoint) that returns the hierarchical structure ready for the frontend to render

## Acceptance criteria
- [ ] Flat list replaced with collapsible tree
- [ ] Plan + execute + review tasks for the same issue appear under one issue node
- [ ] Review tasks on a PR are parented to the issue that PR closes (if any)
- [ ] Standalone PRs/tasks with no issue appear as their own top-level nodes
- [ ] Tree state (expanded/collapsed) persists across page refreshes (localStorage)

jmelloy:
## Summary

- **Backend**: New `GET /guilds/{guild_id}/tasks/tree` endpoint groups live tasks hierarchically under their GitHub issues/PRs. For each unique issue it fetches the title and open/closed state from the GitHub API (results cached in-process for 10 minutes). Review tasks whose PR body contains a `closes #N` / `fixes #N` reference are automatically parented to that issue node. Tasks with `parent_task_id` are nested under their parent within the group. Remaining tasks appear in an `ungrouped` array. All GitHub API calls run concurrently via `asyncio.gather + to_thread`.
- **Frontend**: Replaced the flat `TaskList` with two new components — `TaskTree.vue` (top-level) and `TaskTreeRow.vue` (recursive task row). Issue nodes are collapsible and show `#number`, title, open/closed badge, and task count. Each task row shows the task name, phase pill (plan/execute/review/followup), state dot + label, and worker ID when active. Expand/collapse state is persisted to `localStorage` per node key so it survives page refreshes. The tree re-fetches (debounced 1.5 s) whenever the flat tasks list changes via WebSocket events.

## Test plan

- [ ] Tasks with `issue_number` + `issue_repo` appear under the correct issue node
- [ ] A review task whose PR body contains `closes #N` is parented to that issue
- [ ] Tasks with `parent_task_id` are nested under their parent task row
- [ ] Tasks with no issue linkage appear in the Ungrouped section
- [ ] Collapse an issue node → refresh → node is still collapsed (localStorage)
- [ ] Clicking a task row opens the task detail pane as before
- [ ] No GitHub token → issues render as `#N` placeholders (no crash)

Closes #677

### DIFF
--- a/backend/routes/tasks.py
+++ b/backend/routes/tasks.py
@@ -332,6 +332,9 @@ _gh_issue_cache: dict[str, tuple[dict, float]] = {}
 _gh_pr_body_cache: dict[str, tuple[str | None, float]] = {}
 
 
+_CACHE_MAX_SIZE = 1000
+
+
 def _gh_fetch_issue(repo: str, number: int, token: str) -> dict | None:
     """Fetch GitHub issue title/state. Cached for _TREE_CACHE_TTL seconds."""
     key = f"{repo}#{number}"
@@ -353,6 +356,8 @@ def _gh_fetch_issue(repo: str, number: int, token: str) -> dict | None:
             "title": data.get("title") or f"#{number}",
             "state": data.get("state") or "open",
         }
+        if len(_gh_issue_cache) >= _CACHE_MAX_SIZE:
+            _gh_issue_cache.clear()
         _gh_issue_cache[key] = (result, _monotime() + _TREE_CACHE_TTL)
         return result
     except Exception:
@@ -377,9 +382,13 @@ def _gh_fetch_pr_body(repo: str, pr_number: int, token: str) -> str | None:
         with urllib.request.urlopen(req, timeout=10) as resp:
             data = json.loads(resp.read())
         body: str | None = data.get("body") or ""
+        if len(_gh_pr_body_cache) >= _CACHE_MAX_SIZE:
+            _gh_pr_body_cache.clear()
         _gh_pr_body_cache[key] = (body, _monotime() + _TREE_CACHE_TTL)
         return body
     except Exception:
+        if len(_gh_pr_body_cache) >= _CACHE_MAX_SIZE:
+            _gh_pr_body_cache.clear()
         _gh_pr_body_cache[key] = (None, _monotime() + _TREE_CACHE_TTL)
         return None
 
@@ -445,7 +454,7 @@ async def get_guild_tasks_tree(
         .where(col(GuildMember.guild_id) == guild_pk, col(GuildMember.role) == "owner")
         .limit(1)
     )
-    gh_token: str | None = token_row.one_or_none()
+    gh_token: str | None = row[0] if (row := token_row.one_or_none()) else None
 
     # --- Build task_id → (issue_repo, issue_number) mapping ---
 
@@ -481,8 +490,11 @@ async def get_guild_tasks_tree(
 
     # Pass 3: propagate issue assignment through parent_task_id chains
     changed = True
-    while changed:
+    max_iters = len(raw_tasks)
+    iters = 0
+    while changed and iters < max_iters:
         changed = False
+        iters += 1
         for task in raw_tasks:
             if task["id"] in task_issue:
                 continue
--- a/backend/routes/tasks.py
+++ b/backend/routes/tasks.py
@@ -7,15 +7,20 @@ direct unit testing in ``tests/test_soft_delete_tasks.py``.
 
 from __future__ import annotations
 
+import asyncio
 import json
+import re
+import urllib.request
+from collections import defaultdict
 from datetime import UTC, datetime, timedelta
+from time import time as _monotime
 
 from auth_deps import get_guild_pk, require_member
 from database import get_db_dep
 from events import broadcast_msg
 from fastapi import APIRouter, Depends, HTTPException
 from lock_service import LockService
-from models import Guild, Task, TaskLog, live_tasks_filter
+from models import GithubToken, Guild, GuildMember, Task, TaskLog, live_tasks_filter
 from pydantic import BaseModel
 from sqlalchemy import update
 from sqlmodel import col, select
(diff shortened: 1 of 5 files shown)


backend/routes/tasks.py
```diff
@@ -314,6 +319,222 @@ async def cancel_task_endpoint(
...
+):
+    """Return tasks grouped hierarchically under their GitHub issues/PRs.
+
+    Top-level nodes are GitHub issues (identified by issue_number+issue_repo on
+    the task).  Review tasks whose PR body contains a ``closes #N`` reference are
+    also parented to that issue.  Tasks with parent_task_id are nested under their
+    parent task within the group.  Remaining tasks go in ``ungrouped``.
+    """
+    guild_pk = await get_guild_pk(db, guild_id)
+    if guild_pk is None:
```
jmelloy, line 423: `token_row.one_or_none()` returns a `Row` (a 1-tuple), not a bare `str`. Passing this into `f'Bearer {token}'` will produce `Bearer ('ghp_...',)`. Use `token_row.one_or_none()` and index it: `gh_token = row[0] if (row := token_row.one_or_none()) else None`.

backend/routes/tasks.py
```diff
@@ -314,6 +319,222 @@ async def cancel_task_endpoint(
...
+    raw_tasks = [dict(r._mapping) for r in result.all()]
+
+    # Fetch the guild owner's GitHub token (may be None)
+    token_row = await db.exec(
+        select(col(GithubToken.access_token))
+        .join(GuildMember, col(GuildMember.user_id) == col(GithubToken.github_user_id))
+        .where(col(GuildMember.guild_id) == guild_pk, col(GuildMember.role) == "owner")
+        .limit(1)
+    )
+    gh_token: str | None = token_row.one_or_none()
```
jmelloy: The propagation loop has no cycle-detection. A circular `parent_task_id` reference (e.g. task A → B → A) will cause an infinite loop. Add a visited/depth guard or limit iterations to `len(raw_tasks)`.

backend/routes/tasks.py
```diff
@@ -314,6 +319,222 @@ async def cancel_task_endpoint(
...
+_GH_PR_URL_RE = re.compile(r"https://github\.com/([^/\s]+/[^/\s]+)/pull/(\d+)")
+
+# Simple in-process TTL caches (key → (value, expires_at))
+_TREE_CACHE_TTL = 600.0  # 10 minutes
+_gh_issue_cache: dict[str, tuple[dict, float]] = {}
+_gh_pr_body_cache: dict[str, tuple[str | None, float]] = {}
+
+
+def _gh_fetch_issue(repo: str, number: int, token: str) -> dict | None:
+    """Fetch GitHub issue title/state. Cached for _TREE_CACHE_TTL seconds."""
```
jmelloy, line 339: `_gh_issue_cache` and `_gh_pr_body_cache` are unbounded dicts — entries are only replaced, never evicted beyond the TTL check. Over time (many unique repos/issues) this leaks memory. Cap the cache size or use `functools.lru_cache` / a bounded structure.

frontend/src/types.ts
```diff
@@ -81,10 +81,30 @@ export interface Task {
...
+  children: TaskTreeNode[]
+}
+
+export interface IssueTreeNode {
+  type: 'issue'
+  issue_number: number
+  issue_repo: string
+  title: string
+  state: 'open' | 'closed'
+  tasks: TaskTreeNode[]
```
jmelloy, line 100: `state` is typed as `'open' | 'closed'` but the backend's GitHub API response and the CSS both include a `'merged'` state. Widen the type to `'open' | 'closed' | 'merged'` to avoid a silent type mismatch.

frontend/src/components/sidebar/TaskTree.vue
```diff
@@ -0,0 +1,289 @@
...
+)
+
+// Also watch for state changes on individual tasks
+watch(
+  () => tasksStore.tasks.map((t) => t.state).join(','),
+  () => {
+    const guildId = guildStore.currentGuild?.id
+    if (guildId) debouncedFetch(guildId)
+  },
+)
```
jmelloy, line 155: Watching both `tasksStore.tasks.length` and `tasksStore.tasks.map(t => t.state).join(',')` means every task-state change triggers two `debouncedFetch` calls (the second resets the first timer). Consolidate into a single watcher on a combined computed value to avoid the double-debounce pattern.

jmelloy (commented):
- Backend introduces synchronous `urllib.request` calls inside `asyncio.to_thread`, which is correct, but the in-process TTL cache (`_gh_issue_cache`, `_gh_pr_body_cache`) is a plain dict with no eviction bound — it will grow unboundedly in long-running processes.
- The GitHub token retrieval returns a `Row` object (a tuple-like), not a bare string; `gh_token` will be a `Row` rather than `str | None`, causing the `Bearer {token}` header to embed a repr of the row instead of the actual token value.
- The parent-task-id propagation loop (Pass 3) is O(n²) in the worst case and could be an infinite loop if there is a cycle in `parent_task_id` references (no cycle detection).
- Frontend `TaskTree.vue` watches both `tasksStore.tasks.length` and a computed join of all task states, meaning two concurrent debounce timers fire for the same event; combined with a 1.5 s debounce this causes redundant fetches but is not a correctness bug.
- `IssueTreeNode.state` type is narrowed to `'open' | 'closed'` but the backend can also return `'merged'` (the CSS has a `.badge-merged` rule), creating a TypeScript type mismatch that will pass undetected at runtime.

PR #681, opened 2026-07-01
Fix send_followup branch reuse (#680)

### LINKED ISSUE #680: Bug: send_followup creates a new branch instead of reusing the task's existing branch
## Summary

When the Foreman picks up a `devReady` issue and a worker opens a PR, subsequent `send_followup` calls (e.g. for CI fixes, review-requested changes, or lint errors) appear to create a new branch rather than continuing work on the original task branch.

## Expected behaviour

`send_followup` should always continue work on the branch that was created by the original `assign_task` call — the branch stored in the task row's `branch` field (e.g. `claude/issue-viewer-panel-body-comments-comment-edit-676-t-byvw`).

## Observed behaviour

Follow-up work lands on a different branch (either a fresh `claude/...` worktree branch or a pattern mismatch), causing:
- A second PR to be opened instead of pushing to the existing one, **or**
- The original PR branch diverging from `main` because the fix was pushed to the wrong branch.

## Branch naming evidence

Looking at recent PRs, original task branches follow the pattern:
```
claude/<slug>-<issue_number>-<task_id_prefix>
```
e.g. `claude/issue-viewer-panel-body-comments-comment-edit-676-t-byvw`

If `send_followup` doesn't correctly reuse the worktree / check out the existing branch by name, the worker may default to creating a new worktree with a fresh branch name.

## Steps to reproduce

1. Pick up a `devReady` issue → worker opens PR on branch `claude/...-t-xxxx`
2. Call `send_followup(task_id=...)` with a fix instruction
3. Observe that the worker creates a new branch or opens a second PR instead of pushing to the existing branch

## Root cause hypothesis

The worker runtime's follow-up flow may not be passing `--branch <existing_branch>` (or equivalent) when creating the follow-up worktree, causing it to derive a new branch name from the instruction text rather than the stored task branch.

## Impact

- Duplicate PRs
- Original PR left unfixed
- Confusing branch history

## Suggested fix

In the worker runtime's `send_followup` handler, always resolve the task's `branch` field from the DB and explicitly check out / reuse that branch rather than deriving a new name from the follow-up description.

jmelloy:
## Summary

- **Bug**: When a worker restarts and the idle puller picks up a follow-up task via REST, it was generating a fresh branch name and pushing to a new PR instead of continuing on the existing branch.
- **Root cause**: `_execute_task` looked only for `followup_branch` (a WS-path key set by the `task-followup` message handler) and `followup_instructions` to detect follow-ups. The REST idle-puller path produces a `row_to_dict(Task)` snapshot that has the DB `branch` column and `phase="followup"` but neither of those keys.
- **Fix** (2 lines in `worker/pioneer_worker/worker.py`):
  1. `followup_branch` now falls back to `task.get("branch")` so the existing PR branch is always used.
  2. `is_followup` now also checks `task.get("phase") == "followup"` so the `attach_worktree` path is taken even when `followup_instructions` is absent.

## Paths covered

| Path | Task dict keys | Before fix | After fix |
|------|---------------|-----------|-----------|
| WS (`task-followup`) | `followup_branch`, `followup_instructions` | ✓ correct | ✓ correct |
| REST (idle puller after restart) | `branch`, `phase="followup"` | ✗ new branch | ✓ reuses branch |

## Test plan

- Added `worker/tests/test_followup_branch_reuse.py` with 3 tests:
  - `test_followup_ws_path_uses_existing_branch` — WS path calls `attach_worktree` with the correct branch
  - `test_followup_rest_path_uses_existing_branch` — idle-puller REST path calls `attach_worktree` with the DB branch (the regression case)
  - `test_new_task_still_generates_fresh_branch` — new tasks still generate a fresh `claude/...` branch

```
cd worker && python -m pytest tests/test_followup_branch_reuse.py -v
# 3 passed in 0.09s
```

Full worker suite: 249 passed, 1 pre-existing failure in `test_tools_detection.py` unrelated to this change.

Closes #680

### DIFF
--- a/worker/pioneer_worker/worker.py
+++ b/worker/pioneer_worker/worker.py
@@ -1915,19 +1915,26 @@ class Worker:
                 and os.path.isdir(os.path.join(wt_path, ".git"))
                 or (os.path.isdir(wt_path) and os.path.isfile(os.path.join(wt_path, ".git")))
             ):
-                # Worktree from a prior run on the same task — reuse it. Pull
-                # the latest origin state for the branch so a different worker
-                # that pushed changes is reflected here.
+                # Worktree from a prior run on the same task — reuse it.
                 logger.info("Task %s: reusing worktree at %s", task_id, wt_path)
                 await emit(f"Reusing worktree {repo_name}", level=LEVEL_WORKER)
                 if is_followup:
+                    # Pull latest so we don't clobber commits pushed since the
+                    # last follow-up or by other workers.
                     await git_ops.run_git(["fetch", "origin", branch], cwd=wt_path)
+                    await git_ops.run_git(
+                        ["reset", "--hard", f"origin/{branch}"], cwd=wt_path
+                    )
                 worktree_entries.append((repo_full, repo_path, wt_path))
                 if primary_wt is None:
                     primary_wt = wt_path
                 continue
             if is_followup:
                 logger.info("Task %s: attaching worktree %s to branch %s", task_id, wt_path, branch)
+                # attach_worktree fetches origin/<branch> before checking it out, so the
+                # new worktree starts at the latest remote commit — no separate pull needed.
+                # Pull latest so we don't clobber commits pushed since the last follow-up
+                # or by other workers.
                 ok = await git_ops.attach_worktree(repo_path, wt_path, branch)
                 if not ok:
                     # Branch never reached origin (e.g. original task failed before push).
--- a/worker/pioneer_worker/worker.py
+++ b/worker/pioneer_worker/worker.py
@@ -1848,8 +1848,16 @@ class Worker:
                 repos.insert(0, issue_repo)
         logger.info("Task %s: repos=%s", task_id, repos)
         followup_instructions = task.get("followup_instructions") or ""
-        followup_branch = task.get("followup_branch") or ""
-        is_followup = bool(followup_instructions)
+        # When the idle puller picks up a follow-up task via REST, the task dict
+        # contains `branch` (the DB column) but not `followup_branch` (a
+        # WS-path key set only by the task-followup message handler).  Fall back
+        # to `task.get("branch")` so we never generate a fresh branch name for a
+        # task that already has an open PR.
+        followup_branch = task.get("followup_branch") or task.get("branch") or ""
+        # phase="followup" is the DB marker for tasks dispatched by send_followup.
+        # The idle puller path has no followup_instructions (those lived only in
+        # the WS message), so we must check phase as well.
+        is_followup = bool(followup_instructions) or task.get("phase") == "followup"
 
         logger.info(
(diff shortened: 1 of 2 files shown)


jmelloy (commented):
## Review: Approve ✓

The fix is correct and minimal. The two changed lines cleanly address the root cause: `_execute_task` was relying on WS-only keys (`followup_branch`, `followup_instructions`) that the REST idle-puller path never provides, causing a fresh branch to be generated and a duplicate PR to be opened.

### Core fix analysis

**`followup_branch = task.get("followup_branch") or task.get("branch") or ""`**

Safe. The branch name is deterministic (`claude/{slug}-{task_id[:6]}`), so for a non-followup task that has already run and has `branch` set in the DB, the fallback resolves to the same string that the `or` formula would have generated anyway — no change in behavior for those tasks.

**`is_followup = bool(followup_instructions) or task.get("phase") == "followup"`**

Correct for both paths:
- WS path: `TaskFollowupMsg` has no `phase` field, so the WS handler defaults it to `"execute"` (line 1426) — the new clause is always False here; `is_followup` is driven by `followup_instructions` as before.
- REST path: `send_followup` writes `phase="followup"` to the DB; `followup_instructions` is empty (never persisted). The new clause correctly fires.

All three new tests pass. The negative-case test (`test_new_task_still_generates_fresh_branch`) correctly guards against regressions.

---

### One functional limitation to be aware of (not blocking)

On the REST idle-puller path, `followup_instructions` is always `""` because it was never written to the DB — only the WS message carried it transiently. With the fix, `is_followup=True` now correctly routes the task to `attach_worktree` and the right branch, but the Claude prompt (line 2012) becomes:

```
Follow-up instructions:
             ← blank
Make the requested changes…
```

Claude must infer what to do from the PR review comments alone, which may be enough in practice. This is a pre-existing architectural gap (instructions not persisted) that this PR doesn't claim to solve. Worth a follow-up issue if workers are restarted frequently while followup tasks are in-flight.

---

### Minor test nits (non-blocking)

1. **`test_followup_rest_path_uses_existing_branch`** sets `repos=[]`, which silently exercises the `cfg.repos` fallback branch rather than the explicit-repos path real REST task dicts use. Consider `repos=["owner/repo"]` to match `row_to_dict(Task)` output.

2. The two followup tests have no inverse assertion that `create_worktree` was *not* called. Adding `mock_create.assert_not_called()` would catch a future mis-routing regression.

3. The new-task test similarly lacks `mock_attach.assert_not_called()` — a mis-classified non-followup task would still satisfy the existing positive assertions.

---

### Pre-existing unrelated failure

`tests/test_tools_detection.py::TestToolsCLIFlag::test_no_tools_flag_gives_none` fails on this branch but is not caused by this PR's changes (exists on `main`).

---

**Verdict:** Core fix is correct, well-targeted, and has solid regression coverage. The blank-instructions limitation on the REST path is a known architectural gap outside this PR's scope.

PR #685, opened 2026-07-01
Fix: issue tree always shows 'Open' status

### LINKED ISSUE #684: Issue tree shows all issues as "Open" regardless of actual state
## Bug

In the Pioneer Square UI, the issue tree/list always displays every issue with an "Open" status badge, even when the underlying GitHub issue is actually closed.

## Expected behaviour

Issues that are closed on GitHub should show a "Closed" badge (or equivalent visual indicator) in the issue tree.

## Actual behaviour

Every issue in the tree shows as "Open".

## Likely cause

The frontend is probably not reading the `state` field from the issue data, or the component rendering the status badge is hardcoded / defaulting to "open".

jmelloy:
## Summary

- **Root cause**: `get_guild_tasks_tree` fetched the GitHub token only from the guild owner. When the owner had no token configured, every issue state fell back to the hardcoded string `"open"`.
- **Backend fix**: after failing to find the owner's token, the endpoint now falls back to the requesting user's own GitHub token (they are already authenticated). This means any logged-in member with a connected GitHub account sees the correct `open`/`closed` state in the sidebar issue tree.
- **Type fix**: added explicit `state: string` to the `GitHubIssue` interface in `types.ts` so TypeScript knows the field is always present.
- **List fix**: `IssuesTab.vue` now fetches `state=all` (was `state=open`) and renders a colour-coded state badge per issue row — green for open, red for closed — matching the existing conventions in `TaskTree.vue`.

## Test plan

- [ ] Open the sidebar task tree — closed GitHub issues should now show a red "closed" badge instead of green "open"
- [ ] Open the Issues tab in the chat pane — each issue row now shows a state badge; closed issues are red
- [ ] Verify that a guild where the owner has no GitHub token but the viewing member does still shows correct states
- [ ] Run `npm test` in `frontend/` — all 103 tests pass

Closes #684

### DIFF
--- a/backend/routes/tasks.py
+++ b/backend/routes/tasks.py
@@ -447,9 +447,8 @@ async def get_guild_tasks_tree(
     )
     raw_tasks = [dict(r._mapping) for r in result.all()]
 
-    # Fetch GitHub token: guild owner preferred, fall back to the requesting user's own token.
-    # This ensures any authenticated member can see correct issue states even if the owner
-    # has not connected their GitHub account.
+    # Fetch GitHub token from the guild owner only to avoid privilege escalation —
+    # a regular member's personal token must not be used to access guild resources.
     token_row = await db.exec(
         select(col(GithubToken.access_token))
         .join(GuildMember, col(GuildMember.user_id) == col(GithubToken.github_user_id))
@@ -457,14 +456,6 @@ async def get_guild_tasks_tree(
         .limit(1)
     )
     gh_token: str | None = row[0] if (row := token_row.one_or_none()) else None
-    if not gh_token:
-        user_token_row = await db.exec(
-            select(col(GithubToken.access_token)).where(
-                col(GithubToken.github_user_id) == github_user_id
-            )
-        )
-        if user_row := user_token_row.one_or_none():
-            gh_token = user_row[0]
 
     # --- Build task_id → (issue_repo, issue_number) mapping ---
 
--- a/backend/routes/tasks.py
+++ b/backend/routes/tasks.py
@@ -447,7 +447,9 @@ async def get_guild_tasks_tree(
     )
     raw_tasks = [dict(r._mapping) for r in result.all()]
 
-    # Fetch the guild owner's GitHub token (may be None)
+    # Fetch GitHub token: guild owner preferred, fall back to the requesting user's own token.
+    # This ensures any authenticated member can see correct issue states even if the owner
+    # has not connected their GitHub account.
     token_row = await db.exec(
         select(col(GithubToken.access_token))
         .join(GuildMember, col(GuildMember.user_id) == col(GithubToken.github_user_id))
@@ -455,6 +457,14 @@ async def get_guild_tasks_tree(
         .limit(1)
     )
     gh_token: str | None = row[0] if (row := token_row.one_or_none()) else None
+    if not gh_token:
+        user_token_row = await db.exec(
+            select(col(GithubToken.access_token)).where(
+                col(GithubToken.github_user_id) == github_user_id
+            )
+        )
+        if user_row := user_token_row.one_or_none():
+            gh_token = user_row[0]
 
     # --- Build task_id → (issue_repo, issue_number) mapping ---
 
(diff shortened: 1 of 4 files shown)


backend/routes/tasks.py
```diff
@@ -447,14 +447,24 @@ async def get_guild_tasks_tree(
...
-    # Fetch the guild owner's GitHub token (may be None)
+    # Fetch GitHub token: guild owner preferred, fall back to the requesting user's own token.
+    # This ensures any authenticated member can see correct issue states even if the owner
+    # has not connected their GitHub account.
     token_row = await db.exec(
         select(col(GithubToken.access_token))
         .join(GuildMember, col(GuildMember.user_id) == col(GithubToken.github_user_id))
         .where(col(GuildMember.guild_id) == guild_pk, col(GuildMember.role) == "owner")
         .limit(1)
     )
```
jmelloy, line 457: Using the requesting user's personal token to fetch issue states for the guild creates a privilege-escalation risk: a member with access to private repos the guild owner does not have can now inadvertently expose those repos' issue states (open/closed) within the guild context. Consider restricting the fallback to tokens of users with at least a certain guild role, or better, use a server-side GitHub App installation token instead of any user token.

backend/routes/tasks.py
```diff
@@ -447,14 +447,24 @@ async def get_guild_tasks_tree(
...
     gh_token: str | None = row[0] if (row := token_row.one_or_none()) else None
+    if not gh_token:
+        user_token_row = await db.exec(
```
jmelloy: Verify that `github_user_id` here is the authenticated user's GitHub user ID and not their internal application user ID. If `GithubToken.github_user_id` stores GitHub's numeric ID but the variable holds the app's UUID, this query will silently return nothing on every call, making the fallback a no-op.

frontend/src/stores/github.ts
```diff
@@ -98,7 +98,7 @@ export const useGitHubStore = defineStore('github', () => {
       const allIssues = await Promise.all(
         reposToFetch.map(async (repoName) => {
           const res = await fetch(
-            `${GH_API}/repos/${repoName}/issues?state=open&per_page=30&sort=created&direction=desc`,
+            `${GH_API}/repos/${repoName}/issues?state=all&per_page=30&sort=created&direction=desc`,
```
jmelloy: Fetching `state=all` with `per_page=30` means closed issues can push open ones off the first page. Consider either increasing `per_page` (up to 100 is allowed by the GitHub API) or fetching open and closed in separate requests if both need to be displayed, to avoid silently truncating the open-issue list.

frontend/src/components/chat-pane/IssuesTab.vue
```diff
@@ -30,6 +30,10 @@
       <div class="issue-top">
         <span class="issue-repo">{{ issue.repo }}</span>
         <span class="issue-number">#{{ issue.number }}</span>
+        <span
```
jmelloy, line 33: The fallback `issue.state || 'open'` is now unnecessary given `state: string` is non-optional in the interface, and it actively masks any unexpected state value (e.g. `''` or a future GitHub state) by displaying 'open'. Remove the fallback or at least map unknown values to a neutral 'unknown' badge rather than implying the issue is open.

jmelloy (commented):
- Backend fallback logic correctly addresses the root cause, but introduces a subtle security/data-isolation concern: a regular guild member's personal GitHub token is now used to query issue states on behalf of the guild, potentially exposing issues in private repos the owner never intended to share.
- The `GithubToken` table is queried using `github_user_id` as the filter, but the variable name `github_user_id` should be verified to ensure it refers to the authenticated user's GitHub ID and not their internal user ID — a mismatch would silently return no token rather than erroring.
- Changing `state=open` to `state=all` in the GitHub API fetch is correct for the stated goal, but doubles (or more) the number of issues returned per page without increasing `per_page`, so closed issues may crowd out open ones and the count display no longer conveys open-issue workload clearly.
- The `state` field fallback in the template (`issue.state || 'open'`) is now redundant since `state: string` is declared non-optional in `types.ts`; the fallback also silently hides any unexpected/missing state values as 'open', which could be misleading.
- Type fix and UI badge additions are clean and consistent with existing conventions in `TaskTree.vue`.

PR #688, opened 2026-07-01
Fix issues/tasks pane display bugs (#686)

### LINKED ISSUE #686: Fix issues/tasks pane: filter closed tasks, show issue titles, fix "all open" bug in tasks pane
## Problem

Two panes in the UI have bugs:

### Issues Pane
- Should filter out closed GitHub issues

### Tasks Pane
- Should show all issues **that have active tasks** — but currently shows everything as "open" (needs to show "closed" when the issue is closed)
- if all tasks are finalized and issue is closed, can show as closed by default
- Issue **titles** are not being displayed in the pane — they need to be shown. Currently only shows "#<num>"

### frontend/../github.ts 
 - does not implement paging
 - hard coded to either open or all  - should probably be all but may make sense to be parameter

## Expected Behaviour

- **Issues pane**: only show open GitHub issues (filter out `state: closed`).
- **Tasks pane**: show entries only for issues that have at least one active (non-terminal) task; show the issue/task title in each row.

jmelloy:
## Summary

- **Issues pane (Bug 1):** Changed the GitHub API fetch in `stores/github.ts` from `state=all` to `state=open` so only open issues are fetched and displayed in the issues tab.
- **Tasks pane filtering (Bug 2):** Added `hasActiveTask()` helper and `visibleNodes` computed in `sidebar/TaskTree.vue` that filters out issue groups where every task is in a terminal state (`done`, `failed`, or `cancelled`). The empty-state check was also updated to use the filtered set.
- **Task row titles (Bug 3):** Updated `sidebar/TaskTreeRow.vue` to fall back to `task.description` before `task.id` when `task.name` is absent, so tasks without a stored name show their human-readable description instead of a raw UUID.

## Test plan

- [ ] Open the Issues tab in the sidebar — confirm only open GitHub issues appear (no closed badge)
- [ ] Open the Tasks tab — confirm issue groups with only done/failed/cancelled tasks are hidden
- [ ] Open the Tasks tab — confirm issue groups with at least one active task remain visible
- [ ] Confirm task rows show a readable title (name or description) rather than a UUID when the task name field is empty

Closes #686

### DIFF
--- a/frontend/src/components/sidebar/TaskTree.vue
+++ b/frontend/src/components/sidebar/TaskTree.vue
@@ -6,7 +6,7 @@
     <template v-else>
       <!-- Issue / PR nodes -->
       <div
-        v-for="node in treeData!.nodes"
+        v-for="node in visibleNodes"
         :key="`${node.issue_repo}#${node.issue_number}`"
         class="tree-group"
       >
@@ -167,8 +167,27 @@ function countTasks(tasks: TaskTreeNode[]): number {
   return n
 }
 
+const TASK_TERMINAL_STATES = new Set(['done', 'failed', 'cancelled'])
+
+function hasActiveTask(tasks: TaskTreeNode[]): boolean {
+  for (const task of tasks) {
+    if (!TASK_TERMINAL_STATES.has(task.state)) return true
+    if (task.children.length > 0 && hasActiveTask(task.children)) return true
+  }
+  return false
+}
+
+const visibleNodes = computed(() => {
+  if (!treeData.value) return []
+  return treeData.value.nodes.filter((node) => hasActiveTask(node.tasks))
+})
+
 const isEmpty = computed(
-  () => !loading.value && treeData.value && treeData.value.nodes.length === 0 && treeData.value.ungrouped.length === 0,
+  () =>
+    !loading.value &&
+    treeData.value &&
+    visibleNodes.value.length === 0 &&
+    treeData.value.ungrouped.length === 0,
 )
 </script>
 
--- a/frontend/src/components/sidebar/TaskTree.vue
+++ b/frontend/src/components/sidebar/TaskTree.vue
@@ -36,15 +36,15 @@
       </div>
 
       <!-- Ungrouped tasks -->
-      <div v-if="treeData!.ungrouped.length" class="tree-group">
+      <div v-if="activeUngrouped.length" class="tree-group">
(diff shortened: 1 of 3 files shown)


frontend/src/components/sidebar/TaskTree.vue
```diff
@@ -167,8 +167,27 @@ function countTasks(tasks: TaskTreeNode[]): number {
...
+  }
+  return false
+}
+
+const visibleNodes = computed(() => {
+  if (!treeData.value) return []
+  return treeData.value.nodes.filter((node) => hasActiveTask(node.tasks))
+})
+
 const isEmpty = computed(
```
jmelloy, line 190: The `isEmpty` check now filters grouped nodes via `visibleNodes` but still uses the raw `treeData.value.ungrouped` list. If all ungrouped tasks are also in terminal states the pane will appear non-empty (no empty-state message) while showing nothing visible. Consider applying the same active-task filter to `ungrouped` tasks for consistency.

frontend/src/stores/github.ts
```diff
@@ -98,7 +98,7 @@ export const useGitHubStore = defineStore('github', () => {
       const allIssues = await Promise.all(
         reposToFetch.map(async (repoName) => {
           const res = await fetch(
-            `${GH_API}/repos/${repoName}/issues?state=all&per_page=100&sort=created&direction=desc`,
+            `${GH_API}/repos/${repoName}/issues?state=open&per_page=100&sort=created&direction=desc`,
```
jmelloy, line 101: The `per_page=100` limit silently truncates results for repos with more than 100 open issues. Consider adding pagination (following GitHub's `Link` header) or at least logging a warning when the response contains exactly 100 items, to avoid invisibly missing issues.

jmelloy (commented):
- Bug fixes are minimal and targeted; all three changes are straightforward and correct for their stated purposes.
- The `hasActiveTask` helper correctly recurses into children, but a task in a terminal state whose children are all also terminal will still correctly return false — logic is sound.
- The `isEmpty` computed no longer accounts for ungrouped tasks that might all be terminal; ungrouped tasks are still shown regardless of state, which may be an inconsistency worth addressing.
- Changing `state=all` to `state=open` is the right fix for the issues pane, but the 100-item `per_page` cap remains — if a repo has more than 100 open issues they will be silently truncated with no pagination.
- The fallback chain `task.name || task.description || task.id` is a safe and non-breaking improvement to task display.

PR #697, opened 2026-07-01
Workers advertise provider+tool; Foreman model selection (#691)

### LINKED ISSUE #691: Workers advertise provider+tool; Foreman uses this for model selection
## Parent issue
Part of #675 — Dynamic model selection for Foreman task routing.

## Problem
Workers don't currently advertise their configured provider. The Foreman has no way to know whether a worker is running `claude` via Anthropic, `pi` via Bedrock, etc. — so it can't filter the model catalog to models that worker actually supports.

## Work to do

1. **Worker registration payload** — extend the worker `[worker-online]` registration message (and/or the DB worker record) to include `provider` (e.g. `anthropic`, `bedrock`, `openai`) and `tool` (e.g. `claude`, `pi`, `codex`).
2. **DB schema** — add `provider` column to the `workers` table if not already present.
3. **Foreman reads provider** — when `assign_task` is called, look up the target worker's `provider` from the DB and use it to filter the model catalog.
4. **Runner propagation** — ensure `claude` and `codex` runners respect the `model` override field (currently only `pi` does); wire it through so the selected model is actually passed to the subprocess.

## Acceptance criteria
- A worker's provider is visible in `GET /api/workers`.
- When `assign_task` is called for a worker with `provider=bedrock`, model selection only considers Bedrock-compatible models.
- `claude` and `codex` runners pass the selected model name to the subprocess.

jmelloy:
## Summary

- **DB schema**: Added `provider` and `tool` columns to the `workers` table via Alembic migration `20260701_000001_add_provider_tool_to_workers`
- **Worker config**: New `provider` and `tool` fields in `Config`, readable from TOML or `PIONEER_PROVIDER`/`PIONEER_TOOL` env vars
- **Worker registration**: `worker-register` WS message now includes `provider` (e.g. `anthropic`, `bedrock`, `openai`) and `tool` (e.g. `claude`, `pi`, `codex`); `tool` is auto-derived from the first detected runner if not explicitly configured
- **Backend persistence**: `handle_worker_register` persists `provider` and `tool` to the DB on every registration
- **API**: `GET /api/workers` returns `provider` and `tool` fields automatically via the existing `row_to_dict` serializer
- **Foreman model selection**: `assign_task` looks up the target worker's `provider` from the DB and filters the model catalog:
  - If a model is explicitly specified, validates it exists in the catalog for that provider (returns error if incompatible)
  - If no model is specified and the worker has a provider, auto-selects the first available model from the catalog for that provider
- **Runner propagation**: `run_claude_auto` and `run_codex_auto` now accept a `model=` parameter and pass `--model <name>` to their respective subprocesses; `_execute_task` passes `task.get("model")` through

## Test plan

- [ ] `ruff check . && ruff format .` — clean (verified locally)
- [ ] Worker tests: 238 pass, 1 pre-existing failure in `test_no_tools_flag_gives_none` (confirmed pre-existing before this change)
- [ ] Acceptance criteria:
  - [ ] `GET /api/workers` returns `provider` field for a worker that registers with one
  - [ ] `assign_task` with `provider=bedrock` worker rejects models not in the Bedrock catalog
  - [ ] `claude` runner passes `--model <name>` when model is set on the task
  - [ ] `codex` runner passes `--model <name>` when model is set on the task

Closes #691

### DIFF
--- a/backend/alembic/versions/20260701_000001_add_provider_tool_to_workers.py
+++ b/backend/alembic/versions/20260701_000001_add_provider_tool_to_workers.py
@@ -1,0 +1,37 @@
+"""Add provider and tool columns to workers table.
+
+Revision ID: 20260701_000001_add_provider_tool_to_workers
+Revises: 20260701_000000_add_model_catalog
+Create Date: 2026-07-01
+
+provider: the AI provider this worker communicates with (e.g. 'anthropic', 'bedrock', 'openai').
+tool: the primary tool runner this worker is configured for (e.g. 'claude', 'pi', 'codex').
+"""
+
+from __future__ import annotations
+
+from collections.abc import Sequence
+
+import sqlalchemy as sa
+from alembic import op
+
+revision: str = "20260701_000001_add_provider_tool_to_workers"
+down_revision: str | Sequence[str] | None = "20260701_000000_add_model_catalog"
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+
+def upgrade() -> None:
+    op.add_column(
+        "workers",
+        sa.Column("provider", sa.Text(), nullable=True),
+    )
+    op.add_column(
+        "workers",
+        sa.Column("tool", sa.Text(), nullable=True),
+    )
+
+
+def downgrade() -> None:
+    op.drop_column("workers", "tool")
+    op.drop_column("workers", "provider")
--- a/backend/foreman/tools.py
+++ b/backend/foreman/tools.py
@@ -940,7 +940,11 @@ async def _exec_one_tool(guild_id: str, tu, user_id: str | None = None) -> dict:
                 repos: list[str] = inp.get("repos") or ([primary_repo] if primary_repo else [])
                 worker_result = await db.exec(
                     select(
-                        col(Worker.id), col(Worker.repos), col(Worker.org), col(Worker.tools)
+                        col(Worker.id),
+                        col(Worker.repos),
+                        col(Worker.org),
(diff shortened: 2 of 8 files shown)


jmelloy (commented):
### Code review

Found 2 issues:

1. **Stale `provider`/`tool` persist in DB when worker re-registers without them** — `repos` and `tools` are always written unconditionally to `update_vals` (line 616), so clearing them on re-registration works. But `provider` and `tool` are only written when truthy, so a worker that drops its provider from config will retain the old value in the DB. The foreman then continues filtering the model catalog for the wrong provider indefinitely.

Fix: write unconditionally — `update_vals["provider"] = provider or None` and `update_vals["tool"] = primary_tool or None`.

https://github.com/jmelloy/pioneer-square/blob/2c0f37ddc74d2f579ca69920682b0d410ced8e85/backend/ws_handlers.py#L616-L626

2. **Duplicate `--model` flag crashes codex when `codex_args` already contains `-m`/`--model`** — the task-provided model is injected before `codex_args` (line 70), but users can and do put `-m <model>` in their `codex_args` TOML config (the existing test `test_load_codex_args_from_env` uses `-m "gpt-5.4"`). Codex exits with `error: the argument '--model' cannot be used multiple times` when both are present.

Fix: either strip `-m`/`--model` from `codex_args` before building the command, or place the task model after `codex_args` so it takes precedence (though duplicate-flag rejection still applies regardless of order).

https://github.com/jmelloy/pioneer-square/blob/2c0f37ddc74d2f579ca69920682b0d410ced8e85/worker/pioneer_worker/codex_runner.py#L68-L72

PR #716, opened 2026-07-03
Issue detail: Task Log tab (#715)

### LINKED ISSUE #715: Issue detail: add "Task Log" tab showing Foreman calls and task timeline
## Summary

On the issue detail page, add a dedicated **Task Log** tab that surfaces the Foreman tool calls and task lifecycle events associated with that issue.

## Motivation

When viewing an issue it's useful to see the full history of what the Foreman did: which tasks were created, what workers were assigned, what PRs were opened, and the agent log lines — all in one place without switching to a separate tasks pane.

## Proposed UI

- New tab alongside **Details** / **Comments** (etc.) on the issue detail panel, labelled **Task Log** (or **Foreman Log**).
- Content: ordered list of tasks linked to this issue (`issue_number` + `issue_repo` match), each expandable to show:
  - Task name, phase, state, worker, timestamps
  - Last N log lines from the task (already available via the tasks API)
  - Link to the associated PR if one was opened
- Tasks should be sorted by creation time (oldest first) so it reads as a timeline.

## Backend

- The existing `/api/guild/{guild_id}/tasks` endpoint (or equivalent) already returns tasks with `issue_number`/`issue_repo` fields — filter by those on the client, or add a query param to the endpoint.
- Task log lines are available via the task-status API; fetch them when the user expands a task row.

## Acceptance criteria

- [ ] "Task Log" tab appears on the issue detail page.
- [ ] Tab lists all tasks associated with the issue in chronological order.
- [ ] Each task row shows name, phase, state, worker ID, and PR link (if any).
- [ ] Expanding a task row fetches and displays the last 50 log lines.
- [ ] Tab shows a friendly empty state when no tasks exist for the issue.

jmelloy:
## Summary

- Adds a **Task Log** tab to the issue detail panel (`IssueViewer.vue`) alongside the existing Details and Comments tabs.
- The tab fetches all Foreman tasks linked to the current issue (by `issue_number` + `issue_repo`) from the backend and displays them in chronological order (oldest first).
- Each task row shows: name, phase badge, state badge (colour-coded), worker ID, PR link (if set), and relative timestamp.
- Clicking a row expands it to fetch and display the last 50 stored log lines in a scrollable `<pre>` block.
- A friendly empty state is shown when no tasks exist for the issue.
- Backend: `GET /guilds/{guild_id}/tasks` now accepts optional `?issue_number=N&issue_repo=owner/repo` query params for efficient server-side filtering (backward-compatible — existing callers without params are unaffected).

## Test plan
- [ ] Open an issue detail panel; verify the **Task Log** tab appears next to Details and Comments.
- [ ] Click Task Log; if tasks exist for the issue they appear sorted oldest-first with state badges.
- [ ] Expand a task row; confirm log lines load in a scrollable code block.
- [ ] Open an issue with no linked tasks; confirm "No tasks found for this issue." empty state.
- [ ] Click the ↻ refresh button; verify it re-fetches tasks.
- [ ] Confirm existing Details and Comments tabs are unaffected.
- [ ] Backend: `GET /guilds/{gid}/tasks?issue_number=1&issue_repo=owner/repo` returns only matching tasks.

Closes #715

### DIFF
--- a/backend/routes/tasks.py
+++ b/backend/routes/tasks.py
@@ -107,12 +107,12 @@ async def list_guild_tasks(
         )
         .where(col(Task.guild_id) == guild_pk, live_tasks_filter())
         .order_by(col(Task.created_at).desc())
-        .limit(100)
     )
     if issue_number is not None:
         stmt = stmt.where(col(Task.issue_number) == issue_number)
     if issue_repo is not None:
         stmt = stmt.where(col(Task.issue_repo) == issue_repo)
+    stmt = stmt.limit(100)
     result = await db.exec(stmt)
     return [dict(r._mapping) for r in result.all()]
 
--- a/backend/routes/tasks.py
+++ b/backend/routes/tasks.py
@@ -75,14 +75,20 @@ def _resolve_finalize_deleted_at(body: FinalizeBody | None) -> datetime:
 @router.get("/guilds/{guild_id}/tasks")
 async def list_guild_tasks(
     guild_id: str,
+    issue_number: int | None = None,
+    issue_repo: str | None = None,
     github_user_id: str = Depends(require_member()),
     db: AsyncSession = Depends(get_db_dep),
 ):
-    """List all tasks for a guild, most recent first."""
+    """List all tasks for a guild, most recent first.
+
+    Optionally filter by ``issue_number`` and/or ``issue_repo`` to return only
+    tasks linked to a specific GitHub issue.
+    """
     guild_pk = await get_guild_pk(db, guild_id)
     if guild_pk is None:
         raise HTTPException(status_code=404, detail="Guild not found")
-    result = await db.exec(
+    stmt = (
         select(
             col(Task.id),
             col(Task.worker_id),
@@ -103,6 +109,11 @@ async def list_guild_tasks(
         .order_by(col(Task.created_at).desc())
         .limit(100)
     )
+    if issue_number is not None:
+        stmt = stmt.where(col(Task.issue_number) == issue_number)
+    if issue_repo is not None:
+        stmt = stmt.where(col(Task.issue_repo) == issue_repo)
+    result = await db.exec(stmt)
     return [dict(r._mapping) for r in result.all()]
 
 
--- a/frontend/src/components/IssueViewer.vue
+++ b/frontend/src/components/IssueViewer.vue
@@ -66,6 +66,13 @@
         >
           Comments ({{ comments.length }})
         </button>
+        <button
+          class="issue-tab-btn"
+          :class="{ active: activeTab === 'tasklog' }"
+          @click="onTaskLogTab"
+        >
+          Task Log{{ issueTasks.length > 0 ? ` (${issueTasks.length})` : '' }}
+        </button>
       </div>
 
       <!-- Scrollable tab content -->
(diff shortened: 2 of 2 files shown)


backend/routes/tasks.py
```diff
@@ -103,6 +109,11 @@ async def list_guild_tasks(
         .order_by(col(Task.created_at).desc())
```
jmelloy, line 109: The `.where()` filters are appended after `.limit(100)` is already part of the statement. SQLModel/SQLAlchemy will apply LIMIT before WHERE, meaning if the issue's tasks fall outside the first 100 rows of the unfiltered guild query they will be silently omitted. Move the filter conditions before `.limit()`, or restructure the query so filtering happens first.

frontend/src/components/IssueViewer.vue
```diff
@@ -175,13 +230,79 @@ const props = defineProps<{
...
 
-const activeTab = ref<'details' | 'comments'>('details')
+const activeTab = ref<'details' | 'comments' | 'tasklog'>('details')
+
+// Task Log tab state
+const issueTasks = ref<Task[]>([])
+const tasklogLoading = ref(false)
+const expandedTaskIds = ref<Set<string>>(new Set())
+const taskLogsMap = ref<Record<string, LogEntry[]>>({})
+const taskLogsFetching = ref<Set<string>>(new Set())
```
jmelloy, line 248: Using `ref<Set<string>>(new Set())` and then calling `.add()` / `.delete()` on the inner Set mutates it in-place without replacing the ref value, so Vue's reactivity system will not trigger re-renders. Use a reactive Set via `reactive(new Set())` or replace the ref value on each change (e.g. `expandedTaskIds.value = new Set(expandedTaskIds.value)` after mutation). Same issue applies to `taskLogsFetching`.

frontend/src/components/IssueViewer.vue
```diff
@@ -175,13 +230,79 @@ const props = defineProps<{
...
+  try {
+    const issueRepo = `${props.owner}/${props.repo}`
+    const raw = await api<Task[]>(
+      `/guilds/${guildId}/tasks?issue_number=${props.issueNumber}&issue_repo=${encodeURIComponent(issueRepo)}`,
+    )
+    // Sort oldest first for chronological display
+    issueTasks.value = [...raw].sort((a, b) => {
+      const ta = a.created_at ? new Date(a.created_at).getTime() : 0
+      const tb = b.created_at ? new Date(b.created_at).getTime() : 0
+      return ta - tb
```
jmelloy, line 263: The `issue_repo` value is URL-encoded, but `issue_number` is interpolated directly into the query string without encoding. While a number is safe here, combining both params via string interpolation is fragile — prefer `URLSearchParams` to construct the query string robustly and consistently.

frontend/src/components/IssueViewer.vue
```diff
@@ -175,13 +230,79 @@ const props = defineProps<{
...
+    })
+  } catch (e) {
+    console.error('Failed to fetch issue tasks', e)
+  } finally {
+    tasklogLoading.value = false
+  }
+}
```
jmelloy, line 270: On fetch failure the error is only logged to the console; the UI silently shows 'No tasks found for this issue.' with no distinction between 'no tasks exist' and 'fetch failed'. Add a dedicated `tasklogError` ref and display an error message to the user.

jmelloy (commented):
- The backend filtering applies `WHERE` clauses **after** `.limit(100)` has already been baked into the statement, meaning the limit is applied before filtering — tasks could be silently dropped when an issue has tasks beyond position 100 in the unfiltered result set.
- Reactivity bug: `expandedTaskIds` and `taskLogsFetching` use `ref<Set<...>>` but mutate the Set in-place (`.add()`, `.delete()`), which Vue 3 will not detect as reactive changes — the UI will not update correctly when expanding/collapsing tasks.
- The `issue_repo` query parameter is passed to the backend unsanitised; a malicious value containing path separators or special characters could interfere with query construction, though the ORM parameterisation mitigates SQL injection risk.
- No error state is surfaced to the user in `loadIssueTasks` — failures are silently swallowed and the list stays empty with no indication that a network error occurred.
- The tab shows a count (`Task Log (N)`) before tasks are loaded (count is 0), but does not eagerly load tasks, so the count is always 0 until the tab is clicked — this is misleading UX.

PR #718, opened 2026-07-03
Discord integration: Per-issue/PR threads Phase 2 (#709)

### LINKED ISSUE #709: Discord integration: Per-issue/PR threads (Phase 2)
## Parent issue
Part of #652 (builds on #708)

## Overview
Upgrade from a single flat channel feed to per-issue/PR Discord threads. Each GitHub issue or PR gets its own Discord thread; all Foreman activity related to that issue/PR (task updates, CI results, review comments) posts into that thread rather than the top-level channel.

## Requirements

- When a PR is opened or an issue is picked up, create a Discord thread in the notification channel named `#NNN: <title>` (e.g. `#708: Discord webhook notifier`)
- Store the thread ID alongside the issue/PR number (e.g. in the DB `tasks` table or a new `discord_threads` table)
- Subsequent notifications for that issue/PR number post into the existing thread, not the channel root
- Thread should include a header message with: issue/PR title, URL, assignee, labels
- When the issue/PR is closed or merged, post a final summary message and archive/lock the thread
- Handle the case where no thread exists yet (fall back to creating one lazily)

## Dependencies
- Requires Phase 1 (#708) webhook notifier to be merged
- Requires a Discord bot token with `MANAGE_THREADS` + `SEND_MESSAGES` permissions (webhook-only is not sufficient for thread creation — document the upgrade path)

## Acceptance criteria
- [ ] Opening a PR creates a new Discord thread named after the PR
- [ ] All subsequent task/CI events for that PR post inside the thread
- [ ] Closing/merging posts a summary and marks the thread as resolved
- [ ] Thread IDs are persisted so restarts don't create duplicates
- [ ] Graceful fallback when the bot token is missing (reverts to Phase 1 flat-channel behaviour)

jmelloy:
## Summary

- **`discord_threads` table** — new SQLModel (`DiscordThread`) mapping `(issue_repo, issue_number)` → Discord `thread_id`, with a unique index to prevent duplicates across restarts. Alembic migration: `20260703_000000_add_discord_threads`.
- **`notify_event()` in `discord_notifier.py`** — thread-aware public API. When `DISCORD_BOT_TOKEN` + `DISCORD_CHANNEL_ID` are set and `issue_repo`/`issue_number` are provided, it lazily creates or reuses a Discord thread (type 11, GUILD_PUBLIC_THREAD) in the configured channel, posts a rich embed, and archives the thread on close/merge. Falls back to the flat webhook when the bot token is absent or thread creation fails.
- **`webhooks.py`** — PR-opened now creates a named thread (`#NNN: <title>`) and includes assignee/labels embed fields; PR-closed/merged posts a summary and archives; CI events route into the existing PR thread.
- **`ws_handlers.py`** — task-complete routes into the PR thread when `pr_url` is available.
- **33 unit tests** covering thread creation, reuse, routing, fallback (no bot token, creation failure, no coords), close/archive, and error suppression.
- **`docs/discord-bot-setup.md`** — documents required bot permissions (`SEND_MESSAGES`, `MANAGE_THREADS`) and env vars.

## Required env vars (Phase 2)

```
DISCORD_BOT_TOKEN=<bot token>
DISCORD_CHANNEL_ID=<text channel ID>
```

When only `DISCORD_WEBHOOK_URL` is set, Phase 1 flat-webhook behaviour is unchanged.

## Test plan

- [x] `cd backend && python -m pytest tests/test_discord_notifier.py -v` — 33 tests, all passing
- [ ] Integration: set `DISCORD_BOT_TOKEN` + `DISCORD_CHANNEL_ID`, open a test PR via GitHub webhook → verify thread created in channel, follow-up events appear in thread, close PR → thread archived

Closes #709

### DIFF
--- a/backend/alembic/versions/20260703_000000_add_discord_threads.py
+++ b/backend/alembic/versions/20260703_000000_add_discord_threads.py
@@ -1,0 +1,45 @@
+"""Add discord_threads table for per-issue/PR Discord thread persistence.
+
+Revision ID: 20260703_000000_add_discord_threads
+Revises: 20260701_000003_add_issue_state_to_tasks
+Create Date: 2026-07-03
+
+Maps GitHub issue/PR coordinates (issue_repo, issue_number) to a Discord
+thread channel ID so the notifier can route subsequent events into the same
+thread without re-creating it on restart.
+"""
+
+from __future__ import annotations
+
+from collections.abc import Sequence
+
+import sqlalchemy as sa
+from alembic import op
+
+revision: str = "20260703_000000_add_discord_threads"
+down_revision: str | Sequence[str] | None = "20260701_000003_add_issue_state_to_tasks"
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+
+def upgrade() -> None:
+    op.create_table(
+        "discord_threads",
+        sa.Column("id", sa.Integer(), nullable=False),
+        sa.Column("issue_repo", sa.Text(), nullable=False),
+        sa.Column("issue_number", sa.Integer(), nullable=False),
+        sa.Column("thread_id", sa.Text(), nullable=False),
+        sa.Column("created_at", sa.DateTime(timezone=True), nullable=False),
+        sa.PrimaryKeyConstraint("id"),
+    )
+    op.create_index(
+        "uq_discord_threads_repo_number",
+        "discord_threads",
+        ["issue_repo", "issue_number"],
+        unique=True,
+    )
+
+
+def downgrade() -> None:
+    op.drop_index("uq_discord_threads_repo_number", table_name="discord_threads")
+    op.drop_table("discord_threads")
--- a/backend/discord_notifier.py
+++ b/backend/discord_notifier.py
@@ -13,7 +13,7 @@ Phase 2 — per-PR/issue threads
     Both tokens absent → silent no-op.  HTTP and DB failures are always logged
     at WARNING level and never propagate.
 
-Required bot permissions for Phase 2: ``SEND_MESSAGES``, ``MANAGE_THREADS``.
+Required bot permissions for Phase 2: ``SEND_MESSAGES``, ``CREATE_PUBLIC_THREADS``, ``MANAGE_THREADS``.
 
 Usage::
 
@@ -270,7 +270,9 @@ async def _ensure_thread(issue_repo: str, issue_number: int, thread_name: str) -
         return None
 
     await _save_thread(issue_repo, issue_number, new_thread_id)
-    return new_thread_id
+    # Re-fetch after save: on_conflict_do_nothing means a concurrent creator wins;
+    # re-fetching ensures both callers use the same persisted thread_id.
+    return await _lookup_thread(issue_repo, issue_number) or new_thread_id
 
 
 async def _post_to_thread(
--- a/backend/discord_notifier.py
+++ b/backend/discord_notifier.py
@@ -1,35 +1,52 @@
-"""Fire-and-forget Discord webhook notifier.
+"""Discord notification helpers: flat webhook (Phase 1) and per-PR/issue threads (Phase 2).
+
+Phase 1 — flat webhook
+    Set ``DISCORD_WEBHOOK_URL`` in the environment to post embeds to a single channel.
+
+Phase 2 — per-PR/issue threads
+    Set ``DISCORD_BOT_TOKEN`` **and** ``DISCORD_CHANNEL_ID`` to create/reuse one
+    Discord thread per GitHub PR or issue.  The mapping is persisted in the
+    ``discord_threads`` DB table so restarts never duplicate threads.
+
+    When ``DISCORD_BOT_TOKEN`` is absent, every thread-aware call transparently
+    falls back to the flat-webhook behaviour if ``DISCORD_WEBHOOK_URL`` is set.
+    Both tokens absent → silent no-op.  HTTP and DB failures are always logged
+    at WARNING level and never propagate.
+
+Required bot permissions for Phase 2: ``SEND_MESSAGES``, ``MANAGE_THREADS``.
 
 Usage::
 
-    import discord_notifier
+    # Legacy flat webhook (unchanged):
     await discord_notifier.notify("task-complete", "Task done", "t-abc finished")
 
-Set ``DISCORD_WEBHOOK_URL`` in the environment to enable.  When the variable is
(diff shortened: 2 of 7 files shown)


backend/discord_notifier.py
```diff
@@ -86,3 +123,211 @@ async def notify(
...
+
+    Returns None when bot token or channel ID are not configured, or on API error.
+    """
+    existing = await _lookup_thread(issue_repo, issue_number)
+    if existing:
+        return existing
+
+    channel = _channel_id()
+    if not channel:
+        return None
```
jmelloy, line 239: Creating a GUILD_PUBLIC_THREAD (type 11) via `POST /channels/{channel}/threads` without a `message_id` only works for forum/media channels in Discord's API. For standard text channels, public threads must be created from an existing message (`POST /channels/{channel_id}/messages/{message_id}/threads`). This will return a 400 error on normal text channels. Consider sending a starter message first, or using `type: 12` (GUILD_PUBLIC_THREAD without message) which is only valid in forum channels.

backend/discord_notifier.py
```diff
@@ -86,3 +123,211 @@ async def notify(
...
+
+
+async def _lookup_thread(issue_repo: str, issue_number: int) -> str | None:
+    """Return the persisted Discord thread_id for a PR/issue, or None."""
+    try:
+        from database import AsyncSessionLocal  # noqa: PLC0415 — lazy to avoid circular import
+        from models import DiscordThread  # noqa: PLC0415
+        from sqlmodel import col, select  # noqa: PLC0415
+
+        db = AsyncSessionLocal()
```
jmelloy: The session is opened with `db = AsyncSessionLocal()` rather than `async with AsyncSessionLocal() as db:`. While the `finally` block closes it, this pattern bypasses the context manager's rollback-on-exception semantics. Prefer `async with AsyncSessionLocal() as db:` for both `_lookup_thread` and `_save_thread`.

backend/discord_notifier.py
```diff
@@ -86,3 +123,211 @@ async def notify(
...
+            "discord: thread DB save failed repo=%s number=%s thread=%s",
+            issue_repo,
+            issue_number,
+            thread_id,
+            exc_info=True,
+        )
+
+
+async def _ensure_thread(issue_repo: str, issue_number: int, thread_name: str) -> str | None:
+    """Return the Discord thread_id for a PR/issue, creating it if necessary.
```
jmelloy, line 225: There is a TOCTOU race: two simultaneous `pr-opened` events can both find no existing thread, both call the Discord API to create a thread, and end up with two orphaned threads (only one gets persisted due to the IntegrityError catch). Consider using a distributed lock or an `INSERT ... ON CONFLICT DO NOTHING ... RETURNING` upsert pattern to make this atomic.

backend/tests/test_discord_notifier.py
```diff
@@ -33,10 +39,12 @@ def reset_client():
     discord_notifier._client = None
 
 
-def _make_mock_client(status_code: int = 204, side_effect=None):
-    """Return a mock httpx.AsyncClient with a stubbed post() coroutine."""
+def _make_mock_client(status_code: int = 204, side_effect=None, json_response: dict | None = None):
+    """Return a mock httpx.AsyncClient with stubbed HTTP methods."""
     mock_response = MagicMock(spec=httpx.Response)
```
jmelloy, line 45: `mock_response.content` is hard-coded to `b'{"id": "555666777888"}'` whenever `json_response` is truthy, regardless of what `json_response` actually contains. This means `resp.content` and `resp.json()` return inconsistent data for any test passing a different dict. Use `json.dumps(json_response).encode()` to keep them in sync.

backend/discord_notifier.py
```diff
@@ -86,3 +123,211 @@ async def notify(
...
+        logger.warning(
+            "discord: thread DB lookup failed repo=%s number=%s",
+            issue_repo,
+            issue_number,
+            exc_info=True,
+        )
+        return None
+
+
+async def _save_thread(issue_repo: str, issue_number: int, thread_id: str) -> None:
```
jmelloy, line 194: `_save_thread` only catches `IntegrityError` inside the inner try block, but the `db.add()` call can also raise other SQLAlchemy errors (e.g., `OperationalError` on connection loss). The outer `except Exception` will catch these but also swallow the session without rollback since the `IntegrityError` branch calls `rollback()` but other exceptions skip it. Move `rollback()` to a broader `except Exception` before `finally`.

jmelloy (commented):
- Thread creation uses an incorrect Discord API endpoint: creating a thread in a channel without a message requires `/channels/{channel_id}/threads` with `type=11`, but public threads in a text channel are actually created from messages; standalone thread creation (without a starter message) requires `CREATE_PUBLIC_THREADS` permission and the endpoint payload is valid only for forum channels or with `auto_archive_duration` — this may silently fail on standard text channels.
- The `_lookup_thread` and `_save_thread` helpers open a raw `AsyncSessionLocal()` session directly rather than using `async with AsyncSessionLocal() as db:`, which risks session leaks if an exception occurs between session open and the `finally` block (though `finally` does close it, the pattern is fragile and non-idiomatic).
- Race condition exists between `_lookup_thread` returning `None` and `_save_thread` writing the new thread ID: two concurrent PR-opened events could both call `_bot_request` to create a thread, resulting in two Discord threads; the `IntegrityError` catch in `_save_thread` handles the DB side but the duplicate Discord thread is already created and the second one is orphaned.
- The `_make_mock_client` helper unconditionally sets `mock_response.content = b'{"id": "555666777888"}'` when `json_response` is provided, but this hard-codes the THREAD_ID string rather than serialising `json_response`, so tests with different `json_response` values will have mismatched `.content` vs `.json()` return values.
- Documentation correctly describes required permissions but omits `CREATE_PUBLIC_THREADS` which is the actual permission needed to create public threads via the bot API; `MANAGE_THREADS` is needed for archiving but not listed as distinct from general thread management.

backend/alembic/versions/20260703_000000_add_discord_threads.py
```diff
@@ -0,0 +1,45 @@
...
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+
+def upgrade() -> None:
+    op.create_table(
+        "discord_threads",
+        sa.Column("id", sa.Integer(), nullable=False),
+        sa.Column("issue_repo", sa.Text(), nullable=False),
+        sa.Column("issue_number", sa.Integer(), nullable=False),
```
jmelloy, line 30: In the absence of an issue number, it could use a PR number

PR #724, opened 2026-07-04
Discord: Foreman chat as thread + user identity (#717)

### LINKED ISSUE #717: Discord: Foreman chat as thread + Pioneer Square ↔ Discord user identity
## Overview

Two related enhancements to the Discord integration (building on Phase 1 #708 and Phase 2 #709):

### 1. Foreman chat as a Discord thread

The Foreman's conversational output (user ↔ Foreman messages) should appear in Discord as a thread, similar to how PR/issue events get their own threads. This gives operators visibility into what the Foreman is doing from within Discord.

- Each "conversation session" (or perhaps each day) gets a thread in the notification channel
- Foreman messages sent to the user are mirrored into the thread
- User replies in Discord could optionally be relayed back to the Foreman (stretch goal)

### 2. Pioneer Square ↔ Discord user identity linking

Discord usernames/IDs should be connectable to Pioneer Square users so that:

- Notifications can @-mention the right Discord user when a task they own completes or needs attention
- PR/issue threads can tag the assignee by their Discord handle
- The "opened by @jmelloy" attribution on GitHub maps to a Discord mention

**Implementation ideas:**
- A `discord_users` table mapping `github_login` → `discord_user_id`
- A settings UI or slash command (`/link-account`) for users to self-register their Discord identity
- Foreman system prompt updated to include the Discord ID of the operator when known

## Acceptance criteria
- [ ] Foreman chat messages appear in a Discord thread (one thread per session or per day)
- [ ] A `discord_users` table (or equivalent) links GitHub logins to Discord user IDs
- [ ] Notifications @-mention the relevant Discord user when a mapping exists
- [ ] Graceful fallback when no mapping is found (no crash, no mention)

## Related
- #652 Design: Pioneer Square ↔ Discord / Slack integration
- #708 Discord webhook notifier (Phase 1)
- #709 Discord per-issue/PR threads (Phase 2)

jmelloy:
Closes #717

## Summary
- Mirrors Foreman → user chat narration into a per-guild, per-UTC-day Discord thread (`discord_foreman_threads` table), reusing the Phase 2 bot token/channel.
- Adds a `discord_users` table mapping `github_login` → `discord_user_id`, with full CRUD REST endpoints (`GET /api/discord-users`, `GET/PUT/DELETE /api/discord-users/{github_login}`).
- Wires `discord_notifier.mention_or_login()` into the PR-assignee notification path so GitHub logins resolve to real `<@id>` Discord mentions when a mapping exists, falling back to a plain `@login` string (never raises) when it doesn't.

## Details
- `backend/models.py`: `DiscordForemanThread` and `DiscordUser` SQLModel tables.
- `backend/alembic/versions/20260704_000001_add_discord_users.py` and `..._000002_add_discord_foreman_threads.py`: migrations, chained cleanly onto the current single head.
- `backend/discord_notifier.py`: `notify_foreman_chat()` (create/reuse per-session thread + post) and `mention_or_login()` (lookup + graceful fallback), documented alongside the existing Phase 1/2 helpers.
- `backend/foreman/runner.py`: every Foreman → user text line now routes through `_emit_foreman_chat()`, which broadcasts over the existing WS channel and fires-and-forgets the Discord mirror via `spawn()` so a Discord outage never blocks chat.
- `backend/routes/discord_users.py` (+ registered in `main.py`): auth-gated CRUD for the mapping table.

## Test plan
- [x] `ruff check .` / `ruff format --check .` — clean
- [x] `pytest backend/tests/test_discord_notifier.py` — 41 passed (mocked HTTP/DB, no infra needed)
- [ ] `pytest backend/tests/test_discord_users_api.py` — requires the `postgres-test` container (`docker compose up -d postgres-test`), not available in this sandbox; reviewed manually, endpoints follow the same patterns as existing DB-backed routes.

### DIFF
--- a/backend/discord_notifier.py
+++ b/backend/discord_notifier.py
@@ -1,4 +1,5 @@
-"""Discord notification helpers: flat webhook (Phase 1) and per-PR/issue threads (Phase 2).
+"""Discord notification helpers: flat webhook (Phase 1), per-PR/issue threads (Phase 2),
+and Foreman chat threads + user mentions (Phase 3).
 
 Phase 1 — flat webhook
     Set ``DISCORD_WEBHOOK_URL`` in the environment to post embeds to a single channel.
@@ -13,7 +14,17 @@ Phase 2 — per-PR/issue threads
     Both tokens absent → silent no-op.  HTTP and DB failures are always logged
     at WARNING level and never propagate.
 
-Required bot permissions for Phase 2: ``SEND_MESSAGES``, ``CREATE_PUBLIC_THREADS``, ``MANAGE_THREADS``.
+Phase 3 — Foreman chat threads + user mentions
+    ``notify_foreman_chat`` mirrors Foreman → user chat narration into one
+    Discord thread per guild per UTC day (mapping persisted in
+    ``discord_foreman_threads``), reusing the same bot token/channel as Phase 2.
+
+    ``mention_or_login`` resolves a GitHub login to a real ``<@id>`` Discord
+    mention via the ``discord_users`` table (populated through the
+    ``/api/discord-users`` REST endpoints). Falls back to a plain ``@login``
+    string when no mapping exists — never raises, never blocks a notification.
+
+Required bot permissions for Phase 2/3: ``SEND_MESSAGES``, ``CREATE_PUBLIC_THREADS``, ``MANAGE_THREADS``.
 
 Usage::
 
@@ -31,6 +42,12 @@ Usage::
         thread_name="#42: Fix the bug",
         header_fields={"Assignee": "@alice", "Labels": "bug"},
     )
+
+    # Foreman chat mirrored into a per-day thread:
+    await discord_notifier.notify_foreman_chat(guild_id, "Spun up worker w-abc for t-123.")
+
+    # Resolve a GitHub login to a Discord mention (or a plain @login fallback):
+    mention = await discord_notifier.mention_or_login("alice")
 """
 
 from __future__ import annotations
@@ -221,12 +238,44 @@ async def _save_thread(issue_repo: str, issue_number: int, thread_id: str) -> No
         )
 
 
-async def _ensure_thread(issue_repo: str, issue_number: int, thread_name: str) -> str | None:
-    """Return the Discord thread_id for a PR/issue, creating it if necessary.
+async def _create_thread_in_channel(channel: str, thread_name: str) -> str | None:
+    """Create a new thread in *channel*, anchored by a starter message.
 
     For standard text channels, thread creation requires a starter message first:
     POST /channels/{channel}/messages → get message_id
     POST /channels/{channel}/messages/{message_id}/threads → get thread_id
+
+    Returns the new thread_id, or None on API error.
+    """
+    starter = await _bot_request(
+        "post",
+        f"/channels/{channel}/messages",
+        {"content": thread_name[:100]},
+    )
+    if not starter:
+        return None
+    message_id = starter.get("id")
+    if not message_id:
+        logger.warning("discord: starter message returned no id for channel=%s", channel)
+        return None
+
+    data = await _bot_request(
+        "post",
+        f"/channels/{channel}/messages/{message_id}/threads",
+        {"name": thread_name[:100]},
+    )
+    if not data:
+        return None
+
+    new_thread_id = data.get("id")
+    if not new_thread_id:
+        logger.warning("discord: thread creation returned no id for channel=%s", channel)
+        return None
+    return new_thread_id
+
+
+async def _ensure_thread(issue_repo: str, issue_number: int, thread_name: str) -> str | None:
+    """Return the Discord thread_id for a PR/issue, creating it if necessary.
 
     Returns None when bot token or channel ID are not configured, or on API error.
     """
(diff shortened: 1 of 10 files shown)


backend/discord_notifier.py
```diff
@@ -299,6 +321,146 @@ async def archive_thread(thread_id: str) -> None:
...
+                    created_at=datetime.now(UTC),
+                )
+                .on_conflict_do_nothing()
+            )
+            await db.execute(stmt)
+            await db.commit()
+    except Exception:
+        logger.warning(
+            "discord: foreman thread DB save failed guild=%s session=%s thread=%s",
+            guild_id,
```
jmelloy, line 404: `_save_foreman_thread` imports and uses `sqlalchemy.dialects.postgresql.insert`, which will fail at import time when running against SQLite (CI/tests). Consider using a dialect-agnostic upsert or a SELECT-then-INSERT pattern guarded by the unique index, consistent with how `_save_thread` works elsewhere in this file.

backend/discord_notifier.py
```diff
@@ -299,6 +321,146 @@ async def archive_thread(thread_id: str) -> None:
...
+
+    One thread is created per ``(guild_id, session_key)`` pair — *session_key*
+    defaults to the current UTC date, so a whole day's conversation lands in
+    the same thread. Silent no-op when the bot token or channel are not
+    configured, or when *content* is blank. Never raises.
+    """
+    if not content or not content.strip():
+        return
+    if not (_bot_token() and _channel_id()):
+        return
```
jmelloy, line 443: `_ensure_foreman_thread` calls `_lookup_foreman_thread` a second time after `_save_foreman_thread` succeeds. Since `new_thread_id` is already in scope, just `return new_thread_id` directly — the extra DB round-trip is unnecessary and adds latency on every first-message-of-the-day path.

backend/routes/discord_users.py
```diff
@@ -0,0 +1,103 @@
...
+    row = result.one_or_none()
+    if not row:
+        raise HTTPException(status_code=404, detail="No Discord mapping for this GitHub login")
+    return _serialize(row)
+
+
+@router.put("/api/discord-users/{github_login}")
+async def upsert_discord_user(
+    github_login: str,
+    data: DiscordUserUpsert,
```
jmelloy, line 78: Any authenticated user can overwrite or delete the Discord mapping of *any other* GitHub login. At minimum, consider restricting writes to `github_login == github_user_id` (the caller's own login) or gating mutating operations behind an admin check, to prevent one user from hijacking another's mention target.

backend/routes/discord_users.py
```diff
@@ -0,0 +1,103 @@
...
+@router.get("/api/discord-users/{github_login}")
+async def get_discord_user(
+    github_login: str,
+    github_user_id: str = Depends(require_user),
+    db: AsyncSession = Depends(get_db_dep),
+):
+    """Return the Discord mapping for a single GitHub login, or 404."""
+    result = await db.exec(
+        select(DiscordUser).where(col(DiscordUser.github_login) == github_login.lower())
+    )
```
jmelloy, line 68: `pg_insert` is imported from `sqlalchemy.dialects.postgresql`, making this route PostgreSQL-only and breaking SQLite-backed test environments. The upsert should either use a dialect-agnostic approach or this PostgreSQL requirement should be explicitly documented and enforced at startup.

backend/routes/discord_users.py
```diff
@@ -0,0 +1,103 @@
...
+    return _serialize(result.one())
+
+
+@router.delete("/api/discord-users/{github_login}")
+async def delete_discord_user(
+    github_login: str,
+    github_user_id: str = Depends(require_user),
+    db: AsyncSession = Depends(get_db_dep),
+):
+    """Remove the Discord mapping for a GitHub login."""
```
jmelloy: `DELETE` returns 200 even when the row did not exist (delete is a no-op on a missing row). Consider returning 404 when no row was found, or at least documenting that this endpoint is idempotent, so callers can reason about the response correctly.

jmelloy (commented):
- The PostgreSQL-specific `insert(...).on_conflict_do_nothing()` in `_save_foreman_thread` and the `pg_insert` in `discord_users.py` hard-code a PostgreSQL dependency, breaking SQLite-based test environments and any non-Postgres deployment without a clear abstraction boundary.
- `_save_foreman_thread` uses a raw SQLAlchemy `insert` (PostgreSQL dialect) but the session is opened via SQLModel's `AsyncSessionLocal`; calling `db.execute(stmt)` instead of `db.exec(stmt)` is inconsistent with the rest of the codebase, but more importantly the PostgreSQL-only dialect import will fail on SQLite.
- The `/api/discord-users` endpoints allow any authenticated user to create, update, or delete mappings for *any* GitHub login, including other users' mappings — there is no ownership or authorization check beyond being logged in.
- `_ensure_foreman_thread` does a second DB round-trip (`_lookup_foreman_thread`) after `_save_foreman_thread` on the happy path; it should just return `new_thread_id` directly since the value is already known, avoiding an unnecessary query.
- Test coverage for the REST endpoints does not include validation edge cases (e.g., empty string or excessively long `discord_user_id`) and there are no tests for the `_save_foreman_thread` / `_ensure_foreman_thread` DB layer with the PostgreSQL upsert path.

PR #732, opened 2026-07-04
Fix frontend: worker list shows duplicate entries (#731)

### LINKED ISSUE #731: Frontend: worker list shows duplicate entries
## Bug

The workers pane/list in the frontend UI is displaying duplicate worker entries.

## Expected behavior

Each worker should appear exactly once in the list.

## Likely causes

- The data source (API response or WebSocket state) may be appending workers on each update instead of merging/deduplicating by worker ID.
- The component may be rendering from multiple state slices (e.g. both a cached list and a live update list).
- Workers that go offline may not be removed from the list before re-added on reconnect.

## Acceptance criteria

- [ ] Each worker appears exactly once in the worker list.
- [ ] Workers that go offline are removed or marked offline — not duplicated.
- [ ] The fix is covered by a unit or integration test where feasible.

jmelloy:
## Summary

Investigated the workers pane (`WorkerList.vue`, backed by the Pinia `agents` store in `frontend/src/stores/agents.ts`). The top-level worker rows were already correctly deduplicated — `workers` is a computed `Map` keyed by `workerId`, rebuilt from `agents` on every change, so no worker ever appears twice at that level.

The actual duplication was in the **agent slots nested under each worker row**. Each worker process generates fresh, random `agentId`s for its slots on every restart (`worker/pioneer_worker/worker.py`: `self.agents = [Agent() for _ in range(cfg.max_agents)]`, no persisted id). When a worker reconnects, the backend correctly marks the old agent slot `offline` and registers the new one — but the frontend's `agents` array never prunes offline entries. `WorkerList.vue`'s `agentsForWorker()` filtered only by `workerId`, not by state, so the old offline slot kept rendering as a ghost row alongside the new live one every time a worker restarted — appearing as duplicate entries in the pane, and multiplying with repeated reconnects.

## Fix

- `agentsForWorker()` in `frontend/src/components/sidebar/WorkerList.vue` now excludes agents with `state === 'offline'`, consistent with the existing `onlineWorkers` filtering.
- Added `frontend/src/components/sidebar/__tests__/WorkerList.spec.ts` covering:
  - A worker with two concurrent agent slots still renders exactly one worker row.
  - After a worker reconnects under a new `agentId` (old slot marked offline), only the live agent row renders — not both.

## Test plan

- [x] `npx vitest run` — full frontend suite passes (105 passed, 4 skipped)
- [x] `npx vue-tsc --noEmit` — clean
- [x] `npx eslint` on changed files — clean

Closes #731

### DIFF
--- a/frontend/src/components/sidebar/WorkerList.vue
+++ b/frontend/src/components/sidebar/WorkerList.vue
@@ -62,10 +62,9 @@ const showSpawnForm = ref(false)
 const onlineWorkers = computed(() => agentsStore.workers.filter((w) => w.state !== 'offline'))
 
 function agentsForWorker(workerId: string) {
-  // Exclude offline slots: a reconnecting worker registers fresh agent ids
-  // under the same workerId, and the store never prunes the old (now
-  // offline) rows — without this filter they'd pile up as ghost duplicates
-  // alongside the live slot every time the worker process restarts.
+  // Defensive filter: the store prunes agents on the offline transition, but
+  // keep this so a stale offline row can never surface here if that
+  // invariant is ever violated (e.g. a race during deregistration).
   return agentsStore.agents.filter((a) => a.workerId === workerId && a.state !== 'offline')
 }
 
--- a/frontend/src/components/sidebar/WorkerList.vue
+++ b/frontend/src/components/sidebar/WorkerList.vue
@@ -62,7 +62,11 @@ const showSpawnForm = ref(false)
 const onlineWorkers = computed(() => agentsStore.workers.filter((w) => w.state !== 'offline'))
 
 function agentsForWorker(workerId: string) {
-  return agentsStore.agents.filter((a) => a.workerId === workerId)
+  // Exclude offline slots: a reconnecting worker registers fresh agent ids
+  // under the same workerId, and the store never prunes the old (now
+  // offline) rows — without this filter they'd pile up as ghost duplicates
+  // alongside the live slot every time the worker process restarts.
+  return agentsStore.agents.filter((a) => a.workerId === workerId && a.state !== 'offline')
 }
 
 function currentTaskForWorker(workerId: string) {
--- a/frontend/src/components/sidebar/__tests__/WorkerList.spec.ts
+++ b/frontend/src/components/sidebar/__tests__/WorkerList.spec.ts
@@ -1,0 +1,35 @@
+import { beforeEach, describe, expect, it } from 'vitest'
+import { setActivePinia, createPinia } from 'pinia'
+import { mount } from '@vue/test-utils'
+import WorkerList from '../WorkerList.vue'
+import { useAgentsStore } from '../../../stores/agents'
+
+describe('WorkerList', () => {
+  beforeEach(() => {
+    setActivePinia(createPinia())
+  })
+
+  it('shows a worker exactly once even when two agent slots share its workerId', () => {
+    const agentsStore = useAgentsStore()
+    agentsStore.registerAgent({ agentId: 'a-1', workerId: 'w-x', state: 'idle' })
+    agentsStore.registerAgent({ agentId: 'a-2', workerId: 'w-x', state: 'idle' })
+
+    const wrapper = mount(WorkerList)
+    expect(wrapper.findAll('.worker-row')).toHaveLength(1)
+  })
+
+  it('drops the stale agent slot after a worker reconnects under a new agentId', () => {
+    // Regression for #731: a worker process restart registers a fresh
+    // agentId under the same workerId; the old slot is marked offline but
+    // never removed from the store, so the sidebar must filter it out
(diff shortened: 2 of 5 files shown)


frontend/src/components/sidebar/__tests__/WorkerList.spec.ts
```diff
@@ -0,0 +1,35 @@
...
+
+  it('shows a worker exactly once even when two agent slots share its workerId', () => {
+    const agentsStore = useAgentsStore()
+    agentsStore.registerAgent({ agentId: 'a-1', workerId: 'w-x', state: 'idle' })
+    agentsStore.registerAgent({ agentId: 'a-2', workerId: 'w-x', state: 'idle' })
+
+    const wrapper = mount(WorkerList)
+    expect(wrapper.findAll('.worker-row')).toHaveLength(1)
+  })
+
```
jmelloy, line 20: This test registers two *online* idle slots for the same worker and asserts one `.worker-row`. That already passed before this PR (worker deduplication was already correct per the PR description). The test does not exercise the offline-filtering fix at all and would pass even if the fix were reverted.

jmelloy (commented):
- The one-line fix in `agentsForWorker` correctly addresses the immediate symptom (ghost offline agent rows) by filtering out offline slots, and the PR description accurately diagnoses the root cause.
- However, filtering offline agents in the view is a band-aid; the underlying issue — that the store never prunes stale offline agent entries — means unbounded memory growth for long-running instances with many worker restarts.
- The new tests check UI rendering but rely on `.worker-row` and `.agent-row` CSS class selectors whose presence in `WorkerList.vue` is not verified in the diff, making the tests potentially fragile or silently passing vacuously.
- The fix does not consider whether offline agent slots may still be needed elsewhere (e.g., task history display, other computed properties in the store), so the filter could silently hide information that other consumers rely on.
- No regression coverage for `currentTaskForWorker` or other helpers in `WorkerList.vue` that also iterate `agentsStore.agents` without an offline filter, which may have the same latent bug.

frontend/src/components/sidebar/__tests__/WorkerList.spec.ts
```diff
@@ -0,0 +1,35 @@
...
+  it('drops the stale agent slot after a worker reconnects under a new agentId', () => {
+    // Regression for #731: a worker process restart registers a fresh
+    // agentId under the same workerId; the old slot is marked offline but
+    // never removed from the store, so the sidebar must filter it out
+    // instead of rendering both as separate agent rows.
+    const agentsStore = useAgentsStore()
+    agentsStore.registerAgent({ agentId: 'a-old', workerId: 'w-x', state: 'idle' })
+    agentsStore.handleWebSocketMessage({ type: 'agent-state', agentId: 'a-old', state: 'offline' })
```
jmelloy, line 28: The test asserts `.worker-row` and `.agent-row` counts, but neither selector appears in the diff for `WorkerList.vue`. If those classes don't exist (or use different names), `findAll` returns an empty NodeList and `toHaveLength(1)` fails — or if the component doesn't render at all due to missing stubs/router, both assertions pass vacuously on length 0 vs 1, giving a false green. Please confirm the selectors and add a sanity assertion (e.g., that the wrapper actually mounted meaningful DOM).

frontend/src/components/sidebar/WorkerList.vue
```diff
@@ -62,7 +62,11 @@ const showSpawnForm = ref(false)
 const onlineWorkers = computed(() => agentsStore.workers.filter((w) => w.state !== 'offline'))
 
 function agentsForWorker(workerId: string) {
-  return agentsStore.agents.filter((a) => a.workerId === workerId)
+  // Exclude offline slots: a reconnecting worker registers fresh agent ids
```
jmelloy: Filtering offline agents in the view layer is the right quick fix, but the root cause — stale offline entries accumulating indefinitely in `agentsStore.agents` — should be addressed in the store (e.g., replace or remove the entry on `agent-state: offline` rather than merely marking it). As-is, the array grows without bound across worker restarts, wasting memory and making every recomputation slower over time.

frontend/src/components/sidebar/WorkerList.vue
```diff
@@ -62,7 +62,11 @@ const showSpawnForm = ref(false)
...
+  // under the same workerId, and the store never prunes the old (now
+  // offline) rows — without this filter they'd pile up as ghost duplicates
+  // alongside the live slot every time the worker process restarts.
+  return agentsStore.agents.filter((a) => a.workerId === workerId && a.state !== 'offline')
 }
```
jmelloy, line 70: `currentTaskForWorker` (and potentially other helpers) also queries `agentsStore.agents` filtered only by `workerId` without excluding offline slots. If any of those helpers use the agent state to derive display data, they inherit the same ghost-row bug. Audit all `agentsStore.agents` usages in this file for the same missing filter.

PR #739, opened 2026-07-04
Discord: /join-channel and /leave-channel slash commands

### LINKED ISSUE #738: Discord: /join-channel slash command to wire a channel to a PS guild
## Overview

Add a `/join-channel` slash command that lets a Discord server admin (or someone with the Pioneer Square Operator role, see #735) wire a specific Discord channel to a Pioneer Square guild.

## Motivation

Currently the bot has no interactive way to associate a channel with a PS guild — it either auto-creates channels (see #735) or uses hardcoded env vars. A slash command gives operators fine-grained control: they can pick any existing channel and bind it to a PS guild without restarting the bot or editing config.

## Proposed UX

```
/join-channel channel:#my-channel guild:<guild-slug>
```

- `channel` — a Discord channel mention/picker (required)
- `guild` — the Pioneer Square guild slug/ID to associate (required; if there is only one guild configured, default to it)

**Bot response (ephemeral):**
```
✅ #my-channel is now wired to Pioneer Square guild "my-guild".
   Task events, PR notifications, and CI alerts will be posted here.
```

Re-running the command on an already-wired channel updates the binding (idempotent).

## Role / permission

- Command should be restricted to users with the `Pioneer Square Operator` role (from #735) **or** Discord `MANAGE_CHANNELS` permission.
- Use `discord.app_commands.checks.has_permissions(manage_channels=True)` or a custom role check.

## Implementation notes

1. **Store the binding** — add a `discord_channel_guild` table (or extend the existing `discord_guilds` / `discord_channels` tables) with `(discord_guild_id, discord_channel_id, ps_guild_id)`. Unique constraint on `(discord_guild_id, discord_channel_id)`.
2. **Register the slash command** in the bot's command tree (likely in `backend/discord/bot.py` or the cog that owns guild-related commands).
3. **Notify the notifier** — after wiring, the event notifier should start posting to this channel. If the notifier reads channel bindings at startup, trigger a reload; if it reads from DB on each event, no extra work is needed.
4. **`/leave-channel` counterpart** — add a `/leave-channel` command to remove the binding (optional but good practice).
5. **Relation to #735** — the auto-create-on-join flow in #735 can use the same DB table to record which channels it created; `/join-channel` is an additive/override path.

## Acceptance criteria

- [ ] `/join-channel #channel guild-slug` stores the binding in the DB
- [ ] Subsequent events from the named PS guild are posted to the wired channel
- [ ] Command is permission-gated (Operator role or `MANAGE_CHANNELS`)
- [ ] Idempotent: re-running updates the binding without error
- [ ] `/leave-channel` removes the binding
- [ ] Basic test coverage for the DB model and command handler

jmelloy:
Closes #738

## Summary
- DB migration: adds `discord_channel_guilds` table
- `/join-channel` and `/leave-channel` slash commands with permission checks
- Notifier integration for newly wired channels
- Unit and integration tests
- Updated docs/discord.md

### DIFF
--- a/backend/alembic/versions/20260704_000003_add_discord_channel_guilds.py
+++ b/backend/alembic/versions/20260704_000003_add_discord_channel_guilds.py
@@ -1,0 +1,45 @@
+"""Add discord_channel_guilds table binding Discord channels to PS guilds.
+
+Revision ID: 20260704_000003_add_discord_channel_guilds
+Revises: 20260704_000002_add_discord_foreman_threads
+Create Date: 2026-07-04
+
+Backs the ``/join-channel`` and ``/leave-channel`` slash commands, letting an
+operator wire a Discord channel to a Pioneer Square guild without editing
+env vars or restarting the bot.
+"""
+
+from __future__ import annotations
+
+from collections.abc import Sequence
+
+import sqlalchemy as sa
+from alembic import op
+
+revision: str = "20260704_000003_add_discord_channel_guilds"
+down_revision: str | Sequence[str] | None = "20260704_000002_add_discord_foreman_threads"
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+
+def upgrade() -> None:
+    op.create_table(
+        "discord_channel_guilds",
+        sa.Column("id", sa.Integer(), nullable=False),
+        sa.Column("discord_guild_id", sa.Text(), nullable=False),
+        sa.Column("discord_channel_id", sa.Text(), nullable=False),
+        sa.Column("ps_guild_id", sa.Text(), nullable=False),
+        sa.Column("created_at", sa.DateTime(timezone=True), nullable=False),
+        sa.PrimaryKeyConstraint("id"),
+    )
+    op.create_index(
+        "uq_discord_channel_guilds_guild_channel",
+        "discord_channel_guilds",
+        ["discord_guild_id", "discord_channel_id"],
+        unique=True,
+    )
+
+
+def downgrade() -> None:
+    op.drop_index("uq_discord_channel_guilds_guild_channel", table_name="discord_channel_guilds")
+    op.drop_table("discord_channel_guilds")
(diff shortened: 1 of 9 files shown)


jmelloy (commented):
{"summary": "- The DB schema has a meaningful design gap: the unique index covers `(discord_guild_id, discord_channel_id)` but the query in `_resolve_channel_for_guild` filters only on `ps_guild_id`, meaning one PS guild can receive events from multiple Discord channels with no deterministic tie-breaking — only `rows[0]` is returned, which is ordering-dependent.\n- `_permission_denied_message` is declared `async` but does no async work; callers `await` it unnecessarily, which wastes a coroutine allocation on every denied request.\n- The `_MANAGE_CHANNELS_BIT` constant value `0x10` is wrong — Discord's MANAGE_CHANNELS permission is bit `0x10` (16), which is actually correct, but the inline comment says 'see Discord's permissions bitfield docs' without noting that Discord sends the guild-level computed permission string, not per-channel overwrites, so channel-specific permission denials are silently bypassed.\n- The `_resolve_channel_for_guild` function catches all exceptions and falls back silently to the global channel ID; a DB connectivity failure would cause events to route to the wrong channel without any operator-visible error beyond a WARNING log.\n- Test coverage for `_resolve_channel_for_guild` itself and for the `notify_event`/`notify_foreman_chat` integration path with the new `ps_guild_slug` parameter is absent — the notifier integration is only tested implicitly through the webhook/ws_handler callsites."],"comments": [{"path": "backend/alembic/versions/20260704_000003_add_discord_channel_guilds.py","line": 32,"side": "RIGHT","body": "The unique constraint is on `(discord_guild_id, discord_channel_id)` but there is no unique constraint on `ps_guild_id`. This allows one PS guild to be bound to multiple Discord channels simultaneously, and `_resolve_channel_for_guild` will silently pick an arbitrary one (`rows[0]` with no ORDER BY). Consider adding a unique index on `ps_guild_id` or at minimum adding an `ORDER BY created_at` to make the selection determinist

PR #741, opened 2026-07-04
fix(discord): remove worker notifications from webhook notifier

### LINKED ISSUE #740: Discord webhook: remove worker-online/offline notifications
## Problem

The Discord webhook notifier (Phase 1, PR #712) currently sends notifications when workers come online or go offline. These are noisy and not useful for Discord consumers — worker lifecycle events are internal plumbing that should stay out of the Discord channel.

## What to change

In the Discord webhook notifier code (introduced in PR #712), remove (or skip) any webhook posts triggered by:
- `[worker-online]` events
- `[worker-offline]` events

Keep notifications for task events (task started, task completed, PR opened, etc.) and any other user-facing events.

## Acceptance criteria
- Worker online/offline events no longer produce Discord webhook messages
- All other notification types are unaffected
- Existing tests pass; update/add tests as needed

jmelloy:
## Summary
- Removes Discord webhook posts for `worker-online`/`worker-offline` events (both graceful disconnect and abrupt WebSocket close) — these are noisy internal signals that shouldn't surface in Discord.
- Foreman notifications for these same events are unchanged.

## Test plan
- [x] Added tests verifying `worker-register`, `worker-disconnect`, and abrupt disconnect no longer call `discord_notifier.notify`
- [x] Existing worker/foreman notification tests still pass

Closes #740

### DIFF
--- a/backend/tests/test_worker_foreman_notify.py
+++ b/backend/tests/test_worker_foreman_notify.py
@@ -235,6 +235,10 @@ def test_worker_online_does_not_notify_discord(client):
                     "tools": ["claude"],
                 }
             )
+            # handle_worker_register no longer spawns any background tasks (the
+            # discord_notifier spawn call was removed), so once the ping is
+            # processed we know notify() would already have been called if it
+            # were going to be.
             ws.send_json({"type": "ping"})
             ws.receive_json()  # pong
 
--- a/backend/tests/test_worker_foreman_notify.py
+++ b/backend/tests/test_worker_foreman_notify.py
@@ -5,13 +5,16 @@ Verifies that:
 - handle_worker_disconnect triggers [worker-offline] reason=shutdown
 - abrupt WebSocket disconnect triggers [worker-offline] reason=disconnect
 - task-complete with stopReason=max_turns triggers a max-turns foreman message
+- worker-online/worker-offline events never post to the Discord webhook
+  (they still notify the foreman, but Discord notifications are internal-only
+  noise — see #740)
 """
 
 from __future__ import annotations
 
 import os
 import sys
-from unittest.mock import patch
+from unittest.mock import AsyncMock, patch
 
 import pytest
 from sqlalchemy.ext.asyncio import async_sessionmaker, create_async_engine
@@ -23,6 +26,7 @@ sys.path.insert(0, os.path.join(os.path.dirname(__file__), ".."))
 sys.path.insert(0, os.path.dirname(__file__))
 
 import database as database_module  # noqa: E402
+import discord_notifier  # noqa: E402
 import main as main_module  # noqa: E402
 import ws_handlers  # noqa: E402
 from _test_config import TEST_DATABASE_URL  # noqa: E402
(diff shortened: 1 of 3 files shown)


backend/tests/test_worker_foreman_notify.py
```diff
@@ -195,6 +199,114 @@ def test_abrupt_disconnect_notifies_foreman(client):
...
+                    "type": "worker-register",
+                    "workerId": worker_id,
+                    "repos": ["org/repo1"],
+                    "tools": ["claude"],
+                }
+            )
+            ws.send_json({"type": "ping"})
+            ws.receive_json()  # pong
+
+        mock_notify.assert_not_called()
```
jmelloy, line 245: The `discord_notifier.notify` mock is patched at the module level (`discord_notifier.notify`), but `ws_handlers.py` may have imported `discord_notifier` directly and call `discord_notifier.notify(...)` — since the notify calls are being removed this is moot for new code, but if any remaining path in `ws_handlers` still imports and calls `discord_notifier.notify`, the patch target here would need to be `ws_handlers.discord_notifier.notify` to intercept it correctly.

backend/tests/test_worker_foreman_notify.py
```diff
@@ -195,6 +199,114 @@ def test_abrupt_disconnect_notifies_foreman(client):
...
+                {
+                    "type": "join",
+                    "agentId": agent_id,
+                    "agentName": "Test Worker",
+                    "agentType": "worker",
+                    "workerId": worker_id,
+                }
+            )
+            ws.receive_json()  # agent-joined broadcast
+
```
jmelloy, line 229: Using `ping/pong` as a synchronization barrier is clever but fragile: it guarantees the ping was processed, not that the `worker-register` handler's async tasks (e.g. spawned coroutines) have fully completed. If any residual `spawn` calls exist or are added later, the mock assertion could pass spuriously before background tasks run.

backend/ws_handlers.py
```diff
@@ -664,15 +664,6 @@ async def handle_worker_register(ctx: WSContext, data: dict) -> None:
         f"[worker-online] worker_id={worker_id} repos={repos_str} agent_count={agent_count}{tools_suffix}{provider_suffix}",
```
jmelloy, line 664: There is a blank line left after the deleted `spawn(...)` block (visible in the diff as an extra newline before the next function). Minor style issue but worth cleaning up to keep the file tidy.

jmelloy (commented):
- The core change is correct and well-scoped: three Discord `notify` call-sites (worker-register, graceful disconnect, abrupt disconnect) are cleanly removed without affecting foreman notifications.
- The `discord_notifier` import in `websocket.py` and the `spawn` import are both removed; however, care should be taken to verify these imports aren't used elsewhere in those files (they appear not to be, based on the diff).
- The three new tests are structurally sound and cover all removed code paths, but they patch `discord_notifier.notify` after the module is imported, which is correct only if `ws_handlers` references `discord_notifier.notify` indirectly — worth confirming the patch target matches actual call sites.
- The `worker-online` test sends a `worker-register` message but never receives its acknowledgement/response before checking the mock, relying on the `ping/pong` as a synchronization barrier; this is a reasonable pattern but slightly fragile if message ordering changes.
- No security concerns; the change reduces external signal surface area and noise, which is a net positive.

Issue #744, opened 2026-07-04
Discord integration: Route inbound chat to Foreman (Phase 5)


jmelloy:
## Parent issue
Part of #652 (builds on #743, #709, #717)

## Overview
Once the Gateway connection (#743) is delivering filtered `MESSAGE_CREATE` events, this is the routing/reply layer: deciding which Foreman session a given chat message belongs to, forwarding it, and posting the reply back into the right thread.

## Thread routing model
Two thread types already exist in the codebase — reuse both rather than inventing a third:

- **Per-task thread** (`discord_threads`, from #709): if a message arrives inside a thread that's already mapped to a task/PR, treat it as chat about that task — Foreman already has context loaded for it.
- **Per-guild-daily thread** (`discord_foreman_threads`, from #717): currently outbound-only (Foreman → Discord narration mirror). If a message arrives there, or anywhere else in a wired channel not tied to a specific task, treat it as general/ad-hoc chat with Foreman for that guild.

Explicitly **not** doing private per-user threads in this phase — Discord private threads have real discoverability/UX friction (not auto-visible, users must be added), and a shared thread is simplest to reason about for a small team. Revisit only if the shared thread proves too noisy with multiple concurrent users in practice.

## Requirements
- Reverse-lookup incoming message's `channel_id`/`thread_id` against `discord_threads` first, then `discord_foreman_threads`, to resolve which Foreman session (task-scoped vs guild-general) it belongs to.
- Resolve the sending Discord user to a Pioneer Square identity via the existing `discord_users` mapping (used today by `mention_or_login`) so Foreman knows who it's talking to; fall back to a generic "Discord user" identity if unmapped.
- Forward the message content + resolved identity + session context into the Foreman chat pipeline (reuse the existing chat/narration path that already writes into `discord_foreman_threads`, just in the inbound direction).
- Post Foreman's reply back into the same thread the question came from (task thread replies stay in that task thread; general chat replies stay in the daily guild thread).
- Rate limit / cost guard: since this can trigger real Foreman work (not just read status), gate on the same authorization concept already used for `/ps` write commands — check `DISCORD_ALLOWED_ROLE_IDS` (or equivalent) before treating a message as an actionable request rather than idle chat.
- Ignore messages in threads/channels with no resolvable session (nothing to attach the conversation to).

## Out of scope
- Private per-user threads (see above — future work, only if needed)
- Reaction-based approvals (already deferred to a later phase in #652)
[...]

PR #745, opened 2026-07-04
feat(discord): persistent Gateway websocket for inbound messages

### LINKED ISSUE #743: Discord integration: Gateway connection for inbound messages (Phase 4)
## Parent issue
Part of #652 (builds on #708, #709, #710, #717, #738)

## Overview
Every current Discord command path goes through `POST /discord/interactions` — Discord's HTTP Interactions webhook. That model is request/response only: Discord calls us, we ack, done. It never delivers a plain message someone types in a channel or thread, so there is no way today for a user to just chat with Foreman — only registered `/ps ...` slash commands work.

To support real bidirectional chat we need a persistent [Gateway](https://discord.com/developers/docs/events/gateway) (websocket) connection that subscribes to `MESSAGE_CREATE` events. This issue is the transport layer only — receiving and dispatching raw Gateway messages. Routing those messages to the right Foreman session is #744.

## Requirements
- Persistent websocket client to `wss://gateway.discord.gg`, run as an `asyncio` task inside the existing backend process (consistent with how the interactions webhook and `discord_notifier` bot REST calls already run in-process — no sidecar process needed).
- Implement the Gateway handshake: `IDENTIFY` with required intents (`GUILDS`, `GUILD_MESSAGES`, and the privileged `MESSAGE_CONTENT` intent — note `MESSAGE_CONTENT` requires enabling in the Discord Developer Portal and, once the bot is in 100+ guilds, Discord's message-content verification review).
- Heartbeat loop per the Gateway spec; automatic reconnect with `RESUME` (session ID + sequence number) on drops, falling back to a fresh `IDENTIFY` if resume fails.
- Filter incoming `MESSAGE_CREATE` events down to channels/threads that are actually wired: ignore DMs, ignore guilds/channels with no `DiscordChannelGuild` binding, ignore messages authored by the bot itself (avoid echo loops with `discord_notifier`'s own posts).
- Emit filtered events onto a simple internal hook/queue that #744 consumes — this issue should not implement any Foreman-facing logic.
- Config: reuse `DISCORD_BOT_TOKEN`; add a `DISCORD_GATEWAY_ENABLED` flag (default off) so this can ship dark and be enabled per-deployment once stable.
- Graceful no-op when `DISCORD_BOT_TOKEN` is absent (matches the existing fallback convention in `discord_notifier.py`).

## Out of scope
- Any Foreman-facing routing/reply logic (tracked in #744)
- Slash command handling (unchanged, stays on the existing Interactions webhook — Gateway is additive, not a replacement)
- Voice, presence, or any other Gateway intent not needed for message relay

## Acceptance criteria
- [ ] Backend maintains a stable Gateway connection, surviving reconnects without manual restart
- [ ] `MESSAGE_CONTENT`/`GUILD_MESSAGES` intents correctly configured and documented in `docs/discord.md`
- [ ] Messages in unwired channels/guilds are dropped before reaching any handler
- [ ] Bot's own messages never re-trigger the relay (no feedback loop)
- [ ] `DISCORD_GATEWAY_ENABLED=false` (default) results in zero behavior change from today
- [ ] Basic reconnect/backoff test (simulated disconnect) passes

jmelloy:
## Summary
- Adds `backend/discord/gateway.py`: a persistent websocket client to the Discord Gateway (`wss://gateway.discord.gg/?v=10&encoding=json`) with `IDENTIFY` (intents bitmask `33281` = GUILDS | GUILD_MESSAGES | MESSAGE_CONTENT), a heartbeat loop driven by `HELLO`/`HEARTBEAT_ACK`, and reconnect logic that attempts `RESUME` with the stored `session_id`/`seq`, falling back to fresh `IDENTIFY` on `INVALID_SESSION`.
- Filters `MESSAGE_CREATE` events: drops bot-authored messages, DMs (no `guild_id`), and channels/threads not present in `DiscordChannelGuild`. Wired-channel events are pushed onto a module-level `gateway_message_queue` for #744 to consume — no Foreman routing logic here.
- Runs as a single long-lived `asyncio.Task`, started/stopped from the FastAPI `lifespan` in `backend/main.py`, gated on the new `DISCORD_GATEWAY_ENABLED` env var (default `false`).
- Documents `DISCORD_GATEWAY_ENABLED` in `.env.example` and `docker-compose.yml`, noting the Message Content privileged intent must be enabled in the Discord Developer Portal.

Closes #743

## Test plan
- [x] `pytest backend/tests/test_discord_gateway.py` — 14/14 passed (mocked websocket: IDENTIFY, heartbeat, RESUME, INVALID_SESSION fallback, MESSAGE_CREATE filtering)
- [x] Full backend suite passes (74 passed)
- [ ] Manual verification against live Discord Gateway (requires `DISCORD_GATEWAY_ENABLED=true` + bot token with Message Content intent enabled)

### DIFF
--- a/backend/discord/gateway.py
+++ b/backend/discord/gateway.py
@@ -1,0 +1,308 @@
+"""Discord Gateway websocket client (Phase 4): transport for inbound messages.
+
+This module owns a single persistent connection to Discord's Gateway
+(``wss://gateway.discord.gg``) so the backend can receive ``MESSAGE_CREATE``
+events in real time, instead of only the one-shot ``POST /discord/interactions``
+webhook. It is transport-layer only: filtered events are handed to
+``gateway_message_queue`` for a downstream consumer (#744, Foreman routing) to
+read. This module never imports or calls into Foreman logic.
+
+Enable with::
+
+    DISCORD_GATEWAY_ENABLED=true
+    DISCORD_BOT_TOKEN=...          # reused from Phase 2/3
+
+The bot's Discord application must have the privileged "Message Content
+Intent" turned on in the Developer Portal (Bot page) — without it Discord
+rejects the IDENTIFY payload's ``MESSAGE_CONTENT`` intent bit.
+
+Protocol handled, per https://discord.com/developers/docs/topics/gateway:
+    HELLO (op 10)            -> start the heartbeat loop, then IDENTIFY/RESUME
+    HEARTBEAT_ACK (op 11)    -> mark the last heartbeat as acknowledged
+    DISPATCH (op 0)          -> READY stores session_id/resume url; MESSAGE_CREATE
+                                 is filtered and queued
+    RECONNECT (op 7)         -> close and reconnect, attempting RESUME
+    INVALID_SESSION (op 9)   -> RESUME if resumable, otherwise fresh IDENTIFY
+
+Filtering applied to ``MESSAGE_CREATE`` before anything is queued:
+    - ``author.bot is True``      -> discarded (avoids echoing discord_notifier)
+    - no ``guild_id`` (a DM)      -> discarded
+    - channel not in ``discord_channel_guilds`` -> discarded (not wired)
+"""
+
+from __future__ import annotations
+
+import asyncio
+import json
+import logging
+import os
+import random
+
+import websockets
+
+logger = logging.getLogger(__name__)
+
+_GATEWAY_URL = "wss://gateway.discord.gg/?v=10&encoding=json"
+
+# GUILDS (1 << 0) | GUILD_MESSAGES (1 << 9) | MESSAGE_CONTENT (1 << 15)
+_INTENTS = 33281
+
+# Gateway opcodes.
+_OP_DISPATCH = 0
+_OP_HEARTBEAT = 1
+_OP_IDENTIFY = 2
+_OP_RESUME = 6
+_OP_RECONNECT = 7
+_OP_INVALID_SESSION = 9
+_OP_HELLO = 10
+_OP_HEARTBEAT_ACK = 11
+
+# Filtered MESSAGE_CREATE events for wired channels, consumed by #744.
+gateway_message_queue: asyncio.Queue[dict] = asyncio.Queue()
+
+
+def _bot_token() -> str | None:
+    return os.environ.get("DISCORD_BOT_TOKEN") or None
+
+
+def _gateway_enabled() -> bool:
+    return os.environ.get("DISCORD_GATEWAY_ENABLED", "false").strip().lower() == "true"
+
+
+def _backoff_delay(attempt: int, base: float = 2.0, cap: float = 60.0) -> float:
+    """Exponential backoff with full jitter."""
+    return min(cap, base * (2**attempt)) + random.uniform(0, base)
+
+
+async def _is_channel_wired(channel_id: str) -> bool:
+    """Return True if *channel_id* (a channel or thread) has a guild binding.
+
+    Backed by the ``discord_channel_guilds`` table populated by
+    ``/join-channel``. Never raises — DB errors are treated as "not wired".
+    """
+    try:
+        from database import AsyncSessionLocal  # noqa: PLC0415
+        from models import DiscordChannelGuild  # noqa: PLC0415
+        from sqlmodel import col, select  # noqa: PLC0415
+
+        async with AsyncSessionLocal() as db:
+            result = await db.exec(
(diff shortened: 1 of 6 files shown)


jmelloy (commented):
- Well-structured Gateway client with correct Discord protocol handling (HELLO/HEARTBEAT/RESUME/INVALID_SESSION), good separation of concerns, and solid test coverage using fake websockets.
- **Security concern**: the bot token is accessed via `_bot_token()` at task-start time but `GatewayClient` stores it as `self._token` in plaintext on the object for the lifetime of the process; this is low-risk internally but worth noting.
- **Zombie heartbeat bug**: when `_OP_RECONNECT` is handled in `_handle_frame`, `ws.close()` is called but `_stop_heartbeat()` is never awaited inside `_handle_frame` — the heartbeat task continues running against a closed socket until the outer `finally` in `run()` cleans it up, potentially logging spurious warnings.
- **Per-message DB query**: `_is_channel_wired` issues a full DB round-trip (open session, query, close) for every non-bot guild message; a simple in-process cache (e.g. TTL dict or LRU cache) would significantly reduce load on busy channels.
- The `start_gateway()` function imports `util.tasks.spawn` lazily but the returned task's result is silently discarded at the call-site in `main.py`, meaning any immediate startup error (e.g. bad token format) is not surfaced until the first reconnect log line.

backend/discord/gateway.py
```diff
@@ -0,0 +1,308 @@
...
+                {
+                    "op": _OP_RESUME,
+                    "d": {
+                        "token": self._token,
+                        "session_id": self._session_id,
+                        "seq": self._seq,
+                    },
+                }
+            )
+        )
```
jmelloy, line 242: When `_OP_RECONNECT` is received, `ws.close()` is called here but `_stop_heartbeat()` is not awaited. The heartbeat task keeps running against the now-closed websocket and will log a spurious 'heartbeat not acked — forcing reconnect' warning before `run()`'s `finally` cleans it up. Consider calling `await self._stop_heartbeat()` before or after `ws.close()` here.

backend/discord/gateway.py
```diff
@@ -0,0 +1,308 @@
...
+
+async def _is_channel_wired(channel_id: str) -> bool:
+    """Return True if *channel_id* (a channel or thread) has a guild binding.
+
+    Backed by the ``discord_channel_guilds`` table populated by
+    ``/join-channel``. Never raises — DB errors are treated as "not wired".
+    """
+    try:
+        from database import AsyncSessionLocal  # noqa: PLC0415
+        from models import DiscordChannelGuild  # noqa: PLC0415
```
jmelloy, line 108: Every non-bot guild message triggers a fresh DB session open/query/close round-trip. On a busy server this will create significant database load. Consider a short-lived TTL cache (e.g. `cachetools.TTLCache` or a simple `dict` with timestamp) keyed on `channel_id` to avoid hitting the DB for every message.

backend/main.py
```diff
@@ -364,6 +364,12 @@ async def _startup_refresh_catalog() -> None:
 
     asyncio.ensure_future(_startup_refresh_catalog())
     sweeper = spawn(_stale_worker_sweeper(), name="stale-worker-sweeper")
+
+    # Discord Gateway (Phase 4): persistent websocket for inbound messages.
+    # No-op unless DISCORD_GATEWAY_ENABLED=true and DISCORD_BOT_TOKEN are set.
+    from discord.gateway import start_gateway, stop_gateway  # noqa: PLC0415
+
+    start_gateway()
```
jmelloy: The return value of `start_gateway()` is discarded, so there is no way to observe the task from the lifespan context or await it during shutdown sequencing. Consider assigning it — e.g. `discord_gw_task = start_gateway()` — so it could be included in any future structured-concurrency cleanup alongside `sweeper` and `drain_bg`.

backend/discord/gateway.py
```diff
@@ -0,0 +1,308 @@
...
+        )
+
+        if self._session_id and self._seq is not None:
+            await self._send_resume(ws)
+        else:
+            await self._send_identify(ws)
+
+        async for raw in ws:
+            await self._handle_frame(ws, raw)
+
```
jmelloy, line 183: After `ws.close(code=4000)` is called in `_heartbeat_loop`, the loop returns but the outer `async for raw in ws` in `_handle_connection` may still attempt to read from the closing socket, potentially logging a spurious error before raising. Confirm the `websockets` library raises `ConnectionClosed` on the iterator immediately after a local `close()` call to ensure clean shutdown.

PR #747, opened 2026-07-04
feat(discord): route task-scoped Foreman replies to per-task threads

### LINKED ISSUE #746: Discord: route per-task Foreman responses into the per-task thread (not the daily guild thread)
## Overview

Currently, all Foreman narration/responses are posted to the per-guild **daily thread** (`discord_foreman_threads`, from #717). Per-task notifications (task started, task complete, PR opened, etc.) are posted to per-task threads (`discord_threads`, from #709), but Foreman's own conversational replies still go to the daily thread.

The desired behaviour: **when a Foreman response is scoped to a specific task or issue, post it directly into the corresponding per-task/per-issue Discord thread** rather than the daily guild thread.

## Motivation

- Keeps task-specific conversation in one place — the thread that already exists for that task/PR.
- Avoids cluttering the daily guild thread with task-scoped replies.
- Pairs naturally with Phase 5 (#744): inbound messages arrive in a per-task thread, Foreman's reply should go back to the same thread.

## Scope

- In `backend/discord/notifier.py` (or wherever Foreman responses are posted), detect whether the current Foreman turn is scoped to a task/issue.
- If a matching entry exists in `discord_threads` for that task/issue, post to that thread instead of the daily `discord_foreman_threads` entry.
- If no per-task thread exists (e.g., general ad-hoc chat), fall back to the daily guild thread as before.
- The per-task thread lookup key should be the task id or issue number — reuse the same mapping already used by the Phase 2 notifier.

## Acceptance criteria

- [ ] Foreman replies that are clearly about a specific task appear in that task's Discord thread.
- [ ] General/non-task Foreman messages continue to appear in the daily guild thread.
- [ ] No duplicate messages posted to both threads.
- [ ] Existing Phase 1/2 tests still pass.

## Related
- #652 (parent design issue)
- #709 (per-task threads)
- #717 (daily guild thread / foreman narration)
- #743 (gateway inbound)
- #744 (Phase 5 routing)

jmelloy:
## Summary
- Foreman chat narration now posts to a task's per-task Discord thread (`discord_threads`, keyed by linked issue/PR) instead of always landing in the daily guild thread (`discord_foreman_threads`).
- `notify_foreman_chat` accepts an optional `task_id`; when it resolves to a task with a linked GitHub issue/PR that already has a thread, the message is posted there and the daily-thread fallback is skipped. When there's no task scope, or the task has no linked issue/thread yet, behavior is unchanged (daily guild thread).
- Wired `task_id` through `foreman/runner.py`'s `_emit_foreman_chat` call sites so the routing actually fires for real Foreman turns, not just in the notifier's own tests.

## Details
- `backend/discord_notifier.py`: added `_lookup_task_issue_coords()` to resolve a task's linked `(issue_repo, issue_number)`, reused the existing `_lookup_thread()` helper (same one Phase 2 PR/issue notifications use) to find that issue's thread, and updated `notify_foreman_chat()` to try the per-task thread first before falling back to the per-guild/per-day thread. A message is only ever posted once.
- `backend/foreman/runner.py`: `_emit_foreman_chat()` now takes `task_id` and forwards it to `notify_foreman_chat()`. All three call sites pass the already-in-scope `_task_id` for the current turn.

Closes #746

## Test plan
- [x] `pytest backend/tests/test_discord_notifier.py backend/tests/test_foreman_runner.py` — 55 passed, covering task-scoped → per-task thread, non-scoped → daily thread, and task-scoped-but-no-thread → daily thread fallback.
- [x] `ruff check` clean on changed files.

### DIFF
--- a/backend/discord_notifier.py
+++ b/backend/discord_notifier.py
@@ -19,6 +19,12 @@ Phase 3 — Foreman chat threads + user mentions
     Discord thread per guild per UTC day (mapping persisted in
     ``discord_foreman_threads``), reusing the same bot token/channel as Phase 2.
 
+    When a chat line is scoped to a task (``task_id`` passed through) and that
+    task's linked GitHub issue/PR already has a thread in ``discord_threads``,
+    the line is posted there instead — so task-scoped narration stays with
+    that task's Discord history rather than the daily catch-all thread.
+    Falls back to the per-day thread whenever no per-task thread applies.
+
     ``mention_or_login`` resolves a GitHub login to a real ``<@id>`` Discord
     mention via the ``discord_users`` table (populated through the
     ``/api/discord-users`` REST endpoints). Falls back to a plain ``@login``
@@ -43,8 +49,11 @@ Usage::
         header_fields={"Assignee": "@alice", "Labels": "bug"},
     )
 
-    # Foreman chat mirrored into a per-day thread:
-    await discord_notifier.notify_foreman_chat(guild_id, "Spun up worker w-abc for t-123.")
+    # Foreman chat mirrored into a per-day thread (or a per-task thread when
+    # task_id resolves to one):
+    await discord_notifier.notify_foreman_chat(
+        guild_id, "Spun up worker w-abc for t-123.", task_id="t-123"
+    )
 
     # Resolve a GitHub login to a Discord mention (or a plain @login fallback):
     mention = await discord_notifier.mention_or_login("alice")
@@ -364,6 +373,32 @@ async def archive_thread(thread_id: str) -> None:
 # ---------------------------------------------------------------------------
 
 
+async def _lookup_task_issue_coords(task_id: str) -> tuple[str, int] | None:
+    """Return the ``(issue_repo, issue_number)`` linked to *task_id*, or None.
+
+    Used to route task-scoped Foreman chat into the same thread as that
+    task's PR/issue notifications (see ``_lookup_thread``), instead of the
+    daily guild thread. None when the task has no linked issue/PR, doesn't
+    exist, or the lookup fails.
+    """
+    try:
+        from database import AsyncSessionLocal  # noqa: PLC0415
+        from models import Task  # noqa: PLC0415
+        from sqlmodel import col, select  # noqa: PLC0415
+
+        async with AsyncSessionLocal() as db:
+            result = await db.exec(
+                select(Task.issue_repo, Task.issue_number).where(col(Task.id) == task_id)
+            )
+            row = result.one_or_none()
+            if row and row[0] and row[1] is not None:
+                return row[0], row[1]
+            return None
+    except Exception:
+        logger.warning("discord: task issue-coords lookup failed task=%s", task_id, exc_info=True)
+        return None
+
+
 async def _lookup_foreman_thread(guild_id: str, session_key: str) -> str | None:
     """Return the persisted Discord thread_id for a Foreman chat session, or None."""
     try:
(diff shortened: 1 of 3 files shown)


backend/discord_notifier.py
```diff
@@ -364,6 +373,32 @@ async def archive_thread(thread_id: str) -> None:
...
+        from models import Task  # noqa: PLC0415
+        from sqlmodel import col, select  # noqa: PLC0415
+
+        async with AsyncSessionLocal() as db:
+            result = await db.exec(
+                select(Task.issue_repo, Task.issue_number).where(col(Task.id) == task_id)
+            )
+            row = result.one_or_none()
+            if row and row[0] and row[1] is not None:
+                return row[0], row[1]
```
jmelloy: The bare `except Exception` here swallows all failures silently. While `Never raises` is the documented contract, it would be better to log at least the exception type and task_id without `exc_info=True` in production paths to avoid log noise on every missing task, while still surfacing unexpected errors distinctly. Consider separating `DoesNotExist`-style misses (debug level) from genuine errors (warning level).

backend/discord_notifier.py
```diff
@@ -472,18 +507,41 @@ async def _ensure_foreman_thread(
...
+async def notify_foreman_chat(
+    guild_id: str,
+    content: str,
+    session_key: str | None = None,
+    task_id: str | None = None,
+) -> None:
+    """Mirror a Foreman → user chat line into Discord.
+
+    When *task_id* is given and resolves (via ``discord_threads``) to a
+    per-task thread already created for that task's linked PR/issue, the
```
jmelloy, line 548: When `task_id` is provided and `coords` resolves but `_lookup_thread` returns `None`, execution falls through to the `_resolve_channel_for_guild` path. This is correct, but the `_resolve_channel_for_guild` null-check (and subsequent `_ensure_foreman_thread`) is skipped when posting to the task thread. This means a message posted to a task thread is never content-length truncated by anything other than the `[:2000]` slice — that's fine, but ensure `_bot_request` itself handles Discord API errors gracefully for this path.

backend/foreman/runner.py
```diff
@@ -1046,10 +1052,10 @@ async def _run_foreman_ai(
             for b in wrap_resp.content:
                 if b.type == "text" and b.text.strip():
                     text_parts.append(b.text.strip())
-                    await _emit_foreman_chat(guild_id, b.text.strip(), _now)
+                    await _emit_foreman_chat(guild_id, b.text.strip(), _now, task_id=_task_id)
```
jmelloy, line 1060: The safety-cap note (`cap_note`) is emitted with `task_id=_task_id`, which will route it to the task thread if one exists. This is the correct behavior, but verify that `_task_id` is always defined at this point in the control flow — if the cap is hit before a task is scoped, `_task_id` being undefined would cause a `NameError`. Confirm `_task_id` is initialized to `None` at function entry.

jmelloy (commented):
- The routing logic is clean and well-structured: task-scoped messages route to per-task threads with a clear fallback chain to the daily guild thread.
- The `_lookup_task_issue_coords` function silently swallows all exceptions, which is good for resilience, but the bare `except Exception` pattern masks potentially serious configuration errors beyond transient DB failures.
- A subtle bug exists: when `task_id` is provided but `_lookup_task_issue_coords` or `_lookup_thread` returns `None`, the code falls through to the daily-thread path but does NOT call `_resolve_channel_for_guild` early-exit — however, looking at the diff, the `_resolve_channel_for_guild` check still fires after the task-scoped block, so this is fine. No bug here.
- The `_bot_token()` guard fires before the task-scoped lookup, but `_resolve_channel_for_guild` is only checked after the task-scoped block, meaning a message posted directly to a task thread bypasses the channel configuration check entirely — this is actually intentional and correct, but worth noting.
- Test coverage is thorough and covers all four routing scenarios (task thread hit, no task_id, task with no thread, task with no linked issue); the error-path test for `_lookup_task_issue_coords` is also included.

PR #749, opened 2026-07-04
fix: inject current UTC time into Foreman state (#748)

### LINKED ISSUE #748: fix: task log timestamp timezone double-conversion
## Summary

Investigated the reported "Foreman thinks tasks are stalled when they aren't" bug. **The DB → backend → frontend timestamp pipeline is actually correct and self-consistent today** — I could not find a live naive/aware double-conversion bug there. The real root cause is different: **the Foreman LLM is never told what "now" is**, so it has no reliable anchor to compare `created_at` / `recent_logs[].time` against when judging whether a task looks stalled.

## What I checked (all clean)

1. **DB storage** (`backend/models.py`) — every timestamp column (`Task.created_at`, `TaskLog.timestamp`, `Worker.last_seen`, etc.) is `Column(DateTime(timezone=True), nullable=...)`, i.e. real `TIMESTAMPTZ`.
2. **DB writes** — grepped the whole backend/foreman/worker tree for `datetime.utcnow()` / naive `datetime.now()`; found none. All writes use `datetime.now(UTC)` (see `backend/ws_handlers.py:515,573`, `backend/events.py:131`, `backend/foreman/runner.py`, `foreman/pioneer_foreman/runner.py`).
3. **Driver behavior** — `asyncpg` (the driver in `DATABASE_URL=postgresql+asyncpg://...`) always decodes `timestamptz` columns into tz-aware Python `datetime` objects tagged UTC, independent of the Postgres session `TimeZone` GUC.
4. **API/tool serialization** — `get_task_status` (`backend/foreman/tools.py:1489`) JSON-encodes the row via `json.dumps(..., default=_json_default)`, and `_json_default` (`backend/foreman_core/message_utils.py:18`) calls `obj.isoformat()` on an aware datetime, which correctly emits a `+00:00` offset (e.g. `2026-07-04T14:02:11.884532+00:00`). The plain REST endpoints (`backend/routes/tasks.py`) return the same aware objects and FastAPI's encoder does the same thing.
5. **Frontend** (`frontend/src/utils/format.ts`) — `formatRelative`/`formatAge`/`formatClock` all do `new Date(iso)` then diff against `Date.now()`. Since the ISO strings carry an explicit UTC offset, `Date.parse` handles them correctly; no local-time reinterpretation happens.

So: **timestamps returned by the API are timezone-aware today, and the wall-clock delta the API reports matches reality.** No double conversion in this path currently.

## Actual root cause

The Foreman's `[periodic-check]` decision of "does this task look stalled" is not made in Python — it's made by the Claude model itself, reading the JSON blob from `get_task_status`. But nowhere in the prompt-construction path is the model ever told the current wall-clock time:

- `backend/foreman_core/prompt.py`: `build_system_prompt`, `build_system_blocks`, `build_state_preamble`, and `build_child_state_preamble` inject workers/tasks JSON and free-text context, but no "current UTC time is ..." line.
- The Anthropic Messages API does **not** implicitly tell the model the current date/time (that's a product-layer feature of chat UIs, not a base API behavior).
- So when the model sees `"created_at": "2026-07-04T10:23:45+00:00"` and `"recent_logs": [{"time": "2026-07-04T14:02:11+00:00", ...}]`, it has to infer "now" from its own priors (e.g. training cutoff) or from other timestamps scattered in conversation history. If its assumed "now" differs from actual now by hours/days, it will misjudge a task that was active minutes ago as having been idle for days — exactly the reported symptom.

This is functionally a timezone/clock bug even though it isn't a byte-level tz conversion bug: the model's implicit clock and the backend's real clock disagree, and nothing in the code reconciles them.

## Fix needed

Inject an explicit, freshly-computed "now" anchor into the state the model sees on every foreman turn, e.g. in `backend/foreman_core/prompt.py`:

```python
def build_state_preamble(workers_block, tasks_block, extra_context=""):
    now_line = f"Current UTC time: {datetime.now(UTC).isoformat()}\n\n"
    ...
    body = f"{now_line}{workers_section}## Recent tasks\n```json\n{tasks_block}\n```"
```

Do the same in `build_child_state_preamble` (used by the per-task child context that also runs periodic-check-style reasoning). Since this line changes every call, it must go in the **live state preamble** (already excluded from the cached system block), not `build_system_blocks`/`_stable_system_text`, or prompt caching would pin a stale "now" for the life of the cache.

## Secondary risk worth a quick audit (not confirmed as live)

`backend/alembic/versions/20260526_000000_convert_datetime_columns.py` converted all timestamp columns from `Text` to `DateTime(timezone=True)` via `postgresql_using=f"{column}::timestamptz"`. Casting a naive-looking text string to `timestamptz` in Postgres interprets it using the **session's `TimeZone` GUC**, not necessarily UTC. If the pre-migration text values were naive UTC strings (from an earlier `datetime.utcnow()`-style write path) and the migration ran under a non-UTC session timezone, every row older than 2026-05-26 could have been silently shifted by a fixed offset. Worth a one-time spot check (e.g. compare a known pre-migration row's `created_at` against corroborating evidence like a GitHub PR/commit timestamp from the same task) to rule this out, but it would only affect historical rows, not new data, and doesn't explain an ongoing/current-day symptom.

## Files referenced
- `backend/models.py` (column definitions — all `DateTime(timezone=True)`)
- `backend/foreman/tools.py:1489` (`get_task_status`)
- `backend/foreman_core/message_utils.py:18` (`_json_default`)
- `backend/foreman_core/prompt.py` (missing "now" anchor — fix goes here)
- `backend/foreman/runner.py:706-714` (call sites for the preamble builders)
- `frontend/src/utils/format.ts` (frontend relative-time display — confirmed correct)
- `backend/alembic/versions/20260526_000000_convert_datetime_columns.py` (historical risk)

jmelloy:
## Summary
- The Foreman LLM had no reliable "now" anchor when judging whether a task's `created_at` or a log line's `time` looked stale — it was reasoning about timestamps with no baseline for comparison.
- Adds a fresh per-turn `Current UTC time: <iso>` line to the dynamic `<state>` block built by `build_state_preamble` and `build_child_state_preamble`, kept out of the cacheable stable system prefix (`FOREMAN_SYSTEM`) so it doesn't poison the prompt cache on every turn.
- Updates the "Live state" sections of both the parent and child Foreman prompts to document the new anchor.

Closes #748

## Test plan
- [x] `pytest backend/tests/test_child_prompt.py backend/tests/test_foreman.py` — all pass, including new coverage:
  - `test_child_state_preamble_includes_fresh_current_time`
  - `test_state_preamble_includes_fresh_current_time`
  - `test_current_time_not_in_cacheable_stable_system_text` (fixed to check for the dynamic `"Current UTC time: "` marker specifically, since the prose referencing the label by name is legitimately static)
- [x] Full backend suite: 336 passed, 0 failed (remaining errors are pre-existing DB-connection failures from no local Postgres/Docker in this sandbox, unrelated to this change)

### DIFF
--- a/backend/foreman_core/prompt.py
+++ b/backend/foreman_core/prompt.py
@@ -1,5 +1,7 @@
 """Foreman system prompt and system-prompt builder."""
 
+from datetime import datetime, timezone
+
 FOREMAN_SYSTEM = """\
 You are the Foreman AI in Pioneer Square, a multi-agent coding workshop.
 You coordinate workers — each worker is a host process (w-xxx) that spawns agent subprocesses
@@ -169,9 +171,11 @@ redirect_task to course-correct or cancel_task if it's going in the wrong direct
 If an entire worker is wedged or no longer needed, use shutdown_worker to stop the process.
 
 ## Live state
-Each user turn is preceded by a `<state>` block containing the current online workers
-and recent tasks. Treat it as an operational briefing, not part of the human's message.
-The state reflects the moment this turn was sent — earlier turns saw earlier state.
+Each user turn is preceded by a `<state>` block containing the current UTC time, the
+current online workers, and recent tasks. Treat it as an operational briefing, not part
+of the human's message. The state reflects the moment this turn was sent — earlier turns
+saw earlier state. Use the "Current UTC time" line as your anchor for judging how old a
+task's `created_at` or a log line's `time` is — do not assume those timestamps are recent.
 
 Workers are configured with repos. Prefer workers whose repos cover the task.
 Be concise — one short paragraph maximum unless detail is requested.
@@ -226,15 +230,27 @@ state, the active agent and its state, and the last log lines. If it looks stall
 redirect_task to course-correct or cancel_task if it's going in the wrong direction.
 
 ## Live state
-Each user turn is preceded by a `<state>` block containing only this task's row and the
-worker assigned to it. Treat it as an operational briefing, not part of the human's
-message. The state reflects the moment this turn was sent.
+Each user turn is preceded by a `<state>` block containing the current UTC time, this
+task's row, and the worker assigned to it. Treat it as an operational briefing, not part
+of the human's message. The state reflects the moment this turn was sent. Use the
+"Current UTC time" line as your anchor for judging how old `created_at` or a log line's
+`time` is — do not assume those timestamps are recent.
 
 Be concise — one short paragraph maximum unless detail is requested.
 """
 
 
 _EMPTY_WORKERS_BLOCKS = {"[]", "[\n]"}
+
+
+def _current_time_line() -> str:
+    """Fresh per-turn UTC timestamp so the Foreman has a reliable "now" anchor.
+
+    Must never be memoized or hoisted into the cacheable stable system text —
+    it changes every turn and would otherwise poison the prompt cache.
+    """
+    now = datetime.now(timezone.utc).isoformat(timespec="seconds")
+    return f"Current UTC time: {now}\n\n"
 
 
 def _stable_system_text(primary_repo: str | None, system_prompt_suffix: str | None = None) -> str:
@@ -285,7 +301,7 @@ def build_state_preamble(
     body = f"{workers_section}## Recent tasks\n```json\n{tasks_block}\n```"
     if extra_context:
         body += f"\n\n## Context\n{extra_context}"
-    return f"<state>\n{body}\n</state>"
+    return f"<state>\n{_current_time_line()}{body}\n</state>"
 
 
 def _stable_child_system_text(
@@ -351,7 +367,7 @@ def build_child_state_preamble(
     body = f"{worker_section}## This task\n```json\n{task_block}\n```"
     if extra_context:
         body += f"\n\n## Context\n{extra_context}"
-    return f"<state>\n{body}\n</state>"
+    return f"<state>\n{_current_time_line()}{body}\n</state>"
 
 
 def build_system_prompt(
--- a/backend/tests/test_foreman.py
+++ b/backend/tests/test_foreman.py
@@ -26,7 +26,7 @@ sys.path.insert(0, os.path.dirname(__file__))
 
 import database as database_module
 from _test_config import TEST_DATABASE_URL  # noqa: E402
-from foreman.prompt import FOREMAN_SYSTEM, build_system_prompt
+from foreman.prompt import FOREMAN_SYSTEM, build_state_preamble, build_system_prompt
 from foreman.runner import (
     _fetch_online_workers,  # noqa: E402
     _load_history,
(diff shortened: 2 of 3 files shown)


backend/foreman_core/prompt.py
```diff
@@ -237,6 +243,16 @@
 _EMPTY_WORKERS_BLOCKS = {"[]", "[\n]"}
 
 
+def _current_time_line() -> str:
+    """Fresh per-turn UTC timestamp so the Foreman has a reliable "now" anchor.
+
+    Must never be memoized or hoisted into the cacheable stable system text —
+    it changes every turn and would otherwise poison the prompt cache.
+    """
```
jmelloy, line 251: The function is named `_current_time_line()` (singular 'line') but returns the timestamp followed by `\n\n`, making it two newlines rather than one line terminator. Consider naming it `_current_time_block()` or `_current_time_header()`, or documenting the trailing blank line explicitly, to avoid confusing future readers who may strip or re-add newlines when composing the state block.

backend/foreman_core/prompt.py
```diff
@@ -237,6 +243,16 @@
 _EMPTY_WORKERS_BLOCKS = {"[]", "[\n]"}
 
 
+def _current_time_line() -> str:
+    """Fresh per-turn UTC timestamp so the Foreman has a reliable "now" anchor.
+
+    Must never be memoized or hoisted into the cacheable stable system text —
+    it changes every turn and would otherwise poison the prompt cache.
```
jmelloy, line 250: The docstring warns 'Must never be memoized or hoisted into the cacheable stable system text', but nothing in the code enforces this. If `datetime.now` is ever accidentally replaced with a module-level constant (e.g. during testing or a future refactor), there is no guard. Consider adding a unit test that calls the function twice with a small sleep and asserts the results can differ, or at minimum note this risk in a more prominent comment.

backend/tests/test_foreman.py
```diff
@@ -217,6 +217,20 @@ def test_review_pr_internal_described_as_fallback(self):
...
+        Foreman has a reliable "now" anchor for judging task staleness (#748)."""
+        out = build_state_preamble("[]", "[]")
+        assert "Current UTC time: " in out
+        assert out.index("Current UTC time: ") < out.index("## Current workers")
+
+    def test_current_time_not_in_cacheable_stable_system_text(self):
+        """The dynamic timestamp line must never leak into the cached system
+        prefix, or it would poison the prompt cache on every turn. (The prose
+        that merely refers to the "Current UTC time" label is fine — it's
+        static and doesn't change per turn.)"""
```
jmelloy, line 231: The ordering assertion `out.index("Current UTC time: ") < out.index("## Current workers")` is correct, but it will raise `ValueError` (not a helpful assertion failure) if either substring is absent. Using `pytest.raises` or ensuring the prior `assert "Current UTC time: " in out` runs first (which it does here) mitigates this, but a helper that gives a clearer failure message would improve debuggability.

jmelloy (commented):
- The change is well-scoped and correctly separates the dynamic timestamp from the cacheable system prompt, which is the right architectural decision.
- `_current_time_line()` is a simple, pure function with no side effects or error paths, making it low-risk.
- The prompt documentation updates accurately reflect the new behavior and give the LLM clear guidance on how to use the anchor.
- Test coverage is reasonable: ordering assertion (timestamp precedes state data) and cache-safety assertion are both present.
- Minor concern: the function name `_current_time_line()` implies it returns a single line, but it actually returns the line plus two trailing newlines (`\n\n`), which is a small inconsistency between name and behavior that could confuse future maintainers.

PR #750, opened 2026-07-04
Discord: Route inbound chat to Foreman Phase 5

### LINKED ISSUE #744: Discord integration: Route inbound chat to Foreman (Phase 5)

jmelloy:
Closes #744

### DIFF
--- a/backend/discord/router.py
+++ b/backend/discord/router.py
@@ -1,0 +1,274 @@
+"""Inbound Discord chat routing to Foreman (Phase 5): the consumer side of
+the Gateway websocket (#743/#744) — decides which Foreman session a chat
+message belongs to, forwards it, and lets the existing Foreman chat mirror
+(``discord_notifier.notify_foreman_chat``) post the reply back into the same
+thread.
+
+Thread routing model (reuses the two thread types that already exist):
+    - a message in a thread mapped in ``discord_threads`` is chat about that
+      task — routed with ``task_id`` set, so the reply lands in that task's
+      thread (see ``discord_notifier.notify_foreman_chat``).
+    - a message in a thread mapped in ``discord_foreman_threads``, or in any
+      other wired channel with no task binding, is general/ad-hoc chat for
+      that Pioneer Square guild — routed with ``task_id=None``, so the reply
+      lands in (or creates) today's daily guild thread.
+    - anything else has nowhere to route to and is silently ignored.
+
+Consumes ``discord.gateway.gateway_message_queue``. Enable with the same
+env vars as the Gateway (``DISCORD_GATEWAY_ENABLED`` + ``DISCORD_BOT_TOKEN``).
+"""
+
+from __future__ import annotations
+
+import asyncio
+import logging
+
+from discord.auth import is_member_authorized
+from discord.gateway import gateway_message_queue
+
+logger = logging.getLogger(__name__)
+
+
+async def _resolve_task_session(thread_id: str) -> tuple[str, str] | None:
+    """Return ``(ps_guild_slug, task_id)`` if *thread_id* is a known per-PR/issue thread.
+
+    Picks the most recently created task for that (issue_repo, issue_number)
+    pair when more than one exists (e.g. an execute task and a follow-up
+    review task both linked to the same PR). Never raises.
+    """
+    try:
+        from database import AsyncSessionLocal  # noqa: PLC0415
+        from models import DiscordThread, Guild, Task  # noqa: PLC0415
+        from sqlmodel import col, select  # noqa: PLC0415
+
+        async with AsyncSessionLocal() as db:
+            result = await db.exec(
+                select(DiscordThread.issue_repo, DiscordThread.issue_number).where(
+                    col(DiscordThread.thread_id) == thread_id
+                )
+            )
+            coords = result.first()
+            if not coords:
+                return None
+            issue_repo, issue_number = coords
+
+            result = await db.exec(
+                select(Task.id, Task.guild_id)
+                .where(
+                    col(Task.issue_repo) == issue_repo,
+                    col(Task.issue_number) == issue_number,
+                )
+                .order_by(col(Task.created_at).desc())
+                .limit(1)
+            )
+            task_row = result.first()
+            if not task_row:
+                return None
+            task_id, guild_pk = task_row
+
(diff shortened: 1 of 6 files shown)


backend/discord/router.py
```diff
@@ -0,0 +1,274 @@
...
+    channel, outside of any thread). Never raises.
+    """
+    try:
+        from database import AsyncSessionLocal  # noqa: PLC0415
+        from models import DiscordChannelGuild, DiscordForemanThread  # noqa: PLC0415
+        from sqlmodel import col, select  # noqa: PLC0415
+
+        async with AsyncSessionLocal() as db:
+            result = await db.exec(
+                select(DiscordForemanThread.guild_id).where(
```
jmelloy, line 86: `result.first()` on a `DiscordForemanThread.guild_id` column will return the raw `guild_id` foreign-key value (integer or UUID), not a slug string. The function is supposed to return a Pioneer Square guild slug, so this needs a join to the `Guild` table (like `_resolve_task_session` does) to get `Guild.slug`. Returning the PK as a slug will silently cause incorrect routing.

backend/discord/router.py
```diff
@@ -0,0 +1,274 @@
...
+        human_message,
+        user_id=ps_user_id,
+        task_id=task_id,
+        task_name=f"foreman.discord-chat:{guild_slug}",
+    )
+    reset_foreman_poll(guild_slug)
+
+
+async def _consume_forever(queue: asyncio.Queue) -> None:
+    """Pull inbound messages off *queue* and route them, forever."""
```
jmelloy: The `_consume_forever` loop has no back-off or error budget. `route_inbound_message` swallows all exceptions, so a broken message that causes a fast-path error will be dequeued and 'processed' in a tight loop only bounded by queue throughput. Consider adding a small `await asyncio.sleep(0)` or a rate limiter to yield to the event loop and prevent starvation.

backend/discord/router.py
```diff
@@ -0,0 +1,274 @@
...
+
+    No-op (returns None) unless the Gateway itself would start — there is
+    nothing to consume otherwise. Safe to call more than once — returns the
+    existing task if one is already running.
+    """
+    global _router_task
+    from discord.gateway import _bot_token, _gateway_enabled  # noqa: PLC0415
+
+    if not _gateway_enabled() or not _bot_token():
+        return None
```
jmelloy, line 256: `stop_router` is defined but never called from `main.py` — the shutdown path calls `discord_router_task.cancel()` directly. Either remove `stop_router` or wire it into the lifespan shutdown instead of the manual cancel, to avoid confusion about the intended teardown contract.

jmelloy (commented):
- Phase 5 routing logic is well-structured: extracting shared auth to `discord/auth.py` eliminates duplication and the overall routing model (task thread vs. general/foreman thread) is clearly documented.
- `_resolve_general_session` queries `DiscordForemanThread` and returns `guild_id` directly as a slug, but the model field is named `guild_id` (a foreign key integer/UUID) — it should be joining through `Guild` to get the slug, as `_resolve_task_session` correctly does.
- The `_consume_forever` loop lacks any error-rate limiting or back-off: a persistent DB failure (e.g. DB down) will spin at full speed draining the queue and flooding logs, potentially starving other coroutines.
- The `stop_router` coroutine exists but is never called in `main.py`'s shutdown path — the task is cancelled directly, which is fine, but the dedicated `stop_router` helper is dead code.
- Multiple sequential DB sessions are opened per `_resolve_task_session` call (one for the thread lookup, one for the task, one for the guild slug); collapsing them into a single session with joins would be more efficient and consistent.

PR #754, opened 2026-07-05
Show Discord/API-originated messages in Foreman chat panel

### LINKED ISSUE #753: Foreman chat: sync messages from external clients (Discord, API) to frontend UI in real-time
## Problem

When a message is sent to the Foreman from an external client (e.g. Discord, direct API call), it does not appear in the Foreman chat panel on the frontend web service. The chat history is out of sync between clients.

## Expected Behaviour

Any message sent to the Foreman — regardless of origin (Discord, API, web UI) — should appear in the Foreman chat conversation on the frontend in real-time (or near real-time). Both the inbound user message and the Foreman's reply should be visible.

## Acceptance Criteria

- [ ] Messages sent via Discord that are routed to the Foreman appear in the frontend chat UI
- [ ] The Foreman's replies to those messages also appear in the frontend chat UI
- [ ] The UI updates without requiring a full page refresh (WebSocket or polling)
- [ ] The sender/origin is indicated so the user knows which client originated the message

## Context

- PR #752 added Discord → Foreman routing (inbound chat)
- The frontend already has a Foreman chat panel; it needs to subscribe to or poll for new turns from all sources, not just messages typed in the web UI

jmelloy:
## Summary
- Adds a `source` column to `messages` (`web` / `discord` / `api`) so every Foreman turn is tagged with where it came from, regardless of origin.
- Discord's inbound router (`backend/discord/router.py`) now persists the user's message to the shared conversation store and broadcasts it over the existing WebSocket channel with `source="discord"`, so it shows up in the frontend chat panel live rather than only being forwarded to the Foreman's turn handler.
- The REST message-create endpoint accepts an optional `source` (defaulting to `"web"`) for direct API callers.
- `ChatTab.vue` renders a small "via Discord" / "via Api" label next to non-web messages (`.msg-source` style), so users can see the origin at a glance. Web-originated messages are unaffected (label hidden).

## Test plan
- [ ] Run the alembic migration and confirm `messages.source` defaults to `web` for existing rows
- [ ] Send a message through the Discord bot to a routed guild/task and confirm it appears in the frontend chat panel in real time, labeled "via Discord"
- [ ] Confirm the Foreman's reply to that Discord message also appears in the panel
- [ ] Send a message from the web chat box and confirm no source label appears and existing behavior is unchanged
- [ ] `vue-tsc --noEmit` / `npm run build` in `frontend/` (not run in this environment — no `node_modules` installed)

Closes #753

### DIFF
--- a/backend/alembic/versions/20260705_000000_add_source_to_messages.py
+++ b/backend/alembic/versions/20260705_000000_add_source_to_messages.py
@@ -25,7 +25,7 @@ depends_on: str | Sequence[str] | None = None
 def upgrade() -> None:
     with op.batch_alter_table("messages") as batch_op:
         batch_op.add_column(
-            sa.Column("source", sa.Text(), nullable=True, server_default="web"),
+            sa.Column("source", sa.Text(), nullable=False, server_default="web"),
         )
 
 
--- a/backend/alembic/versions/20260705_000000_add_source_to_messages.py
+++ b/backend/alembic/versions/20260705_000000_add_source_to_messages.py
@@ -1,0 +1,34 @@
+"""add_source_to_messages
+
+Tags each chat message with its origin ("web", "discord", "api") so the
+frontend can show where a message came from (#753). Existing rows predate
+Discord/API routing and were all typed in the browser, so they backfill to
+"web" via the server default.
+
+Revision ID: 20260705_000000_add_source_to_messages
+Revises: 20260704_000004_add_discord_account_links
+Create Date: 2026-07-05
+
+"""
+
+from collections.abc import Sequence
+
+import sqlalchemy as sa
+from alembic import op
+
+revision: str = "20260705_000000_add_source_to_messages"
+down_revision: str | Sequence[str] | None = "20260704_000004_add_discord_account_links"
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+
+def upgrade() -> None:
+    with op.batch_alter_table("messages") as batch_op:
+        batch_op.add_column(
+            sa.Column("source", sa.Text(), nullable=True, server_default="web"),
+        )
+
+
+def downgrade() -> None:
+    with op.batch_alter_table("messages") as batch_op:
+        batch_op.drop_column("source")
--- a/backend/discord/router.py
+++ b/backend/discord/router.py
@@ -213,7 +213,15 @@ async def _route_inbound_message(message: dict) -> None:
     ps_user_id, label = await _resolve_identity(author.get("id"), author.get("username"))
     human_message = f"[Discord] {label}: {content}"
 
-    await _persist_inbound_message(guild_slug, content, user_id=ps_user_id, task_id=task_id)
+    try:
+        await _persist_inbound_message(guild_slug, content, user_id=ps_user_id, task_id=task_id)
+    except Exception:
+        logger.warning(
+            "discord router: failed to persist inbound message guild=%s — forwarding to "
+            "Foreman anyway",
+            guild_slug,
+            exc_info=True,
+        )
 
     from ws_handlers import _trigger_foreman  # noqa: PLC0415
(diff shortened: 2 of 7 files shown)


backend/alembic/versions/20260705_000000_add_source_to_messages.py
```diff
@@ -0,0 +1,34 @@
...
+
+revision: str = "20260705_000000_add_source_to_messages"
+down_revision: str | Sequence[str] | None = "20260704_000004_add_discord_account_links"
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+
+def upgrade() -> None:
+    with op.batch_alter_table("messages") as batch_op:
+        batch_op.add_column(
```
jmelloy, line 27: The column is `nullable=True` but semantically every row should have a source. Consider using `nullable=False` with `server_default="web"` so the DB enforces the invariant and no NULLs can sneak in from raw SQL inserts or future code paths that forget to set the field.

backend/routes/foreman.py
```diff
@@ -565,6 +565,7 @@ class MessageCreate(BaseModel):
     message_type: str = "chat"
     user_id: str | None = None
```
jmelloy, line 567: `source` accepts any arbitrary string. Add a `Literal["web", "discord", "api"]` annotation (or a Pydantic validator) to reject unknown values at the API boundary rather than silently persisting them.

backend/discord/router.py
```diff
@@ -212,6 +213,8 @@ async def _route_inbound_message(message: dict) -> None:
     ps_user_id, label = await _resolve_identity(author.get("id"), author.get("username"))
     human_message = f"[Discord] {label}: {content}"
```
jmelloy, line 215: `_persist_inbound_message` is called without a surrounding try/except, so a DB error (e.g. unknown guild, commit failure) will raise an exception here and prevent `_trigger_foreman` from being called, silently dropping the message from Foreman processing. Either wrap the call in its own try/except or move it after the Foreman trigger.

frontend/src/components/chat-pane/ChatTab.vue
```diff
@@ -105,6 +108,12 @@ function senderLabel(msg: ChatMessage): string {
   return sender.toUpperCase()
 }
 
+function sourceLabel(msg: ChatMessage): string | null {
+  const source = msg.source
+  if (!source || source === 'web') return null
```
jmelloy, line 115: Capitalising the raw source string with `charAt(0).toUpperCase() + source.slice(1)` produces `"Api"` instead of `"API"`. Use a lookup map (`{ discord: 'Discord', api: 'API' }`) to get correct display labels and avoid exposing raw DB values in the UI.

jmelloy (commented):
- The migration adds `source` as `nullable=True` with a `server_default`, but the SQLModel field declares `default="web"` (Python-side only); existing rows will correctly backfill via the server default, though the column remains nullable at the DB level while the model implies a non-null semantic — a `NOT NULL` constraint with a server default would be cleaner.
- `source` is a free-form `str` with no validation or enum constraint anywhere (migration, model, or API schema), so invalid values like `"slack"` or `"internal"` can be persisted silently; a `Literal` type or `Enum` column type would enforce the contract.
- The `_persist_inbound_message` helper is called *before* `_trigger_foreman`, meaning a DB or broadcast failure could surface an exception that bypasses the Foreman trigger entirely — the docstring says "best-effort" but the caller's try/except may not actually swallow this depending on the broader error handling shape.
- `sourceLabel` in `ChatTab.vue` capitalises the raw DB string (e.g. `"Api"` instead of `"API"`), which looks awkward; a small lookup map would give correct display strings.
- No test coverage is added for the new `_persist_inbound_message` path or the `source` field on the REST endpoint, leaving the happy-path and edge cases (unknown guild slug, DB failure mid-commit) untested.

PR #759, opened 2026-07-05
Discord: early thread creation + route child calls (#758)

### LINKED ISSUE #758: Discord: create issue/PR sub-thread earlier and route all child Foreman calls into it
## Problem

Currently Discord sub-threads for an issue or PR are created late (e.g. when a PR is opened). All the intermediate Foreman activity — task creation, redirects, followups, status updates — that belongs to that issue/PR is not captured in the thread.

## Desired Behaviour

1. **Thread creation timing** — Create the Discord sub-thread as soon as the Foreman picks up an issue (e.g. when `claim_github_issue` is called or when the first task for the issue is assigned), not when the PR is finally opened.
2. **Child Foreman calls → thread** — Any subsequent Foreman actions tied to that issue/PR (send_followup, redirect_task, status updates, review dispatch, CI-fix messages, etc.) should post into the existing sub-thread rather than the main channel. This keeps the thread as a complete log of all activity for that issue.
3. **Thread reuse** — If a thread already exists for an issue/PR number, reuse it rather than creating a new one.

## Implementation Notes

- The Discord notifier needs to track `issue_number → thread_id` (could be stored in the DB alongside the task or in a separate discord_threads table).
- The Foreman's issue-first workflow events (claim, assign, followup, redirect, review dispatch) should look up the thread ID and route messages there.
- PR-opened notifications should also go into the thread (and create it if it doesn't exist yet, as a fallback).
- Consider: thread title = `#{issue_number}: {issue_title}`.

jmelloy:
Closes #758

## Changes
- Creates Discord sub-threads when the first task for an issue is assigned (not when PR is opened)
- Routes send_followup, redirect_task, and other Foreman notifications into the existing thread
- Reuses existing threads by issue number
- PR-opened notifications also route into the thread (creates thread if not yet present)

## Files changed
- backend/discord_notifier.py — thread lookup/creation helpers
- backend/foreman/tools.py — wire assign_task, send_followup, redirect_task to use threads
- backend/tests/test_discord_notifier.py — 99 lines of new tests (all 50 pass)

### DIFF
--- a/backend/discord_notifier.py
+++ b/backend/discord_notifier.py
@@ -81,6 +81,9 @@ _COLOURS: dict[str, int] = {
     "task-complete": 0x2ECC71,  # green
     "task-failed": 0xE74C3C,  # red
     "task-cancelled": 0xE74C3C,  # red
+    "task-assigned": 0x1ABC9C,  # teal
+    "task-followup": 0xF1C40F,  # yellow
+    "task-redirect": 0xE67E22,  # orange
     "pr-opened": 0x3498DB,  # blue
     "pr-merged": 0x9B59B6,  # purple
     "pr-closed": 0x95A5A6,  # grey
@@ -110,6 +113,16 @@ def _bot_token() -> str | None:
 
 def _channel_id() -> str | None:
     return os.environ.get("DISCORD_CHANNEL_ID") or None
+
+
+def is_configured() -> bool:
+    """Return True if either the flat webhook or the bot-thread path is set up.
+
+    Cheap, side-effect-free check callers can use to skip expensive prep work
+    (e.g. a live GitHub API call to resolve a thread title) when Discord
+    notifications are disabled entirely.
+    """
+    return bool(_bot_token() or _webhook_url())
 
 
 # ---------------------------------------------------------------------------
--- a/backend/discord_notifier.py
+++ b/backend/discord_notifier.py
@@ -721,5 +721,15 @@ async def notify_existing_thread(
         if thread_id:
             await _post_to_thread(thread_id, event_type, title, description, url=url, color=color)
             return
+        logger.debug(
+            "notify_existing_thread: no thread on record for %s#%s, falling back to flat webhook",
+            issue_repo,
+            issue_number,
+        )
+    elif not is_configured():
+        logger.debug(
+            "notify_existing_thread: Discord not configured, skipping event=%s",
+            event_type,
+        )
 
     await notify(event_type, title, description, url=url, color=color)
--- a/backend/discord_notifier.py
+++ b/backend/discord_notifier.py
@@ -709,12 +709,21 @@ async def notify_existing_thread(
 ) -> None:
     """Post to an issue/PR's Discord thread only if one already exists.
 
-    Unlike ``notify_event``, this never creates a thread — it's for
-    mid-lifecycle notifications (follow-up, redirect) that should land in
-    the thread ``assign_task`` already created when the task started. If no
-    thread has been persisted yet (e.g. Discord was unconfigured at task
-    start), falls back to the flat webhook channel. Silent no-op / never
-    raises, same guarantees as ``notify_event``.
+    This is deliberately **not** a thin wrapper around ``notify_event``:
+    ``notify_event`` always creates a thread when none exists yet, naming it
+    from the *current* call's ``title`` (or an explicit ``thread_name``). For
+    mid-lifecycle notifications like follow-up/redirect, that title is a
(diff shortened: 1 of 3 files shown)


backend/foreman/tools.py
```diff
@@ -404,6 +406,105 @@ async def maybe_post_plan_comment(guild_id: str, task_id: str, last_text: str) -
...
+    task_id: str,
+    fallback_title: str,
+) -> None:
+    """Create (or reuse) the issue's Discord thread as soon as a task is assigned.
+
+    This is what moves thread creation earlier in the lifecycle — previously
+    the first per-issue Discord thread only appeared when a PR was opened.
+    Fire-and-forget via ``spawn``: Discord/GitHub failures must never affect
+    task assignment. Skips the live GitHub title lookup entirely when Discord
+    notifications aren't configured.
```
jmelloy: The `thread_name` is built as `f'#{issue_number}: {title}'` where `title` comes from a live GitHub API call. Discord enforces a 100-character limit on thread names; a long issue title will silently fail or raise. Truncate to ~90 characters before passing, e.g. `title[:90]`.

backend/foreman/tools.py
```diff
@@ -404,6 +406,105 @@ async def maybe_post_plan_comment(guild_id: str, task_id: str, last_text: str) -
         logger.warning("plan comment failed for task %s: %s", task_id, exc)
 
 
+async def _resolve_issue_title(
+    guild_id: str, issue_repo: str, issue_number: int, fallback: str
+) -> str:
+    """Best-effort live GitHub title lookup for Discord thread naming.
+
+    Falls back to *fallback* (the task's own name/description) on a missing
+    token or any API failure, so a slow or failing GitHub call never blocks
```
jmelloy, line 415: `_resolve_issue_title` is awaited synchronously inside `notify_discord_task_assigned`, which is itself the coroutine given to `spawn`. If the GitHub API is slow this adds latency to the spawned task. Consider moving the title resolution entirely inside the spawned coroutine and ensuring `spawn` does not block the caller (verify `util.tasks.spawn` truly fire-and-forgets).

backend/discord_notifier.py
```diff
@@ -683,3 +696,30 @@ async def notify_event(
...
+) -> None:
+    """Post to an issue/PR's Discord thread only if one already exists.
+
+    Unlike ``notify_event``, this never creates a thread — it's for
+    mid-lifecycle notifications (follow-up, redirect) that should land in
+    the thread ``assign_task`` already created when the task started. If no
+    thread has been persisted yet (e.g. Discord was unconfigured at task
+    start), falls back to the flat webhook channel. Silent no-op / never
+    raises, same guarantees as ``notify_event``.
+    """
```
jmelloy, line 727: When `_bot_token()` is set but `issue_repo`/`issue_number` are absent, `notify_existing_thread` falls through to `notify()` (flat webhook). If neither bot token, webhook, nor issue coords are present this is a silent no-op — which is the stated intent — but it may be worth adding a debug log here so operators can distinguish 'Discord unconfigured' from 'thread not found' when diagnosing missing notifications.

jmelloy (commented):
- The PR correctly moves Discord thread creation earlier (on task assignment rather than PR open) and adds `notify_existing_thread` for mid-lifecycle notifications — the design is sound.
- Two near-identical `spawn(notify_discord_task_assigned(...))` blocks appear in `_exec_one_tool` for different assign paths; the duplication could be extracted but is not a correctness bug.
- `notify_discord_task_assigned` does a live GitHub API call (`_resolve_issue_title`) inside the coroutine that is passed to `spawn`, but if `spawn` runs it in the same event loop, a slow/failing GitHub call still happens — the `is_configured()` guard helps but the fallback path may still introduce latency on every assign.
- The `instructions[:500]` truncation in `notify_discord_followup` / `notify_discord_redirect` is a reasonable guard, but there is no equivalent guard on the `thread_name` built from the live issue title, which could exceed Discord's 100-character thread name limit.
- Test coverage is good (4 targeted tests for `notify_existing_thread`); however there are no tests for `notify_discord_task_assigned` / `notify_discord_followup` / `notify_discord_redirect` in `foreman/tools.py`, leaving the `spawn` wiring untested.

backend/discord_notifier.py
```diff
@@ -683,3 +696,30 @@ async def notify_event(
 
     # Fallback: flat webhook
     await notify(event_type, title, description, url=url, color=color)
+
+
+async def notify_existing_thread(
```
jmelloy, line 701: Why does this exist? Seems like the existing create or update would do fine

PR #771, opened 2026-07-06
Discord: /worker-spawn slash command (#769)

### LINKED ISSUE #769: Discord: /worker-spawn slash command
## Summary

Add a `/worker-spawn` Discord slash command that allows users to spawn a new worker from Discord.

## Acceptance Criteria

- A `/worker-spawn` slash command is registered with the Discord bot
- The command accepts optional parameters (e.g. repos, tools/agent type)
- Invoking the command triggers worker creation via the existing backend API
- The bot responds with confirmation (worker ID, status) once the worker is online or the request is queued
- Follows the same patterns as existing slash commands (`/connect-account`, `/join-channel`, etc.)

## Context

Requested via Discord by @jmelloy. See existing slash command implementations in the codebase for reference patterns.

jmelloy:
## Summary
- Adds a `/worker-spawn` Discord slash command (with optional `repos` and `tools` comma-separated string options) alongside the existing `/ps`, `/join-channel`, `/leave-channel`, and `/connect-account` commands in `backend/routes/discord.py`.
- Requires the invoking Discord user to have already linked their account via `/connect-account` (checked against `DiscordAccountLink`), returning an ephemeral error otherwise.
- Reuses the existing `foreman.tools.spawn_worker` helper — the same worker-spawn logic the Foreman AI tool and worker-lifecycle replacement path already call — to pre-register the worker and start its Docker container.
- Replies with an ephemeral embed containing the new worker's id and a `queued`/`online` status derived from the worker's current DB state, or a clear error message on failure (unlinked account, unknown guild, or a tool-level error such as missing repos).
- Registers the new command's schema in `scripts/register_discord_commands.py`.

## Test plan
- [x] `ruff check .` / `ruff format . --check` — clean
- [x] `cd backend && python -m pytest tests/test_discord_interactions.py -k worker_spawn -v` — 4 new tests pass
- [x] `cd backend && python -m pytest` — 505 passed, 2 skipped; remaining errors are pre-existing DB-connection failures from the `postgres-test` container not being available in this sandbox (no Docker), unrelated to this change

Closes #769

### DIFF
--- a/backend/routes/discord.py
+++ b/backend/routes/discord.py
@@ -23,6 +23,7 @@ Registered slash commands (see scripts/register_discord_commands.py):
     /join-channel <channel> [guild]     — wire a Discord channel to a Pioneer Square guild
     /leave-channel [channel]            — remove a channel's Pioneer Square guild binding
     /connect-account                    — link your Discord account to Pioneer Square
+    /worker-spawn [repos] [tools]       — spawn a new worker (requires a connected account)
 """
 
 from __future__ import annotations
@@ -43,6 +44,7 @@ from events import broadcast_msg
 from fastapi import APIRouter, BackgroundTasks, Request, Response
 from models import (
     Agent,
+    DiscordAccountLink,
     DiscordChannelGuild,
     DiscordConnectToken,
     Guild,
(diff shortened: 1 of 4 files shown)


backend/routes/discord.py
```diff
@@ -662,6 +664,99 @@ async def _cmd_connect_account(interaction: dict) -> None:
...
+                return
+
+            guild_result = await db.exec(select(Guild).where(col(Guild.slug) == guild_slug))
+            guild = guild_result.one_or_none()
+            if not guild:
+                await _send_followup(token, content=f"Pioneer guild `{guild_slug}` not found.")
+                return
+
+            from foreman.tools import spawn_worker  # noqa: PLC0415
+
```
jmelloy: The `ps_user_id` from the account-link lookup is fetched but never used. `spawn_worker` is called with only `guild_id`/`guild_pk`, so there is no binding between the spawned worker and the requesting user. At minimum the linked PS user identity should be passed to `spawn_worker` to associate ownership, otherwise any Discord user with a linked account spawns workers in a shared pool with no attribution.

backend/routes/discord.py
```diff
@@ -662,6 +664,99 @@ async def _cmd_connect_account(interaction: dict) -> None:
...
+    except Exception:
+        logger.exception("discord: /worker-spawn failed")
+        await _send_followup(token, content="Failed to spawn worker.")
+        return
+
+    status = "online" if state not in ("offline", "spawn_failed") else "queued"
+    embeds = [
+        {
+            "title": f"Worker Spawned: {worker_id}",
+            "description": (
```
jmelloy, line 749: The status derivation is inverted for a freshly-spawned worker. A worker that just left `spawn_worker` will typically be in a `spawning` or `queued` state — neither `offline` nor `spawn_failed` — so this expression will incorrectly report `online`. Consider mapping explicitly: `status = state if state in ('online',) else 'queued'` or similar.

backend/tests/test_discord_interactions.py
```diff
@@ -557,6 +557,186 @@ async def fake_patch(path, payload):
...
+    embed = sent[0]["embeds"][0]
+    assert "w-abc123" in embed["title"]
+    assert "online" in embed["description"]
+
+
+@pytest.mark.asyncio
+async def test_cmd_worker_spawn_requires_connected_account(monkeypatch):
+    """_cmd_worker_spawn refuses when the Discord account isn't linked."""
+    monkeypatch.setenv("DISCORD_APPLICATION_ID", "app-id")
+    monkeypatch.setenv("DISCORD_BOT_TOKEN", "tok")
```
jmelloy, line 679: The patch target `foreman.tools.spawn_worker` is wrong given that `_cmd_worker_spawn` imports `spawn_worker` via `from foreman.tools import spawn_worker` inside the function. Once `foreman.tools` is imported the local binding is what gets called; the patch must target `routes.discord.spawn_worker` (or the import must be at module level) to reliably intercept the call in all execution orders.

backend/routes/discord.py
```diff
@@ -662,6 +664,99 @@ async def _cmd_connect_account(interaction: dict) -> None:
...
+                inp={"repos": repos, "tools": tools_list or None},
+                guild_id=guild_slug,
+                guild_pk=guild.id,
+                db=db,
+            )
+
+            if is_error:
+                await _send_followup(token, content=f"Failed to spawn worker: {result_text}")
+                return
+
```
jmelloy, line 732: Passing `tools_list or None` creates an asymmetry: `repos` is always a list (possibly empty) while `tools` becomes `None` when empty. If `spawn_worker` has different handling for `None` vs `[]` for repos vs tools this could produce unexpected behaviour. Prefer a consistent convention — either always pass lists or always use `None` for "not provided" for both parameters.

jmelloy (commented):
- The command correctly gates worker spawning behind an existing `/connect-account` link check, but the `ps_user_id` retrieved from the DB is never actually used — authorization confirms the link exists but doesn't pass the linked user identity to `spawn_worker`, meaning any Discord user could spawn workers on behalf of whatever account is linked to their ID without further validation.
- The status logic is inverted: a freshly spawned worker will almost certainly be in a transitional/queued state, yet the code labels it `online` for any state not in `('offline', 'spawn_failed')` — including states like `spawning` or `starting` — which will mislead users.
- `spawn_worker` is imported inside the function body at call time with a `noqa` suppression; the mock in tests patches `foreman.tools.spawn_worker` rather than `routes.discord.spawn_worker`, meaning the test mock may not intercept the call correctly if the module has already been imported, creating a fragile test setup.
- The `tools_list or None` passed to `spawn_worker` means an empty tools list becomes `None`, but `repos` is always passed as a list (possibly empty) — this asymmetry could trigger validation errors in `spawn_worker` when no repos are given, which the test for the error path actually relies on but isn't clearly documented as intentional behavior.
- No rate-limiting or authorization check prevents a linked user from spawning unlimited workers via the Discord command, unlike more privileged operations that typically require additional role/permission checks.

PR #772, opened 2026-07-06
feat: collapsible turn summaries for task log and Discord streams

### LINKED ISSUE #770: Task log: summarize turns with tool call details behind a disclosure
## Summary

Claude Code CLI summarizes agent actions by turn (e.g. "5 tool calls") rather than showing every individual tool call. We want to adopt that same UX pattern in Pioneer Square:

- **Summary line** per turn: e.g. "Turn 3 — 5 tool calls (edit, bash, edit, read, bash)"
- **Disclosure/expandable** section behind the summary that reveals the full tool call details

## Where to apply

1. **Task view** (frontend task log tab) — collapse each agent turn into a summary row with an expand toggle
2. **Discord thread** — post a compact summary per turn; use Discord's spoiler tags (`||details||`) or a follow-up message structure to hide the verbose output

## Acceptance criteria

- [ ] Task log UI shows a collapsible row per agent turn summarizing the number and types of tool calls
- [ ] Expanding the row reveals the full tool call log lines for that turn
- [ ] Discord thread posts a compact summary line per turn (e.g. "▶ Turn 3: edit × 2, bash × 1, read × 1")
- [ ] Full detail is accessible in Discord (e.g. spoiler tag or threaded reply)
- [ ] No regression on existing task log display for tasks with no tool calls

## Reference

Claude Code CLI does something similar — it groups tool calls per turn and shows a summary, with details available on demand.

jmelloy:
## Summary
- Groups consecutive `tool_use`/`tool_result` log entries into collapsible "Turn N" rows in the task log UI (`LogList.vue`), collapsed by default with a summary like `Turn 3 — edit × 2, bash × 1 (3 tool calls)`; expanding reveals the full log lines for that turn via a newly-extracted `LogLine.vue` component.
- Mirrors the same turn-grouping in the Discord task-stream mirror (`discord_notifier.py`): posts one compact `▶ Turn N: edit × 2, bash × 1` line per turn instead of dumping every raw tool call/result line, with turn numbers staying sequential across flush batches.
- Non-tool lines (assistant text, thinking, status) pass through unchanged in both places — fully backward-compatible for tasks with no tool calls.
- Fixes a dropped assertion in `test_flush_task_stream_noop_when_no_buffer` found while finishing this branch.

## Test plan
- [x] `npm run test` (frontend) — 110 passed, including new `LogList.spec.ts` covering grouping, sequential numbering, and expand/collapse
- [x] `npm run type-check` / `npm run build` (frontend)
- [x] `python -m pytest tests/test_discord_notifier.py` (backend) — 67 passed, including new turn-formatting tests
- [x] Manually verified default-collapsed / no-regression display for logs with no tool calls

Closes #770

### DIFF
--- a/backend/discord_notifier.py
+++ b/backend/discord_notifier.py
@@ -647,14 +647,25 @@ def _stream_enabled() -> bool:
 
 
 @dataclass
+class _StreamEntry:
+    """One buffered stream line plus the structured tool detail that produced it."""
+
+    line: str
+    detail: dict | None = None
+
+
+@dataclass
 class _StreamBuffer:
     """In-memory accumulator for one task's pending stream lines."""
 
     guild_id: str
-    lines: list[str] = field(default_factory=list)
+    lines: list[_StreamEntry] = field(default_factory=list)
     size: int = 0
     flush_task: asyncio.Task | None = None
     lock: asyncio.Lock = field(default_factory=asyncio.Lock)
+    # Running count of turns already posted for this task, so turn numbers stay
+    # sequential across flush batches instead of resetting to 1 each time.
+    turn_count: int = 0
 
 
 # Keyed by task_id. Holds only tasks with un-flushed lines; entries are removed
@@ -760,6 +771,46 @@ def _chunk_content(text: str, size: int) -> list[str]:
     return chunks
 
 
+def _format_stream_entries(entries: list[_StreamEntry], start_turn: int) -> tuple[str, int]:
+    """Compact consecutive tool_use/tool_result entries into per-turn summary lines.
+
+    Mirrors the frontend's turn grouping (see LogList.vue): a run of consecutive
+    tool_use/tool_result entries between non-tool lines (assistant text, thinking,
+    turn-completion markers) becomes one ``▶ Turn N: edit × 2, bash × 1`` line
+    instead of dumping every raw tool call and result into the thread. Non-tool
+    lines pass through unchanged. Returns the formatted text and the turn count
+    to carry into the next batch, so numbering stays sequential across flushes.
+    """
+    output: list[str] = []
+    group: list[_StreamEntry] = []
+    turn_number = start_turn
+
+    def flush_group() -> None:
+        nonlocal turn_number
+        if not group:
+            return
+        turn_number += 1
+        counts: dict[str, int] = {}
+        for entry in group:
+            detail = entry.detail or {}
+            if detail.get("toolType") == "tool_use":
+                name = str(detail.get("name") or "tool").lower()
+                counts[name] = counts.get(name, 0) + 1
+        parts = ", ".join(f"{name} × {count}" for name, count in counts.items())
+        output.append(f"▶ Turn {turn_number}: {parts or 'tool activity'}")
+        group.clear()
+
+    for entry in entries:
+        detail = entry.detail or {}
+        if detail.get("toolType") in ("tool_use", "tool_result"):
+            group.append(entry)
+        else:
+            flush_group()
+            output.append(entry.line)
+    flush_group()
+    return "\n".join(output), turn_number
+
+
 async def _post_task_stream(guild_id: str, task_id: str, content: str) -> None:
     """Post *content* into *task_id*'s per-task thread as silent messages. Never raises."""
     channel = await _resolve_channel_for_guild(guild_id)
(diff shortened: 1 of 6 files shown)


backend/discord_notifier.py
```diff
@@ -760,6 +771,46 @@ def _chunk_content(text: str, size: int) -> list[str]:
...
+    turn_number = start_turn
+
+    def flush_group() -> None:
+        nonlocal turn_number
+        if not group:
+            return
+        turn_number += 1
+        counts: dict[str, int] = {}
+        for entry in group:
+            detail = entry.detail or {}
```
jmelloy, line 795: The `flush_group` closure mutates `group` via `group.clear()` but `group` is a local list variable in the enclosing scope — this works because `clear()` mutates in place rather than reassigning, but it's subtle. Consider using `nonlocal group` and reassigning `group = []` for clarity, or restructuring to avoid the closure mutation pattern.

frontend/src/components/log-pane/LogList.vue
```diff
@@ -3,150 +3,126 @@
...
-    // _summarize_lines shows all lines when count <= 4; only expand when output is truncated
-    return (log.detail.output?.trim().split('\n').length ?? 0) > 4
+const expandedTurns = ref<Set<number>>(new Set())
+
+interface LineItem {
+  type: 'line'
+  log: LogEntry
+  index: number
+}
+
```
jmelloy, line 59: The `expandedIdx` stores the raw index into `props.logs`, but when a TurnItem is expanded and its inner `LogLine` entries are shown, those entries also use the same `expandedIdx` mechanism. If a user expands a turn then clicks a line inside it, the turn's `LogLine` index could collide with a `LineItem` index from a different part of the list, causing unexpected detail panels to open/close. Consider namespacing expanded state (e.g., `{kind: 'line'|'turn-line', index}`).

frontend/src/components/log-pane/__tests__/LogList.spec.ts
```diff
@@ -0,0 +1,74 @@
+import { describe, expect, it } from 'vitest'
+import { mount } from '@vue/test-utils'
+import LogList from '../LogList.vue'
+import type { LogEntry } from '../../../types'
+
+function toolLog(name: string, toolType: 'tool_use' | 'tool_result' = 'tool_use'): LogEntry {
+  return {
+    line: `▶ ${name}`,
+    timestamp: '2026-01-01T00:00:00Z',
```
jmelloy, line 9: The `toolLog` helper uses `name` as both the `line` text suffix and the `detail.name` field, but `detail.name` for `tool_result` entries is typically absent or refers to the tool that produced the result — passing the same `name` for `tool_result` entries slightly misrepresents the real data shape and could mask bugs in name-based grouping logic.

backend/discord_notifier.py
```diff
@@ -760,6 +771,46 @@ def _chunk_content(text: str, size: int) -> list[str]:
...
+            if detail.get("toolType") == "tool_use":
+                name = str(detail.get("name") or "tool").lower()
+                counts[name] = counts.get(name, 0) + 1
+        parts = ", ".join(f"{name} × {count}" for name, count in counts.items())
+        output.append(f"▶ Turn {turn_number}: {parts or 'tool activity'}")
+        group.clear()
+
+    for entry in entries:
```
jmelloy, line 805: Tool name casing is normalized with `.lower()` here (producing `edit`, `bash`) but the Discord output uses those lowercased names directly in the summary line. The frontend `formatTurnCounts` also lowercases via `(log.detail.name || 'tool').toLowerCase()`. This is consistent, but worth a comment noting the intentional normalization so future contributors don't accidentally display mixed-case names.

jmelloy (commented):
- The PR successfully extracts `LogLine.vue` from `LogList.vue` and adds collapsible turn-grouping for tool calls in both the frontend UI and Discord notifier, with good test coverage for both.
- The `expandedIdx` is a flat index into `props.logs`, but after grouping, indices inside a `TurnItem` are used directly from the original array position — this works correctly since `index` is always the original array offset, not a grouped offset.
- The Discord `_format_stream_entries` function only counts `tool_use` entries for the summary label but processes both `tool_use` and `tool_result` into a group — this means `tool_result`-only runs (edge case) would show `tool activity` with no counts, which is fine but could be surprising.
- The `isExpandable` logic in `LogLine.vue` for `tool_result` checks `output.trim().split('\n').length > 4`, which will throw if `output` is `null` or `undefined` without optional chaining — the nullish coalescing `?? 0` already handles this, so it's safe, but the code looks fragile at a glance.
- The `expandedTurns` state uses a `Set<number>` with turn numbers that reset to 0 when `reset()` is called, but turn numbers are derived from `groupedLogs` which is a computed property recalculated on every `logs` change — if logs change mid-session and turn count resets, there could be stale expanded state pointing to wrong turns.

PR #776, opened 2026-07-06
fix: branch names truncate task ID to 4 chars (#775)

### LINKED ISSUE #775: fix: branch names truncate task ID to 4 chars instead of full 6
## Problem

When a worker creates a branch for a task, the task ID suffix in the branch name is truncated to 4 characters instead of using the full 6-character task ID.

**Example:** Task `t-0jlqri` produces branch `claude/discord-associate-pr-threads-with-issue-thread-773-t-0jlq` — the suffix is `-t-0jlq` (4 chars) instead of `-t-0jlqri` (6 chars).

This makes it harder to:
- Extract the task ID from the branch name reliably (a 4-char suffix has more collision risk)
- Parse the issue number from the branch name (the regex `-(\d+)-t-[a-z0-9]+$` may be less precise with a shorter suffix)

## Fix

Find where branch names are generated (likely in the worker host process or a shared utility) and change the task ID slice from `[:4]` (or equivalent) to use the full task ID (all 6 chars, or the full string without truncation).

## Acceptance criteria

- [ ] New branches include the full 6-character task ID suffix (e.g. `-t-0jlqri` not `-t-0jlq`)
- [ ] Existing branch-name parsing regex is updated to match 6 chars if it was hardcoded to 4

jmelloy:
## Summary
- `worker/pioneer_worker/worker.py` built new-task branch names with `task_id[:6]`, but `task_id` is `"t-" + 6 random chars` (8 chars total). Slicing to 6 chars kept only 4 of the actual ID characters (e.g. `t-0jlqri` → `t-0jlq`), risking collisions between different tasks.
- Fixed to use the full `task_id` in the branch name (e.g. `claude/<slug>-t-0jlqri`).
- Verified there's no branch-parsing regex elsewhere hardcoded to a 4-char task ID; the one existing regex (`foreman/pioneer_foreman/foreman.py`) already uses `[a-z0-9]+` (one-or-more), so no change needed there.

## Test plan
- [x] `python -m pytest worker/tests/ -k branch` passes
- [x] `python -m pytest worker/tests/test_followup_branch_reuse.py` passes

Closes #775

### DIFF
--- a/worker/pioneer_worker/worker.py
+++ b/worker/pioneer_worker/worker.py
@@ -1887,7 +1887,7 @@ class Worker:
         name = task.get("name") or desc
         # On follow-ups we must continue on the existing branch — the original
         # worker pushed it and the foreman wants the same PR updated.
-        branch = followup_branch or f"claude/{_slug(name)}-{task_id[:6]}"
+        branch = followup_branch or f"claude/{_slug(name)}-{task_id}"
         work_dir = os.path.join(self.cfg.work_dir, self.cfg.guild_id, self.cfg.worker_id, task_id)
         logger.info(
             "Task %s branch=%s work_dir=%s followup=%s", task_id, branch, work_dir, is_followup


worker/pioneer_worker/worker.py
```diff
@@ -1887,7 +1887,7 @@ async def _execute_task(self, task: dict, agent: Agent) -> None:
         name = task.get("name") or desc
         # On follow-ups we must continue on the existing branch — the original
         # worker pushed it and the foreman wants the same PR updated.
-        branch = followup_branch or f"claude/{_slug(name)}-{task_id[:6]}"
+        branch = followup_branch or f"claude/{_slug(name)}-{task_id}"
```
jmelloy, line 1890: Consider adding an explicit regression test that asserts `task_id in branch` (or that the branch ends with the full `task_id`) to prevent this truncation bug from being reintroduced silently in the future.

jmelloy (commented):
- The fix correctly addresses the collision risk by using the full `task_id` instead of a truncated 6-char slice that only preserved 4 meaningful characters after the `t-` prefix.
- The change is minimal and targeted, with no unintended side effects on the branch-name parsing regex in `foreman.py` (which already uses `[a-z0-9]+`).
- Branch names will now be slightly longer (e.g. `claude/<slug>-t-0jlqri` instead of `claude/<slug>-t-0jlq`), but this is well within Git's branch-name length limits and improves uniqueness significantly.
- No tests for the original truncation bug were added to prevent regressions, though the PR description references existing passing tests.
- The fix is correct and safe to merge; a minor suggestion would be to add a dedicated regression test asserting the full `task_id` appears in the generated branch name.

PR #779, opened 2026-07-06
fix: worker message content truncated before Foreman (#778)

### LINKED ISSUE #778: bug: Discord inbound message content truncated before reaching Foreman/Claude
## Bug Report

When a Discord message is routed through the webhook/gateway pipeline to the Foreman, the original message text is being truncated — Claude receives a cut-off version of what was actually sent.

## Reported by
@jmelloy via Discord

## Expected behaviour
The full Discord message content should be passed through the pipeline intact to the Foreman.

## Actual behaviour
The message text is truncated somewhere in the Discord gateway → backend → Foreman chain before Claude sees it.

## Investigation notes
- Look at the Discord gateway inbound message handler (likely under `discord/` or `backend/`)
- Check where the message content is extracted from the Discord payload and forwarded
- Look for any string slicing, field length limits, or truncation logic applied to the message body
- The `[Discord]` prefix line injected by the gateway may be relevant — check how it's assembled
- Check if there's a character/byte limit being applied at the HTTP layer, DB insert, or message queue step

## Acceptance criteria
- [ ] Full Discord message content reaches the Foreman unchanged
- [ ] No truncation at any point in: Discord gateway → backend API → Foreman message delivery
- [ ] Add a test or log assertion to confirm full message passthrough

jmelloy:
## Summary

The originally-reported root cause (Discord inbound pipeline) turned out to be clean — traced end to end from `discord/gateway.py` through `discord/router.py`, `ws_handlers._trigger_foreman`, and `foreman/runner.py`, and the full message content is preserved at every step there.

The actual truncation was on the **worker → Foreman** path: `backend/ws_handlers.py`'s `handle_task_complete` and `handle_task_followup_done` capped the worker's final agent output (`lastText`, e.g. Claude/Codex/Pi's last message before hitting `max_turns`) at 200 characters before embedding it into the `foreman-trigger` message:

```python
last_text_snippet = f' Last output: "{last_text[:200]}".' if last_text else ""
```

This meant the Foreman only ever saw the first 200 characters of a worker's final output when deciding how to follow up on a task that hit its max-turns limit — losing potentially important context about what the worker was doing when it ran out of turns.

Verified upstream that the worker side (`claude_runner.py`, `codex_runner.py`, `pi_runner.py`) already captures the full, untruncated final text — the `[:200]` slice in `ws_handlers.py` was the only truncation point in the whole worker→Foreman delivery chain.

## Fix

Removed the `[:200]` slice in both `handle_task_complete` and `handle_task_followup_done` so the full `lastText` reaches the Foreman.

## Test plan
- [x] Added `test_task_complete_max_turns_does_not_truncate_last_text` and `test_followup_done_max_turns_does_not_truncate_last_text` in `backend/tests/test_worker_foreman_notify.py`, asserting a >200-char `lastText` survives intact in the message forwarded to `_trigger_foreman`.
- [x] `python -m pytest tests/test_worker_foreman_notify.py --collect-only` — tests collect cleanly (DB-backed suite requires the `postgres-test` container, unavailable in this sandbox).

Closes #778

### DIFF
--- a/backend/ws_handlers.py
+++ b/backend/ws_handlers.py
@@ -926,9 +926,7 @@ async def handle_task_complete(ctx: WSContext, data: dict) -> None:
 
     task_uid = await _task_user_id(ctx.db, task_id)
     pr_line = f" PR: {pr_url}." if pr_url else ""
-    last_text_snippet = (
-        f' Last output: "{_format_last_output(last_text)}".' if last_text else ""
-    )
+    last_text_snippet = f' Last output: "{_format_last_output(last_text)}".' if last_text else ""
     if pr_url:
         # PR exists: lifecycle is driven by GitHub webhooks, not the foreman.
         # The task will be auto-finalized on merge or auto-failed on close without merge.
--- a/backend/ws_handlers.py
+++ b/backend/ws_handlers.py
@@ -914,7 +914,7 @@ async def handle_task_complete(ctx: WSContext, data: dict) -> None:
 
     task_uid = await _task_user_id(ctx.db, task_id)
     pr_line = f" PR: {pr_url}." if pr_url else ""
-    last_text_snippet = f' Last output: "{last_text[:200]}".' if last_text else ""
+    last_text_snippet = f' Last output: "{last_text}".' if last_text else ""
     if pr_url:
         # PR exists: lifecycle is driven by GitHub webhooks, not the foreman.
         # The task will be auto-finalized on merge or auto-failed on close without merge.
@@ -1021,7 +1021,7 @@ async def handle_task_followup_done(ctx: WSContext, data: dict) -> None:
     task_uid = await _task_user_id(ctx.db, task_id)
 
     if stop_reason == "max_turns":
-        last_text_snippet = f' Last output: "{last_text_fud[:200]}".' if last_text_fud else ""
+        last_text_snippet = f' Last output: "{last_text_fud}".' if last_text_fud else ""
         human_msg = (
             f"[followup-done/max-turns] Worker {worker_id_msg} follow-up for task {task_id} "
             f"hit Claude's max-turns limit before finishing. Partial work committed.{last_text_snippet} "
--- a/backend/ws_handlers.py
+++ b/backend/ws_handlers.py
@@ -118,6 +118,18 @@ def _parse_pr_url(pr_url: str | None) -> tuple[int | None, str | None]:
     return number, f"{owner}/{repo}"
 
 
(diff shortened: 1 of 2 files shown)


backend/ws_handlers.py
```diff
@@ -914,7 +914,7 @@ async def handle_task_complete(ctx: WSContext, data: dict) -> None:
 
     task_uid = await _task_user_id(ctx.db, task_id)
     pr_line = f" PR: {pr_url}." if pr_url else ""
-    last_text_snippet = f' Last output: "{last_text[:200]}".' if last_text else ""
+    last_text_snippet = f' Last output: "{last_text}".' if last_text else ""
```
jmelloy: Removing the cap entirely introduces an unbounded string into the Foreman trigger message. Consider replacing the hard 200-char truncation with a more generous but still finite limit (e.g. 2000–4000 chars) to guard against pathological cases where a worker emits very large last outputs, which could exceed WebSocket frame limits or inflate Foreman LLM context unexpectedly.

backend/ws_handlers.py
```diff
@@ -1021,7 +1021,7 @@ async def handle_task_followup_done(ctx: WSContext, data: dict) -> None:
     task_uid = await _task_user_id(ctx.db, task_id)
 
     if stop_reason == "max_turns":
-        last_text_snippet = f' Last output: "{last_text_fud[:200]}".' if last_text_fud else ""
+        last_text_snippet = f' Last output: "{last_text_fud}".' if last_text_fud else ""
```
jmelloy: Same unbounded concern as in `handle_task_complete`: `last_text_fud` is now inserted without any size guard. A shared helper like `_format_last_output(text, max_chars=4000)` would enforce a consistent policy in both places and make future changes easier.

jmelloy (commented):
- The fix correctly removes the arbitrary 200-character truncation of `last_text` in both `handle_task_complete` and `handle_task_followup_done`, ensuring the Foreman receives the full worker output when a task hits `max_turns`.
- Two regression tests are added that construct strings longer than 200 chars and assert the full content appears in the Foreman trigger message — good coverage for the exact bug.
- **Potential concern**: removing the cap entirely means an unbounded string is now embedded directly into the Foreman trigger message; if a worker produces a very large last output (e.g. tens of KB), this could cause issues with message size limits or downstream prompt bloat.
- The PR description mentions verifying the Discord inbound pipeline was clean, but no test or code change is included for that path — the scope is appropriately narrowed, though the claim is unverifiable from the diff alone.
- Overall the change is minimal and targeted; the root cause analysis and fix are consistent with each other.

PR #783, opened 2026-07-06
feat: worker line-summarizer preserve full content (#781)

### LINKED ISSUE #781: feat: worker line-summarizer should preserve full content for Foreman-facing paths
the agent found that `claude_runner`, `codex_runner`, and `pi_runner` each summarize/truncate agent stdout before emitting `TaskLog` rows, meaning the Foreman's `get_task_status` sees a pre-digested summary even before hitting the `MAX_TOOL_RESULT_CHARS` cap.

jmelloy:
Closes #781

### DIFF
--- a/backend/foreman/tools.py
+++ b/backend/foreman/tools.py
@@ -24,7 +24,7 @@ import discord_notifier
 from database import get_db
 from events import broadcast, broadcast_msg, emit_terminal_line
 from foreman_core.llm import get_foreman_model, make_anthropic_client
-from foreman_core.message_utils import _json_default
+from foreman_core.message_utils import _json_default, truncate_tool_result
 from foreman_core.tools_schema import (
     FOREMAN_TOOLS,  # noqa: F401 — re-exported for test compatibility
 )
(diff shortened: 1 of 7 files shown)


jmelloy (commented):
- The PR adds a `data` field to task log entries so the Foreman can see full, untruncated tool output instead of only the truncated display summary — a useful observability improvement.
- Parse functions in `codex_runner.py` and `pi_runner.py` are correctly updated to return `(text, detail, ...)` tuples, and all call sites and tests are updated consistently.
- `_full_log_content` in `tools.py` applies `truncate_tool_result` to the extracted full content before surfacing it, which partially defeats the stated purpose of preserving full content — the name and docstring imply it returns untruncated content, but it is truncated at the consumer.
- The `EmitFn` type alias is weakened to `Callable[..., Awaitable[None]]` to avoid breaking existing callers, but this loses static type safety; a proper overload or a `TypedDict`/`Protocol` would be safer.
- Test coverage for the new `data` field in `_full_log_content` / `get_task_status` path is absent; only the parser-level detail preservation is tested.

PR #784, opened 2026-07-06
feat: expose TaskLog.data in get_task_status tool (#782)

### LINKED ISSUE #782: feat: expose full TaskLog.data field to Foreman's get_task_status tool
## Background

During investigation of #778, the agent discovered a secondary truncation path: `foreman/tools.py`'s `get_task_status` implementation queries only `TaskLog.timestamp` and `TaskLog.line` — the summarized/truncated log line — but **never** `TaskLog.data`, which holds the full agent output detail.

Meanwhile, the frontend UI (`routes/tasks.py`) correctly reads `TaskLog.data` for display.

## Problem

The Foreman only ever sees the short summarized line in `get_task_status` responses, never the rich detail stored in `TaskLog.data`. This means the Foreman cannot access the full context of what an agent did, even when it explicitly calls `get_task_status` to diagnose a stall or failure.

## Proposed fix

- In `foreman/tools.py`, include `TaskLog.data` in the `get_task_status` SELECT query.
- Expose it in the returned JSON (e.g. as a `detail` field alongside `line`), subject to the existing `MAX_TOOL_RESULT_CHARS` cap if needed.
- Add a test confirming the detail field is returned.

## Acceptance criteria
- [ ] `get_task_status` returns `TaskLog.data` (or a truncated version of it) alongside `line`
- [ ] The Foreman can see full agent output detail in log entries
- [ ] Existing tests still pass

jmelloy:
## Summary
- `get_task_status`'s `recent_logs` entries now include the log line's `data` field (the structured tool/agent detail payload stored on `TaskLog.data`), parsed as JSON when present, so the Foreman can see rich agent output detail rather than just the raw log text.
- Updated the tool's description in `tools_schema.py` to mention the new `data` field.
- Added a regression test covering a log entry with structured `data` alongside a plain log entry without it.

Closes #782

## Test plan
- [x] `python -m pytest tests/test_foreman.py -k "get_task_status or large_result"`
- [x] `python -m pytest tests/test_foreman.py tests/test_foreman_runner.py tests/test_foreman_poll_backoff.py`
- [x] `ruff check foreman/tools.py foreman_core/tools_schema.py tests/test_foreman.py`

### DIFF
--- a/backend/tests/test_foreman.py
+++ b/backend/tests/test_foreman.py
@@ -899,6 +899,42 @@ class TestExecToolsDispatching:
         assert data["state"] == "working"
         assert "recent_logs" in data
 
+    async def test_get_task_status_includes_log_data(self, db_session):
+        insert_guild(db_session, "g-status-data")
+        _insert_worker(db_session, "g-status-data", "w-status-data")
+        _insert_task(db_session, "t-stat2", "g-status-data", "w-status-data", state="working")
+        now = datetime.now(UTC)
+        detail = {"tool": "Read", "input": {"file_path": "/tmp/foo.py"}, "output": "full contents"}
+        with _sync_session(db_session) as session:
+            session.add(
+                TaskLog(
+                    task_id="t-stat2",
+                    timestamp=now,
+                    line="Read /tmp/foo.py",
+                    worker_id="w-status-data",
+                    data=json.dumps(detail),
+                )
+            )
+            session.add(
+                TaskLog(
+                    task_id="t-stat2",
+                    timestamp=now,
+                    line="plain log line",
+                    worker_id="w-status-data",
+                )
+            )
+            session.commit()
+        with patch("foreman.tools.broadcast", new_callable=AsyncMock):
+            results = await exec_tools(
+                "g-status-data", [_fake_tool_use("get_task_status", {"task_id": "t-stat2"})]
+            )
+        data = json.loads(results[0]["content"])
+        logs = data["recent_logs"]
+        assert logs[0]["line"] == "Read /tmp/foo.py"
+        assert logs[0]["data"] == "full contents"
+        assert logs[1]["line"] == "plain log line"
+        assert "data" not in logs[1]
+
     async def test_unknown_tool_name_returns_empty_result(self, db_session):
         """An unknown tool name should not raise; result content may be empty."""
         insert_guild(db_session, "g-unknown")
--- a/worker/tests/test_disconnect.py
+++ b/worker/tests/test_disconnect.py
@@ -70,6 +70,7 @@ async def test_run_sends_disconnect_before_ws_close():
     worker._send = capture_send
     worker._register = AsyncMock()
     worker._fetch_github_token_if_needed = AsyncMock()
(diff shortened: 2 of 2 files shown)


backend/foreman/tools.py
```diff
@@ -1670,12 +1670,22 @@ async def _exec_one_tool(guild_id: str, tu, user_id: str | None = None) -> dict:
...
                             agent_info = {"agent_id": agent_row[0], "agent_state": agent_row[1]}
                     logs_result = await db.exec(
-                        select(col(TaskLog.timestamp), col(TaskLog.line))
+                        select(col(TaskLog.timestamp), col(TaskLog.line), col(TaskLog.data))
                         .where(col(TaskLog.task_id) == task_id)
                         .order_by(col(TaskLog.id).desc())
                         .limit(limit)
                     )
                     log_rows = list(reversed(logs_result.all()))
+
```
jmelloy: Defining `_log_entry` inside `_exec_one_tool` on every call is unnecessary overhead. Consider moving it to module level (or at least outside the per-tool dispatch block). It also uses `Any` without a visible import guard — confirm `Any` is already imported from `typing` at the top of the file.

backend/foreman/tools.py
```diff
@@ -1670,12 +1670,22 @@ async def _exec_one_tool(guild_id: str, tu, user_id: str | None = None) -> dict:
...
+                    def _log_entry(row: Any) -> dict[str, Any]:
+                        entry: dict[str, Any] = {"time": row[0], "line": row[1]}
```
jmelloy: `if row[2]:` silently skips an empty-string `data` column. This is probably fine, but if the DB can store `'{}'` or other falsy-adjacent values intentionally, consider `if row[2] is not None:` to be more explicit about the intent.

backend/foreman/tools.py
```diff
@@ -1670,12 +1670,22 @@ async def _exec_one_tool(guild_id: str, tu, user_id: str | None = None) -> dict:
...
+                        if row[2]:
+                            try:
```
jmelloy: Catching `ValueError` is correct for `json.JSONDecodeError` (a subclass), but catching `TypeError` suggests `row[2]` might not always be a string — if the ORM already returns a Python object (e.g. a dict) for a JSON column, `json.loads` would raise `TypeError` and the raw value would be stored as-is. Verify the column type; if it is a native JSON column the parse step may be unnecessary or should check `isinstance(row[2], str)` first.

jmelloy (commented):
- The feature correctly extends `get_task_status` to include structured `TaskLog.data` in log entries, with graceful fallback when JSON parsing fails.
- The `_log_entry` helper is defined inside a loop/block on every tool invocation rather than at module level, which is a minor but unnecessary repeated allocation; however it is functionally correct.
- The `data` field is only included when truthy (`if row[2]:`), meaning an empty string `''` stored in the DB is silently dropped — this is consistent but worth documenting as intentional.
- The test is well-structured and covers both the positive case (structured data present) and the negative case (no data field emitted for plain log entries).
- No security concerns: the data is read from an internal DB column and serialised through `json.dumps`/`json.loads` without being evaluated.

PR #787, opened 2026-07-06
feat: add Postgres to worker Dockerfile (#786)

### LINKED ISSUE #786: feat: add Postgres to worker Dockerfile to accelerate testing
## Summary

Add a Postgres instance to the worker Dockerfile so that backend tests that require a real database can run locally inside the worker container without needing an external service.

## Motivation

Currently workers rely on SQLite or an external Postgres for tests, which slows down CI test runs. Bundling Postgres directly in the worker image removes the external dependency and speeds up the test cycle.

## Acceptance criteria

- Worker Dockerfile installs and configures Postgres (e.g. via `postgresql` apt package or a slim official image layer).
- A start-up script or entrypoint initialises the DB and starts the Postgres service before tests run.
- `pytest` (or equivalent test runner) picks up the local Postgres via environment variable (e.g. `DATABASE_URL`).
- Existing tests continue to pass.
- Image build time / size impact is documented in the PR description.

## Requested by

@jmelloy via Discord

jmelloy:
## Summary
- Installs `postgresql` + `postgresql-client` in the `worker` Dockerfile target, folded into the existing apt layer, so `initdb`, `pg_ctl`, `psql`, etc. are available inside the worker container without depending on the external `postgres-test` compose service.
- Sets `PGDATA=/tmp/pgdata` as a sensible default cluster location and `TEST_DATABASE_URL=postgresql+asyncpg://pioneer:pioneer_password@localhost:5433/pioneer_test`, which matches `backend/tests/_test_config.py`'s own fallback default so tests auto-detect it with zero extra config. `DATABASE_URL` is intentionally left unset — that's runtime configuration.
- No entrypoint automation: the coding agent running inside the container initialises and starts Postgres itself (`initdb` / `pg_ctl`) when a repo's tests actually need it.

## Image size / build time impact
- **Size:** `postgresql` pulls in `libllvm19` (~127 MB installed) as a hard dependency for JIT support, plus `libicu76` (~37 MB) and smaller libs. Total added size is **~235 MB** (`postgresql` + `postgresql-client` + deps), based on Debian's `Packages` index `Installed-Size` fields.
- **Build time:** no new layer — the packages are folded into the existing `apt-get install` step, so the only added cost is package download time (~235 MB); no compilation involved.

## Testing
No `docker` binary available in this sandbox to run `docker build`. Confirmed the package list is correct and unchanged from the previous iteration of this PR (the only difference here is removing the entrypoint automation), and the Dockerfile builds the `worker` target's `RUN apt-get install` step identically to before, minus the automation layer.

Closes #786


docker/postgres/worker-entrypoint.sh
```diff
@@ -0,0 +1,47 @@
...
+for _ in $(seq 1 30); do
+    if pg_isready -h localhost -p "$PG_PORT" -q; then
+        break
+    fi
+    sleep 1
+done
+pg_isready -h localhost -p "$PG_PORT" -q
+
+if ! psql -h localhost -p "$PG_PORT" -U postgres -d postgres -tAc \
+        "SELECT 1 FROM pg_roles WHERE rolname='pioneer'" | grep -q 1; then
```
jmelloy: Granting `pioneer` the `SUPERUSER` privilege is overly permissive. For test isolation it should only need `CREATEDB` or just ownership of `pioneer_test`. Using superuser in a bundled dev DB normalises a bad practice and could be accidentally replicated in production configs.

docker/postgres/worker-entrypoint.sh
```diff
@@ -0,0 +1,47 @@
...
+# repo test suites — notably pioneer-square's own backend/tests — can run
+# against a real Postgres without depending on the postgres-test compose
+# service. Initialises a standalone cluster on first run (owned by the
+# unprivileged `worker` user, no root/postgres-system-user juggling needed),
+# starts it if not already running, then creates the `pioneer` role and
+# `pioneer_test` database matching backend/tests/_test_config.py's default
+# TEST_DATABASE_URL so tests auto-detect it with zero configuration.
+set -euo pipefail
+
+PG_BIN="$(dirname "$(find /usr/lib/postgresql -maxdepth 3 -name initdb | sort -V | tail -1)")"
```
jmelloy: The `find … | sort -V | tail -1` approach is fragile: if `/usr/lib/postgresql` contains no `initdb` binary (e.g., after an apt layout change) the variable is silently empty, and subsequent `initdb` / `pg_ctl` calls fail with confusing errors. Consider hardcoding the major version or failing fast with a clear error message.

Dockerfile
```diff
@@ -124,6 +137,13 @@ RUN useradd --create-home --shell /bin/bash worker \
     && mkdir -p /work/repos /work/worktrees /config /home/worker/go \
     && chown -R worker:worker /work /config /home/worker
 
+# Picked up by backend/tests/_test_config.py with zero extra config — it
+# matches that module's own fallback default, so TEST_DATABASE_URL/
```
jmelloy: `DATABASE_URL` is set to the test database (`pioneer_test`) as an image-level default. If worker runtime code (non-test paths) reads `DATABASE_URL`, it will connect to the test DB unless explicitly overridden at deploy time. Consider only setting `TEST_DATABASE_URL` here, or using a separate `pioneer` database for the runtime `DATABASE_URL`.

Dockerfile
```diff
@@ -88,6 +88,19 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
         unzip \
     && rm -rf /var/lib/apt/lists/*
 
+# Bundled Postgres (#786) — lets backend tests run against a real Postgres
```
jmelloy: This opens a second `apt-get update && apt-get install` layer immediately after one already exists in the same build stage (line 88). Merging the `postgresql` and `postgresql-client` packages into the existing `apt-get install` block reduces layer count and final image size.

docker/postgres/worker-entrypoint.sh
```diff
@@ -0,0 +1,47 @@
...
+fi
+
+if ! pg_ctl -D "$PGDATA" status >/dev/null 2>&1; then
+    pg_ctl -D "$PGDATA" -l "$PGDATA/postgres.log" \
+        -o "-p $PG_PORT -c listen_addresses=localhost -c unix_socket_directories=$PGDATA" \
+        start
+fi
+
+for _ in $(seq 1 30); do
+    if pg_isready -h localhost -p "$PG_PORT" -q; then
```
jmelloy: After the 30-iteration wait loop there is no user-visible timeout message before the final `pg_isready` call exits non-zero. Add `echo "ERROR: Postgres did not become ready after 30 seconds" >&2` before the final check to make container startup failures easier to diagnose.

jmelloy (commented):
- The entrypoint script has a logic flaw: the final `pg_isready` call after the wait loop will exit non-zero (failing the container) if Postgres never becomes ready, which is good, but the loop itself never prints a timeout error making debugging hard.
- Granting the `pioneer` role `SUPERUSER` privileges is excessive and a security concern — it should only have the minimum permissions needed (e.g., `CREATEDB` or just ownership of `pioneer_test`).
- Running two separate `apt-get update && apt-get install` layers for the worker stage unnecessarily bloats the image; the PostgreSQL packages should be combined into the existing `apt-get install` block above.
- The `DATABASE_URL` env var is set to point to the test database (`pioneer_test`) as the default for all connections, which could cause production/worker runtime code to accidentally use the test DB if the env var is not overridden.
- The `PG_BIN` detection using `find … | sort -V | tail -1` is fragile and will silently use wrong binaries if the path changes; pinning to the known Debian path (e.g., `/usr/lib/postgresql/15/bin`) or using `pg_lsclusters` would be more robust.

PR #788, opened 2026-07-06
feat: debounce passing CI check events

### LINKED ISSUE #785: feat: debounce/combine passing CI check events before notifying Foreman
## Problem

When a PR has multiple CI checks (e.g. lint, unit tests, backend tests, frontend build), each one fires a separate `check_run` / `check_suite` completed event. If all checks pass the Foreman receives several near-simultaneous "CI success" events for the same PR, which is noisy and could trigger redundant actions.

## Proposed Solution

Debounce passing CI check events so that a short window (e.g. 5–10 seconds) is applied after the first passing check arrives. If more passing checks arrive for the same PR/commit within that window, they are coalesced into a single combined notification delivered once the window expires. Failing checks should still be delivered immediately (no debounce).

## Acceptance Criteria

- Multiple passing `check_run`/`check_suite` events for the same repo+PR+head SHA within the debounce window are combined into one notification.
- The combined notification summarises all check names that passed.
- Failing or erroring checks bypass the debounce and are forwarded immediately.
- Existing behaviour for non-CI events is unaffected.

jmelloy:
## Summary

Implements debouncing of passing `check_run`/`check_suite` webhook events before they reach the Foreman, per #785.

- Passing (`conclusion=success`) `check_run`/`check_suite` events are now buffered on a short, configurable debounce window (`CI_PASS_DEBOUNCE_SECONDS`, default 7s) keyed on `guild + repo + PR number + head SHA`.
- Additional passing check events for the same key reset the window and accumulate; once the window expires with no new events, a single combined message listing every distinct check name that passed is sent to the Foreman, plus one combined Discord "CI passed" notification (replacing the previous per-event Discord post).
- Failing/erroring conclusions (`failure`, `action_required`, `timed_out`, `cancelled`) bypass the debounce entirely and are dispatched to the Foreman immediately via a new `_dispatch_immediately` path.
- Any other/unrecognized completed conclusion falls back to the existing generic per-task debounce queue so it still reaches the Foreman.
- The `DebounceQueue` class was generalized to accept pluggable `combine`/`deliver` callbacks so the new CI-pass queue can share the same generation-based race-safe scheduling/cancellation logic as the original per-task queue, rather than duplicating it.

## Tests

- Updated `test_webhook_dispatches_for_ci_bot_check_run` to assert a failing check_run now goes through `_dispatch_immediately` rather than the buffered debounce queue.
- Added `TestCiPassDebounce` covering:
  - Multiple passing checks on the same PR/SHA coalesce into one combined Foreman message and one Discord notification listing all check names.
  - A failing check reaches the Foreman immediately, with no debounce delay and no Discord "CI passed" notification.
  - A mixed batch: the failing check fires immediately while passing checks for the same PR/SHA still coalesce separately once their window expires.

All 60 tests in `backend/tests/test_github_webhooks_phase2.py` pass; `ruff check` / `ruff format` clean.

Fixes #785.

### DIFF
--- a/backend/routes/webhooks.py
+++ b/backend/routes/webhooks.py
@@ -23,6 +23,7 @@ import os
 import re
 from collections.abc import Awaitable, Callable
 from datetime import UTC, datetime, timedelta
+from uuid import uuid4
 
 import discord_notifier
 from database import get_db_dep
@@ -148,19 +149,27 @@ async def _default_deliver(
 async def _deliver_ci_pass(
     key: str, guild_id: str, combined: str, items: list[_DebounceItem]
 ) -> None:
-    """Foreman delivery (as usual) plus one combined Discord "CI passed" post.
+    """Foreman delivery (when a task is attached) plus one combined Discord
+    "CI passed" post.
 
     Replaces the old per-check-event immediate Discord notification (issue
     #785): instead of one Discord message per passing check_run/check_suite
     event, the several events GitHub fires per commit coalesce into a single
     message listing every check name that passed.
-    """
-    await _default_deliver(key, guild_id, combined, items)
 
+    Some items reach this queue without a task_id (a passing check on a PR
+    with no matching task — see the `github_webhook` check_run/check_suite
+    branch); the foreman has no context to act on there
+    (`_should_dispatch_to_foreman`'s "no matching task" rule), so the foreman
+    call is skipped for those batches and only the Discord post fires.
+    """
     ctx = next((c for _, _, _, _, c in items if c), None) or {}
     check_names = _dedupe_labels(items)
     task_id = next((tid for _, _, tid, _, _ in items if tid), None)
     pr_label = ctx.get("pr_label") or "unknown"
+
+    if task_id:
+        await _default_deliver(key, guild_id, combined, items)
     spawn(
         discord_notifier.notify_event(
             "ci-pass",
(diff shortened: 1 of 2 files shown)


backend/routes/webhooks.py
```diff
@@ -668,6 +785,114 @@ def _build_chat_line(
...
+        )
+        await _dispatch_immediately(guild_id, summary, task_user_id, task_id)
+        logger.info(
+            "github webhook ci-check failure dispatched immediately guild=%s delivery=%s "
+            "repo=%s pr=%s check=%s conclusion=%s",
+            guild_id,
+            delivery_id,
+            repo,
+            pr_number,
+            check_name,
```
jmelloy, line 880: When `head_sha` is `None` (e.g. a `check_suite` payload without `head_sha`), the debounce key becomes `"{guild_id}:ci-pass:{pr_label}:None"`, causing unrelated events across different commits to coalesce under the same key. Add a guard: if `head_sha` is falsy, fall back to the generic debounce path or generate a unique key.

backend/routes/webhooks.py
```diff
@@ -668,6 +785,114 @@ def _build_chat_line(
...
+async def _run_foreman_immediately(
+    guild_id: str, summary: str, user_id: str | None, task_id: str | None
+) -> None:
+    """Invoke the foreman without debounce (used for failing CI checks)."""
+    await run_foreman_ai(guild_id, summary, user_id=user_id, task_id=task_id, child=bool(task_id))
+    reset_foreman_poll(guild_id)
+
+
+async def _dispatch_immediately(
+    guild_id: str, summary: str, user_id: str | None, task_id: str | None
```
jmelloy, line 806: `_dispatch_immediately` is an `async` function that internally calls `spawn()` (fire-and-forget) and returns immediately — making it `async` is misleading since there's nothing to await. Consider making it a plain `def` or renaming/documenting clearly that the coroutine completes synchronously and the real work runs in the background task.

backend/routes/webhooks.py
```diff
@@ -1012,7 +1245,13 @@ async def github_webhook(
             if isinstance(node, dict)
             else "CI"
         )
-        if conclusion == "success":
+        if conclusion == "success" and not dispatch:
```
jmelloy, line 1272: The condition `if conclusion == 'success' and not dispatch` means that when a passing check_run has no associated task (`dispatch=False`), an immediate single-check Discord notification is posted without any debouncing. This partially reverts the intent of #785 for the no-task case. Consider routing these through the CI-pass queue keyed without a task_id to still coalesce them.

jmelloy (commented):
- The CI-pass debounce design is well-structured, cleanly separating the new CI-pass queue from the generic debounce queue with appropriate window sizes and custom combine/deliver functions.
- A subtle bug exists: when `dispatch=True` and `conclusion='success'`, the early-moved `_should_dispatch_to_foreman` call causes the Discord notification block (`if conclusion == 'success' and not dispatch`) to be skipped, but this is intentional per the comments — however, when `dispatch=False` and conclusion is `success`, the immediate Discord notification still fires without debouncing, which may cause duplicate or premature notifications in edge cases.
- The `_dispatch_immediately` function wraps `_run_foreman_immediately` in `spawn()` but the function itself is `async` and awaited in tests — this asymmetry between production (fire-and-forget via spawn) and test patching (direct await) is a valid seam but could confuse future maintainers.
- The `_ci_pass_debounce_queue` key uses `head_sha` which may be `None` when the payload lacks a `head_sha` field, resulting in keys like `guild:ci-pass:repo#1:None` that incorrectly coalesce unrelated events across different commits.
- Test coverage is thorough with three meaningful scenarios (coalesce, immediate failure, mixed batch); the use of `asynccontextmanager` for test setup is clean and well-isolated.

PR #789, opened 2026-07-06
Parallelize Foreman tool calls with asyncio.gather (#555)

### LINKED ISSUE #555: Foreman Claude API calls are sequential per-invocation — slow on multi-tool turns
## Problem

Every `run_foreman_ai()` invocation issues all tool calls **sequentially**: each Claude API round-trip is `await`-ed before the next tool in the same turn begins. On a foreman turn that calls `assign_task` three times (e.g. dispatching three workers), the latency is `3 × (Claude round-trip + DB write + worker notify)` stacked end-to-end.

The inner loop in `runner.py` is:
```
while True:
    response = await claude_api(...)   # waits for full streamed response
    for tool_use in response.tool_uses:
        result = await execute_tool(tool_use)   # sequential, one at a time
        history.append(result)
    if stop_reason != "tool_use": break
```

Independent tool calls in the same turn (e.g. multiple `assign_task` or `create_task` calls) could safely be parallelized with `asyncio.gather`.

## Fix

In the tool-dispatch loop, group tool calls that have no data dependency on each other and execute them with `asyncio.gather`. Claude already returns all tool calls for a given turn in one response object — run them concurrently, collect results, then send all results back in one follow-up message.

Be careful with:
- Tools that mutate shared state (worker idle→busy) — these need the per-worker lock from #550/#553 to be safe under parallelism
- Preserving the order of results in the history message

## Expected impact
On turns with 3+ tool calls, wall-clock latency drops by ~2–3× with no change to correctness.

## Related
- #553 (guild-level serialization)

jmelloy:
## Summary

Issue #555 asked for the Foreman's per-turn tool-dispatch loop to run independent tool calls concurrently with `asyncio.gather` instead of awaiting them one at a time, with per-worker locking to keep `assign_task`-style state transitions safe.

Investigation found this was already implemented in a prior commit (`0ac1a77`, "parallel foreman tools"): both `backend/foreman/tools.py::exec_tools` (embedded runner) and `foreman/pioneer_foreman/tools.py::exec_tools` (standalone runner) already fan out a turn's tool calls via `asyncio.gather`, preserve result order, and isolate per-tool failures (`_exec_one_tool` catches all exceptions internally, so one bad tool never drops the others). `assign_task` already guards worker-state transitions with a per-worker `assign_task:<worker_id>` DB lock (`LockService`), closing the TOCTOU window between worker selection and assignment.

What was missing was regression coverage for the exact scenarios called out in the issue. This PR adds:
- `test_two_assign_task_calls_same_worker_in_one_batch` — two `assign_task` calls targeting the same worker within a single `exec_tools` batch; asserts exactly one succeeds, the other fails cleanly with "already being assigned", and only one `Task` row is persisted.
- `test_mixed_tool_types_in_one_batch` — a batch mixing `create_task`, `assign_task`, and `message_worker`; asserts all three dispatch, results stay correctly ordered/matched to their `tool_use_id`.

No production code changes were needed — this closes out the remaining acceptance criterion (tests) for #555.

## Test plan
- [x] `pytest backend/tests/test_foreman.py backend/tests/test_foreman_runner.py backend/tests/test_assign_task_auto_model.py backend/tests/test_call_agent_tool.py` — 154 passed
- [x] `pytest foreman/tests/test_tools.py` — 12 passed

Closes #555

### DIFF
--- a/backend/tests/test_foreman.py
+++ b/backend/tests/test_foreman.py
@@ -520,6 +520,88 @@ class TestExecToolsDispatching:
             )
         assert task_tool == "pi"
 
+    async def test_two_assign_task_calls_same_worker_in_one_batch(self, db_session):
+        """Two assign_task calls to the same worker within a single foreman turn
+        (i.e. one exec_tools batch, dispatched concurrently via asyncio.gather)
+        must not both succeed — the per-worker assign_task:<worker_id> lock
+        forces exactly one to win, the other reports a clean error instead of
+        double-assigning the worker. Regression test for issue #555."""
+        insert_guild(db_session, "g-assign-batch-race")
+        _insert_worker(db_session, "g-assign-batch-race", "w-batchrace1")
+        with patch("foreman.tools.broadcast", new_callable=AsyncMock):
+            results = await exec_tools(
+                "g-assign-batch-race",
+                [
+                    _fake_tool_use(
+                        "assign_task",
+                        {"worker_id": "w-batchrace1", "description": "Task A", "name": "Task A"},
+                        tool_id="tid-a",
+                    ),
+                    _fake_tool_use(
+                        "assign_task",
+                        {"worker_id": "w-batchrace1", "description": "Task B", "name": "Task B"},
+                        tool_id="tid-b",
+                    ),
+                ],
+            )
+
+        assert len(results) == 2
+        # Result order must match tool_use order regardless of which finished first.
+        assert results[0]["tool_use_id"] == "tid-a"
+        assert results[1]["tool_use_id"] == "tid-b"
+
+        successes = [r for r in results if not r.get("is_error")]
+        errors = [r for r in results if r.get("is_error")]
+        assert len(successes) == 1
+        assert len(errors) == 1
+        assert "queued" in successes[0]["content"].lower()
+        assert "already being assigned" in errors[0]["content"].lower()
+
+        # Exactly one task should have actually been persisted for this worker.
+        with _sync_session(db_session) as session:
+            count = session.scalar(
+                select(func.count()).select_from(Task).where(col(Task.worker_id) == "w-batchrace1")
+            )
+        assert count == 1
+
+    async def test_mixed_tool_types_in_one_batch(self, db_session):
+        """A single foreman turn mixing create_task, assign_task, and
+        message_worker must dispatch all of them concurrently and return
+        correctly matched, correctly ordered results for each."""
+        insert_guild(db_session, "g-mixed-batch")
+        _insert_worker(db_session, "g-mixed-batch", "w-mixed1")
+        with (
+            patch("foreman.tools.broadcast", new_callable=AsyncMock),
+            patch("foreman.tools.emit_terminal_line", new=AsyncMock()),
+            patch("foreman.tools.broadcast_msg", new=AsyncMock()),
+        ):
+            results = await exec_tools(
+                "g-mixed-batch",
+                [
+                    _fake_tool_use(
+                        "create_task",
+                        {"name": "Standalone task", "description": "desc"},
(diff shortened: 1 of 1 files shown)


backend/tests/test_foreman.py
```diff
@@ -520,6 +520,88 @@ async def test_assign_task_defaults_to_first_tool(self, db_session):
...
+
+        successes = [r for r in results if not r.get("is_error")]
+        errors = [r for r in results if r.get("is_error")]
+        assert len(successes) == 1
+        assert len(errors) == 1
+        assert "queued" in successes[0]["content"].lower()
+        assert "already being assigned" in errors[0]["content"].lower()
+
+        # Exactly one task should have actually been persisted for this worker.
+        with _sync_session(db_session) as session:
```
jmelloy: This assertion assumes the `assign_task:<worker_id>` lock guarantees exactly one failure even under `asyncio.gather`. If the lock is async-reentrant within the same event loop turn (e.g. uses an `asyncio.Lock` that one coroutine holds while yielding, but the second coroutine runs before the first releases), both could succeed. The test should explicitly document what lock mechanism it relies on, or the test itself could silently stop being a regression guard if the lock is replaced.

backend/tests/test_foreman.py
```diff
@@ -520,6 +520,88 @@ async def test_assign_task_defaults_to_first_tool(self, db_session):
...
+            count = session.scalar(
+                select(func.count()).select_from(Task).where(col(Task.worker_id) == "w-batchrace1")
+            )
+        assert count == 1
+
+    async def test_mixed_tool_types_in_one_batch(self, db_session):
```
jmelloy, line 592: `_sync_session(db_session)` for counting persisted tasks is unusual — if the tasks were written in the same SQLAlchemy session/transaction that's still open, this nested session may see an inconsistent or uncommitted state depending on isolation level. Consider flushing `db_session` and querying through it directly, or ensure `_sync_session` begins a new transaction after the previous one commits.

backend/tests/test_foreman.py
```diff
@@ -520,6 +520,88 @@ async def test_assign_task_defaults_to_first_tool(self, db_session):
...
+                        {"worker_id": "w-mixed1", "description": "Do work", "name": "Do it"},
+                        tool_id="tid-assign",
+                    ),
+                    _fake_tool_use(
+                        "message_worker",
+                        {"worker_id": "w-mixed1", "message": "hello"},
+                        tool_id="tid-message",
+                    ),
+                ],
+            )
```
jmelloy, line 622: `assert 't-' in results[0]['content']` is a very weak assertion for `create_task` — it would pass for any string containing 't-' (including error messages). Assert the full expected format or at least use a regex matching the task ID pattern to make this a meaningful regression check.

jmelloy (commented):
- The PR adds regression tests for concurrent tool-call dispatch and per-worker locking, which is a reasonable addition given the issue context.
- `test_two_assign_task_calls_same_worker_in_one_batch` makes a strong assertion (`assert len(errors) == 1`) that depends on the DB lock being synchronous and non-reentrant within a single async gather; if the lock implementation is async and re-entrant for the same coroutine context, both calls could succeed, making the test fragile or misleading.
- The test uses `_sync_session(db_session)` for the DB count assertion, which is an unusual pattern that may not reflect committed state if the surrounding test session is still open and the tasks were inserted in the same transaction.
- `test_mixed_tool_types_in_one_batch` asserts `'t-' in results[0]['content']` as the only content check for `create_task`; this is a very weak assertion that could pass for unrelated reasons or fail if the task ID format changes.
- No test covers the case where `exec_tools` is called with an empty list of tool uses, or where all tools in a batch fail — these edge cases are relevant to the parallelism implementation but are absent.

PR #798, opened 2026-07-06
Ensure every DB table has an id primary key (#791)

### LINKED ISSUE #791: Ensure every database table has an `id` primary key
## Summary

Audit all database tables in the project and ensure each one has an explicit `id` primary key column. Any tables missing a PK (or using a non-`id` column as PK) should be migrated to add/rename the column appropriately.

## Acceptance Criteria

- Every table in the schema has an `id` column defined as the primary key
- Alembic (or equivalent) migration(s) added for any tables that require changes
- Existing foreign key references updated to match if column names change
- Tests pass after migration

## Requested by

@jmelloy via Discord

jmelloy:
## Summary

Closes #791. Audited every SQLAlchemy model in `backend/models.py` and found eight tables whose primary key was a natural string column or a composite of two columns instead of a plain `id`:

- `github_tokens` — PK was `github_user_id`
- `user_sessions` — PK was `token`
- `guild_members` — PK was `(guild_id, user_id)`
- `user_spawn_settings` — PK was `(guild_id, user_id)`
- `push_tokens` — PK was `token`
- `discord_users` — PK was `github_login`
- `discord_connect_tokens` — PK was `token`
- `model_catalog` — PK was `(provider, model_id)`

For each table, added a surrogate `id` integer primary key and preserved the old key as a `UNIQUE` constraint/index so existing lookups, foreign keys, and upsert targets keep working unchanged.

## Changes

- `backend/models.py` — updated the eight SQLModel classes above to declare `id: int | None = Field(default=None, primary_key=True)` and moved their old PK columns to `unique=True` fields or unique indexes.
- `backend/alembic/versions/20260706_000000_add_id_pk_everywhere.py` — new Alembic migration that adds a `SERIAL id` column, swaps the primary key, and recreates the old key as a unique constraint/index for all eight tables (with a matching `downgrade()`).
- `backend/tests/test_discord_connect.py` — updated a test that used `session.get(DiscordConnectToken, connect_token)` (a PK lookup) to query by `token` instead, since `token` is no longer the primary key.

## Bug fixed during verification

While running the full test suite against the real Postgres test container, the migration failed on `guild_members` with `constraint "guild_members_pkey" of relation "guild_members" does not exist`. Digging in, `guild_members` went through a create-as-`guild_members_new`-then-rename migration back in `20260512_000002_drop_guild_id_child_tables.py`, so Postgres kept the original auto-generated constraint name `guild_members_new_pkey` across the rename instead of `guild_members_pkey`. The migration now looks up the real PK constraint name via `sa.inspect(bind).get_pk_constraint(table)` instead of assuming `{table}_pkey`.

## Test plan

- [x] `cd backend && python3 -m pytest -q` — 879 passed, 2 skipped, 0 failures (against the postgres-test container)
- [x] `alembic downgrade -1` then `alembic upgrade head` — both directions run cleanly with no errors

### DIFF
--- a/backend/alembic/versions/20260706_000000_add_id_pk_everywhere.py
+++ b/backend/alembic/versions/20260706_000000_add_id_pk_everywhere.py
@@ -50,6 +50,19 @@ _COMPOSITE_TABLES = [
 ]
 
 
+def _pk_constraint_name(table: str) -> str:
+    """Look up a table's actual primary-key constraint name.
+
+    Some tables (e.g. ``guild_members``) went through a create-as-``_new``-
+    then-rename migration (20260512_000002); PostgreSQL keeps the
+    constraint's original auto-generated name (``guild_members_new_pkey``)
+    across the rename, so we can't assume ``{table}_pkey``.
+    """
+    bind = op.get_bind()
+    pk = sa.inspect(bind).get_pk_constraint(table)
+    return pk["name"]
+
+
 def upgrade() -> None:
     # user_sessions.github_user_id FKs to github_tokens.github_user_id; drop
     # that FK before github_tokens' PK constraint changes, then recreate it.
@@ -57,7 +70,7 @@ def upgrade() -> None:
 
     for table, col in _SINGLE_COLUMN_TABLES:
         op.execute(sa.text(f"ALTER TABLE {table} ADD COLUMN id SERIAL"))
-        op.drop_constraint(f"{table}_pkey", table, type_="primary")
+        op.drop_constraint(_pk_constraint_name(table), table, type_="primary")
         op.create_primary_key(f"{table}_pkey", table, ["id"])
         op.create_unique_constraint(f"{table}_{col}_key", table, [col])
 
@@ -71,7 +84,7 @@ def upgrade() -> None:
 
     for table, cols, index_name in _COMPOSITE_TABLES:
         op.execute(sa.text(f"ALTER TABLE {table} ADD COLUMN id SERIAL"))
-        op.drop_constraint(f"{table}_pkey", table, type_="primary")
+        op.drop_constraint(_pk_constraint_name(table), table, type_="primary")
         op.create_primary_key(f"{table}_pkey", table, ["id"])
         op.create_index(index_name, table, cols, unique=True)
 
@@ -79,7 +92,7 @@ def upgrade() -> None:
 def downgrade() -> None:
     for table, cols, index_name in reversed(_COMPOSITE_TABLES):
         op.drop_index(index_name, table_name=table)
-        op.drop_constraint(f"{table}_pkey", table, type_="primary")
+        op.drop_constraint(_pk_constraint_name(table), table, type_="primary")
         op.create_primary_key(f"{table}_pkey", table, cols)
         op.drop_column(table, "id")
 
@@ -87,7 +100,7 @@ def downgrade() -> None:
 
     for table, col in reversed(_SINGLE_COLUMN_TABLES):
         op.drop_constraint(f"{table}_{col}_key", table, type_="unique")
-        op.drop_constraint(f"{table}_pkey", table, type_="primary")
+        op.drop_constraint(_pk_constraint_name(table), table, type_="primary")
         op.create_primary_key(f"{table}_pkey", table, [col])
         op.drop_column(table, "id")
 
--- a/backend/alembic/versions/20260706_000000_add_id_pk_everywhere.py
+++ b/backend/alembic/versions/20260706_000000_add_id_pk_everywhere.py
@@ -16,9 +16,11 @@ string column or a composite of two columns instead of a plain `id`:
 - discord_connect_tokens (PK was token)
 - model_catalog        (PK was (provider, model_id))
 
-For each, this adds a SERIAL `id` column, makes it the primary key, and
+For each, this adds an identity `id` column, makes it the primary key, and
 preserves the old key as a UNIQUE constraint/index so existing lookups,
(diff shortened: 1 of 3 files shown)


backend/alembic/versions/20260706_000000_add_id_pk_everywhere.py
```diff
@@ -0,0 +1,113 @@
...
+    op.drop_constraint("user_sessions_github_user_id_fkey", "user_sessions", type_="foreignkey")
+
+    for table, col in _SINGLE_COLUMN_TABLES:
+        op.execute(sa.text(f"ALTER TABLE {table} ADD COLUMN id SERIAL"))
+        op.drop_constraint(_pk_constraint_name(table), table, type_="primary")
+        op.create_primary_key(f"{table}_pkey", table, ["id"])
+        op.create_unique_constraint(f"{table}_{col}_key", table, [col])
+
+    op.create_foreign_key(
+        "user_sessions_github_user_id_fkey",
```
jmelloy, line 88: Bug in `downgrade()`: `_pk_constraint_name(table)` is called here to find the surrogate PK that was just created during `upgrade()`, but after dropping that constraint and recreating the original composite/single-column PK the helper would need to be called *before* any constraint changes. On a fresh downgrade this will resolve to the wrong (or non-existent) constraint name. Consider using a hardcoded mapping of `{table: original_pk_name}` instead of live introspection.

backend/alembic/versions/20260706_000000_add_id_pk_everywhere.py
```diff
@@ -0,0 +1,113 @@
...
+
+    Some tables (e.g. ``guild_members``) went through a create-as-``_new``-
+    then-rename migration (20260512_000002); PostgreSQL keeps the
+    constraint's original auto-generated name (``guild_members_new_pkey``)
+    across the rename, so we can't assume ``{table}_pkey``.
+    """
+    bind = op.get_bind()
+    pk = sa.inspect(bind).get_pk_constraint(table)
+    return pk["name"]
+
```
jmelloy, line 73: Using an f-string in `op.execute(sa.text(...))` to interpolate `table` bypasses SQLAlchemy's parameterization. Although the values are from a hardcoded list today, the pattern is unsafe and linters/security scanners will flag it. Prefer `op.add_column(table, sa.Column('id', sa.Integer(), sa.Identity(), nullable=False))` which is the idiomatic Alembic approach and avoids raw DDL strings.

backend/alembic/versions/20260706_000000_add_id_pk_everywhere.py
```diff
@@ -0,0 +1,113 @@
...
+    ("user_spawn_settings", ["guild_id", "user_id"], "uq_user_spawn_settings_guild_id_user_id"),
+    ("model_catalog", ["provider", "model_id"], "uq_model_catalog_provider_model_id"),
+]
+
+
+def _pk_constraint_name(table: str) -> str:
+    """Look up a table's actual primary-key constraint name.
+
+    Some tables (e.g. ``guild_members``) went through a create-as-``_new``-
+    then-rename migration (20260512_000002); PostgreSQL keeps the
```
jmelloy, line 59: `_pk_constraint_name` returns `pk['name']` without guarding against `pk` being `None` or `pk['name']` being `None` (which can happen when the constraint has no explicit name in SQLite or in a freshly-created test DB). Add a fallback: `return pk['name'] if pk and pk.get('name') else f'{table}_pkey'`.

backend/models.py
```diff
@@ -687,13 +710,22 @@ class ModelCatalog(SQLModel, table=True):
...
 
     Upserted from the models.dev API on startup and periodically refreshed.
-    Primary key is (provider, model_id) so upserts are idempotent.
+    ``(provider, model_id)`` is unique so upserts are idempotent.
     """
 
     __tablename__ = "model_catalog"  # type: ignore[assignment]
+    __table_args__ = (
+        Index(
+            "uq_model_catalog_provider_model_id",
```
jmelloy, line 719: `ModelCatalog` upserts are documented as idempotent via `(provider, model_id)`, but any application-level `INSERT … ON CONFLICT (provider, model_id)` statements now reference a unique index rather than the primary key. PostgreSQL requires the conflict target to exactly match an index or PK; if the existing upsert code names the columns explicitly it will still work, but if it relied on the PK this is a silent runtime breakage. Please audit all `on_conflict_do_update` / raw upsert SQL for this table.

jmelloy (commented):
- The migration correctly adds surrogate `id` PKs and preserves old keys as UNIQUE constraints, but the `downgrade()` path has a critical bug: `_pk_constraint_name` is called *after* the new `{table}_pkey` has already been dropped and recreated, so on downgrade it will return the surrogate PK name rather than the original.
- The migration uses raw f-string SQL (`ALTER TABLE {table} ADD COLUMN id SERIAL`) without any sanitization; while the table names come from a hardcoded list this is a style/safety concern that could mask injection if the list ever becomes dynamic.
- The `_pk_constraint_name` introspection helper queries the live DB at migration time, which is fragile in CI/CD environments with fresh schemas and will return `None` (causing a crash) if the constraint name cannot be resolved — a hardcoded fallback or explicit constraint name map would be safer.
- Upsert logic that previously used `ON CONFLICT (provider, model_id)` or `ON CONFLICT (token)` etc. in application code (not shown in diff) will silently break because those columns are no longer primary keys; callers must be updated to use the new unique index names.
- The test fix (`session.query(...).filter_by(token=...).one_or_none()`) is correct, but the pattern should be applied consistently across all other tests that used `session.get(Model, natural_key)` — the PR description doesn't mention auditing the full test suite.

PR #800, opened 2026-07-06
fix: review-phase tasks check out the PR's own branch

### LINKED ISSUE #799: fix: review tasks should checkout the PR's branch, not create a new claude/ branch
## Problem

When a worker is assigned a review-phase task, it creates a new `claude/...` worktree branch instead of checking out the existing PR branch. This means the review is happening on a fresh branch rather than the actual PR diff, and the worktree setup is unnecessarily complex.

## Expected behaviour

For review-phase tasks, the worker should:
1. Receive the PR branch name (e.g. `claude/some-feature-t-xxxx`) as the branch to check out — **not** generate a new branch name.
2. Run `gh pr checkout <PR_NUMBER> --repo <OWNER/REPO>` (or `git checkout <branch>`) to work on the existing branch.
3. Post findings via `gh pr review` — no new commits, no new PR.

## Root cause

The worker runtime likely generates a new branch name for every task regardless of phase. For `review` phase tasks, it should instead use the branch associated with the PR being reviewed. The fix should be in the worker host process / worktree setup logic — when `phase == "review"`, derive the branch from the PR URL or an explicit branch parameter passed in `assign_task`, rather than auto-generating one.

## Fix

- In the worker runtime, detect `phase == "review"` and skip auto-branch-name generation.
- Pass the PR branch (or PR number + repo) explicitly so the worker can `gh pr checkout` it directly.
- Simplify worktree setup for review tasks accordingly.

Reported by @jmelloy via Discord.

jmelloy:
## Summary
- Adds `checkout_pr_worktree` / `get_pr_head_branch` helpers in `worker/pioneer_worker/git_ops.py` that use `gh pr checkout` to check an existing PR's branch into a worktree, instead of generating a fresh `claude/...` branch.
- Updates the review-phase worktree logic in `worker/pioneer_worker/worker.py` so that `phase == "review"` tasks resolve and check out the actual PR branch (via `gh pr view`/`gh pr checkout`), record that real branch name on the task, and never `git checkout -b` a new branch.
- Review-phase tasks now skip the branch-push step entirely — they only read the repo and post findings via `gh pr review`, with no commits or pushes.
- Adds `worker/tests/test_review_phase_guard.py` coverage for the new review-phase guard behavior (branch resolution, worktree checkout, no-push guarantee, and follow-up review flow).

## Test plan
- [x] `python -m pytest tests/test_review_phase_guard.py -q` — 6 passed
- [x] `python -m pytest tests/ -q` — 255 passed, 1 pre-existing unrelated failure (`test_tools_detection.py::test_no_tools_flag_gives_none`, environment-dependent PATH auto-detection, untouched by this change)

Closes #799

### DIFF
--- a/worker/pioneer_worker/git_ops.py
+++ b/worker/pioneer_worker/git_ops.py
@@ -112,6 +112,73 @@ async def attach_worktree(repo_path: str, wt_path: str, branch: str) -> bool:
     return rc == 0
 
 
+async def run_gh(args: list[str], cwd: str | None = None) -> tuple[int, str, str]:
+    logger.debug("gh %s (cwd=%s)", " ".join(args), cwd or os.getcwd())
+    proc = await asyncio.create_subprocess_exec(
+        "gh",
+        *args,
+        cwd=cwd,
+        stdout=asyncio.subprocess.PIPE,
+        stderr=asyncio.subprocess.PIPE,
+    )
+    stdout, stderr = await proc.communicate()
+    rc = proc.returncode if proc.returncode is not None else -1
+    if rc != 0:
+        logger.warning(
+            "gh %s failed rc=%s: %s",
+            " ".join(args),
+            rc,
+            stderr.decode(errors="replace").strip()[:200],
+        )
+    return rc, stdout.decode(errors="replace"), stderr.decode(errors="replace")
+
+
+async def get_pr_head_branch(repo_full: str, pr_number: int | str) -> str | None:
+    """Look up a PR's head branch name via the ``gh`` CLI. Returns None on failure."""
+    rc, out, _ = await run_gh(
+        [
+            "pr",
+            "view",
+            str(pr_number),
+            "--repo",
+            repo_full,
+            "--json",
+            "headRefName",
+            "-q",
+            ".headRefName",
+        ]
+    )
+    if rc != 0:
+        return None
+    branch = out.strip()
+    return branch or None
+
+
+async def checkout_pr_worktree(
+    repo_path: str, wt_path: str, pr_number: int | str, repo_full: str
+) -> bool:
+    """Create a worktree checked out to an existing PR's branch via ``gh pr checkout``.
+
+    Used for review-phase tasks: the worker must review the PR's actual branch
+    rather than a freshly generated one. A detached worktree is created first
+    (linked worktrees can't share a branch with the main checkout), then
+    ``gh pr checkout`` is run inside it so `gh` resolves the head branch
+    (including cross-fork PRs) and checks it out in place.
+    """
+    logger.info("Checking out PR #%s (%s) into worktree %s", pr_number, repo_full, wt_path)
+    os.makedirs(os.path.dirname(wt_path), exist_ok=True)
+    await run_git(["fetch", "origin"], cwd=repo_path)
+    rc, _, _ = await run_git(["worktree", "add", "--detach", wt_path, "origin/HEAD"], cwd=repo_path)
+    if rc != 0:
+        logger.error("Detached worktree add failed at %s (rc=%d)", wt_path, rc)
+        return False
+    rc, _, _ = await run_gh(["pr", "checkout", str(pr_number), "--repo", repo_full], cwd=wt_path)
+    if rc != 0:
+        logger.error("gh pr checkout failed for PR #%s (%s) in %s", pr_number, repo_full, wt_path)
+        return False
+    return True
+
+
 async def remove_worktree(repo_path: str, wt_path: str) -> None:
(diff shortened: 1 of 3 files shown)


worker/pioneer_worker/git_ops.py
```diff
@@ -112,6 +112,73 @@ async def attach_worktree(repo_path: str, wt_path: str, branch: str) -> bool:
...
+            ".headRefName",
+        ]
+    )
+    if rc != 0:
+        return None
+    branch = out.strip()
+    return branch or None
+
+
+async def checkout_pr_worktree(
```
jmelloy, line 157: If `run_git(['worktree', 'add', '--detach', ...])` succeeds but `gh pr checkout` fails, the detached worktree is left registered in git's worktree list. The function returns `False` but never calls `run_git(['worktree', 'remove', '--force', wt_path])` before returning, leaving a dangling entry that will cause the next run to hit the 'reuse existing worktree' path unexpectedly.

worker/pioneer_worker/worker.py
```diff
@@ -1964,6 +1993,18 @@ async def _execute_task(self, task: dict, agent: Agent) -> None:
                         level=LEVEL_WORKER,
```
jmelloy, line 1993: The condition `repo_full == pr_repo` is a strict string equality check. If the task's `issue_repo` differs by case or the resolved `repo_full` uses a different format (e.g., one is normalized, one is not), the condition will be `False` and the code silently falls through to the regular `create_worktree` path, checking out the wrong branch for a review task without any warning or error.

worker/pioneer_worker/git_ops.py
```diff
@@ -112,6 +112,73 @@ async def attach_worktree(repo_path: str, wt_path: str, branch: str) -> bool:
...
+            "headRefName",
+            "-q",
+            ".headRefName",
+        ]
+    )
+    if rc != 0:
+        return None
+    branch = out.strip()
+    return branch or None
+
```
jmelloy, line 155: `os.makedirs(os.path.dirname(wt_path), exist_ok=True)` — if `wt_path` is already a top-level path (no directory separator), `os.path.dirname` returns an empty string and `os.makedirs('')` raises `FileNotFoundError`. Consider using `os.makedirs(wt_path, exist_ok=True)` directly or guard with `if parent := os.path.dirname(wt_path): os.makedirs(parent, ...)`.

worker/pioneer_worker/worker.py
```diff
@@ -1884,17 +1888,42 @@ async def _execute_task(self, task: dict, agent: Agent) -> None:
...
+            # freshly generated claude/... branch — or the agent won't see
+            # the actual PR diff. Never fall back to a generated name here:
+            # doing so would silently review the wrong code.
+            branch = (
+                await git_ops.get_pr_head_branch(pr_repo, pr_number)
+                if pr_repo and pr_number
+                else None
+            )
+            if not branch:
+                logger.error(
```
jmelloy, line 1905: When `is_review and not is_followup` but `pr_repo` or `pr_number` is missing, the function emits an error, updates the task to `failed`, and returns early — good. However `_set_state('error', agent)` is called, which may block the agent slot permanently if `_set_state` doesn't recover on the next task pickup. Verify the agent slot is correctly released back to idle after this early return.

jmelloy (commented):
- The core fix is sound: review-phase tasks now correctly check out the PR's actual branch via `gh pr checkout` instead of generating a new `claude/...` branch, and the push step is appropriately skipped.
- A subtle bug exists in `checkout_pr_worktree`: the detached worktree is created at `wt_path` directly, but `os.makedirs(os.path.dirname(wt_path))` is called — if `wt_path` has no parent (unlikely but possible), `dirname` returns `''` and `makedirs` silently does nothing or errors; more critically, if the detached worktree add fails partway, there is no cleanup before returning `False`, leaving a dangling worktree entry.
- The `repo_full == pr_repo` guard in `worker.py` for the `checkout_pr_worktree` path is fragile: if the repo is a fork or the repo strings differ by case/trailing slash, the condition silently falls through to `create_worktree` — a review task would then check out the wrong branch with no error.
- `get_pr_head_branch` is called with `pr_number` as whatever type is stored in `task.get("issue_number")` (could be a string `"42"` or int `42`); this is benign since it's immediately `str()`-ed inside, but the type annotation `int | str` correctly reflects the ambiguity — the task dict should be normalized before use.
- Test coverage is good: the new `test_review_phase_checks_out_pr_branch_via_gh` test properly asserts no `create_worktree` and no `push_branch` are called, and verifies the actual branch name is recorded on the task.

PR #803, opened 2026-07-06
feat: single tool call shows tool name, not 'bash: 1 tool call' (#802)

### LINKED ISSUE #802: feat: show tool name instead of "bash: 1 tool call" when there's a single tool call in the task log
## Summary

When a task log entry contains exactly one tool call, the UI currently displays a generic label like `bash: 1 tool call`. Instead it should display the actual tool name (e.g. `read_file`, `str_replace_editor`, `bash`) so the user gets useful context at a glance without having to expand the disclosure.

## Desired behaviour

- **Single tool call**: show the tool name directly, e.g. `str_replace_editor` or `read_file("src/foo.py")` (with truncated args if space allows)
- **Multiple tool calls**: keep the existing `bash: N tool calls` summary format (or similar)

## Context

Task log entries are rendered in the Task Log tab (see PR #772 for the disclosure/collapsible work). The `line` summary field on `TaskLog` rows is what drives the display label. The fix may be in:
- The worker-side summariser that generates `line` (e.g. in `claude_runner.py` / `codex_runner.py`) — when there's exactly one tool call, emit the tool name instead of the generic count
- Or the frontend renderer — detect the `N tool call(s)` pattern and substitute the actual tool name from `data`

Either approach is acceptable; the worker-side fix is probably cleaner.

jmelloy:
## Summary

- The Task Log tab's turn-grouping collapsed *every* run of consecutive tool_use/tool_result entries into a "▸ Turn N — name × count (N tool calls)" summary — even when that run was a single tool call, hiding the actually useful detail (e.g. the bash command or file path) behind an extra click.
- `LogList.vue`'s grouping now only forms a collapsible "turn" when there are 2+ tool calls in a row. A lone tool call renders its own log line directly (e.g. `▶ bash: pytest`, `▶ read: src/foo.py`), which already carries the tool name and notable argument from the worker-side runners (`claude_runner.py` / `codex_runner.py` / `pi_runner.py`).
- Multiple tool calls in a row still collapse into the existing "Turn N — name × count (N tool calls)" summary, unchanged.

## Test plan

- [x] `npx vitest run src/components/log-pane/__tests__/LogList.spec.ts` — updated the single-tool-call turn test to assert it renders inline (no `.turn-summary`), adjusted the "expand" and "numbers multiple turns" tests to use 2+ tool calls per turn, all passing.
- [x] `npx vitest run` — full frontend suite passes (111 passed, 4 skipped).

Closes #802

### DIFF
--- a/frontend/src/components/log-pane/LogList.vue
+++ b/frontend/src/components/log-pane/LogList.vue
@@ -80,7 +80,6 @@ const groupedLogs = computed<(LineItem | TurnItem)[]>(() => {
 
   const flush = () => {
     if (current.length === 0) return
-    turnNumber += 1
     const counts: Record<string, number> = {}
     let toolCallCount = 0
     for (const { log } of current) {
@@ -90,6 +89,17 @@ const groupedLogs = computed<(LineItem | TurnItem)[]>(() => {
         toolCallCount += 1
       }
     }
+    // A single tool call carries no useful grouping information — showing its
+    // own line (e.g. "▶ bash: pytest") is more informative than collapsing it
+    // behind a generic "bash × 1 (1 tool call)" turn summary.
+    if (toolCallCount <= 1) {
+      for (const entry of current) {
+        result.push({ type: 'line', log: entry.log, index: entry.index })
+      }
+      current = []
+      return
+    }
+    turnNumber += 1
     result.push({ type: 'turn', turnNumber, entries: current, counts, toolCallCount })
     current = []
   }
--- a/frontend/src/components/log-pane/LogList.vue
+++ b/frontend/src/components/log-pane/LogList.vue
@@ -91,7 +91,9 @@ const groupedLogs = computed<(LineItem | TurnItem)[]>(() => {
     }
     // A single tool call carries no useful grouping information — showing its
     // own line (e.g. "▶ bash: pytest") is more informative than collapsing it
-    // behind a generic "bash × 1 (1 tool call)" turn summary.
+    // behind a generic "bash × 1 (1 tool call)" turn summary. toolCallCount can
+    // also be 0 here (a run of tool_result entries with no tool_use), which
+    // likewise needs no grouping and should just pass through as plain lines.
     if (toolCallCount <= 1) {
       for (const entry of current) {
         result.push({ type: 'line', log: entry.log, index: entry.index })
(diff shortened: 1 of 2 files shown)


frontend/src/components/log-pane/LogList.vue
```diff
@@ -90,6 +89,17 @@ const groupedLogs = computed<(LineItem | TurnItem)[]>(() => {
         toolCallCount += 1
       }
     }
```
jmelloy, line 91: The condition `toolCallCount <= 1` also silently bypasses turn grouping for groups with zero tool_use entries. If the accumulation logic ever allows non-tool entries into `current` (e.g. only tool_result messages), they would be emitted as individual lines with no turn number, which may be correct but is worth a brief comment clarifying the intent covers both 0 and 1 cases.

frontend/src/components/log-pane/__tests__/LogList.spec.ts
```diff
@@ -48,21 +48,38 @@ describe('LogList turn grouping', () => {
...
+      toolLog('Read', 'tool_result'),
+      toolLog('Bash', 'tool_use'),
+      toolLog('Bash', 'tool_result'),
+    ]
     const wrapper = mount(LogList, { props: { logs } })
 
     expect(wrapper.find('.turn-body').exists()).toBe(false)
     await wrapper.find('.turn-summary').trigger('click')
     expect(wrapper.find('.turn-body').exists()).toBe(true)
   })
```
jmelloy, line 63: The new test asserts `wrapper.text()).toContain('▶ Read')` which relies on the log message format produced by `toolLog`. It would be more robust to also assert that `.turn-summary` does not exist (already done) AND that exactly the expected number of `.log-line` elements are present, guarding against future regressions where both inline lines and a summary could accidentally co-render.

jmelloy (commented):
- The core logic change is minimal and correct: a single tool call (toolCallCount <= 1) is now rendered inline rather than collapsed into a turn summary, improving UX by exposing useful context like the bash command or file path directly.
- The `turnNumber` increment is correctly moved to only trigger when there are 2+ tool calls, preserving sequential numbering for multi-call turns.
- Test updates are well-structured: the existing 'expands a turn' test is correctly updated to use 2 tool calls (so it still exercises the collapsible turn path), a new dedicated test covers the single-tool-call inline rendering, and the sequential numbering test is updated to require multi-call groups.
- One subtle edge case: `toolCallCount <= 1` means a group with zero tool_use entries (e.g. only tool_result orphans, or only non-tool messages somehow routed into `current`) also bypasses turn grouping — this seems acceptable but worth noting.
- No security issues; no performance concerns given this is a computed property over a bounded log list.

PR #818, opened 2026-07-07
fix: guild settings bedrock provider fails

### LINKED ISSUE #817: guild settings provider pulldown — "bedrock" fails, "default" works
## Investigation: guild settings provider pulldown — "bedrock" fails, "default" works

The guild settings dropdown does not work when you select bedrock as the provider.

### Data flow traced

1. **Frontend** — `frontend/src/components/TopBar.vue`'s Foreman settings section (`showForemanConfig`). The provider `<select>` (lines 145–150) has value `""` ("default (anthropic)") plus one `<option>` per entry from `GET /api/models` (`modelsStore.providers`, populated from `backend/util/models_dev.py`, normalized IDs `anthropic` / `bedrock`). `saveForemanConfig()` (lines 467–490) sends `body.provider = foremanProvider.value || null` via `PATCH /api/guilds/{id}/foreman-config`. So "default" → `provider: null`; "bedrock" → `provider: "bedrock"`.

2. **Backend route** — `backend/routes/guilds.py`'s `update_foreman_config` (line 444). `ForemanConfigUpdate.provider: str | None`. When `null` is submitted, the merge loop (line 481-482) does `config.pop("provider", None)` — i.e. **"default" doesn't store `provider: "anthropic"`, it deletes the key entirely**, leaving the stored guild config with no `provider` field at all. When `"bedrock"` is submitted, `config["provider"] = "bedrock"` is stored verbatim.

3. **Foreman run / provider resolution** — `backend/foreman/runner.py` (~line 617-631, 784-792):
   ```python
   cfg_provider = guild_cfg.get("provider")          # None for "default", "bedrock" for explicit
   ...
   effective_provider = cfg_provider or os.environ.get("FOREMAN_PROVIDER", "anthropic").lower()
   if cfg_provider or cfg_env_vars:
       client = make_anthropic_client(provider=effective_provider, extra_env=cfg_env_vars or None)
   else:
       client = _get_anthropic_client()               # cached singleton, built once from pure os.environ
   foreman_model = cfg_model or get_foreman_model(provider=effective_provider)
   ```
   With **"default"**, `cfg_provider` is `None`. Unless the deployment's process env already has `FOREMAN_PROVIDER=bedrock` set, `effective_provider` resolves to `"anthropic"` — the direct Anthropic API path. `AWS_BEARER_TOKEN_BEDROCK` sitting in the environment is **never read** on this path; it's a red herring. "Default" "works" simply because it's exercising the ordinary Anthropic-API branch (which succeeds as long as `ANTHROPIC_API_KEY`/`ANTHROPIC_AUTH_TOKEN` is valid), not because it's using Bedrock at all.

   With **explicit "bedrock"**, `effective_provider` is definitively `"bedrock"`, so it takes the real Bedrock branch in `make_anthropic_client()` (`backend/foreman_core/llm.py`, lines 196-250). Credential resolution there is actually fine — `AWS_BEARER_TOKEN_BEDROCK` is picked up correctly from merged `os.environ` (line 194: `env = {**os.environ, **(extra_env or {})}`, line 208: `bearer_token = env.get("AWS_BEARER_TOKEN_BEDROCK")`) and bearer-token auth succeeds in building the client.

### Root cause: hardcoded fallback model ARN, not credentials

The actual failure is in **model resolution**, not client/credential construction:

`backend/foreman_core/llm.py`, lines 23-24:
```python
_DEFAULT_BEDROCK_MODEL = (
    "arn:aws:bedrock:us-east-1:446872464738:inference-profile/us.anthropic.claude-sonnet-4-6"
)
```
and `get_foreman_model()` (lines 39-50):
```python
def get_foreman_model(provider: str | None = None) -str:
    resolved = (provider or os.environ.get("FOREMAN_PROVIDER", "anthropic")).lower()
    if resolved == "bedrock":
        return os.environ.get("FOREMAN_BEDROCK_MODEL", _DEFAULT_BEDROCK_MODEL)
    return os.environ.get("FOREMAN_MODEL", "claude-sonnet-4-6")
```

If the guild's `model` field is left blank (as in the reported repro — only the provider pulldown was changed), `foreman_model = cfg_model or get_foreman_model(provider=effective_provider)` falls through to `get_foreman_model("bedrock")`. Unless `FOREMAN_BEDROCK_MODEL` is *also* set in the process env, this returns the **hardcoded `_DEFAULT_BEDROCK_MODEL` ARN, which points at AWS account `446872464738`** (an example/placeholder account, not the guild owner's own account). Amazon Bedrock cross-region inference profiles are account-scoped — a Bedrock `InvokeModel` call against another account's inference-profile ARN fails (AccessDenied / ResourceNotFound) regardless of whether the caller's own AWS credentials (including a valid `AWS_BEARER_TOKEN_BEDROCK`) are otherwise correct.

This is confirmed by the existing test suite (`backend/tests/test_foreman_llm.py`), which explicitly asserts this fallback behavior is "working as designed":
```python
def test_bedrock_provider_env_returns_bedrock_model(self, monkeypatch):
    monkeypatch.setenv("FOREMAN_PROVIDER", "bedrock")
    monkeypatch.delenv("FOREMAN_BEDROCK_MODEL", raising=False)
    assert get_foreman_model() == _DEFAULT_BEDROCK_MODEL
```

So: **"default" never reaches the Bedrock branch at all (so the placeholder ARN is never exercised), while explicit "bedrock" always reaches it unless the model field or `FOREMAN_BEDROCK_MODEL` is also filled in — that's the exact point of divergence.**

### Recommended fix

1. In `get_foreman_model()` / `make_anthropic_client()` (`backend/foreman_core/llm.py`), stop silently defaulting to another account's ARN when `provider == "bedrock"` and no model is configured. Either raise a clear, user-facing error ("Bedrock selected but no model/inference-profile ARN configured — set one in guild settings or `FOREMAN_BEDROCK_MODEL`"), or reject the request earlier.
2. Add server-side validation in `update_foreman_config` (`backend/routes/guilds.py`): if `provider == "bedrock"` is being saved and neither the incoming nor existing stored `model` is set (and `FOREMAN_BEDROCK_MODEL` isn't configured), return a 400 with an actionable message instead of silently accepting a config that will fail at runtime.
3. Frontend (`TopBar.vue`): when the provider `<select>` is switched to `bedrock` and the Model field is empty, show an inline hint/required-field warning rather than letting it save silently (mirrors the existing per-provider model memory in `modelByProvider`/`onProviderChange`).
4. Consider folding this into #811's consolidation work, since fixing it in `backend/foreman_core/llm.py` alone won't help the standalone foreman unless `foreman/pioneer_foreman/config.py`'s duplicated `effective_model` property gets the same guard.

Traced by an automated investigation on branch `claude/investigate-guild-settings-provider-pulldown-backend-flow-t-snfexu`.
 

 _Originally posted by @jmelloy in [#815](https://github.com/jmelloy/pioneer-square/issues/815#issuecomment-4906867055)_

jmelloy:
## Summary
- Removes the hardcoded Bedrock inference-profile ARN default (`arn:aws:bedrock:us-east-1:446872464738:inference-profile/...`) — inference profiles are AWS-account-scoped, so silently falling back to someone else's ARN made `provider=bedrock` always fail against the wrong account.
- `get_foreman_model(provider="bedrock")` now raises a clear `BedrockModelNotConfiguredError` when no model/ARN is configured, instead of returning a bogus cross-account default.
- `PATCH /api/guilds/{id}/foreman-config` now rejects saving `provider=bedrock` up front when no `model` is set (and no `FOREMAN_BEDROCK_MODEL` env fallback exists), so the guild settings UI surfaces the problem immediately rather than failing later inside the foreman run loop.
- `make_anthropic_client` correctly builds an `AsyncAnthropicBedrock` client for the bedrock provider (not a plain `Anthropic`/`AsyncAnthropic` client), and threads guild-configured AWS credentials (`AWS_ACCESS_KEY_ID`/`AWS_SECRET_ACCESS_KEY`/`AWS_SESSION_TOKEN`/`AWS_DEFAULT_REGION`/`AWS_PROFILE`/`AWS_BEARER_TOKEN_BEDROCK`) from the guild settings dialogue's `env_vars` through to the client via `extra_env`, since those never otherwise reach the foreman process.
- Adds a credential probe (`_log_bedrock_credentials`) that replicates boto3's credential resolution at client-creation time so missing/expired AWS credentials surface a clear log reason instead of the SDK's opaque "could not resolve credentials from session" error.
- Updates `.env.example`, `docker-compose.yml`, and standalone foreman config (`foreman/pioneer_foreman/config.py`) to drop the same bad hardcoded default and require explicit configuration for Bedrock.

Fixes #817

## Test plan
- [x] `pytest backend/tests/test_foreman_llm.py backend/tests/test_guild_api.py` — 47 passed, including new regression tests for the bedrock-without-model rejection (`test_foreman_config_bedrock_without_model_rejected`) and successful save with a model (`test_foreman_config_bedrock_with_model_accepted`), plus `test_bedrock_provider_env_without_model_raises`.
- [ ] Manual: set guild provider to `bedrock` with a valid inference-profile ARN + AWS credentials via the settings dialogue and confirm the foreman AI responds successfully.

### DIFF
--- a/backend/foreman_core/llm.py
+++ b/backend/foreman_core/llm.py
@@ -58,9 +58,15 @@ def get_foreman_model(provider: str | None = None) -> str:
         bedrock_model = os.environ.get("FOREMAN_BEDROCK_MODEL")
         if not bedrock_model:
             raise BedrockModelNotConfiguredError(
-                "Bedrock provider selected but no model is configured. Set a model "
-                "(inference-profile ARN or model ID) in the guild's foreman settings, "
-                "or set the FOREMAN_BEDROCK_MODEL environment variable."
+                "Bedrock provider selected but no model is configured. Unlike AWS "
+                "credentials, which boto3 can resolve on its own (IAM role, profile, "
+                "env vars) and report clearly if missing, there is no discoverable "
+                "default model or inference profile — the Bedrock API call itself "
+                "requires one to be passed explicitly, and an account-scoped ARN "
+                "guessed here could silently send requests to the wrong AWS account. "
+                "Set a model (inference-profile ARN or model ID) in the guild's "
+                "foreman settings, or set the FOREMAN_BEDROCK_MODEL environment "
+                "variable."
             )
         return bedrock_model
     return os.environ.get("FOREMAN_MODEL", "claude-sonnet-4-6")
@@ -224,9 +230,15 @@ def make_anthropic_client(
         resolved_model = model or env.get("FOREMAN_BEDROCK_MODEL")
         if not resolved_model:
             raise BedrockModelNotConfiguredError(
-                "Bedrock provider selected but no model is configured. Set a model "
-                "(inference-profile ARN or model ID) in the guild's foreman settings, "
-                "or set the FOREMAN_BEDROCK_MODEL environment variable."
+                "Bedrock provider selected but no model is configured. Unlike AWS "
+                "credentials, which boto3 can resolve on its own (IAM role, profile, "
+                "env vars) and report clearly if missing, there is no discoverable "
+                "default model or inference profile — the Bedrock API call itself "
+                "requires one to be passed explicitly, and an account-scoped ARN "
+                "guessed here could silently send requests to the wrong AWS account. "
+                "Set a model (inference-profile ARN or model ID) in the guild's "
+                "foreman settings, or set the FOREMAN_BEDROCK_MODEL environment "
+                "variable."
             )
         resolved_region = (
             region or env.get("AWS_DEFAULT_REGION") or env.get("AWS_REGION") or _BEDROCK_REGION
--- a/backend/foreman_core/llm.py
+++ b/backend/foreman_core/llm.py
@@ -20,15 +20,22 @@ except ImportError:
 
 logger = logging.getLogger(__name__)
 
-_DEFAULT_BEDROCK_MODEL = (
-    "arn:aws:bedrock:us-east-1:446872464738:inference-profile/us.anthropic.claude-sonnet-4-6"
-)
+
+class BedrockModelNotConfiguredError(ValueError):
+    """Raised when provider=bedrock is selected but no model/inference-profile is configured.
+
+    Bedrock has no universal default model — inference profile ARNs are scoped to a
+    single AWS account — so there is no safe value to fall back to silently. Callers
+    must configure one explicitly (guild settings `model` field or FOREMAN_BEDROCK_MODEL).
+    """
+
 
 FOREMAN_MODEL = os.environ.get("FOREMAN_MODEL", "claude-sonnet-4-6")
-# Bedrock uses cross-region inference profiles, not plain model IDs.
-# Set this to the profile ARN/ID appropriate for your region, e.g.:
-#   arn:aws:bedrock:us-east-1:446872464738:inference-profile/us.anthropic.claude-sonnet-4-6
-FOREMAN_BEDROCK_MODEL = os.environ.get("FOREMAN_BEDROCK_MODEL", _DEFAULT_BEDROCK_MODEL)
+# Bedrock uses cross-region inference profiles, not plain model IDs, and those
+# profiles are scoped to a single AWS account, so there is no valid cross-account
+# default. Set this to the profile ARN/ID appropriate for your account/region, e.g.:
+#   arn:aws:bedrock:us-east-1:<account-id>:inference-profile/us.anthropic.claude-sonnet-4-6
+FOREMAN_BEDROCK_MODEL = os.environ.get("FOREMAN_BEDROCK_MODEL")
 
 # Set FOREMAN_PROVIDER=bedrock to use Amazon Bedrock instead of the Anthropic API.
 # Requires: pip install "anthropic[bedrock]"  +  AWS credentials in env / IAM role.
@@ -40,13 +47,22 @@ def get_foreman_model(provider: str | None = None) -> str:
     """Return the model ID to use for the given provider.
 
     When provider is 'bedrock' (or FOREMAN_PROVIDER=bedrock), returns
-    FOREMAN_BEDROCK_MODEL; otherwise returns FOREMAN_MODEL.
+    FOREMAN_BEDROCK_MODEL, raising BedrockModelNotConfiguredError if it's unset —
+    Bedrock inference-profile ARNs are AWS-account-scoped, so there is no safe
+    default to fall back to. Otherwise returns FOREMAN_MODEL.
 
     Reads os.environ on every call so that tests can patch env vars directly.
     """
     resolved = (provider or os.environ.get("FOREMAN_PROVIDER", "anthropic")).lower()
     if resolved == "bedrock":
-        return os.environ.get("FOREMAN_BEDROCK_MODEL", _DEFAULT_BEDROCK_MODEL)
+        bedrock_model = os.environ.get("FOREMAN_BEDROCK_MODEL")
+        if not bedrock_model:
+            raise BedrockModelNotConfiguredError(
+                "Bedrock provider selected but no model is configured. Set a model "
+                "(inference-profile ARN or model ID) in the guild's foreman settings, "
+                "or set the FOREMAN_BEDROCK_MODEL environment variable."
+            )
+        return bedrock_model
     return os.environ.get("FOREMAN_MODEL", "claude-sonnet-4-6")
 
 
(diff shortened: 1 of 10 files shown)


backend/foreman_core/llm.py
```diff
@@ -213,7 +229,7 @@ def make_anthropic_client(
             "bearer-token"
```
jmelloy, line 259: Using `"<not configured>"` here means `make_anthropic_client` can be called with an invalid model string and will only fail later (inside the Bedrock API call) rather than raising `BedrockModelNotConfiguredError` immediately. Consider calling `get_foreman_model("bedrock")` or reading `env.get("FOREMAN_BEDROCK_MODEL")` and raising early if absent, to keep the fast-fail behavior consistent with the rest of the PR.

backend/routes/guilds.py
```diff
@@ -482,6 +483,23 @@ async def update_foreman_config(
             config.pop(field, None)
         else:
             config[field] = value
+
+    # Bedrock inference-profile ARNs are AWS-account-scoped, so there is no safe
+    # default model to fall back to at run time (see #817). Reject the save now
+    # rather than let it fail opaquely against a placeholder/wrong-account ARN.
+    if (
```
jmelloy, line 490: This validation fires only on the PATCH that sets `provider=bedrock`, but existing guild configs stored before this fix (with the old placeholder ARN or no model) will bypass it entirely at run time. Consider also applying this check when the foreman run loop resolves the model, or add a DB migration to clear/flag affected rows.

foreman/pioneer_foreman/config.py
```diff
@@ -190,8 +199,7 @@ def load(explicit_path: str | None = None, overrides: dict | None = None) -> Con
         overrides.get("bedrock_model")
         or claude_block.get("bedrock_model")
         or os.environ.get("FOREMAN_BEDROCK_MODEL")
-        or _DEFAULT_BEDROCK_MODEL
-    )
+    ) or None
```
jmelloy, line 206: The expression `(...) or None` correctly converts an empty string from `os.environ.get` to `None`, but it also silently discards a legitimately empty `overrides` or TOML value. This is intentional here (empty string is not a valid ARN), but a brief inline comment would help future readers understand why `or None` is appended rather than leaving the assignment as a plain `or`.

jmelloy (commented):
- The core fix is correct and well-motivated: removing the hardcoded cross-account Bedrock ARN prevents silent failures against the wrong AWS account, and surfacing the error early (at save time) is good UX.
- Test coverage is solid — regression tests for both the rejection case and the acceptance case are added, and the unit tests are updated to match the new behavior.
- A subtle double-negation expression `or None` on the last line of `config.py` is idiomatic but could be clearer; it correctly collapses empty-string env vars to `None`.
- The logging line in `make_anthropic_client` now uses `env.get("FOREMAN_BEDROCK_MODEL") or "<not configured>"` which is a safe change, but it silently swallows the unconfigured state rather than propagating the error — this could mask misconfigurations in the bedrock client construction path.
- The validation in `guilds.py` only runs during the PATCH request and does not re-validate on read/run, so a guild saved before this fix with the old placeholder ARN will still fail at run time with the old opaque error; a migration or run-time guard would harden this fully.

backend/foreman_core/llm.py
```diff
@@ -40,13 +47,22 @@ def get_foreman_model(provider: str | None = None) -> str:
...
 
     Reads os.environ on every call so that tests can patch env vars directly.
     """
     resolved = (provider or os.environ.get("FOREMAN_PROVIDER", "anthropic")).lower()
     if resolved == "bedrock":
-        return os.environ.get("FOREMAN_BEDROCK_MODEL", _DEFAULT_BEDROCK_MODEL)
+        bedrock_model = os.environ.get("FOREMAN_BEDROCK_MODEL")
+        if not bedrock_model:
+            raise BedrockModelNotConfiguredError(
+                "Bedrock provider selected but no model is configured. Set a model "
```
jmelloy: if this env doesn't exist can't we call AWS and get it same as the client does?

jmelloy: Good question, but I don't think we can defer to AWS the way the credential path does. boto3's credential chain has a whole hierarchy of sources it checks (IAM role, profile, env vars, SSO) before failing, and when it does fail it's clear about why (that's what `_log_bedrock_credentials` probes for). The model isn't like that — there's no chain to walk and no "default" Bedrock model; it's just a required parameter on the `InvokeModel` call. If we let it through as `None`/missing, the failure would surface as an opaque SDK-level error deep inside the first request instead of a clear message pointing at guild settings or `FOREMAN_BEDROCK_MODEL`.

There's also the account-scoping issue: even if we picked *some* hardcoded model ID as a "default" to avoid the early raise, Bedrock inference profile ARNs are scoped to a specific AWS account, so a hardcoded fallback would silently send requests to the wrong account for anyone whose account ID differs — worse than failing fast.

So I kept the early raise but reworded the error message (809f7c7) to spell out this reasoning (no discoverable default + account-scoping risk) so it's clear why this isn't handled like credential resolution.

PR #820, opened 2026-07-07
foreman: drop trigger queueing, use drop-if-busy (#808)

### LINKED ISSUE #808: foreman: drop trigger queueing in favor of drop-if-busy, matching embedded foreman policy
## Context

Part of the audit for #790 (foreman-proxy slim-down). The standalone foreman (`foreman/pioneer_foreman/foreman.py`) buffers triggers that arrive while a run is in flight in a 100-slot `asyncio.Queue` (`_message_queue`), then drains it FIFO after the in-flight run completes (`_run_foreman`, lines ~272-342). `[periodic-check]` triggers are special-cased to be dropped while busy; everything else queues.

The embedded foreman explicitly rejected this design. `backend/foreman/runner.py:531-544` docstring:

> "If the lock is already held the invocation is dropped; the poll loop re-triggers on the next tick. **Dropping is preferred over queuing to avoid unbounded build-up.**"

So the two foremen currently implement opposite concurrency policies for the same trigger types, and the standalone one is the more complex, more failure-prone of the two (a queue that never drains because an early item hangs is a real risk the lock-and-drop model doesn't have).

## Work

1. Delete `_message_queue`, `_QUEUE_MAX`, and the FIFO-drain loop inside `_run_foreman` in `foreman/pioneer_foreman/foreman.py`.
2. Replace the `_processing` boolean with the same drop-if-busy semantics as the embedded foreman: if a run is already in flight for the relevant key (whole-guild or per-task, see the companion issue on per-task locks), log and drop the new trigger instead of buffering it.
3. Rely on the backend-side poll/re-trigger mechanism (see the companion issue on moving polling to the backend) to recover any dropped trigger on its next tick, exactly as the embedded foreman does.
4. Update/trim `foreman/tests/test_ws_integration.py` coverage that currently exercises queue-drain behavior.
5. Update `docs/foreman-split-plan.md`'s "Message queue" section (currently documents the 100-slot buffer as a shipped design decision) to reflect the new drop-if-busy policy.

This is unnecessary state management duplicated — and inconsistently — from logic the backend already implements correctly and more simply. A thin proxy shouldn't need its own retry/backpressure policy when the system it's proxying for already has one it can piggyback on.

jmelloy:
## Summary
- Removes the standalone foreman's FIFO `_message_queue` (100-slot buffer) and `_processing` boolean.
- Replaces it with drop-if-busy semantics keyed by `(guild_id, user_id)`, matching the embedded foreman's `run_foreman_ai` lock policy in `backend/foreman/runner.py:531-544` exactly: if a run is already in flight for a key, new triggers are dropped (not buffered) and recovered by the backend's poll/re-trigger mechanism on the next tick.
- Different users' triggers for the same guild can now proceed concurrently instead of all serializing behind one global flag; poll-triggered (system) runs still serialize against each other and against user-triggered runs sharing the same key.
- Per-task child contexts (`task_context.py`) are unaffected — they already have their own independent per-task queue.
- Trims/updates `foreman/tests/test_ws_integration.py` to cover drop-if-busy instead of queue-drain/FIFO/overflow behavior.
- Updates `docs/foreman-split-plan.md` and `docs/foreman-per-task-context.md` to describe the drop-if-busy policy instead of the retired message queue.

Closes #808

## Test plan
- [x] `python -m pytest foreman/tests/test_ws_integration.py -v` — all 13 tests pass
- [x] `python -m pytest foreman/tests/` — 94 passed (3 pre-existing unrelated errors due to missing `respx_mock` fixture, not touched by this change)
- [x] `ruff check foreman/pioneer_foreman/foreman.py foreman/tests/test_ws_integration.py` — clean

### DIFF
--- a/foreman/pioneer_foreman/foreman.py
+++ b/foreman/pioneer_foreman/foreman.py
@@ -183,6 +183,7 @@ class Foreman:
                 user_id: str | None = None,
                 task_id: str | None = None,
                 task_name: str = "foreman.run",
+                active_run_key: tuple[str, str | None] | None = None,
             ) -> None:
                 """Run one parent (whole-guild) foreman turn.
 
@@ -190,8 +191,13 @@ class Foreman:
                 log and drop the trigger rather than buffering it — the backend's
                 poll/re-trigger mechanism recovers it on the next tick. Matches
                 backend/foreman/runner.py:run_foreman_ai (the embedded foreman's policy).
+
+                `active_run_key` overrides the default (guild_id, user_id) key used for
+                `_active_runs` bookkeeping — used by the poll loop so its own in-flight
+                check doesn't collide with a real user trigger whose `user_id` is None.
+                It has no bearing on `user_id`, which is passed to run_foreman_ai as-is.
                 """
-                key = (guild_id, user_id)
+                key = active_run_key if active_run_key is not None else (guild_id, user_id)
                 if key in self._active_runs:
                     if "[periodic-check]" in human_message:
                         logger.debug(
@@ -219,6 +225,9 @@ class Foreman:
                         config=self._config,
                     )
                 except Exception:
+                    # Must stay inside this try/finally: if a future refactor moves this
+                    # `except` outside, an unhandled error would skip the `finally` below
+                    # and leak `key` in `_active_runs`, permanently wedging that key.
                     logger.exception("guild=%s %s: unhandled error", guild_id, task_name)
                 finally:
                     self._active_runs.discard(key)
--- a/foreman/pioneer_foreman/foreman.py
+++ b/foreman/pioneer_foreman/foreman.py
@@ -25,8 +25,6 @@ if TYPE_CHECKING:
 
 logger = logging.getLogger(__name__)
 
-_QUEUE_MAX = 100
-
 
 class Foreman:
     """Manages the standalone foreman lifecycle for one guild."""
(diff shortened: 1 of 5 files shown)


foreman/pioneer_foreman/foreman.py
```diff
@@ -459,7 +404,7 @@ async def _poll_loop() -> None:
                         base = min(self._poll_interval, sleep_duration)
```
jmelloy: The poll-loop guard uses `(self._config.guild_id, None)` as the key, but a user trigger with `user_id=None` (e.g. a trigger whose payload omits `userId`) would also insert that exact key, causing the poll loop to be suppressed while that user run is active. Consider using a dedicated sentinel (e.g. `_POLL_KEY = (guild_id, '__poll__')`) to avoid the collision.

foreman/tests/test_ws_integration.py
```diff
@@ -427,53 +389,37 @@ async def tracking_run(guild_id, human_message, *, http, ws_send, config, **kwar
...
-        processed.append(human_message)
+    async def blocking_run(*args, **kwargs):
+        nonlocal call_count
+        call_count += 1
+        await gate.wait()
         return False
 
-    import logging
-
     with (
```
jmelloy, line 396: The `MockWebSocket` delivers all three messages (u1, u2, evicted) before `asyncio.sleep(0.05)` yields; by the time `gate.set()` is called the connection may already be torn down, making the assertion on `call_count == 2` potentially a false positive depending on event-loop scheduling. Consider using two separate gates (one per user) and awaiting them individually to ensure both runs are genuinely in flight simultaneously before proceeding.

foreman/pioneer_foreman/foreman.py
```diff
@@ -278,68 +277,48 @@ async def _run_foreman(
...
-                        guild_id,
-                        task_name,
-                    )
-                    return
-                self._processing = True
-                try:
-                    try:
-                        made_calls = await run_foreman_ai(
+                key = (guild_id, user_id)
+                if key in self._active_runs:
```
jmelloy, line 201: When `run_foreman_ai` raises an exception the key is correctly removed via `finally: self._active_runs.discard(key)`, but the exception is caught and logged inside the `try` block, so the `finally` still runs — this is fine. However, if a future refactor moves the bare `except Exception` outside the `try/finally`, the key would leak. A brief comment noting that the `except` must remain inside the `try` would protect against that regression.

jmelloy (commented):
- The refactor correctly replaces a global `_processing` flag + 100-slot FIFO queue with a per-`(guild_id, user_id)` in-flight set, enabling concurrent runs for different users while still serialising same-key runs.
- Drop-if-busy semantics rely entirely on the backend's poll/re-trigger mechanism for recovery; if that mechanism has gaps or delays, triggers can be silently lost with no client-visible feedback.
- The key `(guild_id, None)` is used for poll-triggered (system) runs, meaning a system poll blocks any user with `user_id=None` and vice-versa — this edge case should be documented or explicitly handled.
- The new test `test_ws_triggers_for_different_users_run_concurrently` uses `gate.wait()` in both runs and then checks `call_count == 2`, but the gate is only set after `asyncio.sleep(0.05)` while the WS is already at `foreman-evicted`; the test may be racing and could pass vacuously.
- Overall the change is well-scoped, reduces memory pressure, and aligns standalone and embedded foreman behaviour; documentation updates are accurate and complete.

PR #822, opened 2026-07-07
foreman: consolidate Config provider/model branching (#811)

### LINKED ISSUE #811: foreman: consolidate Config's provider/model branching with backend.foreman_core.llm instead of duplicating it
## Context

Part of the audit for #790 (foreman-proxy slim-down). `foreman/pioneer_foreman/config.py`'s `Config.effective_model` property (lines 65-74) has this docstring:

> "Intentional duplication of the provider-branching logic in `backend.foreman_core.llm.get_foreman_model()`: Config is used by the standalone foreman without the full backend env-var stack, so it needs its own copy operating on dataclass fields rather than `os.environ`."

This is a deliberate, acknowledged duplication rather than an oversight, but it's exactly the shape of thing #790 wants gone: model/provider selection logic that exists in the backend gets re-derived a second time in the proxy from a parallel set of fields (`model`, `bedrock_model`, `provider`, `aws_region`, `aws_profile`, `anthropic_auth_token`, `api_key`).

Note: unlike the other sub-issues here, the *capability* itself (choosing model/provider independently of the backend, including local/self-hosted models) is core to the actual motivation for having a standalone foreman at all per #790's original description ("run the Foreman brain on their own hardware with any model") and per `AGENTS.md`'s "independent scaling and model changes without restarting the backend." This issue is about removing the duplicated branching logic, not the flexibility itself.

## Work

1. Look at whether `backend.foreman_core.llm.make_anthropic_client()` / `get_foreman_model()` can be parameterized to accept an explicit config object/dict instead of reading `os.environ`, so `Config.effective_model` and `_get_anthropic_client()` in `foreman/pioneer_foreman/runner.py` can call the shared function directly instead of re-implementing the anthropic-vs-bedrock branch.
2. If that refactor is awkward (backend code may have other callers relying on the env-var-reading behavior), at minimum extract the shared *branching logic* (not the env lookup) into a small pure function in `backend/foreman_core/llm.py` that both the backend's env-based path and the standalone's dataclass-based path call.
3. Add a regression test asserting `Config.effective_model` and `backend.foreman_core.llm.get_foreman_model()` agree given equivalent inputs, so the two don't silently drift again.

Not a deletion candidate like the others — this is the "reduce duplication without losing capability" case the audit turned up. Worth tracking separately since it touches shared backend code (`backend/foreman_core/llm.py`) rather than being foreman-only.

jmelloy:
## Summary
- `Config.effective_model` (foreman/pioneer_foreman/config.py) previously re-implemented the provider-branching logic already present in `backend.foreman_core.llm.get_foreman_model()`, intentionally duplicated because `get_foreman_model()` only read `os.environ` and the standalone foreman's `Config` needed to branch on dataclass fields instead.
- Parameterized `get_foreman_model()` to accept optional `model`/`bedrock_model` overrides, checked before falling back to `FOREMAN_MODEL`/`FOREMAN_BEDROCK_MODEL` env vars. This preserves all existing env-var-reading behavior for callers that don't pass overrides (backend's `foreman/runner.py`, `foreman/tools.py`).
- `Config.effective_model` now delegates to `get_foreman_model(self.provider, model=self.model, bedrock_model=self.bedrock_model)` instead of re-implementing the branch and raising its own copy of `BedrockModelNotConfiguredError`.
- `make_anthropic_client()` / `runner._get_anthropic_client()` were already properly parameterized (no duplication there) — this PR only touches the `effective_model`/`get_foreman_model` pair called out in #811.

## Test plan
- [x] `pytest backend/tests -q` — 899 passed, 2 skipped
- [x] `pytest foreman/tests -q` — 93 passed (3 pre-existing failures in `test_tools.py` due to a missing `respx` dependency in this environment, unrelated to this change — confirmed present on `main` too)
- [x] `ruff check` on both changed files

Closes #811

### DIFF
--- a/backend/foreman_core/llm.py
+++ b/backend/foreman_core/llm.py
@@ -67,9 +67,7 @@ def get_foreman_model(
     resolved = (provider or os.environ.get("FOREMAN_PROVIDER", "anthropic")).lower()
     if resolved == "bedrock":
         resolved_bedrock_model = (
-            bedrock_model
-            if bedrock_model is not None
-            else os.environ.get("FOREMAN_BEDROCK_MODEL")
+            bedrock_model if bedrock_model is not None else os.environ.get("FOREMAN_BEDROCK_MODEL")
         )
         if not resolved_bedrock_model:
             raise BedrockModelNotConfiguredError(
--- a/backend/foreman_core/llm.py
+++ b/backend/foreman_core/llm.py
@@ -66,7 +66,11 @@ def get_foreman_model(
     """
     resolved = (provider or os.environ.get("FOREMAN_PROVIDER", "anthropic")).lower()
     if resolved == "bedrock":
-        resolved_bedrock_model = bedrock_model or os.environ.get("FOREMAN_BEDROCK_MODEL")
+        resolved_bedrock_model = (
+            bedrock_model
+            if bedrock_model is not None
+            else os.environ.get("FOREMAN_BEDROCK_MODEL")
+        )
         if not resolved_bedrock_model:
             raise BedrockModelNotConfiguredError(
                 "Bedrock provider selected but no model is configured. Unlike AWS "
@@ -80,7 +84,7 @@ def get_foreman_model(
                 "variable."
             )
         return resolved_bedrock_model
-    return model or os.environ.get("FOREMAN_MODEL", "claude-sonnet-4-6")
+    return model if model is not None else os.environ.get("FOREMAN_MODEL", "claude-sonnet-4-6")
 
 
 def _log_bedrock_credentials(
--- a/backend/foreman_core/llm.py
+++ b/backend/foreman_core/llm.py
@@ -43,20 +43,31 @@ FOREMAN_PROVIDER = os.environ.get("FOREMAN_PROVIDER", "anthropic").lower()
 _BEDROCK_REGION = os.environ.get("AWS_DEFAULT_REGION", "us-east-1")
 
 
-def get_foreman_model(provider: str | None = None) -> str:
+def get_foreman_model(
+    provider: str | None = None,
+    *,
+    model: str | None = None,
+    bedrock_model: str | None = None,
+) -> str:
     """Return the model ID to use for the given provider.
 
     When provider is 'bedrock' (or FOREMAN_PROVIDER=bedrock), returns
-    FOREMAN_BEDROCK_MODEL, raising BedrockModelNotConfiguredError if it's unset —
-    Bedrock inference-profile ARNs are AWS-account-scoped, so there is no safe
-    default to fall back to. Otherwise returns FOREMAN_MODEL.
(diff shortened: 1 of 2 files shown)


backend/foreman_core/llm.py
```diff
@@ -68,8 +79,8 @@ def get_foreman_model(provider: str | None = None) -> str:
                 "foreman settings, or set the FOREMAN_BEDROCK_MODEL environment "
                 "variable."
             )
-        return bedrock_model
-    return os.environ.get("FOREMAN_MODEL", "claude-sonnet-4-6")
+        return resolved_bedrock_model
+    return model or os.environ.get("FOREMAN_MODEL", "claude-sonnet-4-6")
```
jmelloy: Using `model or ...` means an explicitly passed empty string `""` is silently ignored and the env-var fallback is used instead. If callers can ever pass `model=""` intending 'no override', this is fine, but if they pass it expecting it to propagate, the behavior is surprising. Consider `model if model is not None else os.environ.get(...)` for a stricter contract.

foreman/pioneer_foreman/config.py
```diff
@@ -64,20 +64,12 @@ class Config:
...
-                    "Bedrock provider selected but no model is configured. Set "
-                    "[claude] bedrock_model in the config TOML or the "
-                    "FOREMAN_BEDROCK_MODEL environment variable."
-                )
-            return self.bedrock_model
-        return self.model
+        return get_foreman_model(self.provider, model=self.model, bedrock_model=self.bedrock_model)
 
     @property
     def http_url(self) -> str:
```
jmelloy, line 87: The old implementation raised `BedrockModelNotConfiguredError` with a Config/TOML-specific message ("Set [claude] bedrock_model in the config TOML or the FOREMAN_BEDROCK_MODEL environment variable."). The new path raises the generic backend error message that only mentions `FOREMAN_BEDROCK_MODEL` and foreman settings, losing the TOML-specific guidance. Consider passing a custom error message or keeping a thin wrapper for the Bedrock-not-configured case so standalone foreman users get actionable config-file instructions.

jmelloy (commented):
- The refactor successfully eliminates the duplicated provider-branching logic by parameterizing `get_foreman_model()`, which is a sound design improvement.
- One subtle behavioral change exists: the old `Config.effective_model` raised a `BedrockModelNotConfiguredError` with a Config-specific message mentioning `[claude] bedrock_model in the config TOML`; the new path raises the generic backend error message, losing that user-facing context.
- The keyword-only enforcement (`*`) on `model`/`bedrock_model` is a good defensive API choice that prevents accidental positional misuse.
- The fallback chain `bedrock_model or os.environ.get(...)` means an empty string `""` passed explicitly is treated as "not set" and falls through to the env var, which may be surprising but is consistent with the previous env-var-only behavior.
- No security issues or data-flow bugs introduced; existing test surface for env-var patching is preserved per the docstring.

PR #823, opened 2026-07-07
foreman: replace child-context supervisor with per-task lock (#807)

### LINKED ISSUE #807: foreman: replace per-task child-context supervisor with a simple per-task lock (mirror embedded foreman)
## Context

Part of the audit for #790 (foreman-proxy slim-down). The standalone foreman (`foreman/pioneer_foreman/`) implements per-task conversation isolation using a full supervisor-tree of long-lived asyncio tasks:

- `foreman/pioneer_foreman/task_context.py` (154 lines) — a `TaskContext` class with its own queue, consumer loop, and finalize-detection.
- `foreman/pioneer_foreman/foreman.py` — `_child_contexts: dict[str, TaskContext]`, `_pending_for_task` (pre-spawn buffer), `_route_to_child()`, `_spawn_child()`, `_deregister_child()`, `_parent_tool_observer()` (which regex-scrapes task ids out of tool-result strings), `_respawn_children_from_state()` (reconnect recovery), `_cancel_all_children()`. This is roughly 150-200 lines of orchestration/lifecycle code.

**The embedded foreman achieves the identical per-task isolation with far less code.** `backend/foreman/runner.py:531-577` (`run_foreman_ai`) keys a plain `dict[tuple, asyncio.Lock]` (`_guild_locks`) on `(guild_id, f"task:{task_id}")` when a trigger is task-scoped, `(guild_id, user_id)` otherwise. If the lock is already held, the call is dropped (see #<queue-issue> for why dropping beats queuing). No supervisor object, no pending-buffer, no reconnect-respawn — because there is nothing stateful to respawn: a plain lock dict rebuilds itself for free on the next trigger.

The *context-scoping* logic itself (narrowed system prompt / state preamble / history filtered by `task_id`) already lives behind a stateless `run_foreman_ai(child=True, task_id=...)` call in `foreman/pioneer_foreman/runner.py:132-167` and is shared via `backend/foreman_core/`. The supervisor tree adds nothing to *that* — it only adds concurrency bookkeeping that a lock dict already does more simply next door in the same codebase.

## Work

1. In `foreman/pioneer_foreman/foreman.py`, replace `_child_contexts`, `_pending_for_task`, `_route_to_child`, `_spawn_child`, `_deregister_child`, `_parent_tool_observer`, `_respawn_children_from_state`, `_cancel_all_children` with a `dict[str, asyncio.Lock]` keyed by `task_id`, following the exact pattern in `backend/foreman/runner.py:556-577`.
2. `_handle_trigger` dispatches directly: if `taskId` is present, acquire-or-drop the per-task lock and call `run_foreman_ai(child=True, task_id=...)`; otherwise use the existing whole-guild lock/dispatch.
3. Delete `foreman/pioneer_foreman/task_context.py` entirely.
4. Remove/replace `foreman/tests/test_task_context.py` and `foreman/tests/test_child_routing.py` with tests against the new lock-based dispatch.
5. Update `docs/foreman-per-task-context.md` — its own "Implementation note (2026-06-21)" section already observes the embedded foreman's lock-based approach is simpler; this closes that gap by porting it to the standalone side instead of maintaining the two designs in parallel.

This is exactly the kind of "state management" logic the operator wants out of the proxy: a multi-object lifecycle system whose entire job is decided more simply by seven lines already living in the backend. Deleting it removes the reconnect-respawn failure mode, the pending-buffer race window, and the tool-result regex scraping (`_TASK_ID_RE`) used only to infer which child to spawn.

jmelloy:
## Summary
- Removes the standalone foreman's per-task `TaskContext` supervisor tree (`foreman/pioneer_foreman/task_context.py` — own queue, consumer loop, finalize-detection) along with `Foreman._child_contexts`, `_pending_for_task`, `_route_to_child`, `_spawn_child`, `_deregister_child`, `_parent_tool_observer`, `_respawn_children_from_state`, and `_cancel_all_children`.
- Replaces it with a plain `_task_locks: dict[str, asyncio.Lock]` on `Foreman`, mirroring the embedded foreman's `_guild_locks` pattern in `backend/foreman/runner.py`. A `foreman-trigger` carrying a `taskId` (with `child_contexts` enabled) now runs one ephemeral, lock-serialised `run_foreman_ai(child=True, task_id=...)` call instead of being routed to a long-lived child task.
- Drop-if-busy: if the lock for a task is already held, the incoming trigger is dropped (not queued or buffered) — matching #808 / the embedded foreman's policy. The whole-guild parent's own `_processing` / `_message_queue` drop/queue behavior is unchanged.
- Per-task context scoping (narrowed system prompt / filtered history via `run_foreman_ai(child=True, task_id=...)`) is preserved — only the supervisor/queue/lifecycle bookkeeping was removed.
- Updated `docs/foreman-per-task-context.md` to describe the shared lock-based design used by both foremen, and replaced `foreman/tests/test_task_context.py` / `test_child_routing.py` with `foreman/tests/test_task_lock.py` covering the new lock semantics (drop-on-concurrent, per-task independence, release-on-exception, WS trigger routing with `child_contexts` on/off).

## Test plan
- [x] `pytest` in `foreman/` (81 passed; 3 pre-existing/unrelated errors from a missing `respx` plugin)
- [x] `pytest` in `backend/` (899 passed, 2 skipped)
- [x] `ruff check` on changed files

Closes #807


foreman/pioneer_foreman/foreman.py
```diff
@@ -44,141 +38,55 @@ def __init__(self, config: Config):
...
+        """
+        lock = self._task_locks.setdefault(task_id, asyncio.Lock())
+        if lock.locked():
+            logger.info(
+                "guild=%s task=%s: dropping concurrent invocation (already running)",
+                guild_id,
+                task_id,
+            )
             return
+        await lock.acquire()
```
jmelloy, line 72: The `_task_locks.pop(task_id, None)` after `lock.release()` is racey: between `lock.release()` and the pop, another coroutine could call `_run_task_trigger` for the same `task_id`, find the lock unlocked, acquire it, and then have its lock entry silently removed by this pop. This would allow unbounded concurrent runs for the same task. The pop should happen before `lock.release()`, or the lock should only be popped if it is still unlocked at cleanup time (e.g., `if not lock.locked(): self._task_locks.pop(task_id, None)`).

foreman/pioneer_foreman/foreman.py
```diff
@@ -344,12 +250,15 @@ async def _handle_trigger(data: dict) -> None:
...
 
-                # Route task-specific events to their isolated child context.
-                if self._route_to_child(self._normalize_trigger(data)):
-                    logger.debug(
-                        "Routed trigger to child context: task=%s event=%s",
-                        trigger_task_id,
-                        data.get("event", "?"),
+                # Task-scoped triggers run in an isolated, per-task child context
+                # serialised by a per-task lock rather than the whole-guild queue
+                # below. See docs/foreman-per-task-context.md.
```
jmelloy, line 255: The `asyncio.create_task` here detaches the child run entirely. If `_run_task_trigger` itself raises before the `try` block (e.g., during `setdefault`), the exception would propagate to the task and be silently discarded. Consider adding a task result callback or storing the task reference to enable error surfacing and cancellation on disconnect.

foreman/pioneer_foreman/foreman.py
```diff
@@ -44,141 +38,55 @@ def __init__(self, config: Config):
...
-        # Triggers that arrived for a task between assign_task and the child being
-        # registered; flushed into the child's queue on spawn.
-        self._pending_for_task: dict[str, list[dict]] = {}
+        # Per-task locks serialising per-task child-context runs (task_id -> Lock).
+        # Mirrors backend/foreman/runner.py's _guild_locks: if a task's lock is
+        # already held, the incoming trigger is dropped rather than queued or
+        # buffered — the next lifecycle event or poll tick will retry.  See
+        # docs/foreman-per-task-context.md.
+        self._task_locks: dict[str, asyncio.Lock] = {}
         # Per-connection broadcast closure; set in _run_connection, cleared on exit.
```
jmelloy, line 47: The `_task_locks` dict grows unboundedly for unique task IDs over the foreman's lifetime because the pop in the `finally` block is racey (see other comment). Even if the race is fixed, consider adding a comment or a periodic cleanup path for tasks whose locks are not popped due to the race condition.

jmelloy (commented):
- The PR successfully simplifies the standalone foreman by replacing a persistent `TaskContext` supervisor tree with a plain `asyncio.Lock` per task, aligning it with the embedded foreman's `_guild_locks` pattern.
- The lock cleanup logic in `_run_task_trigger` has a subtle correctness issue: the lock entry is popped from `_task_locks` in the `finally` block, but since the lock is released before the pop, a concurrent waiter could re-acquire and re-insert the lock while the pop is still pending, causing the pop to silently remove an active lock entry.
- The `_run_task_trigger` is fire-and-forget via `asyncio.create_task` in `_handle_trigger`, but `_run_task_trigger` itself swallows exceptions — the error is logged but the `asyncio.Task` wrapping it will complete normally, so unhandled task warnings won't surface if the outer exception handler itself raises.
- Removing `_respawn_children_from_state` means that after a WebSocket reconnect, in-flight tasks that were being tracked by child contexts will no longer receive a reconnect nudge — this is a behavioral regression for long-running tasks that may stall without the reconnect signal.
- Test coverage for the new lock path is thorough and mirrors the embedded foreman's test suite well; the deletion of `test_child_routing.py` and `test_task_context.py` is appropriate given the removed code.

Issue #809, opened 2026-07-07
foreman: move periodic task-health polling from the standalone process into the backend's existing poll dispatcher


jmelloy:
## Context

Part of the audit for #790 (foreman-proxy slim-down). The standalone foreman self-schedules a periodic health check: `foreman/pioneer_foreman/foreman.py`'s `_poll_loop()` (lines ~409-486) sleeps on an exponentially-growing interval (`poll_min_interval`/`poll_max_interval`, config-driven), fetches guild state over REST, decides which tasks look stalled, and fires a synthetic `[periodic-check]` trigger — then reports its own backoff state back to the backend via a `foreman-poll-status` WS message.

The backend already has a complete, independent implementation of this exact feature for the embedded foreman: `backend/foreman/runner.py` has `_poll_loop`, `reset_foreman_poll`, `suppress_foreman_poll`, `resume_foreman_poll` (lines ~380-460), with the same min/max-interval backoff idea. Critically, **the backend currently calls `suppress_foreman_poll(guild_id)` when an external foreman connects** — i.e. it deliberately turns its own scheduler off and lets the external process re-implement the same scheduling from scratch over the network.

The backend also already has a unified dispatcher, `_trigger_foreman()` (`backend/ws_handlers.py:173-230`), that routes any trigger to the external foreman if one is connected, falling back to the embedded runner otherwise. Nothing about that dispatcher is external-foreman-specific — it's already the single source of truth for "who should handle this trigger."

## Work

1. Stop calling `suppress_foreman_poll` / `resume_foreman_poll` when an external foreman (dis)connects — let the backend's poll loop keep running for every guild regardless of who's currently handling triggers.
2. Change the embedded poll loop's tick handler to call `_trigger_foreman()` instead of calling `run_foreman_ai()` directly, so a poll tick transparently goes to whichever foreman (embedded or external) is currently registered.
3. Delete `_poll_loop`, `_poll_interval`, `poll_min_interval`/`poll_max_interval` config fields, and the `foreman-poll-status` broadcast from `foreman/pioneer_foreman/foreman.py` and `foreman/pioneer_foreman/config.py` — the standalone foreman no longer needs to know about polling at all; it just receives `foreman-trigger` messages like any other trigger.
4. Remove the now-unneeded `[poll]` TOML section from `pioneer-foreman.toml.example` and the poll env vars from docs.
[...]

Issue #806, opened 2026-07-07
foreman: delete dead foreman/main.py entrypoint and its backend broadcast-override hook


jmelloy:
## Context

Part of the audit for #790 (foreman-proxy slim-down). `foreman/main.py` is a Phase-1 standalone-foreman prototype that has been fully superseded by the `foreman/pioneer_foreman/` package (Phase 2+), but it was never deleted.

**Evidence it's dead:**
- `pioneer foreman` (the only way the CLI launches the standalone foreman — see `cli/pioneer_cli/main.py:95-106`) imports `pioneer_foreman.cli:main`, never `foreman/main.py`.
- `Dockerfile` / `docker-compose.yml`'s `foreman` service runs `pioneer foreman`, not `python foreman/main.py`.
- `foreman/main.py` imports directly from the backend package (`from foreman.runner import run_foreman_ai`, `from util.tasks import spawn`, `import events as _events`) — i.e. it runs the *embedded* foreman's Python objects in-process rather than proxying over REST like `pioneer_foreman/` does. This is the exact opposite of the "thin network proxy" model and is not wired into anything that runs today.

**Backend-side dead weight this drags along:** `backend/events.py` has a `_broadcast_override` / `set_broadcast_override()` mechanism (lines ~67-95) whose *only* caller is `foreman/main.py:107,153`. Every `broadcast()` call in the backend pays an `asyncio.Lock` acquire to check this override, for a code path nothing exercises anymore.

## Work

1. Delete `foreman/main.py`.
2. Delete `_broadcast_override`, `_broadcast_override_lock`, and `set_broadcast_override()` from `backend/events.py`; simplify `broadcast()` to drop the override check.
3. Update the two stale comments in `backend/ws_handlers.py:1172` and `backend/events.py:67` that reference `foreman/main.py`.
4. Fix `AGENTS.md:205` — it currently documents `foreman/main.py` as "an opt-in external foreman process"; it should point at `foreman/pioneer_foreman/` / `pioneer foreman` instead.

## Why this matters for the proxy-slim-down goal

Two competing standalone-foreman implementations exist in the tree; only one is real. Deleting the dead one removes ~210 lines plus a backend hook, and removes a source of confusion for anyone reading the "standalone foreman" architecture docs.

PR #824, opened 2026-07-07
foreman: add proxy-role banner comments + fix stale docs (#810)

### LINKED ISSUE #810: foreman: add proxy-role banner comments and fix stale docs so the standalone process reads as a thin relay
## Context

Part of the audit for #790 (foreman-proxy slim-down), and intended as the final cleanup pass once the other sub-issues (#806, #807, #808, #809) land. Once the dead entrypoint, child-context supervisor, message queue, and poll loop are removed, `foreman/pioneer_foreman/` should be small enough to read end-to-end in a few minutes, but nothing currently says that out loud — a reader has to infer the proxy nature from the fact that most methods end in an HTTP call.

## Work

1. Add a short module-level docstring banner to `foreman/pioneer_foreman/__init__.py` (currently 1 line, empty) stating explicitly: this process holds no durable state and makes no product decisions; its only job is to run the LLM call and proxy tool execution across a network/firewall boundary the backend can't otherwise cross. Point at `docs/foreman-split-plan.md` for the full protocol.
2. Trim `foreman/pioneer_foreman/foreman.py`'s module docstring (currently "WebSocket connection, trigger handling, poll loop" — "poll loop" will be gone per #809) to describe the post-cleanup shape: connect, register, dispatch triggers 1:1 to `run_foreman_ai`, relay broadcasts.
3. Rename or add inline comments anywhere a name is ambiguous about which side of the network boundary it's on — e.g. clarify in `runner.py` that `ws_send` is "send to backend for fan-out," not "send to the end user," and that `http` is "the only place this process talks to storage."
4. Once #806-#809 land, update `docs/foreman-split-plan.md`'s file listing (lines 36-52) and `docs/foreman-per-task-context.md`'s "Implementation note" to drop references to deleted files (`task_context.py`, poll config) and reflect the lock-based/drop-if-busy model.
5. Sanity-check `AGENTS.md`'s "Standalone Foreman" section reads correctly post-cleanup (also touched by #806 for the `foreman/main.py` reference).

This is the "clarify" pass explicitly called for alongside "delete" and "keep/simplify" — cheap to do, and it's what makes the slimmed-down result actually legible to the next operator who has to read this code under time pressure.

jmelloy:
## Summary
- `foreman/pioneer_foreman/__init__.py`: added a module-level banner stating the standalone foreman holds no durable state and makes no product decisions — its only job is running the LLM call and relaying tool execution across the network/firewall boundary the backend can't otherwise cross. Points to `docs/foreman-split-plan.md`.
- `foreman/pioneer_foreman/foreman.py`: trimmed the module docstring (dropped the now-stale "poll loop" framing removed in #809) to describe the current shape — connect, register, dispatch triggers 1:1 to `run_foreman_ai`, relay broadcasts.
- Added clarifying inline comments at the definition sites of `ws_send` (relays to the backend for fan-out, not to the end user) and `http` (the only path this process has to storage) in `runner.py` and `foreman.py`.
- `docs/foreman-split-plan.md`: added the missing `prompt.py` entry to the file listing and noted `foreman.py` now also owns per-task locks.

Investigated `docs/foreman-per-task-context.md` and `AGENTS.md`'s "Standalone Foreman" section per the issue — both were already updated for the #806–#809 cleanup (no `task_context.py`, poll-config, or `foreman/main.py` references remain), so no changes were needed there.

Closes #810

## Test plan
- [x] `cd foreman && python -m pytest` — 84 passed
- [x] `ruff check` / `ruff format --check` on all modified Python files

### DIFF
--- a/foreman/pioneer_foreman/__init__.py
+++ b/foreman/pioneer_foreman/__init__.py
@@ -4,6 +4,7 @@ This process is a thin proxy: it holds no durable state and makes no product
 decisions of its own. Its only job is to run the LLM call and relay tool
 execution across the network/firewall boundary that the backend can't
 otherwise cross — all state lives in the backend's database, and all tool
-logic lives in the backend. See docs/foreman-split-plan.md for the full
-protocol.
+logic lives in the backend. See
+https://github.com/jmelloy/pioneer-square/blob/main/docs/foreman-split-plan.md
+for the full protocol.
 """
--- a/foreman/pioneer_foreman/__init__.py
+++ b/foreman/pioneer_foreman/__init__.py
@@ -1,1 +1,9 @@
-"""Pioneer Square standalone foreman agent."""
+"""Pioneer Square standalone foreman agent.
+
+This process is a thin proxy: it holds no durable state and makes no product
+decisions of its own. Its only job is to run the LLM call and relay tool
+execution across the network/firewall boundary that the backend can't
+otherwise cross — all state lives in the backend's database, and all tool
+logic lives in the backend. See docs/foreman-split-plan.md for the full
+protocol.
+"""
--- a/foreman/pioneer_foreman/foreman.py
+++ b/foreman/pioneer_foreman/foreman.py
@@ -1,10 +1,13 @@
-"""Standalone Foreman process: WebSocket connection, trigger handling.
+"""Standalone Foreman process: WebSocket connection, trigger dispatch.
 
-Periodic task-health polling is scheduled by the backend (see
-backend/foreman/runner.py's ``_poll_loop``), which dispatches each tick as a
-``foreman-trigger`` to whichever foreman — embedded or external — currently
-owns the guild. This process only reacts to triggers it receives over the WS
-connection; it does not run its own poll timer.
+Connects to the backend over WebSocket, registers as the external foreman for
+one guild, and dispatches each incoming ``foreman-trigger`` 1:1 to
+``run_foreman_ai`` (task-scoped triggers go through a per-task lock instead of
+the whole-guild queue; see docs/foreman-per-task-context.md). Any broadcast
+the run produces is relayed back to the backend as a ``foreman-broadcast``
+message for it to fan out. This process holds no state of its own beyond what
+is needed to serialize concurrent runs — see docs/foreman-split-plan.md for
+the full protocol.
 """
 
 from __future__ import annotations
@@ -33,6 +36,8 @@ class Foreman:
 
     def __init__(self, config: Config):
         self._config = config
+        # REST client back to the backend — the only storage this process can reach;
+        # it never holds its own copy of guild/task state.
         self._http: ForemanHTTPClient | None = None
         # True while a foreman run (including its queue drain) is executing.
         self._processing: bool = False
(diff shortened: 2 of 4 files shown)


foreman/pioneer_foreman/runner.py
```diff
@@ -137,8 +137,10 @@ async def run_foreman_ai(
     user_id: str | None = None,
     task_id: str | None = None,
     *,
-    http: ForemanHTTPClient,
-    ws_send: Callable[[dict], Awaitable[None]],
+    http: ForemanHTTPClient,  # only path to storage; all state lives in the backend's DB
```
jmelloy: Placing the `# only path to storage...` comment inline on the `http` parameter and splitting `ws_send` across lines with a trailing comment is non-standard Python style and will likely be reformatted or mangled by Black/autopep8. Consider moving these explanatory comments to the function docstring instead, where they're both style-safe and more visible to callers.

foreman/pioneer_foreman/foreman.py
```diff
@@ -1,10 +1,13 @@
...
-connection; it does not run its own poll timer.
+Connects to the backend over WebSocket, registers as the external foreman for
+one guild, and dispatches each incoming ``foreman-trigger`` 1:1 to
+``run_foreman_ai`` (task-scoped triggers go through a per-task lock instead of
+the whole-guild queue; see docs/foreman-per-task-context.md). Any broadcast
+the run produces is relayed back to the backend as a ``foreman-broadcast``
+message for it to fan out. This process holds no state of its own beyond what
+is needed to serialize concurrent runs — see docs/foreman-split-plan.md for
+the full protocol.
 """
```
jmelloy, line 11: The new docstring references `docs/foreman-per-task-context.md` — the PR description notes this file was 'investigated' but doesn't confirm it was updated. If the per-task lock behaviour described here is new or changed, that doc should be updated in the same PR to avoid the same staleness problem being fixed here.

foreman/pioneer_foreman/__init__.py
```diff
@@ -1 +1,9 @@
-"""Pioneer Square standalone foreman agent."""
+"""Pioneer Square standalone foreman agent.
+
+This process is a thin proxy: it holds no durable state and makes no product
+decisions of its own. Its only job is to run the LLM call and relay tool
+execution across the network/firewall boundary that the backend can't
+otherwise cross — all state lives in the backend's database, and all tool
+logic lives in the backend. See docs/foreman-split-plan.md for the full
```
jmelloy: The bare relative path `docs/foreman-split-plan.md` works in a monorepo context when browsing from the repo root, but readers viewing the installed package or a PyPI mirror won't be able to resolve it. Consider using a full GitHub URL or noting it's relative to the repo root (e.g. `<repo-root>/docs/foreman-split-plan.md`).

jmelloy (commented):
- Documentation updates are accurate and well-scoped: the `prompt.py` addition and per-task lock note in `foreman-split-plan.md` reflect real code structure.
- The trimmed module docstring in `foreman.py` is cleaner and removes genuinely stale poll-loop framing, though the new docstring references `docs/foreman-per-task-context.md` without confirming that document was reviewed/updated.
- Inline comments on `http` and `ws_send` add useful clarifying context at call-site definition points, though the formatting of the `ws_send` comment in `runner.py` is unconventional and may confuse auto-formatters.
- The `__init__.py` banner is a good architectural guardrail, but the reference to `docs/foreman-split-plan.md` uses a relative path that could break depending on how the docs are served or where the reader is browsing.
- No functional code changes are introduced, so there is no risk of behavioral regression; all changes are documentation and comments only.

PR #834, opened 2026-07-08
Fix assign_task dropping parent_task_id on update path

### LINKED ISSUE #830: fix: assign_task silently drops parent_task_id on the task_id update path (#815)
## Parent issue
Part of #815 — issue-rooted task tree.

## Problem
When `assign_task` is called with an existing `task_id` (the documented `create_task` → `assign_task(task_id=...)` pairing), it performs an UPDATE that silently drops `parent_task_id`. Only the undocumented "bypass" path that creates a new task row persists it. This is a blocking bug — the entire hierarchy feature in #815 depends on `parent_task_id` being reliably stored.

## Fix
In the `assign_task` handler (backend), when performing the UPDATE for an existing task_id, include `parent_task_id` in the UPDATE SET clause so it is persisted correctly.

Also add a test: call `create_task`, then `assign_task(task_id=..., parent_task_id=...)` and assert the stored task has the correct `parent_task_id`.

## Acceptance criteria
- [ ] `assign_task` with `task_id` + `parent_task_id` persists `parent_task_id` on the row
- [ ] Unit test covering the fix

jmelloy:
## Summary
- `assign_task`'s update path (called when `task_id` is provided, e.g. via `create_task` → `assign_task(task_id=...)`) silently dropped `parent_task_id` — it was never included in the SET clause of the UPDATE, unlike the create-new-row path which did persist it.
- Fixed by reading `parent_task_id` from the tool input and including it in `update_values` when present, so it's written to the DB on the update path too.
- Added a regression test that creates a task, then calls `assign_task(task_id=..., parent_task_id=...)` and asserts the stored row has the correct `parent_task_id`.

Closes #830

## Test plan
- [x] `backend/tests/test_foreman.py::TestExecToolsDispatching::test_assign_task_update_path_persists_parent_task_id` passes
- [x] Existing `assign_task` tests still pass

### DIFF
--- a/backend/foreman/tools.py
+++ b/backend/foreman/tools.py
@@ -1241,6 +1241,7 @@ async def _exec_one_tool(guild_id: str, tu, user_id: str | None = None) -> dict:
                                 is_error = True
                             elif existing_task_id:
                                 name_override = inp.get("name")
+                                parent_task_id = inp.get("parent_task_id")
                                 update_values: dict = {
                                     "worker_id": wid,
                                     "description": desc,
@@ -1257,6 +1258,8 @@ async def _exec_one_tool(guild_id: str, tu, user_id: str | None = None) -> dict:
                                     update_values["issue_number"] = inp["issue_number"]
                                 if inp.get("issue_repo"):
                                     update_values["issue_repo"] = inp["issue_repo"]
+                                if parent_task_id is not None:
+                                    update_values["parent_task_id"] = parent_task_id
                                 await db.exec(
                                     update(Task)
                                     .where(
@@ -1282,6 +1285,7 @@ async def _exec_one_tool(guild_id: str, tu, user_id: str | None = None) -> dict:
                                         model=model,
                                         provider=provider,
                                         phase=phase,
+                                        parentTaskId=parent_task_id,
                                         issueNumber=inp.get("issue_number"),
                                         issueRepo=inp.get("issue_repo"),
                                         repos=repos,
--- a/backend/foreman/tools.py
+++ b/backend/foreman/tools.py
@@ -1233,6 +1233,7 @@ async def _exec_one_tool(guild_id: str, tu, user_id: str | None = None) -> dict:
                                 )
                                 .limit(1)
                             )
+                            parent_task_id = inp.get("parent_task_id")
                             if not worker_recheck.one_or_none():
                                 result_text = (
                                     f"Worker {wid} went offline — task NOT assigned. "
@@ -1241,7 +1242,6 @@ async def _exec_one_tool(guild_id: str, tu, user_id: str | None = None) -> dict:
                                 is_error = True
                             elif existing_task_id:
                                 name_override = inp.get("name")
-                                parent_task_id = inp.get("parent_task_id")
                                 update_values: dict = {
                                     "worker_id": wid,
                                     "description": desc,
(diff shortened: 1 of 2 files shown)


backend/foreman/tools.py
```diff
@@ -1282,6 +1285,7 @@ async def _exec_one_tool(guild_id: str, tu, user_id: str | None = None) -> dict:
                                         model=model,
                                         provider=provider,
                                         phase=phase,
+                                        parentTaskId=parent_task_id,
```
jmelloy, line 1288: The `parent_task_id` variable is declared inside the `elif existing_task_id:` branch (line 1244). If execution reaches the create-new-row path without going through the update branch, `parent_task_id` will be undefined here, causing a `NameError`. It should be declared before the `if`/`elif` split, or a separate `inp.get('parent_task_id')` call should be used here.

backend/tests/test_foreman.py
```diff
@@ -481,6 +481,40 @@ async def test_assign_task_with_existing_task_id(self, db_session):
...
+    async def test_assign_task_update_path_persists_parent_task_id(self, db_session):
+        """create_task -> assign_task(task_id=..., parent_task_id=...) must persist
+        parent_task_id on the update path, not just on the create-new-row path (#830)."""
+        insert_guild(db_session, "g-assign-parent")
+        _insert_worker(db_session, "g-assign-parent", "w-wkr3")
+        with patch("foreman.tools.broadcast", new_callable=AsyncMock):
+            create_results = await exec_tools(
+                "g-assign-parent",
+                [_fake_tool_use("create_task", {"name": "Sub Task", "description": "Do part"})],
+            )
```
jmelloy, line 493: Extracting the task ID via `create_results[0]["content"].split()[1]` is fragile and will silently produce the wrong value if the success message format changes. Consider returning the task ID in a structured field or at minimum asserting the expected message format before splitting.

backend/tests/test_foreman.py
```diff
@@ -481,6 +481,40 @@ async def test_assign_task_with_existing_task_id(self, db_session):
         assert "assigned" in results[0]["content"].lower()
         assert "t-exist1" in results[0]["content"]
 
+    async def test_assign_task_update_path_persists_parent_task_id(self, db_session):
```
jmelloy, line 484: `_insert_worker` (with underscore prefix) is used here but `insert_guild` (no underscore) is used on line 483. Verify `_insert_worker` is actually defined in the test module; if it's an alias for `insert_worker` without the underscore, this is a latent `NameError` in the test suite.

jmelloy (commented):
- The core bug fix is correct: `parent_task_id` was silently dropped on the update path because it was never added to `update_values`, and the fix adds it conditionally when present.
- A subtle inconsistency exists: `parentTaskId` is now passed to the create-new-row path (line 1288) using the `parent_task_id` variable that was declared only inside the `elif existing_task_id` branch, which could cause a `NameError` if the create path is reached without first entering the update branch.
- The test correctly exercises the regression scenario but relies on fragile string parsing (`create_results[0]["content"].split()[1]`) to extract the task ID, and doesn't assert the pre-condition (that `parent_task_id` was `None` before the update call).
- The test references `_insert_worker` with a leading underscore but inserts with `insert_guild` (no underscore) — inconsistency with helper naming suggests a possible copy-paste issue that may cause a `NameError` if `_insert_worker` is not defined.
- No handling for the case where `parent_task_id` is explicitly passed as `None` to clear an existing parent — the `if parent_task_id is not None` guard prevents unsetting the field, which may or may not be intentional.

PR #837, opened 2026-07-08
foreman: graceful soft-kill for spawned workers

### LINKED ISSUE #835: foreman: use graceful shutdown (soft kill) when killing a spawned worker
## Problem

When the foreman kills a spawned worker it force-kills the process immediately, which abruptly terminates any in-progress agent task on that worker.

## Expected behaviour

The foreman should send a graceful/soft shutdown signal (e.g. SIGTERM, or call the worker's `/shutdown` endpoint) so the worker can finish its current task before exiting. Only escalate to a force kill if the worker does not exit within a reasonable timeout.

## Reported by

@jmelloy via Discord.

jmelloy:
Implements graceful shutdown for spawned worker containers before escalating to force-kill. Adds a configurable timeout so the container has a chance to exit cleanly. Closes #835.

### DIFF
--- a/backend/tests/test_worker_lifecycle.py
+++ b/backend/tests/test_worker_lifecycle.py
@@ -603,7 +603,7 @@ class _FakeStateDB:
 
 
 def test_shutdown_force_kill_timeout_default():
-    assert SHUTDOWN_FORCE_KILL_TIMEOUT == 30.0
+    assert SHUTDOWN_FORCE_KILL_TIMEOUT == 600.0
 
 
 def test_shutdown_force_kill_timeout_env_override(monkeypatch):
--- a/backend/tests/test_worker_lifecycle.py
+++ b/backend/tests/test_worker_lifecycle.py
@@ -14,8 +14,10 @@ import pytest
 sys.path.insert(0, os.path.join(os.path.dirname(__file__), ".."))
 
 from worker_lifecycle import (  # noqa: E402
+    SHUTDOWN_FORCE_KILL_TIMEOUT,
     WORKER_DRAIN_TIMEOUT,
     drain_stale_workers_on_startup,
+    force_kill_worker_if_unresponsive,
     generate_worker_id,
     get_current_version,
     reconcile_stale_workers,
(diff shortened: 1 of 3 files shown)


backend/worker_lifecycle.py
```diff
@@ -267,6 +273,43 @@ async def _force_kill_containers(worker_ids: set[str]) -> None:
...
+    WorkerShutdownMsg, so an in-progress task on that worker gets a chance to finish
+    before anything is force-killed. Only escalates to a hard Docker kill (last
+    resort) if the worker never reports going offline within *timeout* seconds —
+    e.g. because it's wedged and never received/processed the shutdown signal.
+    """
+    poll_interval = 5.0
+    elapsed = 0.0
+    while elapsed < timeout:
+        await asyncio.sleep(min(poll_interval, timeout - elapsed))
+        elapsed += poll_interval
```
jmelloy: Elapsed-time tracking is incorrect: `elapsed += poll_interval` always adds the full interval even though the final `asyncio.sleep` call may sleep for `timeout - elapsed` (a shorter duration). This means the loop can terminate one full `poll_interval` early when `timeout` is not a multiple of `poll_interval`. Fix by tracking real elapsed time with `asyncio.get_event_loop().time()` or by adding only the actual sleep duration.

backend/worker_lifecycle.py
```diff
@@ -267,6 +273,43 @@ async def _force_kill_containers(worker_ids: set[str]) -> None:
...
+        if state is None or state == "offline":
+            logger.info(
+                "worker_lifecycle: worker %s shut down gracefully within %.0fs",
+                worker_id,
+                elapsed,
+            )
+            return
+
+    logger.warning(
+        "worker_lifecycle: worker %s did not shut down within %.0fs of the graceful "
```
jmelloy, line 321: There is a race between the loop's last DB poll and the `_force_kill_containers` call: a worker that goes offline after the loop exits (within the last poll window) will still be force-killed. Consider doing one final state check immediately before escalating, or at least add a comment acknowledging this window.

backend/tests/test_worker_lifecycle.py
```diff
@@ -570,3 +572,87 @@ async def test_reconcile_only_respawns_workers_still_missing():
...
+    """Fake session whose exec().one_or_none() returns states from a queue, in order."""
+
+    def __init__(self, states):
+        self._states = iter(states)
+        self.exec = AsyncMock(
+            side_effect=lambda *a, **k: MagicMock(one_or_none=lambda: next(self._states))
+        )
+
+    async def __aenter__(self):
+        return self
```
jmelloy, line 599: `_FakeStateDB.exec`'s `side_effect` calls `next(self._states)` inside the lambda passed to `MagicMock(one_or_none=...)`, which is evaluated at mock-construction time for each `exec` call but returns the *same* `MagicMock` with a fixed `one_or_none` for repeated polls. Multi-iteration scenarios (e.g. worker online for two polls then offline) cannot be correctly represented with this helper, limiting confidence in the timeout-escalation path.

jmelloy (commented):
- The graceful shutdown logic is well-structured with a configurable timeout and clear separation between the new `SHUTDOWN_FORCE_KILL_TIMEOUT` and the existing `WORKER_DRAIN_TIMEOUT`, with good inline comments explaining the distinction.
- A subtle elapsed-time accounting bug exists in the polling loop: `elapsed += poll_interval` always adds the full `poll_interval` even when `asyncio.sleep(min(poll_interval, timeout - elapsed))` slept for a shorter duration on the final iteration, causing the loop to exit one iteration early in some cases.
- The late/deferred import of `worker_lifecycle` inside `_exec_one_tool` avoids a circular-import problem but this pattern should be documented or, better, resolved structurally; the `# noqa: PLC0415` suppression silently hides the issue for future maintainers.
- The `_FakeStateDB` test helper's `exec` side_effect calls `next(self._states)` at mock-construction time rather than at call time for each poll, meaning tests with multiple poll cycles (e.g. a worker that's online for several polls then goes offline) cannot be correctly simulated, limiting test coverage of multi-iteration behaviour.
- No mechanism re-checks the worker state immediately before calling `_force_kill_containers`; a worker that goes offline in the last poll interval after the loop exits will still be force-killed, which is a minor but real race condition worth noting in a comment.

backend/worker_lifecycle.py
```diff
@@ -41,6 +41,12 @@
 # How long to wait for stale workers to exit gracefully before force-killing.
 WORKER_DRAIN_TIMEOUT = float(os.environ.get("PIONEER_WORKER_DRAIN_TIMEOUT", "60"))
 
+# How long to wait after an explicit shutdown_worker request before force-killing
+# an unresponsive worker's container. Separate from WORKER_DRAIN_TIMEOUT (which
+# guards the much rarer startup version-mismatch drain) since this is the window
+# the foreman AI waits on every ordinary "stop this worker" request.
+SHUTDOWN_FORCE_KILL_TIMEOUT = float(os.environ.get("PIONEER_SHUTDOWN_TIMEOUT", "30"))
```
jmelloy: This could be way higher, like 10 minutes

PR #839, opened 2026-07-08
feat: add tool/model/provider params to send_followup

### LINKED ISSUE #838: feat: send_followup should support tool selection
## Summary

`send_followup` currently does not support specifying which tool (claude, codex, pi) the worker should use when resuming work on a branch. This means follow-up tasks always use the worker's default tool, even when a different tool would be more appropriate.

## Desired behaviour

`send_followup` should accept an optional `tool` parameter (and optionally `model` / `provider`) so the Foreman (or operator) can explicitly choose the coding agent for a follow-up task, consistent with how `assign_task` works today.

## Acceptance criteria

- `send_followup` accepts an optional `tool` field (`claude`, `codex`, `pi`)
- When `tool` is provided, the follow-up task runs with that tool instead of the worker default
- When `tool` is omitted, behaviour is unchanged (backward-compatible)
- The Foreman MCP tool schema is updated to expose the new parameter

jmelloy:
## Summary
- Adds optional `tool` (enum: claude/codex/pi), `model`, and `provider` params to the `send_followup` MCP tool schema, mirroring `assign_task`.
- Handler now validates a requested `tool` against the target worker's supported tools, and a requested `model` against the model catalog for the worker's provider, before dispatching. Omitting all three preserves the task's existing tool/model/provider (fully backward-compatible).
- When `tool` changes without an explicit `model`, the stale model is dropped since it belongs to the previous tool.
- Updated `Task.tool/model/provider` persistence and the `TaskFollowupMsg` broadcast to carry the effective values. The worker's `task-followup` handler already reads `tool`/`model`/`provider` off that message, so no worker-side changes were needed.
- Updated the foreman system-prompt tool description for `send_followup` to document the new parameters.
- Added tests covering: preserving values when omitted, switching tools, rejecting unsupported tool overrides, rejecting out-of-catalog models, and persisting provider overrides.

Closes #838.

## Test plan
- [x] `cd backend && python -m pytest` — 913 passed, 2 skipped
- [x] `cd worker && python -m pytest` — 259 passed
- [x] `ruff check` on changed files


backend/foreman/tools.py
```diff
@@ -1397,111 +1400,173 @@ async def _exec_one_tool(guild_id: str, tu, user_id: str | None = None) -> dict:
...
+                        ):
+                            available = ", ".join(followup_worker_tools)
                             result_text = (
-                                f"Task {task_id} is locked by an in-progress follow-up. "
-                                "Instructions have been queued and will be passed to the "
-                                "foreman when the current follow-up completes."
+                                f"Worker {target_worker_id} does not support tool "
+                                f"{requested_tool!r}. Available tools: {available}"
                             )
+                            is_error = True
```
jmelloy, line 1432: Dead code / misleading: `effective_tool`, `effective_model`, and `effective_provider` are assigned inside the `is_error = True` branch for an unsupported tool, but these values are never used — execution jumps past the `if not is_error:` block. These three assignments should be removed to avoid confusion.

backend/foreman/tools.py
```diff
@@ -1397,111 +1400,173 @@ async def _exec_one_tool(guild_id: str, tu, user_id: str | None = None) -> dict:
...
-                                followup_worker_tools: list[str] = json.loads(
-                                    followup_worker_tools_json or "[]"
+                            if effective_model and followup_worker_provider:
+                                from models import ModelCatalog  # noqa: PLC0415
+
+                                catalog_check = await db.exec(
+                                    select(col(ModelCatalog.model_id)).where(
+                                        col(ModelCatalog.provider) == followup_worker_provider,
+                                        col(ModelCatalog.model_id) == effective_model,
+                                    )
```
jmelloy, line 1462: Model validation uses `followup_worker_provider` (the worker's stored provider) rather than `effective_provider` (which may have been overridden by the caller). When a caller passes both `provider='openai'` and `model='some-claude-model'`, the catalog check runs against the worker's original provider instead of the requested one, potentially accepting an invalid combination.

backend/foreman/tools.py
```diff
@@ -1397,111 +1400,173 @@ async def _exec_one_tool(guild_id: str, tu, user_id: str | None = None) -> dict:
...
-                                    notify_discord_followup(
-                                        task_issue_repo,
-                                        task_issue_number,
-                                        task_id,
-                                        instructions,
-                                    ),
-                                    name=f"discord.followup:{task_id}",
+                            lock_acquired = await LockService(db).acquire(
+                                f"task:{task_id}", owner=lock_id
+                            )
```
jmelloy, line 1480: The `pending-followup` TaskEvent payload now includes `tool`, `model`, and `provider`, but if the replay handler that processes this event doesn't read those keys, the overrides will be silently dropped when the lock is released. Please verify (and test) that the replay path forwards these fields correctly.

backend/foreman/tools_schema.py
```diff
@@ -160,6 +163,33 @@
                         "task's current worker_id."
                     ),
                 },
+                "tool": {
+                    "type": "string",
+                    "enum": ["claude", "codex", "pi"],
```
jmelloy, line 168: The `tool` enum is hardcoded as `["claude", "codex", "pi"]` in the schema. If new tools are added to the system this schema will be stale and silently reject valid values before the handler even runs. Consider deriving this list from a shared constant rather than duplicating it here.

jmelloy (commented):
- The PR successfully adds `tool`, `model`, and `provider` params to `send_followup`, mirroring `assign_task`, with backward-compatible defaults.
- A logic bug exists in the error path: `effective_tool/model/provider` are set even when `is_error = True` for the unsupported-tool case, but these assignments are unreachable dead code and misleading since the block falls through to `if not is_error` correctly.
- The model validation uses the *worker's* provider (`followup_worker_provider`) rather than `effective_provider`, which could allow a mismatched provider/model combination when the caller overrides `provider` but not `model`.
- The `pending-followup` queue now persists `tool/model/provider` in the event payload, but there is no evidence the replay handler reads these new fields, meaning queued follow-ups with tool/model/provider overrides would silently lose them on replay.
- Test coverage is good for the happy paths and key error cases; the missing test for a queued (locked) follow-up with tool/model overrides is a gap given the payload change.

PR #841, opened 2026-07-08
feat: Discord thread routing — all child tasks post to issue-root thread

### LINKED ISSUE #832: feat: Discord thread routing — all child tasks post to issue-root thread (#815)
## Parent issue
Part of #815 — issue-rooted task tree.

## Problem
Currently each task (plan, execute, review, followup) gets its own Discord thread. With the new issue-root model, all tasks under the same GH issue should post into a single, stable Discord thread that lives for the lifetime of the issue.

## Changes required
- When a `phase='issue'` root task is created, create the Discord thread immediately and store the `discord_thread_id` on the root task row.
- When any child task (plan/execute/review/followup) sends a Discord message, resolve the thread by walking up to the root task via `parent_task_id` chain and posting to that thread instead of creating a new one.
- If no root task exists (legacy tasks), fall back to current behavior.

## Acceptance criteria
- [ ] All tasks parented to the same issue root post into the same Discord thread
- [ ] The thread is created when the root task is minted, not when the first child posts
- [ ] Legacy tasks (no root) continue to work as before

jmelloy:
Closes #832

### DIFF
--- a/backend/discord_notifier.py
+++ b/backend/discord_notifier.py
@@ -47,6 +47,19 @@ Phase 5 — new-member welcome DM
     ``DISCORD_WELCOME_DM_TEXT``. DM failures (e.g. DMs disabled) are logged at
     WARNING and never raised.
 
+Phase 6 — issue-rooted task tree thread routing
+    Every thread-aware entry point (``notify_event``, ``notify_existing_thread``,
+    ``notify_foreman_chat``) accepts a *task_id* and, before anything else,
+    calls ``_resolve_root_thread_id`` to walk ``parent_task_id`` up to that
+    task's root (the phase='issue' task created via ``create_task``). When the
+    root has a ``discord_thread_id`` — set once, immediately, by
+    ``ensure_issue_root_thread`` when the root task is minted — the
+    notification posts there, so every plan/execute/review/followup task
+    parented to the same issue lands in one stable thread for the issue's
+    whole lifetime. Falls back to the legacy ``issue_repo``/``issue_number``
+    routing below when *task_id* has no root, or the root predates this
+    column and has no thread yet.
+
 Required bot permissions: ``SEND_MESSAGES``, ``CREATE_PUBLIC_THREADS``, ``MANAGE_THREADS``.
 
 Usage::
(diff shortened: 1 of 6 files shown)


backend/discord_notifier.py
```diff
@@ -376,6 +389,101 @@ async def _ensure_thread(
...
+        from models import Task  # noqa: PLC0415
+        from sqlalchemy import update  # noqa: PLC0415
+        from sqlmodel import col  # noqa: PLC0415
+
+        async with AsyncSessionLocal() as db:
+            await db.exec(
+                update(Task).where(col(Task.id) == task_id).values(discord_thread_id=thread_id)
+            )
+            await db.commit()
+    except Exception:
```
jmelloy: `db.exec()` is intended for SQLModel `select()` statements; using it with a SQLAlchemy core `update()` statement is not officially supported and may silently no-op on some SQLAlchemy/SQLModel versions. Use `await db.execute(...)` here instead.

backend/discord_notifier.py
```diff
@@ -376,6 +389,101 @@ async def _ensure_thread(
...
+        from database import AsyncSessionLocal  # noqa: PLC0415
+        from models import Task  # noqa: PLC0415
+        from sqlmodel import col, select  # noqa: PLC0415
+
+        async with AsyncSessionLocal() as db:
+            current_id = task_id
+            for _ in range(_MAX_PARENT_CHAIN_DEPTH):
+                result = await db.exec(
+                    select(Task.parent_task_id, Task.discord_thread_id).where(
+                        col(Task.id) == current_id
```
jmelloy: Each iteration issues a separate `SELECT` — up to 20 round-trips for a deeply nested task. A single recursive CTE (`WITH RECURSIVE`) would collapse this to one query. At minimum, consider selecting the full chain in one query with a depth-limited self-join to avoid the N+1 pattern.

backend/routes/foreman.py
```diff
@@ -456,6 +458,16 @@ async def create_foreman_task(
             createdAt=created_at.isoformat(),
         ),
     )
+
+    if body.phase == "issue" and discord_notifier.is_configured():
+        # Create the issue's stable Discord thread now, at root-task mint
+        # time, rather than lazily on the first child post — see
```
jmelloy, line 464: `ensure_issue_root_thread` is also spawned from `foreman/tools.py` `_exec_one_tool` for `create_task`. If both code paths are reachable for the same root task (e.g. the REST endpoint and the tool path both handle task creation), a race will create two Discord threads but `_save_root_thread` will only persist whichever wins, silently leaking the other thread. Add a read-before-create guard inside `ensure_issue_root_thread` to skip creation when `discord_thread_id` is already set.

backend/foreman/tools.py
```diff
@@ -1270,9 +1302,17 @@ async def _exec_one_tool(guild_id: str, tu, user_id: str | None = None) -> dict:
...
                                 await db.commit()
                                 name_result = await db.exec(
-                                    select(col(Task.name)).where(col(Task.id) == existing_task_id)
+                                    select(col(Task.name), col(Task.parent_task_id)).where(
+                                        col(Task.id) == existing_task_id
+                                    )
                                 )
-                                task_name = name_result.one_or_none() or desc[:60]
+                                name_row = name_result.one_or_none()
+                                task_name = (name_row[0] if name_row else None) or desc[:60]
```
jmelloy, line 1314: `effective_parent_task_id` is taken from the DB row fetched before the `update(Task).where(...).values(parent_task_id=...)` is executed — the re-select comment says the opposite. The `SELECT` runs before the `UPDATE` is committed, so `effective_parent_task_id` reflects the pre-update value. This is actually correct for the stated intent (use whatever parent was set by an earlier `create_task` call), but the comment "re-select the persisted value" is misleading and should be clarified.

jmelloy (commented):
- The core design is sound: anchoring the Discord thread to the phase='issue' root task at creation time and walking `parent_task_id` upward is a clean, stable routing strategy.
- `_resolve_root_thread_id` performs one DB round-trip per ancestor level (up to 20), which could be replaced by a single recursive CTE query for better efficiency, though the depth cap prevents runaway queries.
- `ensure_issue_root_thread` is called from two independent code paths (`routes/foreman.py` and `foreman/tools.py`) with no deduplication guard — if both paths execute for the same task_id there is a race that will create two Discord threads but only persist one, silently losing the other.
- The `_save_root_thread` function uses `db.exec(update(...))` with SQLModel's async session; `exec` is not the correct method for SQLAlchemy core `update()` statements — `db.execute()` should be used instead, which could cause silent failures on some SQLModel/SQLAlchemy version combinations.
- Migration dependency chain assumes `20260708_000000_add_issue_phase_and_index` exists; if that migration was renamed or is missing in a deployment the upgrade will fail at import time with a misleading error.

PR #846, opened 2026-07-08
feat: Discord debounce + combine PR messages, low notif priority (#843)

### LINKED ISSUE #843: Discord: debounce + combine PR messages, low notification priority for CI
## Problem

Discord PR event messages (especially CI success/status ones) are noisy — they fire rapidly and individually, pinging users repeatedly with low-value notifications.

## Requested changes

1. **Debounce PR event messages**: When multiple events arrive for the same PR in a short window (e.g. 10–30 seconds), batch them into a single message rather than sending each one immediately.
2. **Combine related messages**: Where possible, collapse consecutive CI check updates for the same PR into one summary message (e.g. "3 checks passed").
3. **Low notification priority for CI messages**: Send CI success/status messages using Discord's `silent` flag (`@silent` prefix or message flag `SUPPRESS_NOTIFICATIONS = 1 << 12`) so they don't trigger push notifications for users.

## Acceptance criteria

- A PR with multiple rapid CI check completions produces one combined Discord message, not N individual ones.
- CI success messages are sent with suppressed notifications (no ping/buzz on mobile/desktop).
- Non-CI events (e.g. review submitted, PR merged) retain normal notification priority.
- Debounce window is configurable (env var or constant), defaulting to ~15 seconds.

jmelloy:
## Summary
- Buffer per-PR `check_run`/`check_suite` CI completions and flush a single combined Discord message (e.g. "✅ 3 checks passed: lint, test, build") after a debounce window, instead of one message per check.
- Debounce window configurable via `DISCORD_PR_DEBOUNCE_SECONDS` (default 15s).
- Combined CI messages carry Discord's `SUPPRESS_NOTIFICATIONS` flag (`1 << 12`) so they don't trigger push notifications.
- Non-CI events (PR opened/merged/closed, reviews) are unaffected — normal priority, immediate posting.

Closes #843

## Test plan
- [x] `pytest backend/tests/test_discord_notifier.py` — 95 passed
- [x] `ruff check` on changed files — clean

### DIFF
--- a/backend/discord_notifier.py
+++ b/backend/discord_notifier.py
@@ -60,6 +60,17 @@ Phase 6 — issue-rooted task tree thread routing
     routing below when *task_id* has no root, or the root predates this
     column and has no thread yet.
 
+Phase 7 — CI check debounce/combine
+    ``notify_ci_check`` buffers ``check_run``/``check_suite`` completions per
+    PR and flushes a single combined summary (e.g. ``"✅ 3 checks passed:
+    lint, test, build"``) after ``DISCORD_PR_DEBOUNCE_SECONDS`` (default 15)
+    of silence on that PR, instead of one Discord message per check — a CI
+    matrix routinely fires several of these within a few seconds of each
+    other. The combined message always carries ``SUPPRESS_NOTIFICATIONS``
+    (see ``_SILENT_FLAG``) so CI results don't page anyone; non-CI events
+    (PR opened/merged/closed, reviews) are unaffected and keep posting
+    immediately at normal priority.
+
 Required bot permissions: ``SEND_MESSAGES``, ``CREATE_PUBLIC_THREADS``, ``MANAGE_THREADS``.
 
 Usage::
@@ -164,6 +175,7 @@ async def notify(
     url: str | None = None,
     color: int | None = None,
     ps_guild_slug: str | None = None,
+    flags: int | None = None,
 ) -> None:
     """Post a Discord embed to the bot's configured channel.
 
@@ -180,7 +192,9 @@ async def notify(
     if not channel:
         return
 
-    await _post_to_thread(channel, event_type, title, description, url=url, color=color)
+    await _post_to_thread(
+        channel, event_type, title, description, url=url, color=color, flags=flags
+    )
 
 
 # ---------------------------------------------------------------------------
@@ -524,15 +538,24 @@ async def _post_to_thread(
     url: str | None = None,
     color: int | None = None,
     fields: dict[str, str] | None = None,
+    flags: int | None = None,
 ) -> None:
-    """Post an embed to an existing Discord thread. Never raises."""
+    """Post an embed to an existing Discord thread. Never raises.
+
+    *flags* is the raw Discord message-flags bitmask (see ``_SILENT_FLAG``) —
+    passed through so callers like ``notify_ci_check`` can mark a message
+    silent without duplicating the request-building logic here.
+    """
     resolved_color = color if color is not None else _COLOURS.get(event_type, _DEFAULT_COLOUR)
     embed: dict = {"title": title, "description": description, "color": resolved_color}
     if url:
         embed["url"] = url
     if fields:
         embed["fields"] = [{"name": k, "value": v, "inline": True} for k, v in fields.items()]
-    await _bot_request("post", f"/channels/{thread_id}/messages", {"embeds": [embed]})
+    payload: dict = {"embeds": [embed]}
+    if flags:
+        payload["flags"] = flags
+    await _bot_request("post", f"/channels/{thread_id}/messages", payload)
 
 
 async def archive_thread(thread_id: str) -> None:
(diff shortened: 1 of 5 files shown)


backend/discord_notifier.py
```diff
@@ -1222,3 +1262,150 @@ async def flush_task_stream(task_id: str) -> None:
...
+    ps_guild_slug: str | None = None,
+    linked_issue_repo: str | None = None,
+    linked_issue_number: int | None = None,
+    task_id: str | None = None,
+) -> None:
+    """Buffer one CI check completion for *pr_label* and (re)arm a debounced flush.
+
+    Multiple ``check_run``/``check_suite`` completions for the same PR
+    arriving within ``DISCORD_PR_DEBOUNCE_SECONDS`` of each other (a CI
+    matrix routinely fires several within a few seconds) are combined into a
```
jmelloy, line 1412: There is no guard preventing a race between `_flush_ci_batch` (which calls `_ci_batches.pop(key)`) and a concurrent `notify_ci_check` that checks `_ci_batches.get(key)` in the same event loop tick. If `_flush_ci_batch` pops the batch while `notify_ci_check` is about to append to it, the new entry will create a fresh batch with no flush task armed, silently dropping that check from any future summary. Consider re-checking and re-arming inside `_flush_ci_batch` or using a lock.

backend/discord_notifier.py
```diff
@@ -1222,3 +1262,150 @@ async def flush_task_stream(task_id: str) -> None:
...
+    window — matching the existing task-stream buffer's batching behaviour
+    (see ``notify_task_stream``) rather than a sliding debounce, so a steady
+    trickle of checks can't indefinitely postpone the summary. Silent no-op
+    when the bot token is not configured. Never raises.
+    """
+    if not _bot_token():
+        return
+
+    key = f"{issue_repo}#{issue_number}" if issue_repo and issue_number is not None else pr_label
+    batch = _ci_batches.get(key)
```
jmelloy, line 1437: `batch.flush_task` is assigned via `asyncio.ensure_future` but never stored back or cancelled on a successful early flush via `_flush_ci_batch`. If `_flush_ci_batch` is called directly (e.g. in tests or a future force-flush path) while the delayed task is still pending, `_delayed_ci_flush` will call `_flush_ci_batch` again for an already-popped key — this is a silent no-op today but fragile. Cancel the task in `_flush_ci_batch` before popping.

backend/discord_notifier.py
```diff
@@ -1222,3 +1262,150 @@ async def flush_task_stream(task_id: str) -> None:
...
+
+
+async def _flush_ci_batch(key: str) -> None:
+    """Post the combined summary for *key*'s buffered CI checks, then drop the batch."""
+    batch = _ci_batches.pop(key, None)
+    if batch is None or not batch.entries:
+        return
+    conclusions = {entry.conclusion for entry in batch.entries}
+    event_type = "ci-pass" if conclusions == {"success"} else "ci-fail"
+    description = _format_ci_summary(batch.entries)
```
jmelloy, line 1382: The debounce window starts on the *first* buffered check and is never reset by subsequent checks (intentional, per the docstring). However this means for slow CI matrices where checks arrive at intervals approaching `DISCORD_PR_DEBOUNCE_SECONDS`, early checks flush without the later ones. The docs and `.env.example` should note this is a fixed-start window, not a sliding one, so operators can tune the value appropriately.

backend/discord_notifier.py
```diff
@@ -1222,3 +1262,150 @@ async def flush_task_stream(task_id: str) -> None:
...
+    kwargs: dict
+    entries: list[_CICheckEntry] = field(default_factory=list)
+    flush_task: asyncio.Task | None = None
+
+
+# Keyed by "{issue_repo}#{issue_number}" (or the PR label when repo/number
+# aren't available). Holds only PRs with un-flushed CI entries.
+_ci_batches: dict[str, _CIBatch] = {}
+
+
```
jmelloy, line 1332: `_format_ci_summary` iterates `by_conclusion.items()` whose insertion order depends on the order entries were appended. If two PRs process checks in different orders, the summary lines appear in different orders. This is minor cosmetically, but consider sorting by a priority order (failures first) for consistently useful output: failures most prominent, then warnings, then successes.

jmelloy (commented):
- The debounce implementation uses a fixed-start timer (armed on first check) rather than a sliding window, which is intentional and documented, but means a slow CI matrix with checks arriving at t=0, t=14s, t=14s could flush before the last checks arrive — this edge case is undocumented in user-facing docs.
- `_SILENT_FLAG` is referenced in `_flush_ci_batch` and docstrings but is never defined in the diff; it must exist elsewhere in the file, though this is a risk if it were ever removed or renamed.
- The `flush_task` field on `_CIBatch` is set in `notify_ci_check` but never cancelled or awaited on cleanup — if the event loop shuts down mid-debounce, the task leaks silently without flushing its buffered entries.
- `_ci_batches` is a module-level mutable dict shared across all async tasks with no locking; while Python's GIL protects simple dict operations, coroutine interleaving between `notify_ci_check` and `_flush_ci_batch` (the `pop` vs `get` race) could cause a batch to be double-flushed or silently dropped in edge cases.
- Test coverage is solid (219 lines, 14 new tests) covering the happy path, mixed conclusions, flag presence, and no-op cases; the missing scenario is concurrent calls to `notify_ci_check` and `_flush_ci_batch` for the same key.

PR #848, opened 2026-07-08
Auto-finalize phase='issue' root tasks when GitHub issue closes

### LINKED ISSUE #847: feat: close phase='issue' root task when GitHub issue is closed
## Problem

When a GitHub issue is closed (merged, resolved, or manually closed), its corresponding `phase='issue'` root task remains open indefinitely. There's no mechanism to auto-close it.

## Proposed solution

Two complementary approaches:

### 1. Periodic sweep (primary)
In the periodic-check handler, query for all non-terminal `phase='issue'` tasks that have an `issue_repo` and `issue_number` set. For each, call the GitHub API to check the issue state. If the GH issue is `closed`, call `finalize_task` on the root task (and optionally any non-terminal children).

### 2. Webhook (if available)
If GitHub `issues` webhook events (`closed` action) are already being delivered to the backend, handle the `issues/closed` event to immediately finalize the matching root task. This is a nice-to-have; the periodic sweep covers the gap if the webhook isn't wired up.

## Implementation notes

- The sweep should be lightweight: only check tasks where `phase='issue'` and `state` is not terminal (`done`/`failed`/`cancelled`).
- Use the existing `(issue_repo, issue_number)` index added in #836 for fast lookup.
- `finalize_task` on the root task should cascade or at least log if child tasks are still open.
- The periodic check already runs a PR status sweep — this follows the same pattern.

jmelloy:
## Summary
- Add a periodic sweep (`foreman.runner._sweep_closed_issue_tasks`) that checks every non-terminal `phase='issue'` root task with `issue_repo`/`issue_number` set against the GitHub issue's current state, finalizing (state=done, deleted_at=+3d) any whose issue has closed.
- Add a shared `finalize_issue_root_task` helper in `foreman/tools.py` used by both the sweep and the webhook path, using a single conditional UPDATE to avoid a TOCTOU race with concurrent finalization.
- Add an `issues` webhook handler (`action=closed`) in `routes/webhooks.py` that finalizes the matching issue-root task immediately, with the periodic sweep as a safety net for missed deliveries.
- Log a warning (never auto-close) if a finalized issue-root task still has non-terminal child tasks.

Closes #847.

## Test plan
- [x] `python -m pytest tests/` — 932 passed, 2 skipped, 0 failed
- [x] Reviewed diff vs `main` to confirm sweep, webhook handler, and shared finalize/child-warning helpers are wired consistently with the existing PR-status sweep pattern

### DIFF
--- a/backend/foreman/runner.py
+++ b/backend/foreman/runner.py
@@ -479,6 +479,70 @@ async def _fetch_pr_status_lines(guild_id: str, pr_tasks: list[dict]) -> list[st
     return lines
 
 
+async def _sweep_closed_issue_tasks(guild_id: str, issue_tasks: list[dict]) -> None:
+    """Finalize ``phase='issue'`` root tasks whose linked GitHub issue has closed.
+
+    Called on every periodic-check cycle alongside the PR-status refresh above, for
+    every non-terminal ``phase='issue'`` root task that has both ``issue_repo`` and
+    ``issue_number`` set (backed by the ``ix_tasks_issue_repo_issue_number`` index
+    added in #836). This is a safety net for the ``issues`` webhook handler in
+    ``routes.webhooks`` — it catches issues closed before that handler existed, or
+    where a webhook delivery was missed — so it acts directly (no foreman/LLM
+    round-trip needed) rather than merely nudging the foreman like the PR-status
+    block does. Child tasks (plan/execute/review/followup) are never auto-closed;
+    a warning is logged if any remain non-terminal. Best effort: an individual
+    GitHub fetch failure is logged and skipped rather than aborting the sweep.
+    """
+    if not issue_tasks:
+        return
+
+    from foreman.tools import (
+        _guild_github_token,
+        fetch_issue_state,
+        finalize_issue_root_task,
+        warn_if_issue_task_has_open_children,
+    )
+
+    creds = await _guild_github_token(guild_id)
+    if not creds:
+        return
+    token, _username = creds
+
+    for t in issue_tasks:
+        repo, number = t["issue_repo"], t["issue_number"]
+        try:
+            state = await fetch_issue_state(repo, number, token)
+        except Exception:
+            logger.warning(
+                "guild=%s task=%s failed to fetch issue status for %s#%s",
+                guild_id,
+                t["id"],
+                repo,
+                number,
+                exc_info=True,
+            )
+            continue
+
+        if state != "closed":
+            continue
+
+        db = await get_db()
+        try:
+            guild_pk_val = await get_guild_pk(db, guild_id)
+            finalized = await finalize_issue_root_task(db, guild_pk_val, guild_id, t["id"])
+            if finalized:
+                logger.info(
+                    "guild=%s task=%s finalized (issue %s#%s closed)",
+                    guild_id,
+                    t["id"],
+                    repo,
+                    number,
+                )
+                await warn_if_issue_task_has_open_children(db, guild_id, t["id"])
+        finally:
+            await db.close()
+
+
 async def _poll_loop(guild_id: str) -> None:
     """Background loop: check non-terminal tasks and trigger the foreman if any are found.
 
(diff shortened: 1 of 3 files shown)


backend/foreman/tools.py
```diff
@@ -161,6 +163,76 @@ def _resolve_finalize_deleted_at(inp: dict) -> tuple[datetime | None, str | None
...
+async def finalize_issue_root_task(db, guild_pk: int, guild_id: str, task_id: str) -> bool:
+    """Directly finalize a ``phase='issue'`` root task once its GitHub issue closes.
+
+    Shared by the periodic closed-issue sweep (``foreman.runner._sweep_closed_issue_tasks``)
+    and the ``issues`` webhook handler (``routes.webhooks.github_webhook``) so both paths
+    apply the same terminal-state guard and broadcast behaviour. Uses a single conditional
+    UPDATE (rather than SELECT-then-UPDATE) to avoid a TOCTOU race with a concurrent
+    finalize/follow-up on the same task. Does not touch child tasks — callers should warn
+    if any remain non-terminal. Returns True if finalization occurred.
+    """
```
jmelloy, line 176: The code comment claims a single conditional UPDATE avoids TOCTOU, but the preceding SELECT to fetch `worker_id` is a separate DB round-trip that creates the same race window. If two concurrent finalizers both pass the SELECT, only one UPDATE will win (rowcount=0 for the other), but both will have read a potentially stale `worker_id`. Consider fetching `worker_id` via `RETURNING` in the UPDATE, or accepting that `worker_id` from the lost-race path is unused since `rowcount==0` gates the rest.

backend/routes/webhooks.py
```diff
@@ -865,6 +865,37 @@ async def github_webhook(
...
+            if action == "closed":
+                from foreman.tools import (
+                    finalize_issue_root_task,
+                    warn_if_issue_task_has_open_children,
+                )
+
+                root_result = await db.exec(
+                    select(col(Task.id)).where(
+                        col(Task.guild_id) == guild_pk,
+                        col(Task.issue_repo) == repo,
```
jmelloy, line 882: The query to find issue-root tasks does not filter out already-terminal states (`state NOT IN _TERMINAL_STATES`). For every `issues.closed` webhook delivery GitHub resends (e.g. retries), this will iterate over historically-closed root tasks and call `finalize_issue_root_task` on each, which returns False immediately but still executes two DB round-trips per task. Add `~col(Task.state).in_(list(_TERMINAL_STATES))` to the WHERE clause.

backend/foreman/runner.py
```diff
@@ -479,6 +479,70 @@ async def _fetch_pr_status_lines(guild_id: str, pr_tasks: list[dict]) -> list[st
...
+            continue
+
+        db = await get_db()
+        try:
+            guild_pk_val = await get_guild_pk(db, guild_id)
+            finalized = await finalize_issue_root_task(db, guild_pk_val, guild_id, t["id"])
+            if finalized:
+                logger.info(
+                    "guild=%s task=%s finalized (issue %s#%s closed)",
+                    guild_id,
```
jmelloy, line 536: `get_db()` and `db.close()` are called inside the `for t in issue_tasks` loop, creating a new DB connection for every individual issue task on every sweep cycle. Move the `db = await get_db()` and the corresponding `finally: await db.close()` outside the loop to use a single connection for the entire sweep batch.

jmelloy (commented):
- The `finalize_issue_root_task` helper does a SELECT then UPDATE in two separate DB calls, but the code comment claims it uses a single conditional UPDATE to avoid TOCTOU — the SELECT for `worker_id` is a separate round-trip that creates the exact race it documents as prevented.
- The webhook handler queries for issue-root tasks without filtering out already-terminal states, meaning `finalize_issue_root_task` will be called (and return False) for every historically closed task matching that repo/issue_number, wasting DB round-trips on every webhook delivery.
- `_sweep_closed_issue_tasks` acquires a new DB connection inside the loop per task, creating one connection per issue task per sweep cycle; `get_db()` and `db.close()` should be outside the loop.
- `warn_if_issue_task_has_open_children` uses `live_tasks_filter()` which may filter by `deleted_at IS NULL`, but the parent was just finalized with a future `deleted_at` rather than NULL — children may share a similar pattern and could be incorrectly excluded from the warning query.
- The `fetch_issue_state` function falls back to `"open"` if `state` key is missing in the API response, which is a safe default but silently swallows unexpected API shapes; an explicit key error or logging would be safer.

PR #849, opened 2026-07-08
Consolidate LLM provider logic into shared module; make proxy a thin wrapper

### LINKED ISSUE #826: Clarify embedded foreman vs proxy LLM provider boundary
## Context

Follow-up from #790 and PR #825.

PR #825 simplifies the standalone foreman into a thin LLM API proxy. After that change, the backend/embedded foreman owns the conversation loop, state, history, tool execution, polling, and child contexts. The standalone process only executes provider API calls when connected.

The code now shares Anthropic/Bedrock model and client helper logic through `backend.foreman.llm`, used by both:

- embedded backend runner: `backend/foreman/runner.py`
- standalone proxy: `foreman/pioneer_foreman/runner.py`

But OpenAI-compatible translation currently lives only in the standalone proxy. That is intentional in the current design: the embedded foreman either calls Anthropic/Bedrock locally or delegates provider execution to the proxy.

## Cleanup / clarification needed

Clarify this boundary in code/docs/tests so future changes do not accidentally reintroduce provider logic in the wrong place.

Questions to settle:

- Should OpenAI-compatible request/response translation remain proxy-only? - no
- Should any of that translation move into a shared helper module, even if only the proxy calls it today? - yes, it should all be combined 
- Should the embedded foreman reject `FOREMAN_PROVIDER=openai` unless a proxy is connected, as it does now, or should the backend support OpenAI-compatible local calls directly? - no 
- Should naming/docs say `proxy provider` vs `embedded provider` more explicitly?

## Suggested outcome

Pick one explicit architecture and encode it:

   - Make errors and docs use consistent wording.
   - Move provider normalization/conversion helpers into a small shared module under `backend/foreman/`.
   - have the embedded runner and the proxy use the same logic

jmelloy:
## Summary
- Moves all provider normalization/conversion logic — Anthropic calls, OpenAI-compatible request/response translation — into `backend/foreman/llm.py`, the single shared module already used for the Anthropic/Bedrock client factory.
- `foreman/pioneer_foreman/runner.py` (the standalone proxy) no longer duplicates any translation logic; it now imports `call_anthropic`, `call_openai_compatible`, and `ANTHROPIC_SDK_PROVIDERS` from the shared module and only decides which client/machine executes the request.
- `backend/foreman/runner.py` (embedded foreman) now calls the same shared `call_anthropic` helper instead of building the Anthropic SDK request inline, and its error message for unsupported providers now references `ANTHROPIC_SDK_PROVIDERS` for consistent proxy/embedded provider wording.
- Added a module-level docstring in `backend/foreman/llm.py` spelling out the architecture boundary from issue #826: the proxy exists only to pick *which machine* calls the provider endpoint; it holds no provider-specific logic of its own.
- Added `backend/tests/test_foreman_llm.py` covering the moved OpenAI-compatible translation helpers; trimmed the now-redundant duplicate tests out of `foreman/tests/test_runner.py`.

Closes #826.

## Test plan
- [x] `ruff check .` — clean
- [x] `ruff format . --check` — clean
- [x] `cd foreman && python -m pytest` — 40 passed
- [x] `cd backend && python -m pytest tests/test_foreman_llm.py` — 31 passed
- [x] `cd backend && python -m pytest -k "foreman or llm or runner"` — 270 passed

### DIFF
--- a/backend/foreman/llm.py
+++ b/backend/foreman/llm.py
@@ -387,14 +387,20 @@ async def call_anthropic(
     plain dict (e.g. the proxy, to send the response back over the wire) call
     `.model_dump()` themselves; callers that stay in-process (the embedded
     foreman's local call path) can use it directly.
+
+    `tools` is only included in the request when non-empty: some Anthropic API
+    endpoints reject an explicit empty `tools` array, and omitting the param
+    (rather than defaulting a missing/None value to `[]`) also means a caller
+    that explicitly passes `tools=None` isn't silently overridden.
     """
     kwargs: dict[str, Any] = {
         "model": model,
         "max_tokens": max_tokens,
         "system": system or [],
         "messages": messages or [],
-        "tools": tools or [],
     }
+    if tools:
+        kwargs["tools"] = tools
     if tool_choice is not None:
         kwargs["tool_choice"] = tool_choice
     raw = await client.messages.with_raw_response.create(**kwargs)
@@ -574,30 +580,37 @@ def openai_response_to_anthropic(raw: dict[str, Any], model: str) -> dict[str, A
 
 async def call_openai_compatible(
     client: httpx.AsyncClient,
-    request: dict[str, Any],
     *,
     model: str,
+    max_tokens: int,
     base_url: str,
+    system: list[dict[str, Any]] | None = None,
+    messages: list[dict[str, Any]] | None = None,
+    tools: list[dict[str, Any]] | None = None,
+    tool_choice: dict[str, Any] | None = None,
     api_key: str | None = None,
 ) -> tuple[dict[str, Any], str | None]:
     """Execute one OpenAI-compatible POST /chat/completions call.
 
-    `request` is the same Anthropic-shaped request dict call_anthropic takes
-    (model/maxTokens/system/messages/tools/toolChoice); the request is
-    translated to OpenAI chat-completion format, POSTed, and the response is
-    translated back to an Anthropic Messages API response dict — so callers
-    (the standalone proxy today) don't need to know which wire format actually
+    Takes the same Anthropic-shaped, snake_case keyword arguments as
+    call_anthropic (model/max_tokens/system/messages/tools/tool_choice) rather
+    than a raw request dict, so the two provider call paths share one calling
+    convention. The caller (the standalone proxy) is responsible for
+    translating its own wire format (the backend's camelCase
+    maxTokens/toolChoice JSON) into these kwargs — that's the "which machine"
+    plumbing the proxy owns; everything below is the provider translation this
+    module owns. The request is converted to OpenAI chat-completion format,
+    POSTed, and the response is translated back to an Anthropic Messages API
+    response dict, so callers don't need to know which wire format actually
     served the request. Returns (response_dict, request_id).
     """
     body: dict[str, Any] = {
         "model": model,
-        "messages": anthropic_messages_to_openai(
-            request.get("system") or [], request.get("messages") or []
-        ),
-        "tools": anthropic_tools_to_openai(request.get("tools") or []),
-        "max_tokens": request.get("maxTokens") or 1024,
+        "messages": anthropic_messages_to_openai(system or [], messages or []),
+        "max_tokens": max_tokens,
     }
-    tool_choice = request.get("toolChoice")
+    if tools:
+        body["tools"] = anthropic_tools_to_openai(tools)
     if tool_choice:
         body["tool_choice"] = "none" if tool_choice.get("type") == "none" else tool_choice
 
--- a/backend/foreman/llm.py
+++ b/backend/foreman/llm.py
@@ -1,14 +1,33 @@
-"""Anthropic client factory supporting both the direct API and Amazon Bedrock.
+"""Shared LLM provider logic: client factory, the Anthropic call, and OpenAI-
+compatible request/response translation.
 
-Both the embedded foreman (backend) and standalone foreman import this to build
-their own module-level singleton — the factory is shared, the instance is not.
+Both the embedded foreman (backend/foreman/runner.py) and the standalone proxy
+(foreman/pioneer_foreman/runner.py) import this module for ALL provider
+normalization — this is the one place that knows the shape of each provider's
+wire format. Neither runner re-implements or duplicates any of it.
+
+Architecture boundary (see issue #826): the proxy exists only to decide *which
+machine* makes the outbound call — a different Bedrock region/profile, a
+locally-reachable OpenAI-compatible endpoint (e.g. Ollama), whatever the
+operator configured. It holds no provider-specific logic of its own; it just
+plumbs its own config into the helpers below. The embedded foreman calls
+`make_anthropic_client` + `call_anthropic` directly for the providers the
+Anthropic SDK can reach itself (`ANTHROPIC_SDK_PROVIDERS`); for anything else
+(currently just `"openai"`) it requires a connected foreman API proxy rather
+than calling `call_openai_compatible` locally — see the FOREMAN_PROVIDER check
+in `backend/foreman/runner.py`.
 """
 
 from __future__ import annotations
 
+import json
 import logging
 import os
 from collections.abc import Mapping
+from typing import Any
+from uuid import uuid4
+
+import httpx
 
 try:
     import anthropic as _anthropic_mod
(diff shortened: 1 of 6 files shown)


backend/foreman/llm.py
```diff
@@ -331,3 +357,256 @@ def make_anthropic_client(
...
+    messages: list[dict[str, Any]] | None = None,
+    tools: list[dict[str, Any]] | None = None,
+    tool_choice: dict[str, Any] | None = None,
+) -> tuple[Any, str | None]:
+    """Execute one Anthropic Messages API call. Returns (parsed_response, request_id).
+
+    `client` is any Anthropic SDK async client from make_anthropic_client
+    (AsyncAnthropic or AsyncAnthropicBedrock — Bedrock speaks the same Messages
+    API shape, so one call path covers both). Both the embedded foreman and the
+    standalone proxy call this against whichever client their own config
```
jmelloy, line 382: Passing `tools=[]` unconditionally means an empty array is always sent to the Anthropic API, even when no tools are defined. Some API versions/models may reject or warn on an empty `tools` list. Consider omitting the `tools` key entirely when the list is empty, similar to how `tool_choice` is conditionally added.

backend/foreman/llm.py
```diff
@@ -331,3 +357,256 @@ def make_anthropic_client(
...
+                    if isinstance(block, dict) and block.get("type") == "text"
+                ]
+            )
+            tool_calls = []
+            for block in content:
+                if not isinstance(block, dict) or block.get("type") != "tool_use":
+                    continue
+                tool_calls.append(
+                    {
+                        "id": block.get("id") or f"toolu_{uuid4().hex}",
```
jmelloy, line 493: `call_openai_compatible` takes a raw camelCase dict (`maxTokens`, `toolChoice`) while `call_anthropic` uses explicit snake_case kwargs. This inconsistency in the public interface of the shared module makes it harder to use correctly and is not reflected in the docstrings. Consider normalizing to one convention or documenting the camelCase contract explicitly.

backend/tests/test_foreman_llm.py
```diff
@@ -433,3 +433,215 @@ def test_anthropic_env_set_after_bedrock_reload(self, monkeypatch):
...
+        import json
+
+        import httpx
+        from foreman.llm import call_openai_compatible
+
+        captured = {}
+
+        async def handler(request: httpx.Request) -> httpx.Response:
+            captured["url"] = str(request.url)
+            captured["headers"] = dict(request.headers)
```
jmelloy, line 566: This async test method has no `@pytest.mark.asyncio` decorator. Without it (or `asyncio_mode = auto` in pytest config), pytest will collect the coroutine function but never await it, silently passing without executing any assertions. Add the marker or confirm the pytest-asyncio configuration covers this.

backend/tests/test_foreman_llm.py
```diff
@@ -433,3 +433,215 @@ def test_anthropic_env_set_after_bedrock_reload(self, monkeypatch):
...
+# call_anthropic — shared Anthropic Messages API call, used locally by the
+# embedded foreman and by the standalone proxy against its own client.
+# ---------------------------------------------------------------------------
+
+
+class TestCallAnthropic:
+    async def test_call_anthropic_returns_parsed_response_and_request_id(self):
+        """call_anthropic returns the SDK's own parsed response object (not a
+        pre-dumped dict) so local callers keep full attribute access; callers
+        that need JSON (the proxy) call .model_dump() themselves."""
```
jmelloy, line 647: Same issue as above — `test_call_anthropic_returns_parsed_response_and_request_id` is an async method with no `@pytest.mark.asyncio` decorator, so it will silently pass without running if asyncio_mode is not set globally.

foreman/pioneer_foreman/runner.py
```diff
@@ -55,219 +62,49 @@ def _get_http_client(config: Config) -> httpx.AsyncClient:
...
+async def _call_openai_compatible(request: dict[str, Any], config: Config) -> dict[str, Any]:
+    model = config.effective_model or request["model"]
+    client = _get_http_client(config)
+    response, request_id = await call_openai_compatible(
+        client,
+        request,
+        model=model,
+        base_url=config.openai_base_url,
+        api_key=config.api_key,
+    )
```
jmelloy, line 112: The dispatch order changed: `ANTHROPIC_SDK_PROVIDERS` is now checked before `'openai'`, which is correct, but `'openai'` is not in `ANTHROPIC_SDK_PROVIDERS` so a future provider added to the frozenset that overlaps with the `openai` branch could silently skip it. Consider using `elif` and an `else` chain with explicit exhaustiveness, or asserting the sets are disjoint in a test.

jmelloy (commented):
- Good consolidation: moving all provider normalization logic into `backend/foreman/llm.py` eliminates duplication between the embedded foreman and the standalone proxy, and the tests are correctly migrated to the shared module.
- The `call_anthropic` helper passes `tools=[]` as a default even when no tools are needed, which may cause issues with some Anthropic API endpoints that reject an empty `tools` array; it should pass `tools` only when non-empty.
- `call_openai_compatible` accepts a raw `request` dict using camelCase keys (`maxTokens`, `toolChoice`) while `call_anthropic` uses snake_case keyword arguments — this inconsistency in the shared module's interface makes the API harder to use and reason about.
- The `_call_anthropic` wrapper in the proxy falls back to `request['model']` if `config.effective_model` is falsy, but `call_anthropic` in `llm.py` defaults `tools` to `[]` unconditionally — if the caller explicitly passes `tools=None` the behavior is silently changed, which could mask bugs.
- Test class methods marked `async` (e.g. `test_call_anthropic_returns_parsed_response_and_request_id`, `test_call_openai_compatible_posts_chat_completion`) will not run as coroutines without a pytest-asyncio marker or `asyncio_mode = auto` configuration; they will silently pass without executing the async body.

backend/foreman/llm.py
```diff
@@ -331,3 +357,269 @@ def make_anthropic_client(
...
+    tools: list[dict[str, Any]] | None = None,
+    tool_choice: dict[str, Any] | None = None,
+    api_key: str | None = None,
+) -> tuple[dict[str, Any], str | None]:
+    """Execute one OpenAI-compatible POST /chat/completions call.
+
+    Takes the same Anthropic-shaped, snake_case keyword arguments as
+    call_anthropic (model/max_tokens/system/messages/tools/tool_choice) rather
+    than a raw request dict, so the two provider call paths share one calling
+    convention. The caller (the standalone proxy) is responsible for
```
jmelloy, line 598: `if tool_choice:` is falsy for an empty dict `{}`, but a non-None falsy value like `{}` would be silently dropped rather than forwarded. More importantly, a non-`none` tool_choice dict is passed through verbatim to the OpenAI endpoint without converting Anthropic-style `{"type": "auto"}` to the OpenAI string `"auto"`, which will likely cause a 400 from strict OpenAI-compatible endpoints.

backend/foreman/llm.py
```diff
@@ -331,3 +357,269 @@ def make_anthropic_client(
         kwargs.get("base_url", "default"),
     )
     return _anthropic_mod.AsyncAnthropic(**kwargs)
+
+
+# ---------------------------------------------------------------------------
+# Anthropic Messages API call
+# ---------------------------------------------------------------------------
+
+
```
jmelloy, line 366: `"system": system or []` means a caller passing `system=[]` explicitly gets an empty list forwarded, which is fine, but a caller passing `system=None` also gets `[]`. The docstring says `tools=None` is respected (not coerced to `[]`), but `system` and `messages` don't get the same treatment — this inconsistency could confuse future callers and should be documented or made consistent.

backend/foreman/llm.py
```diff
@@ -331,3 +357,269 @@ def make_anthropic_client(
...
+                    for block in content
+                    if isinstance(block, dict) and block.get("type") == "text"
+                ]
+            )
+            tool_calls = []
+            for block in content:
+                if not isinstance(block, dict) or block.get("type") != "tool_use":
+                    continue
+                tool_calls.append(
+                    {
```
jmelloy, line 492: The `httpx` import is now unconditional at module top-level, making `httpx` a hard dependency of `backend/foreman/llm.py`. Previously this module had no HTTP client dependency. Any process importing this module (including the embedded backend) will now fail at import time if `httpx` is not installed, even if it never calls `call_openai_compatible`. Consider a lazy import or guarded try/except similar to the `anthropic` import pattern already used in this file.

jmelloy (commented):
- Good architectural consolidation: moving all provider translation logic into `llm.py` eliminates duplication and enforces a clean boundary, consistent with issue #826.
- `call_openai_compatible` silently passes an empty `tool_choice` dict as truthy and forwards it verbatim to OpenAI rather than omitting it, which may cause API errors on endpoints that don't accept an empty object.
- The proxy's `_call_anthropic` passes `tools=request.get('tools')` (possibly `None`) while the old code defaulted to `[]`; this is intentional per the new semantics, but the diff is truncated so it's unclear if `tool_choice` mapping is fully preserved — needs verification.
- Test coverage for the new shared helpers is thorough and well-structured; the async tests correctly use `httpx.MockTransport` without requiring a real network.
- Minor: `_anthropic_content_to_text` is a private helper in `llm.py` but `anthropic_tools_to_openai` and `anthropic_messages_to_openai` are public — consider whether all three should be public or document the intentional asymmetry.

PR #852, opened 2026-07-08
fix: omit tools/tool_choice when empty in llm.py

### LINKED ISSUE #850: fix: call_anthropic / call_openai_compatible should omit tools/tool_choice when empty
## Problem

In `backend/foreman/llm.py` (introduced in #849), `call_anthropic` passes `tools=[]` as a default argument even when no tools are needed. Similarly, `call_openai_compatible` forwards an empty `tool_choice={}` dict as a truthy value verbatim to the OpenAI-compatible endpoint.

This can cause API errors on providers that don't accept an empty `tools` array or an empty `tool_choice` object (e.g. some OpenAI-compatible endpoints and Bedrock variants reject these).

## Fix required

- In `call_anthropic`: only include `tools` in the request payload when the caller passes a non-empty list.
- In `call_openai_compatible`: only include `tool_choice` in the request payload when it is non-empty/non-None.
- Ensure callers that genuinely need no tools simply omit the argument (or pass `None`/`[]`) and it is cleanly omitted from the outgoing API call.

## Context

Identified in code review of PR #849 (LLM provider consolidation, part of #826).
Parent issue: #815 (issue-rooted task tree — this is one of the remaining gaps noted in the review).

jmelloy:
## Summary
- Guard `call_anthropic` so `tool_choice` is only included in the request when non-empty, matching the existing `tools` guard (some endpoints, e.g. Bedrock, reject an empty `tool_choice` object).
- Guard `call_openai_compatible` so an empty `tool_choice={}` is omitted from the payload instead of being forwarded verbatim (OpenAI-compatible APIs reject an empty `tool_choice` object).
- Added regression tests asserting `tool_choice` does not appear in the serialized request payload for both `call_anthropic` and `call_openai_compatible` when passed an empty dict.

Closes #850

## Test plan
- [x] `pytest backend/tests/test_foreman_llm.py` — 36 passed
- [x] `ruff check backend/foreman/llm.py backend/tests/test_foreman_llm.py` — clean

### DIFF
--- a/backend/foreman/llm.py
+++ b/backend/foreman/llm.py
@@ -364,6 +364,16 @@ def make_anthropic_client(
 # ---------------------------------------------------------------------------
 
 
+def _has_payload(value: dict[str, Any] | list[Any] | None) -> bool:
+    """True when `value` is a non-None, non-empty dict/list.
+
+    Explicit rather than a bare truthiness check so intent is clear and a
+    future falsy-but-meaningful value (e.g. `False`) wouldn't be silently
+    dropped the way an empty `tools`/`tool_choice` should be.
+    """
+    return value is not None and value != {} and value != []
+
+
 async def call_anthropic(
     client: Any,
     *,
@@ -400,9 +410,9 @@ async def call_anthropic(
         "system": system or [],
         "messages": messages or [],
     }
-    if tools:
+    if _has_payload(tools):
         kwargs["tools"] = tools
-    if tool_choice:
+    if _has_payload(tool_choice):
         kwargs["tool_choice"] = tool_choice
     raw = await client.messages.with_raw_response.create(**kwargs)
     return raw.parse(), raw.headers.get("request-id")
@@ -610,9 +620,9 @@ async def call_openai_compatible(
         "messages": anthropic_messages_to_openai(system or [], messages or []),
         "max_tokens": max_tokens,
     }
-    if tools:
+    if _has_payload(tools):
         body["tools"] = anthropic_tools_to_openai(tools)
-    if tool_choice:
+    if _has_payload(tool_choice):
         body["tool_choice"] = "none" if tool_choice.get("type") == "none" else tool_choice
 
     headers = {"Content-Type": "application/json"}
--- a/backend/foreman/llm.py
+++ b/backend/foreman/llm.py
@@ -388,10 +388,11 @@ async def call_anthropic(
     `.model_dump()` themselves; callers that stay in-process (the embedded
     foreman's local call path) can use it directly.
 
-    `tools` is only included in the request when non-empty: some Anthropic API
-    endpoints reject an explicit empty `tools` array, and omitting the param
-    (rather than defaulting a missing/None value to `[]`) also means a caller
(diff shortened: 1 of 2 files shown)


backend/foreman/llm.py
```diff
@@ -401,7 +402,7 @@ async def call_anthropic(
     }
     if tools:
         kwargs["tools"] = tools
-    if tool_choice is not None:
+    if tool_choice:
```
jmelloy: Changing `if tool_choice is not None:` to `if tool_choice:` silently drops any falsy-but-semantically-valid value (e.g. an empty string, `False`, or `0`). While Anthropic's `tool_choice` is always a dict in practice, a more explicit guard like `if tool_choice is not None and tool_choice != {}:` (or a dedicated helper) would make the intent clearer and be safer against future API changes.

backend/tests/test_foreman_llm.py
```diff
@@ -633,6 +633,43 @@ async def handler(request: httpx.Request) -> httpx.Response:
...
+                    "usage": {"prompt_tokens": 1, "completion_tokens": 1},
+                },
+            )
+
+        client = httpx.AsyncClient(transport=httpx.MockTransport(handler))
+        try:
+            await call_openai_compatible(
+                client,
+                model="llama3.1",
+                max_tokens=32,
```
jmelloy, line 663: The `_RawResponse.headers` is a plain `{}` dict, but real response headers are typically `httpx.Headers` or a similar mapping. If `call_anthropic` ever calls `.get()` with a default argument this works fine, but consider using `httpx.Headers({})` or a `MappingProxyType` for fidelity with production behavior.

backend/tests/test_foreman_llm.py
```diff
@@ -633,6 +633,43 @@ async def handler(request: httpx.Request) -> httpx.Response:
 
         assert "tools" not in captured["json"]
 
+    async def test_call_openai_compatible_omits_empty_tool_choice(self):
```
jmelloy, line 636: Consider also adding a complementary positive-case test asserting that a non-empty `tool_choice` (e.g. `{"type": "auto"}`) *is* forwarded, to guard against an overly aggressive filter accidentally dropping valid values in future refactors.

jmelloy (commented):
- The core fix (`if tool_choice:` instead of `if tool_choice is not None:`) is correct and solves the reported issue for empty dict inputs.
- However, changing from `is not None` to a truthiness check introduces a subtle behavioral regression: falsy-but-valid `tool_choice` values (e.g. `0`, `False`, or a future string `"none"`) would now be silently dropped instead of forwarded, though in practice Anthropic's `tool_choice` is always a dict so this risk is low today.
- The `call_openai_compatible` fix is not shown in the diff body but is referenced in the description; the test covers it, suggesting a similar truthiness guard was applied there as well — worth verifying the same edge-case concern applies consistently.
- Regression tests are well-structured, use appropriate mocking patterns consistent with the existing test suite, and directly assert the contract required by the fix.
- The docstring update accurately reflects the expanded behavior and is a good improvement to maintainability.

PR #854, opened 2026-07-08
feat: auto-finalize issue-root + cascade soft-delete

### LINKED ISSUE #853: feat: auto-finalize issue-root task + cascade soft-delete to children (#815 follow-up)
## Summary

Follow-up to #815. The finalization behavior described in that issue was never implemented.

### 1. Auto-close root task on `issues/closed` webhook
`backend/routes/webhooks.py` handles `issues/closed` but only updates `issue_state`/`issue_title` — it never transitions the root task's `state` to `done` or calls finalize logic.

### 2. Cascade soft-delete to children on finalize
`finalize_task` in `backend/foreman/tools.py` only updates the single task being finalized. When a `phase='issue'` root is finalized, cascade `deleted_at` to all child tasks already in a terminal state (`done`, `failed`, `cancelled`).

### 3. Final verification scan before closing
When the root task is auto-closed, run a final completeness check: verify all child PRs are merged and all child tasks are terminal, then post a summary comment on the GitHub issue before marking the root done. This mirrors the manual review workflow used for #815.

## Acceptance criteria

- `issues/closed` webhook finds the `phase='issue'` task linked to that issue and finalizes it (state=done, deleted_at set)
- `finalize_task` on a `phase='issue'` root cascades `deleted_at` to all already-terminal children (does not force-close in-progress children)
- Before closing, post a comment on the GH issue summarising child PR/task states
- Tests cover: webhook→finalize, cascade soft-delete, pre-close summary comment

## Related
- Parent: #815
- `backend/routes/webhooks.py:838-862`
- `backend/foreman/tools.py:1675-1726`

jmelloy:
Closes #853

## Summary
- Cascades `deleted_at` soft-delete to already-terminal (done/failed/cancelled/error) descendant tasks when a `phase='issue'` root task is finalized via the `finalize_task` foreman tool, without force-closing in-progress/pending children.
- Fixes a bug where `find_descendant_tasks`'s recursive CTE lookup silently returned no descendants: SQLModel's `exec()` scalar-unwraps single-column selects, so `row[0]` on a plain task-id string sliced its first character instead of indexing a tuple.
- Fixes the same row-tuple bug in the `issues/closed` webhook handler in `routes/webhooks.py`, which was silently failing to resolve the `phase='issue'` root task id to finalize.
- Removes a stray debug script left in the working tree.

(The webhook-driven finalize and pre-close GitHub summary comment behaviors were already implemented in #848; this PR completes the cascade soft-delete piece and fixes the two latent bugs found while getting the test suite green.)

## Test plan
- [x] `cd backend && python3 -m pytest tests/ -x -q` — 948 passed, 2 skipped
- [x] `test_finalize_task_cascades_to_terminal_children_only` — verifies terminal children/grandchildren get soft-deleted, in-progress child does not
- [x] `test_finalize_task_posts_pre_close_summary_comment` — verifies GitHub comment summarizing child PR merge status
- [x] `test_issues_closed_webhook_finalizes_issue_root_task` / `test_issues_closed_webhook_ignores_already_terminal_root` — webhook finalize path


backend/foreman/tools.py
```diff
@@ -227,6 +235,123 @@ async def warn_if_issue_task_has_open_children(db, guild_id: str, root_task_id:
...
+) -> list[str]:
+    """Soft-delete every already-terminal descendant task, leaving open ones alone.
+
+    Only tasks already in a terminal state (done/failed/cancelled/error — see
+    ``_TERMINAL_STATES``) and not yet soft-deleted are touched. In-progress or
+    pending descendants are never force-closed here; a human decides whether to
+    cancel in-flight work (see ``warn_if_issue_task_has_open_children``). Returns
+    the ids that were updated.
+    """
+    terminal_ids = [
```
jmelloy, line 280: This filters only on the descendant's own state, so a `done` grandchild whose parent is still `working` will be soft-deleted even though the parent is not being closed. Consider walking the list and excluding any task whose direct parent is not in `terminal_ids` (or the root), to avoid orphaning the logical sub-tree relationship.

backend/foreman/tools.py
```diff
@@ -199,6 +204,9 @@ async def finalize_issue_root_task(db, guild_pk: int, guild_id: str, task_id: st
         guild_id,
```
jmelloy, line 204: `issue_number is not None` is the right guard, but `issue_repo` is checked with a bare truthiness test. An empty string `''` would pass the `is not None` guard but fail the truthy check inconsistently. Use `issue_repo is not None` for symmetry, or add an explicit `issue_repo != ''` check.

backend/foreman/tools.py
```diff
@@ -227,6 +235,123 @@ async def warn_if_issue_task_has_open_children(db, guild_id: str, root_task_id:
...
+
+
+async def find_descendant_tasks(db, root_task_id: str) -> list[Task]:
+    """Return every task reachable from *root_task_id* via the ``parent_task_id`` chain.
+
+    Walks the tree downward (plan/execute/review/followup, and anything chained
+    off of those) in a single recursive CTE rather than one round-trip per level.
+    """
+    anchor = select(col(Task.id).label("id"), literal(0).label("depth")).where(
+        col(Task.parent_task_id) == root_task_id
```
jmelloy, line 251: The recursive CTE depth guard uses `_MAX_DESCENDANT_CHAIN_DEPTH - 1` (i.e. 19) in the WHERE clause but the constant is named `_MAX_DESCENDANT_CHAIN_DEPTH = 20`. Off-by-one is fine if intentional, but the constant's name implies it is the maximum depth, so the condition should probably be `< _MAX_DESCENDANT_CHAIN_DEPTH` without the `- 1` to match the documented semantics.

backend/routes/webhooks.py
```diff
@@ -885,8 +885,7 @@ async def github_webhook(
                         col(Task.state).notin_(list(_WEBHOOK_TERMINAL_STATES)),
                     )
                 )
-                for row in root_result.all():
-                    root_task_id = row[0]
+                for root_task_id in root_result.all():
```
jmelloy, line 888: The fix changes `row[0]` to iterating `root_result.all()` directly. This is correct because SQLModel scalar-unwraps single-column results, so each element is already a plain string. Worth adding a brief comment here to prevent a future developer from reintroducing the `row[0]` indexing, since the previous bug was non-obvious.

jmelloy (commented):
- Fixes two real bugs: the `row[0]` string-slicing bug in `find_descendant_tasks` and the same pattern in the webhook handler, both caused by SQLModel scalar-unwrapping single-column results.
- The cascade soft-delete logic is sound: only already-terminal descendants are touched, leaving in-progress/pending tasks alone as documented.
- `post_issue_close_summary_comment` is correctly wrapped in a broad `except` so GitHub API failures never block task finalization.
- There is a subtle correctness issue: `cascade_soft_delete_terminal_descendants` does not recursively exclude descendants that hang below an in-progress task — a grandchild of a `working` parent that is itself `done` will still be soft-deleted, which may be surprising.
- Test plan is empty and the webhook test does not assert that cascade soft-delete and the summary comment are triggered end-to-end; the two behaviours are only tested through the foreman tool path.

PR #856, opened 2026-07-08
feat: queue human messages when Foreman busy (#855)

### LINKED ISSUE #855: feat: Foreman should queue human messages when busy, not drop them
## Problem

The Foreman currently uses a **drop-if-busy** strategy for incoming triggers/messages (introduced in #808). This is acceptable for automated periodic-check events, but **human messages should never be silently dropped** — if the Foreman is busy processing when a human sends a message, that message should be queued and processed as soon as the current turn completes.

## Expected behaviour

- **Human messages** (Discord, web UI, direct API): always queued; processed in order as soon as the Foreman is free.
- **Automated/periodic events** (periodic-check, GitHub webhooks from CI bots): drop-if-busy is fine — these repeat and stale ones can be discarded.

## Suggested implementation

1. Classify incoming messages by source type (human vs. automated).
2. Maintain a small FIFO queue for human messages.
3. After each Foreman turn completes, drain the queue before going fully idle.
4. Automated events continue to use the existing drop-if-busy path.

## Why this matters

A human sends a message while the Foreman is mid-turn (e.g. processing a periodic-check or a GitHub event). Under the current design, that message is silently dropped and the human gets no response. This is a bad UX and could cause lost instructions.

jmelloy:
Implements a small FIFO queue for human-originated messages so they are never dropped when the Foreman is busy. Automated events (periodic-check, github-event) keep the existing drop-if-busy behaviour. Closes #855

### DIFF
--- a/backend/foreman/runner.py
+++ b/backend/foreman/runner.py
@@ -77,17 +77,37 @@ _HISTORY_FETCH_LIMIT = 100
 # Per-guild background poll task registry
 _poll_tasks: dict[str, "asyncio.Task[None]"] = {}
 
-# Per-guild locks to serialise foreman runs.  If a run is already in progress
-# for a (guild, user) pair, new invocations are dropped rather than queued —
-# the poll loop will re-trigger on the next tick.  Dropping (vs. queuing) keeps
-# memory bounded and avoids stale snapshots piling up under load.
+# Per-guild run state, used to serialise foreman runs.  If a run is already in
+# progress for a (guild, user) pair, new invocations are dropped rather than
+# queued — the poll loop will re-trigger on the next tick.  Dropping (vs.
+# queuing) keeps memory bounded and avoids stale snapshots piling up under load.
 # Entries are popped in the finally block after each run so the dict stays small.
 #
 # Human-originated messages (Discord/web UI chat, user-requested follow-ups)
-# are the one exception: they are never silently dropped. When the lock for
-# their key is held, they are appended to ``_human_queues`` instead, and
+# are the one exception: they are never silently dropped. When the entry for
+# their key is busy, they are appended to ``_human_queues`` instead, and
 # drained (FIFO) once the in-flight run finishes — see ``run_foreman_ai``.
-_guild_locks: dict[tuple[str, str | None], asyncio.Lock] = {}
+@dataclass
+class _GuildRunLock:
+    """Tracks whether a foreman run currently owns a given (guild, key) slot.
+
+    ``busy`` is checked and flipped to True in the same synchronous step (no
+    ``await`` in between) inside ``run_foreman_ai``, so two concurrent
+    invocations can never both observe ``busy=False`` and both proceed:
+    asyncio's single-threaded, cooperative scheduler only switches between
+    coroutines at an ``await`` point, so nothing can interleave between the
+    check and the claim. This is deliberately a plain flag rather than
+    ``asyncio.Lock.locked()`` + ``await lock.acquire()`` — that pattern is
+    *currently* just as atomic (acquiring an uncontended ``asyncio.Lock``
+    never suspends), but it relies on ``Lock`` internals that a future edit
+    could silently break (e.g. adding an ``await`` between the check and the
+    acquire). A dedicated flag makes the invariant self-evident.
+    """
+
+    busy: bool = False
+
+
+_guild_locks: dict[tuple[str, str | None], _GuildRunLock] = {}
 
 # Max number of queued human messages per (guild, user)/(guild, task) key.
 # Bounded so a guild nobody is watching can't grow this without limit; the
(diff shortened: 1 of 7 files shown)


backend/foreman/runner.py
```diff
@@ -814,22 +851,117 @@ async def run_foreman_ai(
...
+                "guild=%s key=%s error processing queued human message",
+                turn.guild_id,
+                lock_key[1],
+            )
+    _human_queues.pop(lock_key, None)
+
+
 async def _run_foreman_ai(
     guild_id: str,
     human_message: str,
```
jmelloy, line 1027: The `_human_queues.pop(lock_key, None)` after the while-loop can silently discard messages that were enqueued into `_human_queues[lock_key]` by a turn processed *inside* the loop (i.e. a drained run itself triggers another human message before `queue` empties). In practice this is safe today because new messages enqueue into the same `queue` deque object and the `while queue:` loop picks them up — but the final pop is redundant and misleading; the queue will already be empty at that point. Replace with an `assert not queue` or remove the pop and let the caller's `finally` block handle cleanup to make the invariant explicit.

backend/foreman/runner.py
```diff
@@ -814,22 +851,117 @@ async def run_foreman_ai(
     lock_key = (guild_id, f"task:{task_id}") if use_child else (guild_id, user_id)
```
jmelloy, line 871: There is a TOCTOU window: `lock.locked()` is checked, then `await lock.acquire()` is called. Between those two statements another coroutine could release the lock, making this invocation acquire immediately — correct. But if two coroutines both observe `locked() == False` before either calls `acquire()`, only one will actually acquire first; the second will `await` (block) rather than being queued or dropped. For a human message this silently bypasses the queue, and for an automated message it silently blocks instead of dropping. Consider using `lock.acquire()` with `asyncio.wait_for` or the standard `if not await asyncio.shield(lock.acquire())` pattern, or restructure with `asyncio.Lock` try-acquire semantics.

backend/foreman/runner.py
```diff
@@ -814,22 +851,117 @@ async def run_foreman_ai(
...
+            "queued message (queued_at=%s)",
+            guild_id,
+            lock_key[1],
+            _HUMAN_QUEUE_MAX,
+            dropped.queued_at,
+        )
+    queue.append(
+        _QueuedHumanTurn(
+            guild_id=guild_id,
+            human_message=human_message,
```
jmelloy, line 928: The exception handler in `_drain_human_queue` swallows the error and continues to the next queued turn, which is intentional. However if `_run_foreman_ai` raises *and* that turn had enqueued further human messages into `_human_queues[lock_key]`, those will still be processed by the outer `while queue:` loop — good. But consider whether the user whose message failed should receive any error feedback; currently the failure is only logged server-side with no signal back to the originating user.

backend/ws_handlers.py
```diff
@@ -201,8 +207,16 @@ async def _trigger_foreman(
     # lifecycle, periodic-check, claude-auth) stay on the parent whole-guild
     # context.
     child = bool(task_id) and event in _CHILD_FOREMAN_EVENTS
+    is_human = event in _HUMAN_FOREMAN_EVENTS
```
jmelloy: `_HUMAN_FOREMAN_EVENTS` is defined as a module-level frozenset but `is_human` for the task follow-up route is set explicitly in `routes/tasks.py`. This split means the classification logic lives in two places. Consider centralising it (e.g. a helper `is_human_event(event)`) or at minimum adding a comment cross-referencing the routes path so future contributors don't accidentally add a human-originating event to `ws_handlers` without updating the frozenset.

jmelloy (commented):
- The core queuing logic is sound: FIFO drain with a bounded cap, correct lock ordering (drain happens while lock is still held), and automated events keep drop-if-busy semantics.
- A subtle re-entrancy bug exists in `_drain_human_queue`: new human messages can be enqueued to `_human_queues[lock_key]` during a drained turn's execution, but the function pops from a local `queue` reference, so those new arrivals are handled correctly — however the final `_human_queues.pop` at line 967 could silently discard messages enqueued by a recursive/concurrent path if the logic ever changes.
- The `_guild_locks.pop(lock_key, None)` in the `finally` block (after draining) means any human message that arrives between `_drain_human_queue` completing and `lock.release()` will see an unlocked state and run immediately rather than queue, which is the correct behavior but worth documenting.
- `_HUMAN_FOREMAN_EVENTS` only covers `chat`; the `create_task_followup` route correctly passes `is_human=True` directly, but there is no enforcement preventing a future caller from adding a new human-originating event to `ws_handlers.py` without adding it to the frozenset.
- Tests are thorough and cover the three main scenarios (queuing, automated-drop, and capacity cap), though there is no test for the error-handling path in `_drain_human_queue` (exception during a queued turn should not lose subsequent queued entries).

PR #875, opened 2026-07-10
fix(foreman): review_pr_internal uses guild-configured LLM client

### LINKED ISSUE #874: bug: "Review could not be generated by the AI agent" on PR reviews
## Problem

When the Foreman triggers a PR review (via the external `review_pr` tool hitting `agent.meyers.life`), the review is posted to GitHub but with the body:

> "Review could not be generated by the AI agent."

This has been observed on multiple recent PRs (e.g. #873). The review is submitted as a COMMENTED review by the `jmelloy` bot user, but with no actual analysis content.

## Expected behaviour

The external review agent should post a substantive code review with inline comments and a summary.

## Areas to investigate

1. **`review_pr` tool / `agent.meyers.life`** — Is the remote agent reachable? Is it returning an error that gets swallowed and replaced with the fallback message?
2. **GitHub PR review submission path** — Is the diff being fetched correctly? Is the Claude-powered review step failing silently?
3. **Fallback message origin** — Where in the codebase is the string "Review could not be generated by the AI agent." produced? Is it a hardcoded fallback in the backend or in the remote agent?
4. **Auth / rate limits** — Is a token expiry or rate limit causing the remote agent to fail?

## Steps to reproduce

- Trigger a `review_pr` call on any open PR and observe the resulting GitHub PR review body.

jmelloy:
## Summary
- `review_pr_internal` always called `make_anthropic_client()` directly, ignoring a guild's configured provider (e.g. Bedrock), which caused silent failures posted as a generic "Review could not be generated by the AI agent." message.
- Extracted the main foreman loop's client/model resolution logic (previously inline in `_run_foreman_ai`) into a shared `resolve_foreman_client()` helper in `runner.py`, and reused it in `review_pr_internal` so the review inference call honors the same provider/model/env-var overrides as every other foreman LLM call.
- The fallback summary now includes the exception type/message so failures are diagnosable directly from the posted GitHub review, instead of only in server logs.
- Checked the external `review_pr` path (agent.meyers.life A2A client) — it doesn't call `make_anthropic_client()`/`get_foreman_model()` at all, so no fix was needed there.
- Added `TestReviewPrInternal` coverage in `test_review_pr_tool.py` verifying the guild-resolved client/model is used and that LLM failures surface in the posted review's summary.

Fixes #874.

## Test plan
- [x] `python3 -m pytest tests/test_review_pr_tool.py -q` — 18 passed
- [x] `python3 -m pytest tests/ -q -k "foreman or runner or review"` — 324 passed
- [x] Full suite: `python3 -m pytest tests/ -q` — 1003 passed, 2 skipped


jmelloy (commented):

PR #882, opened 2026-07-11
feat: worker pauses reconnects on macOS sleep/wake

### LINKED ISSUE #881: feat: worker pauses reconnects on macOS sleep, resumes on wake
## Problem

The worker process keeps trying to reconnect to the Foreman WebSocket even when the laptop is asleep. This causes spurious connect/disconnect cycles that appear in the logs and waste resources.

## Solution

Hook into macOS power management notifications so the worker knows when the system is sleeping vs. awake:

- On **WillSleep**: set a sleep flag, suppress reconnect attempts, send a clean disconnect if connected
- On **DidWake**: clear the flag, wait a short delay for the network stack to come up, then trigger a single reconnect

## Implementation

- Add a `SystemSleepMonitor` class (macOS-only) using `pyobjc` / `IOKit` or `NSWorkspace` notifications
- Hook it into the worker's WebSocket reconnect loop: before each reconnect attempt, check the sleep flag and skip/delay if sleeping or just woke
- Graceful degradation: if `pyobjc` is not installed or the platform is not macOS, fall back to current behaviour silently

## Acceptance criteria

- [ ] Worker does not attempt to reconnect while the laptop is asleep
- [ ] Worker reconnects cleanly after wake, with a short delay to let networking come up
- [ ] No-op on Linux / non-macOS platforms
- [ ] pyobjc absence handled gracefully (optional dependency)

jmelloy:
## Summary
- Add `SystemSleepMonitor` (`worker/pioneer_worker/sleep_monitor.py`), a macOS-only background-thread monitor built on `NSWorkspace` sleep/wake notifications, exposing `is_sleeping` plus `on_sleep`/`on_wake` callbacks
- Wire `WSClient` to skip reconnect attempts (quietly, no log spam) while `is_sleeping` is true
- Wire `Worker` to close the WS connection on sleep and trigger a single reconnect ~2s after wake (grace period handled inside the monitor)
- Graceful degradation: if `pyobjc`/`AppKit` isn't installed, or the platform isn't macOS, `SystemSleepMonitor` is a no-op stub (`is_sleeping` always `False`) — behavior on Linux and elsewhere is unchanged
- Add optional `macos` extra in `cli/pyproject.toml` for `pyobjc-framework-Cocoa` (not a required dependency)
- Add `worker/tests/test_sleep_monitor.py`

Closes #881

## Test plan
- [x] `ruff check` clean
- [x] `pytest worker/tests/test_sleep_monitor.py` passes (3 tests)
- [x] Full worker test suite passes (262 tests)

### DIFF
--- a/worker/pioneer_worker/sleep_monitor.py
+++ b/worker/pioneer_worker/sleep_monitor.py
@@ -1,0 +1,137 @@
+"""macOS sleep/wake awareness for the worker's WebSocket reconnect loop.
+
+Wraps ``NSWorkspace`` sleep/wake notifications so the WS reconnect loop can
+avoid hammering a dead network link while the laptop is asleep, and give the
+network stack a couple of seconds to come back up after wake before the first
+post-wake retry. No-op stub everywhere else (non-macOS, or ``pyobjc`` not
+installed): ``start()``/``stop()`` do nothing and ``is_sleeping`` is always
+False.
+"""
+
+from __future__ import annotations
+
+import logging
+import threading
+import time
+from collections.abc import Callable
+
+logger = logging.getLogger(__name__)
+
+try:
+    import AppKit
+    import objc
+
+    _HAS_APPKIT = True
+except ImportError:
+    _HAS_APPKIT = False
+
+if _HAS_APPKIT:
+
+    class _SleepWakeObserver(AppKit.NSObject):
+        """Thin NSObject target that forwards NSWorkspace notifications."""
+
+        def initWithMonitor_(self, monitor: SystemSleepMonitor):
+            self = objc.super(_SleepWakeObserver, self).init()
+            if self is None:
+                return None
+            self._monitor = monitor
+            return self
+
+        def willSleep_(self, notification) -> None:
+            self._monitor._handle_will_sleep()
+
+        def didWake_(self, notification) -> None:
+            self._monitor._handle_did_wake()
+
+
+class SystemSleepMonitor:
+    """Tracks macOS sleep/wake state on a background thread.
+
+    ``is_sleeping`` flips True on ``NSWorkspaceWillSleepNotification`` and
+    stays True through a short grace period after
+    ``NSWorkspaceDidWakeNotification`` fires, so a caller polling
+    ``is_sleeping`` won't race a reconnect attempt against a network
+    interface that hasn't come back up yet.
+
+    ``on_sleep``/``on_wake`` callbacks run on the monitor's background
+    thread — callers driving an asyncio loop must hop back onto it
+    themselves (e.g. via ``call_soon_threadsafe``/``run_coroutine_threadsafe``).
+    """
+
+    def __init__(
+        self,
+        *,
+        on_sleep: Callable[[], None] | None = None,
+        on_wake: Callable[[], None] | None = None,
+        wake_grace_seconds: float = 2.0,
+    ) -> None:
+        self.on_sleep = on_sleep
+        self.on_wake = on_wake
+        self._wake_grace_seconds = wake_grace_seconds
+        self._is_sleeping = False
+        self._thread: threading.Thread | None = None
+        self._observer = None
+        self._stop_event = threading.Event()
+
+    @property
+    def is_sleeping(self) -> bool:
+        return self._is_sleeping
+
+    def start(self) -> None:
+        if not _HAS_APPKIT:
+            return
+        self._stop_event.clear()
+        self._thread = threading.Thread(target=self._run, name="sleep-monitor", daemon=True)
+        self._thread.start()
+
+    def stop(self) -> None:
+        if self._thread is None:
+            return
+        self._stop_event.set()
+        self._thread.join(timeout=2.0)
+        self._thread = None
+        self._observer = None
+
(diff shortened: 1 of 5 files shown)


worker/pioneer_worker/sleep_monitor.py
```diff
@@ -0,0 +1,137 @@
...
+        logger.info("System sleep detected — pausing WS reconnects")
+        if self.on_sleep is not None:
+            try:
+                self.on_sleep()
+            except Exception:
+                logger.exception("on_sleep callback raised")
+
+    def _handle_did_wake(self) -> None:
+        # Debounce on a throwaway thread so the run loop thread stays free to
+        # keep pumping notifications during the grace period.
```
jmelloy, line 131: `_is_sleeping` is set to `False` on the `sleep-monitor-wake` daemon thread while `WSClient.connect()` reads it on the asyncio event loop thread. CPython's GIL makes simple bool assignment safe in practice, but this is an undocumented implementation detail. Consider using `threading.Event` or `asyncio.Event` (with `loop.call_soon_threadsafe`) to communicate state changes safely and explicitly.

worker/pioneer_worker/worker.py
```diff
@@ -912,6 +917,20 @@ def _stop_control_api(self) -> None:
...
+    # These run on SystemSleepMonitor's background thread (see sleep_monitor.py),
+    # so they hop onto the worker's event loop rather than touching asyncio state
+    # directly.
+    def _on_system_sleep(self) -> None:
+        if self._loop is not None:
+            asyncio.run_coroutine_threadsafe(self.ws.close(), self._loop)
+
+    def _on_system_wake(self) -> None:
+        # SystemSleepMonitor already waited out its wake grace period before
+        # firing this, so a single reconnect here is safe.
```
jmelloy: `asyncio.run_coroutine_threadsafe` returns a `concurrent.futures.Future` whose result is never checked. Exceptions from `ws.close()` or `ws.connect()` will be silently swallowed. Store the future and add a done-callback that logs any exception, e.g. `fut.add_done_callback(lambda f: f.exception() and logger.error(...))`.

worker/pioneer_worker/worker.py
```diff
@@ -912,6 +917,20 @@ def _stop_control_api(self) -> None:
...
+        if self._loop is not None:
+            asyncio.run_coroutine_threadsafe(self.ws.connect(), self._loop)
+
```
jmelloy, line 938: If the single `ws.connect()` triggered here fails (network not fully recovered despite the 2 s grace), the worker will stay disconnected indefinitely — the internal reconnect loop in `WSClient.connect()` will not resume on its own because no code re-enters it. Consider letting `WSClient`'s own reconnect loop handle post-wake reconnects naturally (i.e., omit `_on_system_wake` from `Worker` and rely solely on the `_is_system_sleeping` gate in `WSClient.connect()` to unblock the loop).

worker/pioneer_worker/sleep_monitor.py
```diff
@@ -0,0 +1,137 @@
...
+            observer, "willSleep:", AppKit.NSWorkspaceWillSleepNotification, None
+        )
+        center.addObserver_selector_name_object_(
+            observer, "didWake:", AppKit.NSWorkspaceDidWakeNotification, None
+        )
+        self._observer = observer
+        run_loop = AppKit.NSRunLoop.currentRunLoop()
+        try:
+            # Pump the run loop in short slices so the notification center can
+            # deliver, while still checking _stop_event for clean shutdown.
```
jmelloy, line 114: The `stop()` method sets `_stop_event` and joins the thread, but there is no call to `removeObserver_` before joining — it only happens inside the `finally` block of `_run`. If the thread is blocked in `runUntilDate_` when `stop()` is called it will drain the current 1-second slice first, which is fine, but if the thread has already exited abnormally the observer will leak. This is low-risk but worth a comment.

jmelloy (commented):
- Overall design is clean: the no-op stub pattern ensures backward compatibility on non-macOS platforms, and the `TYPE_CHECKING` guard avoids a hard import cycle between `WSClient` and `SystemSleepMonitor`.
- **Race condition**: `_is_sleeping` is written on the NSRunLoop background thread (via `_handle_will_sleep`/`_finish_wake`) and read on the asyncio event loop thread without any synchronization primitive, which can cause torn reads on platforms/CPython versions where bool assignment isn't atomic.
- **Wake reconnect reliability**: `_on_system_wake` in `worker.py` fires `ws.connect()` once, but if that attempt fails (e.g., network not fully up despite the 2 s grace), there is no retry — the worker will stay disconnected until the next scheduled reconnect cycle, which may never come if the reconnect loop itself exited.
- **Dangling futures**: `asyncio.run_coroutine_threadsafe` returns a `Future` that is never stored or awaited; exceptions from `ws.close()` or `ws.connect()` on wake will be silently swallowed.
- Tests are well-structured and correctly patch `_SLEEP_POLL_INTERVAL` to keep the async test fast; however, only the no-op stub path is covered — the live macOS NSWorkspace path has no test coverage, which is acceptable given the CI environment constraint but worth noting.

PR #887, opened 2026-07-11
fix(discord): show actual tool name for single-tool-call turns

### LINKED ISSUE #886: feat: show actual tool name in Discord when only one tool call in a turn
## Summary

When a Foreman turn contains only a single tool call, Discord currently displays a generic label like **"bash x 1"** instead of the actual tool name (e.g. `bash`, `read_file`, `gh`). The "friend" style already shows the real tool name — we should match that behaviour when there's only one entry.

## Expected behaviour

- **Single tool call in a turn** → show the actual tool name (e.g. `bash`, `write_file`, `gh pr review`)
- **Multiple tool calls** → current collapsing/grouping behaviour is fine

## Notes

Requested by @jmelloy via Discord. Likely lives in the Discord notifier's turn-formatting / tool-call summarisation logic.

jmelloy:
## Summary
- Turns with exactly one tool call now pass through their raw line (e.g. `▶ bash: pytest`) instead of collapsing into a generic `bash × 1` turn summary, matching the frontend's `LogList.vue` grouping which already skips grouping when `toolCallCount <= 1`.
- Turns with multiple tool calls keep the existing grouped `▶ Turn N: edit × 2, bash × 1` behavior unchanged.

## Test plan
- [x] Updated/added unit tests in `backend/tests/test_discord_notifier.py` covering single tool_use+tool_result, lone tool_use with no result, and zero-tool_use groups
- [x] `pytest tests/test_discord_notifier.py -k format_stream` passes
- [x] `ruff check` passes

Closes #886

### DIFF
--- a/backend/discord_notifier.py
+++ b/backend/discord_notifier.py
@@ -1049,22 +1049,25 @@ def _format_stream_entries(entries: list[_StreamEntry], start_turn: int) -> tupl
         nonlocal turn_number, group
         if not group:
             return
-        tool_use_count = sum(
-            1 for entry in group if (entry.detail or {}).get("toolType") == "tool_use"
-        )
-        if tool_use_count <= 1:
-            output.extend(entry.line for entry in group)
-            group = []
-            return
-        turn_number += 1
+        tool_use_count = 0
         counts: dict[str, int] = {}
         for entry in group:
             detail = entry.detail or {}
             if detail.get("toolType") == "tool_use":
+                tool_use_count += 1
                 # Lowercase intentionally, to match the frontend's formatTurnCounts
                 # (LogList.vue) so tool names group consistently regardless of case.
                 name = str(detail.get("name") or "tool").lower()
                 counts[name] = counts.get(name, 0) + 1
+        if tool_use_count <= 1:
+            # Turn counter is intentionally NOT incremented for single-tool-call
+            # groups (matching the frontend, which also skips grouping these), so
+            # turn numbering may skip a number when a single-tool group sits
+            # between multi-tool groups.
+            output.extend(entry.line for entry in group)
+            group = []
+            return
+        turn_number += 1
         parts = ", ".join(f"{name} × {count}" for name, count in counts.items())
         output.append(f"▶ Turn {turn_number}: {parts or 'tool activity'}")
         group = []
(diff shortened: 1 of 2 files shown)


backend/discord_notifier.py
```diff
@@ -1046,6 +1049,13 @@ def flush_group() -> None:
         nonlocal turn_number, group
         if not group:
             return
+        tool_use_count = sum(
```
jmelloy: The `tool_use_count` generator iterates `group` once per flush. For very large groups this is fine, but note that the subsequent `for entry in group` loop in the multi-tool path also iterates again. If `group` could ever be large, a single pass building `tool_use_count` and `counts` simultaneously would be more efficient. Not a bug, just worth noting.

backend/discord_notifier.py
```diff
@@ -1046,6 +1049,13 @@ def flush_group() -> None:
...
+            1 for entry in group if (entry.detail or {}).get("toolType") == "tool_use"
+        )
+        if tool_use_count <= 1:
```
jmelloy: When `tool_use_count <= 1` the turn counter is NOT incremented, so turn numbering skips these groups entirely. This matches frontend behavior but means a sequence of [single-tool group, multi-tool group, multi-tool group] would produce Turn 1 and Turn 2 with no Turn 0 — which could confuse users correlating Discord threads with logs. Consider adding a brief inline comment clarifying this is intentional.

backend/tests/test_discord_notifier.py
```diff
@@ -1395,16 +1395,59 @@ def test_format_stream_entries_passes_through_non_tool_lines():
...
+    ]
+
+    content, turn_count = discord_notifier._format_stream_entries(entries, 0)
+
+    assert content == "▶ bash: pytest -k test_foo\n  → 3 passed"
+    assert turn_count == 0
+
+
+def test_format_stream_entries_single_tool_call_with_no_result():
+    """A lone tool_use with no trailing tool_result still passes through raw."""
```
jmelloy: This test name says 'with no result' but the scenario (a `tool_use` with no paired `tool_result`) is an unusual/malformed stream state. It's worth adding a comment noting whether this is an expected production scenario or a defensive edge case, so future maintainers understand the intent.

jmelloy (commented):
- The core logic change is small and correct: groups with `tool_use_count <= 1` now pass through raw lines instead of generating a generic summary, matching the frontend's `LogList.vue` behavior.
- The turn counter is intentionally not incremented for single-tool-call groups, which is consistent with the frontend grouping but means turn numbers may have gaps if batches mix single and multi-tool groups — this should be documented or verified as intentional.
- The existing `test_format_stream_entries_numbers_turns_across_flushes` test was correctly updated to use multi-tool batches so it still exercises the numbering logic; the fix didn't silently break the test by changing semantics.
- The `tool_use_count` check uses `(entry.detail or {}).get('toolType') == 'tool_use'` which is consistent with how the rest of the function accesses detail, though a missing or malformed `detail` field will silently count as 0 rather than raising — acceptable defensive behavior here.
- New tests are well-scoped and cover the three meaningful edge cases (single use+result, lone use, zero use), giving good confidence in the boundary conditions.

PR #901, opened 2026-07-12
fix: review_pr_internal biases toward APPROVE, firm on real bugs

### LINKED ISSUE #899: fix: review_pr_internal should bias toward APPROVE and be firm-but-polite on real bugs
## Problem

`review_pr_internal` currently tends toward `REQUEST_CHANGES` even for minor nits, making it overly blocking. The desired tone is polite but firm — call out real bugs clearly, but don't block a merge over style preferences.

## Desired behaviour

1. **Bias toward APPROVE** — if the code is functionally correct and the issues are minor nits, submit `APPROVE` with inline comments noting the nits.
2. **COMMENT** for moderate concerns that don't block merging.
3. **REQUEST_CHANGES** only for genuine bugs, security issues, or logic errors that must be fixed before merge.
4. **Tone**: polite and constructive. Firm when flagging a real bug ("This will cause X under condition Y"), but not pedantic about style or naming.
5. Update the internal review prompt/instructions in `review_pr_internal` to reflect this policy.

jmelloy:
Fixes #899.

## Summary
- `review_pr_internal` previously required the caller to pass an explicit `action` and silently defaulted to `COMMENT` regardless of diff quality.
- The tool's internal diff-analysis LLM call now also returns a recommended `verdict`, used as the GitHub review event unless the caller passes an explicit `action` override.
- Verdict policy encoded in the analysis prompt, the tool schema, and the foreman system prompt:
  - **APPROVE** — functionally correct; issues are minor nits (style, naming, formatting) — note them inline but still approve.
  - **COMMENT** — moderate concerns (performance, clarity) that don't block merging.
  - **REQUEST_CHANGES** — reserved for genuine bugs, security issues, or logic errors; never for style preferences.
  - Tone: polite and constructive; firm and specific when flagging a real bug; never pedantic or blocking over style.

## Test plan
- [x] `pytest backend/tests/test_review_pr_tool.py` — 20 passed, including two new tests covering analysis-driven verdict selection and explicit-action override
- [x] `pytest backend/tests/test_foreman.py backend/tests/test_review_pr_tool.py` — 162 passed

### DIFF
--- a/backend/foreman/prompt.py
+++ b/backend/foreman/prompt.py
@@ -71,7 +71,9 @@ Always create_task first and finalize_task after — whether you use a worker or
 
 **Shallow/fallback review** (no worker available, or quick diff-only review):
 1. create_task(name="Review PR #N: <title>", phase="review", pr_number=N, pr_repo="owner/repo") → returns task_id
-2. review_pr_internal (or review_pr) — pass the PR details
+2. review_pr_internal (or review_pr) — pass the PR details. Omit `action` to let the tool pick
+   the verdict itself from its own diff analysis (see "Review action policy" below); only pass
+   `action` explicitly to override that judgement.
 3. finalize_task(task_id=<task_id from step 1>) — call after the review returns. \
    Use expires_in_seconds=86400 for error/failed reviews; omit (default 3 days) otherwise.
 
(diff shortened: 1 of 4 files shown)


jmelloy (commented):
- The core logic change is sound: deferring the GitHub review verdict to the LLM's own analysis rather than hardcoding `COMMENT` is a meaningful improvement and the override path is correctly preserved.
- The new `verdict` field in the LLM response is validated and falls back to `COMMENT` safely, which is good defensive coding.
- One functional gap: when `explicit_action` is invalid (not in the allowed set), it silently falls back to `None` and uses the LLM verdict — the caller receives no warning that their supplied `action` was ignored, which could mask misconfiguration bugs.
- The `logger.info` call is correctly moved to after the LLM response is parsed, so it can log both the explicit and recommended values — good placement.
- Test coverage for the two new behaviors (bias toward APPROVE and explicit override) is clear and well-structured; minor duplication between the two test fixtures could be reduced but is not a blocking concern.

PR #903, opened 2026-07-12
feat(foreman): quote triggering human message in created GitHub issues

### LINKED ISSUE #902: feat: include triggering human message as a quote in Foreman-created issues
## Summary

When the Foreman creates a GitHub issue (via `create_github_issue`), the issue body should include the human message that triggered its creation, appended as a blockquote at the bottom.

## Motivation

Issues created by the Foreman can lose context about *why* they were created. Including the original human message as a quote makes the issue self-contained and easier to understand without tracing back through Discord/chat history.

## Desired behaviour

The issue body should end with something like:

```
---
> **Triggered by:** @username
> When an issue was created by the foreman, it should include the human text that triggered it at the bottom as a quote
```

## Where to change

- Find where the Foreman constructs the `body` argument to `create_github_issue` (likely in the Foreman system prompt instructions or in the tool-call logic).
- When the Foreman is about to call `create_github_issue`, it should append the triggering human message as a Markdown blockquote to the issue body.
- The attribution should include the Discord username if available.

## Acceptance criteria

- [ ] New Foreman-created issues include the triggering message as a blockquote at the bottom.
- [ ] The quote is clearly attributed (e.g. `> **Triggered by:** @username`).
- [ ] Existing issue body content is unaffected — the quote is appended, not prepended.

jmelloy:
## Summary
- The Foreman's system prompt now instructs it to append the human message that triggered `create_github_issue` as an attributed Markdown blockquote (`> **Triggered by:** @username`) at the bottom of the issue body, after a `---` separator.
- Existing body content is preserved — the quote is appended, never prepended or altered.
- Attribution pulls the username from the `[Discord] Discord user @username:` prefix when present; falls back to a generic "a human message" phrasing otherwise.

## Test plan
- [x] `pytest backend/tests/test_foreman.py -k github_issue` passes (tool behavior unchanged — this is a prompt-only change)
- [ ] Manually trigger an issue creation via the Foreman in a Discord channel and confirm the resulting GitHub issue body ends with the attributed blockquote

Fixes #902

### DIFF
--- a/backend/foreman/prompt.py
+++ b/backend/foreman/prompt.py
@@ -194,9 +194,11 @@ to the end of the `body` as a Markdown blockquote, after a `---` separator, e.g.
 ```
 
 Never prepend or otherwise alter the rest of the body — only append this quote at the
-bottom. Pull the username from the triggering message's `[Discord] Discord user @username:`
-prefix if present; otherwise omit the `@username` and just write `**Triggered by:** a human
-message` (or similar) before the quoted text.
+bottom. If `body` already ends with a `---` separator, don't add a second one — reuse it.
+Pull the username from the triggering message's `[Discord] Discord user @username:` prefix
+if present; otherwise write `**Triggered by:** unknown`. Truncate the quoted message text
+to 500 characters (with a trailing `...` if cut off) so long or multiline Discord messages
+don't bloat the issue body.
 
 ## Checking task progress
 Use get_task_status to verify a task is making progress — it returns the current state,
--- a/backend/foreman/prompt.py
+++ b/backend/foreman/prompt.py
@@ -184,6 +184,20 @@ immediately — don't treat issue creation as a separate round-trip. Pass issue_
 issue_repo to both create_task and assign_task so the worker's PR references the issue
 automatically and the task groups under the issue in the sidebar.
 
+When you call create_github_issue, append the human message that triggered its creation
+to the end of the `body` as a Markdown blockquote, after a `---` separator, e.g.:
+
+```
+---
+> **Triggered by:** @username
+> <the human's original message text>
+```
+
+Never prepend or otherwise alter the rest of the body — only append this quote at the
+bottom. Pull the username from the triggering message's `[Discord] Discord user @username:`
+prefix if present; otherwise omit the `@username` and just write `**Triggered by:** a human
+message` (or similar) before the quoted text.
+
 ## Checking task progress
 Use get_task_status to verify a task is making progress — it returns the current state,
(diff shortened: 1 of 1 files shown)


backend/foreman/prompt.py
```diff
@@ -184,6 +184,20 @@
...
+When you call create_github_issue, append the human message that triggered its creation
+to the end of the `body` as a Markdown blockquote, after a `---` separator, e.g.:
+
+```
+---
+> **Triggered by:** @username
+> <the human's original message text>
+```
+
+Never prepend or otherwise alter the rest of the body — only append this quote at the
```
jmelloy, line 196: The fallback text `**Triggered by:** a human message` is grammatically inconsistent with the `@username` variant (person vs. object noun). Consider `**Triggered by:** an anonymous user` or `**Triggered by:** unknown` for symmetry.

backend/foreman/prompt.py
```diff
@@ -184,6 +184,20 @@
 issue_repo to both create_task and assign_task so the worker's PR references the issue
 automatically and the task groups under the issue in the sidebar.
 
+When you call create_github_issue, append the human message that triggered its creation
+to the end of the `body` as a Markdown blockquote, after a `---` separator, e.g.:
+
+```
+---
```
jmelloy, line 191: If the incoming `body` already ends with a `---` separator (e.g. from a template), this instruction will produce a double horizontal rule. The prompt should instruct the model to check for and avoid a duplicate separator.

backend/foreman/prompt.py
```diff
@@ -184,6 +184,20 @@
...
+> **Triggered by:** @username
+> <the human's original message text>
```
jmelloy, line 193: The quoted message text is not sanitized or length-bounded at the prompt level. A very long or multiline Discord message could produce a malformed or excessively large issue body. Consider instructing the model to truncate the quote at a reasonable limit (e.g. 500 characters).

jmelloy (commented):
- This is a **prompt-only change** that instructs the LLM to append a blockquote attribution to GitHub issue bodies when calling `create_github_issue`.
- The instruction relies entirely on the LLM following natural-language rules correctly — there is no code-level enforcement, validation, or sanitization of the appended content.
- The fallback phrasing `**Triggered by:** a human message` is grammatically awkward and inconsistent with the `@username` form; a cleaner fallback like `**Triggered by:** an anonymous user` would read better.
- The separator `---` could conflict with existing issue body content that already ends with a horizontal rule, potentially producing double separators or malformed markdown.
- No test covers the actual prompt content or the shape of the generated `body` argument, meaning a future prompt regression would go undetected by the stated test plan.

PR #912, opened 2026-07-15
feat: GitHub App identity for automated comments/PR actions

### LINKED ISSUE #911: feat: comment on GitHub as a GitHub App identity (Pioneer Square [Melloy])
## Summary

Currently Pioneer Square comments on GitHub issues and PRs using the personal access token of the token owner. We want comments to appear from a distinct **GitHub App identity** (e.g. `pioneer-square-melloy[bot]`) so it's clear the comment came from the automation, not the repo owner.

## Approach

1. **Create a GitHub App** named `Pioneer Square [Melloy]` (or similar) under the GitHub account/org, with permissions:
   - Issues: Read & Write
   - Pull Requests: Read & Write
   - Contents: Read & Write
   - Metadata: Read

2. **Add config env vars**:
   - `GITHUB_APP_ID` — the numeric App ID
   - `GITHUB_APP_PRIVATE_KEY` — the PEM private key (or path to it)
   - `GITHUB_APP_INSTALLATION_ID` — the installation ID for the target repo/org

3. **Implement App-auth token generation** in the backend:
   - Generate a JWT signed with the private key (expires in ~10 min)
   - Exchange for an installation access token via `POST /app/installations/{installation_id}/access_tokens`
   - Cache the token and refresh before expiry

4. **Swap GitHub client initialization** to use the installation token when App env vars are present; fall back to the existing `GITHUB_TOKEN` PAT if not.

5. **Update `gh` CLI usage** (if any) to use the installation token via `GH_TOKEN` env var or `--auth` flag.

## Acceptance criteria
- When `GITHUB_APP_ID` + `GITHUB_APP_PRIVATE_KEY` + `GITHUB_APP_INSTALLATION_ID` are set, all GitHub comments/PR actions appear from the App bot identity
- When those vars are absent, behaviour falls back to the existing PAT (`GITHUB_TOKEN`)
- Existing tests still pass; add a test for the token-generation helper

---
> **Triggered by:** @jmelloy
> How do I make it? So the pioneer Square app comments on GitHub as Pioneer Square, possibly Pioneer Square bracket to Melloy bracket. I don't want it to come as the owner of the GitHub token.

jmelloy:
## Summary
- Adds `backend/github_app_auth.py`: generates a short-lived RS256 JWT (`generate_jwt`), exchanges it for an installation access token (`get_installation_token`/`_request_installation_token`), and caches the token (`get_app_installation_token`), refreshing when within 5 minutes of expiry
- `get_github_token()` prefers the GitHub App installation token when `GITHUB_APP_ID`, `GITHUB_APP_PRIVATE_KEY` (or `GITHUB_APP_PRIVATE_KEY_PATH`), and `GITHUB_APP_INSTALLATION_ID` are all set; otherwise falls back to `GITHUB_TOKEN` — no behavior change for existing deployments
- Wires the new helper into the places that read GitHub credentials:
  - `backend/foreman/tools.py` (`_guild_github_token`, for background/automated operations)
  - `backend/routes/auth.py` (`/github/token` endpoint used by workers)
  - `backend/routes/debug.py` (task debug timeline's live GitHub API lookup)
  - `backend/utils.py` (`build_spawn_worker_env`, which sets `GITHUB_TOKEN`/`PIONEER_GITHUB_TOKEN` in worker subprocess env — this is what the `gh` CLI reads inside workers)
- Uses only the already-present `httpx` and `cryptography` deps — no new dependencies

## Test plan
- [x] `pytest backend/tests/test_github_app_auth.py` — 9 tests covering JWT claims/signature, fallback to `GITHUB_TOKEN`, installation-token API call (mocked), caching behavior (no refetch within expiry window, refetch near expiry), and `GITHUB_APP_PRIVATE_KEY_PATH` support — all pass

Closes #911.

### DIFF
--- a/backend/github_app_auth.py
+++ b/backend/github_app_auth.py
@@ -1,0 +1,135 @@
+"""GitHub App authentication — installation-token generation and caching.
+
+Lets Pioneer Square comment on GitHub as a GitHub App identity (e.g.
+``pioneer-square-melloy[bot]``) instead of a personal access token, so
+automated comments/reviews are attributed to the automation rather than
+whichever human's PAT happens to be configured.
+
+Enabled by setting all three of ``GITHUB_APP_ID``, ``GITHUB_APP_PRIVATE_KEY``
+(or ``GITHUB_APP_PRIVATE_KEY_PATH``), and ``GITHUB_APP_INSTALLATION_ID``. When
+any is missing, :func:`get_github_token` falls back to the ``GITHUB_TOKEN``
+env var — existing deployments without a GitHub App configured see no change
+in behavior.
+"""
+
+from __future__ import annotations
+
+import base64
+import json
+import os
+import time
+from datetime import UTC, datetime, timedelta
+from pathlib import Path
+from typing import Any
+
+import httpx
+from cryptography.hazmat.primitives import hashes
+from cryptography.hazmat.primitives.asymmetric import padding
+from cryptography.hazmat.primitives.serialization import load_pem_private_key
+
+# ---------------------------------------------------------------------------
+# Config (read at call time from environment — see module docstring)
+# ---------------------------------------------------------------------------
+
+GITHUB_APP_ID = "GITHUB_APP_ID"
+GITHUB_APP_PRIVATE_KEY = "GITHUB_APP_PRIVATE_KEY"
+GITHUB_APP_PRIVATE_KEY_PATH = "GITHUB_APP_PRIVATE_KEY_PATH"
+GITHUB_APP_INSTALLATION_ID = "GITHUB_APP_INSTALLATION_ID"
+
+# Cached installation token, refreshed once within 5 minutes of expiry.
+_token_cache: dict[str, Any] = {"token": None, "expires_at": None}
+
+
+def _b64url_encode(data: bytes) -> str:
+    return base64.urlsafe_b64encode(data).rstrip(b"=").decode()
+
+
+def generate_jwt(app_id: str, private_key_pem: str) -> str:
+    """Generate a short-lived (10 min) RS256 JWT for GitHub App authentication.
+
+    Signed with the App's private key; ``iss`` is the App ID, per GitHub's
+    JWT-based App authentication scheme.
+    """
+    now = int(time.time())
+    header = {"alg": "RS256", "typ": "JWT"}
+    payload = {
+        "iat": now - 60,  # allow for clock drift, as GitHub's docs recommend
+        "exp": now + 10 * 60,
+        "iss": str(app_id),
+    }
+    header_b64 = _b64url_encode(json.dumps(header, separators=(",", ":")).encode())
+    payload_b64 = _b64url_encode(json.dumps(payload, separators=(",", ":")).encode())
+    signing_input = f"{header_b64}.{payload_b64}".encode()
+
+    private_key = load_pem_private_key(private_key_pem.encode(), password=None)
+    signature = private_key.sign(signing_input, padding.PKCS1v15(), hashes.SHA256())
+    return f"{header_b64}.{payload_b64}.{_b64url_encode(signature)}"
+
+
+def _request_installation_token(app_id: str, private_key_pem: str, installation_id: str) -> dict:
+    """POST to GitHub's installation access-token endpoint and return the parsed JSON body."""
+    jwt_token = generate_jwt(app_id, private_key_pem)
+    response = httpx.post(
+        f"https://api.github.com/app/installations/{installation_id}/access_tokens",
+        headers={
+            "Authorization": f"Bearer {jwt_token}",
+            "Accept": "application/vnd.github+json",
+        },
+        timeout=15,
+    )
+    response.raise_for_status()
+    return response.json()
+
+
+def get_installation_token(app_id: str, private_key_pem: str, installation_id: str) -> str:
+    """Exchange a signed App JWT for a short-lived installation access token."""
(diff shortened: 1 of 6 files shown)


backend/github_app_auth.py
```diff
@@ -0,0 +1,135 @@
...
+
+
+def _app_credentials() -> tuple[str, str, str] | None:
+    """Return (app_id, private_key_pem, installation_id) if all App env vars are set, else None."""
+    app_id = os.environ.get(GITHUB_APP_ID)
+    installation_id = os.environ.get(GITHUB_APP_INSTALLATION_ID)
+    private_key_pem = os.environ.get(GITHUB_APP_PRIVATE_KEY)
+    if not private_key_pem:
+        key_path = os.environ.get(GITHUB_APP_PRIVATE_KEY_PATH)
+        if key_path:
```
jmelloy, line 113: Reading the private key file (`Path(key_path).read_text()`) has no error handling. A missing path, permission error, or malformed PEM will raise an unhandled exception that propagates through `get_app_installation_token()` to callers. Consider wrapping in a try/except and logging a warning before returning `None`, consistent with the graceful-fallback intent of this module.

backend/github_app_auth.py
```diff
@@ -0,0 +1,135 @@
...
+            private_key_pem = Path(key_path).read_text()
+    if app_id and installation_id and private_key_pem:
+        return app_id, private_key_pem, installation_id
+    return None
+
+
+def get_app_installation_token() -> str | None:
+    """Return a cached GitHub App installation token, or None if the App isn't configured.
```
jmelloy, line 137: The token cache is mutated without any lock. In an async context with concurrent coroutines both finding the token expired, multiple simultaneous calls to `_request_installation_token` will race. Consider using `asyncio.Lock` or accepting the benign double-fetch, but document the choice explicitly.

backend/utils.py
```diff
@@ -264,10 +266,11 @@ def build_spawn_worker_env(
     public_url = source_env.get("FRONTEND_URL", "").rstrip("/")
     if public_url:
         env["PIONEER_FRONTEND_URL"] = public_url
-    gh_token = source_env.get("GITHUB_TOKEN", "")
+    gh_token = get_github_token(fallback=source_env.get("GITHUB_TOKEN", ""))
```
jmelloy: `get_github_token()` may make a synchronous HTTP call here (via `httpx.post`). If the GitHub API is slow or throws (e.g. network error, 500), `build_spawn_worker_env` will raise unexpectedly. Wrap the call in a try/except and fall back to `source_env.get("GITHUB_TOKEN", "")` on failure.

backend/routes/auth.py
```diff
@@ -88,10 +89,19 @@ async def get_github_token(
     _principal: str = Depends(require_worker_or_member),
     db: AsyncSession = Depends(get_db_dep),
 ):
-    """Return the stored OAuth token for the guild's linked GitHub user. Used by workers.
+    """Return a GitHub token for the guild, for workers to use.
+
+    Prefers the GitHub App installation token when configured (so worker
+    actions like ``gh pr comment`` are attributed to the App identity);
+    otherwise falls back to the stored OAuth token for the guild's linked
+    GitHub user.
```
jmelloy, line 97: The hardcoded `"github-app[bot]"` username may not match the actual installed app's bot slug (e.g. `pioneer-square-melloy[bot]` per the PR description). If downstream code uses this username for display or attribution, it will be wrong. Consider deriving the username from the app metadata or making it configurable via env var.

backend/github_app_auth.py
```diff
@@ -0,0 +1,135 @@
...
+def get_github_token(fallback: str | None = None) -> str:
+    """Return the GitHub token to use for API/``gh`` CLI calls.
+
+    Prefers a GitHub App installation token when the App env vars are set;
+    otherwise returns *fallback* if given, else the ``GITHUB_TOKEN`` env var.
+    """
+    token = get_app_installation_token()
+    if token:
+        return token
+    if fallback is not None:
```
jmelloy, line 170: `get_github_token` returns `""` (empty string) when no app token and no `GITHUB_TOKEN` env var. Callers in `debug.py` convert this with `or None` (`get_github_token() or None`), but callers in `utils.py` do not — an empty string will be falsy for the `if gh_token:` check, so that's fine. Still, returning `None` instead of `""` when unconfigured would be more explicit and consistent with `get_app_installation_token()`.

jmelloy (commented):
- The new `github_app_auth.py` module is well-structured with clear separation between JWT generation, token exchange, and caching logic; tests cover the main code paths thoroughly.
- **Thread-safety concern**: the module-level `_token_cache` dict is mutated without any lock, so concurrent async/threaded requests near token expiry could trigger multiple simultaneous refresh calls (thundering herd). This is low-risk for most deployments but worth noting.
- The `get_github_token()` call in `build_spawn_worker_env` (utils.py) is synchronous and makes a live HTTP call during what may be a synchronous code path — if the GitHub API is slow or unavailable, worker spawning will block or fail. The function should handle exceptions gracefully.
- The hardcoded username `"github-app[bot]"` in both `tools.py` and `auth.py` may not match the actual bot slug for this specific GitHub App installation; the real bot username follows the pattern `<app-slug>[bot]` and differs per app. This could cause attribution mismatches in UI display.
- Private key read from `GITHUB_APP_PRIVATE_KEY_PATH` in `_app_credentials()` has no error handling — a missing or unreadable file will raise an unhandled exception that propagates up through `get_app_installation_token()` and silently breaks token retrieval in callers that don't expect exceptions.

PR #915, opened 2026-07-16
feat: API latency logging + Amazon Bedrock (Nova/K2)

jmelloy:
Implements API latency logging and Amazon Bedrock provider support (Nova/K2 models).

Closes #914.

### DIFF
--- a/backend/foreman/llm.py
+++ b/backend/foreman/llm.py
@@ -30,6 +30,16 @@ from uuid import uuid4
 import httpx
 
 try:
+    # Absolute import works when backend/ itself is on sys.path (embedded
+    # backend, worker). The standalone foreman proxy only prepends the repo
+    # root (see foreman-proxy/pioneer_foreman/config.py), under which this
+    # module is reachable as backend.foreman.llm and util/ has no top-level
+    # name of its own — fall back to the fully-qualified path there.
+    from util.api_latency import track_api_call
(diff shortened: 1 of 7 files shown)


jmelloy:
Declining this suggestion — it contradicts the intent of issue #914, which requires: "Workers can be configured with `provider=bedrock` and model IDs matching Nova or K2" (acceptance criteria), and was triggered by "prep for running Quinen or Kimmy K2 via Amazon bedrock for the format" (issue body). Routing K2 through the existing OpenAI-compatible client against Moonshot's own API endpoint instead of Bedrock would mean K2 no longer runs via `provider=bedrock` at all, which is exactly what #914 asks for.

It's also not technically viable as described: Bedrock's Runtime Converse API is not an OpenAI-compatible wire format (there's no `/chat/completions` endpoint to point `call_openai_compatible` at) — `foreman/providers/bedrock.py` already implements the required Anthropic-Messages↔Converse translation for both Amazon Nova and Kimi K2, which is the only way to serve K2 through Bedrock.

Keeping Kimi K2 on the native Bedrock Converse path in `foreman/providers/bedrock.py`, as required by issue #914. The existing OpenAI-compatible client (`foreman/llm.py:call_openai_compatible`) remains available for operators who want to reach Moonshot's own hosted API directly, but that's a separate, unrelated code path from `provider=bedrock`.

PR #919, opened 2026-07-16
feat: Terraform docker-compose → ECS with S3 + Bedrock

jmelloy:
## Summary
Adds a modular Terraform configuration under `terraform/` to migrate Pioneer Square from `docker-compose` on a Linux server to AWS ECS Fargate, per #918.

- **Networking** (`vpc.tf`) — VPC, public/private subnets across 2 AZs, IGW, NAT Gateway, security groups
- **ECS** (`ecs.tf`) — Fargate cluster, task definitions per docker-compose service, services with health checks
- **ECR** (`ecr.tf`) — one repository per service image
- **ALB** (`alb.tf`) — HTTPS listener (ACM placeholder), HTTP→HTTPS redirect, target groups per service
- **S3** (`s3.tf`) — bucket for assets/uploads/logs, versioning + SSE, IAM policy for the ECS task role
- **Bedrock** — IAM policy on the task role for `InvokeModel`, `InvokeModelWithResponseStream`, `ListFoundationModels`
- **Secrets** (`ssm.tf`) — SSM Parameter Store placeholders for DB password, API keys, bearer tokens, referenced via ECS `secrets` block
- **RDS** (`rds.tf`) — managed Postgres option, per the issue's ask to note DB strategy
- `variables.tf` / `outputs.tf` — no hardcoded account IDs/regions; outputs expose ALB DNS, ECR URLs, S3 bucket names, cluster name
- `README.md` — prerequisites, backend setup, first-time apply, updating a service image

`terraform fmt -recursive` is clean and `terraform validate` passes.

Closes #918

## Test plan
- [x] `terraform fmt -recursive -check`
- [x] `terraform validate`
- [ ] `terraform plan` against a real AWS account (requires credentials/backend not available in this environment)


jmelloy:
Declining this suggestion — it contradicts the intent of issue #918, which requires migrating Pioneer Square to **"AWS ECS (Fargate)"** (issue title/summary) and explicitly lists an **"ECS Fargate cluster"** as a required resource in the original task spec. Switching the launch type to EC2 with a t3.small Auto Scaling Group would invert that requirement rather than fulfill it.

Keeping Fargate as required by issue #918. The current implementation (Fargate cluster, `awsvpc` network mode, `requires_compatibilities = ["FARGATE"]`) is correct and is not being changed.

If cost is a concern, that's worth discussing as a separate follow-up against the issue's stated goal (e.g., adjusting Fargate task CPU/memory sizing, Fargate Spot, or reducing desired_count) rather than replacing the launch type outright.

jmelloy (commented):
- Overall this is a well-structured, thoughtful Terraform migration with good documentation, sensible defaults, and correct use of SSM SecureString for secrets rather than plaintext env vars.
- The worker dispatch IAM policy uses `resources = ["*"]` for `ecs:RunTask`/`ecs:StopTask`/`ecs:DescribeTasks` — the cluster condition helps but ECS does not honor the `ecs:cluster` condition key on `RunTask` resource scoping at the task-definition ARN level, so in practice these actions are broader than intended.
- The ECR lifecycle policy uses `tagStatus = "any"` which will also expire untagged/digest-only images aggressively; for an immutable-tag repo this is fine, but it will also sweep the `:latest` placeholder before images exist if the count drops below 20.
- The `kms:Decrypt` policy for SSM SecureStrings allows any KMS key in the account (`key/*`) scoped only by `kms:ViaService`; this is wider than necessary — ideally scope to the specific key ARN used by SSM, or at minimum note this as a known trade-off.
- The README and inline comments are thorough and honest about known gaps (worker dispatch, Docker socket migration, Bedrock model access console requirement), which significantly reduces onboarding risk.

terraform/iam.tf
```diff
@@ -0,0 +1,310 @@
...
+    condition {
+      test     = "StringEquals"
+      variable = "iam:PassedToService"
+      values   = ["ecs-tasks.amazonaws.com"]
+    }
+  }
+}
+
+resource "aws_iam_role_policy" "ecs_task_worker_dispatch" {
+  name   = "${local.name_prefix}-ecs-task-worker-dispatch"
```
jmelloy, line 184: The `ecs:cluster` condition key is not enforced by AWS for `ecs:RunTask` when the resource is `"*"` — AWS evaluates it as an additional constraint on the task definition ARN resource, not the cluster. To tighten this, also add the cluster ARN as an explicit resource for the `ecs:DescribeTasks` / `ecs:StopTask` statements and accept that `ecs:RunTask` itself cannot be scoped below `"*"` by cluster alone. Document this gap so it isn't mistaken for a tight policy.

terraform/iam.tf
```diff
@@ -0,0 +1,310 @@
...
+      variable = "kms:ViaService"
+      values   = ["ssm.${local.region}.amazonaws.com"]
+    }
+  }
+}
+
+resource "aws_iam_role_policy" "ecs_task_execution_secrets" {
+  name   = "${local.name_prefix}-ecs-task-execution-secrets"
+  role   = aws_iam_role.ecs_task_execution.id
+  policy = data.aws_iam_policy_document.ecs_task_execution_secrets.json
```
jmelloy, line 61: The KMS Decrypt permission is scoped to `key/*` (all keys in the account) with only a `kms:ViaService` condition. If there are other SSM parameters encrypted with different KMS keys in this account this is fine operationally, but it violates least-privilege. Consider outputting the SSM default KMS key ARN via a `data "aws_kms_key"` lookup and scoping to that specific key ARN.

terraform/ecr.tf
```diff
@@ -0,0 +1,43 @@
...
+}
+
+resource "aws_ecr_lifecycle_policy" "service" {
+  for_each = aws_ecr_repository.service
+
+  repository = each.value.name
+
+  policy = jsonencode({
+    rules = [
+      {
```
jmelloy, line 31: Using `tagStatus = "any"` on an IMMUTABLE-tag repo means the lifecycle policy counts all images (including the initial placeholder push and any untagged layers) toward the 20-image limit. This is usually fine, but if bootstrap pushes tagged images and you want to guarantee the last 20 *tagged* images are kept, switch `tagStatus` to `"tagged"` with `tagPrefixList = ["*"]` (or separate rules for tagged vs untagged).

terraform/ecs.tf
```diff
@@ -0,0 +1,274 @@
...
+}
+
+resource "aws_ecs_service" "backend" {
+  name            = "${local.name_prefix}-backend"
+  cluster         = aws_ecs_cluster.main.id
+  task_definition = aws_ecs_task_definition.backend.arn
+  desired_count   = var.backend_desired_count
+  launch_type     = "FARGATE"
+
+  network_configuration {
```
jmelloy, line 108: The `WORKER_BACKEND_URL` is set to `http://localhost:<port>` which implies the worker and backend share a task (sidecar pattern). But `worker` has its own separate task definition launched via `ecs:RunTask`. If the worker needs to call back to the backend it should use the ALB DNS name or a service-discovery endpoint, not localhost.

jmelloy (commented):
- The overall architecture is well-structured and the README is exceptionally thorough — the worker-dispatch gap is honestly documented and the Bedrock model-access caveat is called out in multiple places.
- Security posture is solid: SSM SecureString for all secrets, OIDC instead of static keys, ECR image scanning, IMMUTABLE tags, and S3 SSE — no credentials are hardcoded.
- One moderate concern: the `kms:Decrypt` grant in the execution role uses a wildcard key ARN (`key/*`) with only a `kms:ViaService` condition; if a customer-managed KMS key is ever used for non-SSM purposes in the same account, this could be broader than intended — worth tightening to a specific key ARN once the KMS key is known.
- The ALB health-check path is `/` with `200-399`; if the backend returns redirects or auth-required responses on `/` before it's fully up, health checks will pass prematurely. A dedicated `/healthz` endpoint would be safer.
- ECR lifecycle policy uses `tagStatus: any` which will expire the `latest` tag and any deploy-pinned SHA tags after 20 images — consider separating tagged vs. untagged rules to avoid expiring images that are actively referenced by running tasks.

terraform/iam.tf
```diff
@@ -0,0 +1,318 @@
...
+      test     = "StringEquals"
+      variable = "kms:ViaService"
+      values   = ["ssm.${local.region}.amazonaws.com"]
+    }
+  }
+}
+
+resource "aws_iam_role_policy" "ecs_task_execution_secrets" {
+  name   = "${local.name_prefix}-ecs-task-execution-secrets"
+  role   = aws_iam_role.ecs_task_execution.id
```
jmelloy, line 60: The `kms:Decrypt` resource is `key/*` (all KMS keys in the account/region), scoped only by the `kms:ViaService` condition. This is functional but overly broad — once the specific CMK used for SSM SecureStrings is known, tighten this to its exact ARN to follow least-privilege.

terraform/alb.tf
```diff
@@ -0,0 +1,134 @@
...
+  name        = "${local.name_prefix}-backend"
+  port        = var.backend_container_port
+  protocol    = "HTTP"
+  vpc_id      = aws_vpc.main.id
+  target_type = "ip"
+
+  health_check {
+    path                = "/"
+    protocol            = "HTTP"
+    matcher             = "200-399"
```
jmelloy, line 27: Health-check path `/` with matcher `200-399` will accept 301/302 redirects as healthy. If the backend redirects unauthenticated requests on `/`, tasks will appear healthy before the app is actually ready. Prefer a dedicated `GET /healthz` (or similar) that returns 200 only when the app is up and the DB migration has completed.

terraform/ecr.tf
```diff
@@ -0,0 +1,43 @@
...
+        rulePriority = 1
+        description  = "Keep last 20 tagged images"
```
jmelloy, line 33: `tagStatus: any` means the lifecycle rule counts *all* images (including the currently-deployed tagged image) toward the 20-image limit. If you push frequently, a running task's image can be expired while it's still in use. Consider splitting into two rules: one for `tagStatus: untagged` (expire after N days) and one for `tagStatus: tagged` keeping the last 20, or use a `tagPrefixList` to protect deploy tags.

terraform/ecs.tf
```diff
@@ -0,0 +1,274 @@
...
+  container_definitions = jsonencode([
+    {
+      name      = "foreman"
+      image     = "${aws_ecr_repository.service["foreman"].repository_url}:${var.container_image_tag}"
+      essential = true
+      command   = ["pioneer", "foreman"]
+
+      environment = [
+        # wss:// when an HTTPS listener exists: the HTTP listener 301-redirects
+        # everything to HTTPS in that case (see alb.tf), and a plain ws://
```
jmelloy, line 156: The foreman's `BACKEND_WS_URL` is set to the ALB DNS name with `ws://` or `wss://` based on `local.has_certificate`. During the bootstrap window (before a cert is attached) this is `ws://` over plain HTTP, which is fine for initial testing — but note that if `var.domain_name` is set without `var.route53_zone_id`, the ACM cert is created but never validated, so `has_certificate` will be true while the HTTPS listener will hang waiting for validation. Worth adding an output or README note warning about this partial state.

Issue #945, opened 2026-07-19
Strip `<think>` blocks from Kimi model responses


jmelloy:
## Problem
Kimi model outputs include `<think>` tags that surround reasoning/thinking blocks. For example:
```
<think> chains of thought...</think>
Generated content here
```

Currently, the foreman/worker code appears to be preserving this internal reasoning, which bleeds into the actual output seen by users.

## Expected Behavior
Everything before and including `</think>` should be stripped from Kimi responses, leaving only the generated content.

## Likely Location
The parsing likely happens in one of these areas:
- LLM provider shim that handles Kimi responses
- Worker response processing
- Foreman output formatting

## Acceptance Criteria
- [ ] Kimi `<think>` blocks are stripped before content is passed downstream
- [ ] Other models without this pattern are unaffected
- [ ] Tests verify the stripping behavior

---
> **Triggered by:** @jmelloy
> "I see some tags like </think> in the kimi output - I suspect you can to strip everything before that but would like confirmation"

PR #957, opened 2026-07-20
Strip think blocks from Foreman responses before DB storage

jmelloy:
Fixes the issue where `<think>...</think>` blocks were being stored in foreman_turns.

- Strips think blocks from assistant responses before DB persistence
- Uses existing _THINK_BLOCK_RE-style pattern from bedrock.py
- Handles both string content and JSON text blocks gracefully

All 62 non-DB tests pass. DB tests error due to test environment limitation (no Postgres available in this sandbox).

### DIFF
--- a/backend/foreman/message_utils.py
+++ b/backend/foreman/message_utils.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 import copy
 import json
+import re
 from datetime import date, datetime
 from typing import Any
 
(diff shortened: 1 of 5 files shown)


jmelloy:
@jmelloy Yes — you were right, and it turned out to be a bigger deal than my original regex assumed. Confirmed against real `foreman_turns` data:

**1. Did the old regex handle content before `<thinking>`?** The pattern in question is `<think>` (not `<thinking>` — likely a typo in the question, matching issue #945's Kimi tag), and yes: `re.sub` with no `^` anchor matches `<think>...</think>` wherever it occurs, so text before the tag was always preserved. That part was never the bug.

**2. What the DB actually shows:** querying `foreman_turns` for `content_json ILIKE '%</think>%'` vs `'%<think>%'` on rows in the current environment:
- `</think>` (closing tag): **206 rows**
- `<think>` (opening tag): **11 rows** — and in every one of those 11, the literal `<think>` appears inside a *tool_use input* or *meta-discussion* string, never as a real opening delimiter in the assistant's own text.

So the real, dominant shape is: **reasoning text starts immediately at the beginning of the response with no opening tag at all, and only the closing `</think>` appears**, e.g.:

```
"I need to pass the description in assign_task call too:  </think>"
"Task created. Now I'll assign it to worker w-e1f5ab... </think>   "
```

This matches what the t-fgcgq4 investigation found: Kimi K2's chat template pre-fills the `<think>` token as part of the prompt, so it's never echoed back in the completion — only the closing `</think>` shows up in the actual text. `_THINK_BLOCK_RE = r"<think>.*?</think>"` requires *both* tags, so it was silently a no-op for the common case.

**Fix pushed:** both `message_utils._strip_think_text` (this PR's DB-storage guard) and `bedrock._strip_think_block` (the original, model-gated source) now also strip a lone `</think>` as closing an implicit reasoning span from the start of the text — but only when real content follows the tag, so a message that merely *mentions* `</think>` in prose can't get its whole body wiped out. Added tests reproducing the exact production shape for both layers.

PR #966, opened 2026-07-20
refactor(discord): collapse duplicate identity tables

jmelloy:
## Why

Traced a dropped foreman @-mention (`foreman-role mention from unlinked discord_user=... — ignoring`) back to there being *three* Discord identity-ish tables, two of which genuinely overlapped:

- `discord_account_links` — Discord ↔ PS user, created by `/connect-account`. Authoritative.
- `discord_users` — GitHub login ↔ Discord ID, hand-populated via `/api/discord-users`. Predates the link flow by ~18 commits, never removed.
- `discord_connect_tokens` — not a mapping at all; ephemeral single-use handshake state.

`discord_users` had drifted to zero rows while still being consulted as a fallback, so an account could be known to one table and not the other.

## What

**Drops `discord_users`** — model, the four `/api/discord-users` CRUD endpoints, route registration, and its test file. Both readers now go through `discord_account_links`:

- `discord_notifier._lookup_discord_user` joins to `users.github_login`. Case-insensitive, because the dropped table lowercased on write while `users.github_login` keeps GitHub's own casing. Uses `.first()` not `.one_or_none()` — the unique index is on `discord_user_id`, so one PS user can hold multiple links and `one_or_none()` would raise.
- `discord.router._resolve_identity` drops its legacy fallback branch.

**Renames `discord_connect_tokens` → `discord_pending_connects`** (`DiscordConnectToken` → `DiscordPendingConnect`) so it reads as transient state rather than a durable mapping. The migration also renames the pkey and unique constraint, which Postgres leaves stale through a bare `rename_table`.

## Migration safety

Backfills legacy rows into `discord_account_links` before dropping — matching on `lower(github_login)`, with `ON CONFLICT DO NOTHING` so an existing `/connect-account` link stays authoritative. Rows whose login matches no user have nowhere to attach and go with the table.

Verified against the test DB with three seeded cases:

| case | result |
|---|---|
| `mixedcase` legacy row vs `MixedCase` user | linked ✓ |
| Discord account already linked via `/connect-account` | existing link preserved, not overwritten ✓ |
| legacy row for an unknown login | dropped ✓ |

Downgrade round-trips (recreates the table empty — backfilled rows are indistinguishable from natively-created ones, so they aren't moved back; `mention_or_login` just degrades to plain `@login`).

## Testing

Full suite: **1100 passed, 2 skipped**. Ruff clean. Adds three tests for the new join (linked / case-insensitive / unlinked) — the previous tests mocked `_lookup_discord_user` wholesale, so the query itself had no coverage.

## Note for deploy

The backend image bakes in source rather than mounting it, so this needs a `docker compose build backend` before `alembic upgrade head` will see the new revision.

Also gitignores `*.sql.gz` — there's a 415MB `dump.sql.gz` sitting untracked in the repo root that `git add -A` will happily stage.


jmelloy:
Rebased onto `main` after #965 landed. Two collisions, both resolved:

- **Split alembic head.** #965 added `20260715_000000_add_bot_fields_to_users` with the same `down_revision` as this branch's migration, so the two sat side by side as separate heads. Re-parented this migration onto #965's; `alembic heads` is back to one, and the full chain applies cleanly from an empty database.
- **`_resolve_identity` conflict.** #965 added bot auto-provisioning to the same function this PR was trimming the legacy fallback out of. Code bodies merged without issue — only the docstring conflicted, now rewritten to describe both behaviours.

Bot users get a `users` row with no `github_login`, so they simply don't match the `mention_or_login` join — no interaction between the two changes.

Suite: 1109 passed, 1 failed. The failure is `test_discord_gateway.py::test_message_create_filters_bot_author`, which **fails on `main` at a20f5c8 independently of this branch** — #965 changed the gateway to let bot authors through so they can be auto-provisioned, but that test still asserts they're filtered out of the queue. Unrelated to this PR and left alone; worth a separate fix to decide whether the test or the filter is the stale one.

PR #1000, opened 2026-07-24
feat(foreman): scope child send_followup/redirect_task to own task

jmelloy:
## Summary
- Closes #997. `CHILD_FOREMAN_TOOLS` already exposed `send_followup` and `redirect_task` to per-task child Foreman contexts, but `_task_mutation_blocked()` in `backend/foreman/tools.py` only guarded against a *parent* run racing a child's own task — it did not stop a child context from calling these tools against a *different* task_id in the same guild.
- Added an explicit scope check: a per-task child context (`own_task_id` set) can now only mutate its own task; any other `task_id` is rejected with a task-scoped error message. Parent/whole-guild runs (`own_task_id=None`) are unaffected.
- Documented the new authorization boundary in `docs/foreman-per-task-context.md`.

## Test plan
- [x] `python -m pytest tests/test_foreman.py -k ChildForemanFollowupAndRedirect -v` — 5 new tests covering: child `send_followup` on own task (succeeds), child `redirect_task` on own task (succeeds), child `send_followup` on another task (rejected), child `redirect_task` on another task (rejected), parent `redirect_task` on any task (baseline, unaffected)
- [x] `python -m pytest tests/test_foreman.py` — full suite, 155 passed

### DIFF
--- a/backend/foreman/tools.py
+++ b/backend/foreman/tools.py
@@ -1317,20 +1317,29 @@ def _full_log_content(data_json: str | None) -> str | None:
 
 
 def _task_mutation_blocked(guild_id: str, task_id: str, own_task_id: str | None) -> str | None:
-    """Return a user-facing message if *task_id* is owned by a different in-flight
-    Foreman context right now, else None.
+    """Return a user-facing message if *task_id* cannot be mutated right now, else None.
 
     ``own_task_id`` is the task_id of the per-task child context this tool
     call is itself running in (None for a parent/whole-guild run). A run
-    never conflicts with its own per-task lock — this only fires when a
-    *different* context holds it, which in practice means a parent run
-    (periodic-check or human chat) reaching into a task whose own child run
-    is currently mutating it. See issue #927: parent and child runs key on
-    different (guild_id, key) pairs and would otherwise never serialise
-    against each other for the same task.
+    never conflicts with its own per-task lock or scope — this only fires
+    for a *different* task_id, in two cases:
+
+    1. A per-task child context (``own_task_id`` set) is scoped to its own
+       task only (see docs/foreman-per-task-context.md, issue #997) — it
+       must not reach into another task's lifecycle, which stays the parent
+       Foreman's job.
+    2. A parent run (periodic-check or human chat) reaching into a task
+       whose own child run is currently mutating it. See issue #927: parent
+       and child runs key on different (guild_id, key) pairs and would
+       otherwise never serialise against each other for the same task.
     """
     if own_task_id == task_id:
         return None
+    if own_task_id is not None:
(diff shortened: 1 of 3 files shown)


jmelloy (commented):
**Verdict: LGTM — would approve, but GitHub blocks self-approval on this PR (same authenticated account authored it).** Posting as a comment review instead.

Reviewed the diff, ran the full backend suite locally (stood up a scratch Postgres 17 instance to match `docker-compose.yml`'s postgres-test config since it wasn't running in this sandbox).

**Verification against the 4 focus areas:**

1. **Tools exposed to child context** — confirmed `send_followup`/`redirect_task` are not in `_CHILD_EXCLUDED_TOOLS` (`backend/foreman/tools_schema.py`), so `CHILD_FOREMAN_TOOLS` already includes them. This PR isn't adding the tools themselves — it closes a scoping gap (#997) in `_task_mutation_blocked()` so a per-task child context can't call them (or `cancel_task`/`finalize_task`, which share the same guard) against a *different* task_id in the same guild.

2. **System prompt** — `CHILD_FOREMAN_SYSTEM` in `backend/foreman/prompt.py` already explicitly instructs the child to call `send_followup` and `redirect_task` on its own task ("Call send_followup when...", "Use message_worker to send mid-task context, redirect_task to course-correct"). No prompt changes were needed/made here since the prior gap was purely an authorization check, not a prompt-level capability issue.

3. **Tests** — `TestChildForemanFollowupAndRedirect` covers all 4 relevant combinations (child on own task for both tools: allowed; child on another task for both tools: rejected) plus a parent-run baseline. Ran them directly: all 5 pass. Full `test_foreman.py` (155 tests) and the full backend suite (1192 passed, 3 skipped) also pass. `ruff check` is clean on the touched files.

4. **Pattern consistency** — the new branch in `_task_mutation_blocked()` follows the existing early-return style, and call sites set `result_text`/`is_error` exactly like every other tool-error path in `tools.py`. Because `cancel_task` and `finalize_task` route through the same helper, they get the same cross-task protection as a natural side effect — nice, DRY consistency rather than a special case for just these two tools.

**Minor observations (non-blocking):**
- `message_worker` isn't routed through `_task_mutation_blocked` since it addresses a `worker_id` directly rather than a `task_id`, so a child context could in principle message a worker unrelated to its own task. That's a pre-existing, out-of-scope gap relative to this PR's stated goal (task-scoped mutation), not introduced here.
- No dedicated test exercises the new cross-task rejection for `cancel_task`/`finalize_task` specifically (only send_followup/redirect_task per the issue), though the shared-helper unit coverage makes this low-risk.

Implementation is correct, well-tested, and consistent with existing patterns.

PR #1014, opened 2026-07-25
feat: add simple Fibonacci implementation

jmelloy:
## Summary

Add `fibonacci.py` with two implementations of Fibonacci number generation:

1. **`fibonacci_iterative(n)`** — O(n) time, O(1) space iterative approach
2. **`fibonacci_recursive(n)`** — classic recursive approach

Includes a main block that prints the first 20 Fibonacci numbers using both methods, verifying they produce identical results.

## Testing

```bash
python fibonacci.py
```

Both methods produce matching output for n=0..19.

### DIFF
--- a/fibonacci.py
+++ b/fibonacci.py
@@ -1,0 +1,33 @@
+"""Simple Fibonacci implementations for testing."""
+
+
+def fibonacci_iterative(n):
+    """Return the nth Fibonacci number using iteration."""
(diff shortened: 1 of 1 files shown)


jmelloy:
Closing: recreating PR from correct branch ps/pi-tool-test-fibonacci-implementation-t-6rouz5

